PR #37, opened 2015-10-08
Pxc 432

ronin13:


sql/lock.cc
```diff
@@ -1095,7 +1095,8 @@ bool Global_read_lock::make_global_read_lock_block_commit(THD *thd)
     thd->mdl_context.release_lock(m_mdl_blocks_commits_lock);
     m_mdl_blocks_commits_lock= NULL;
     wsrep_locked_seqno= WSREP_SEQNO_UNDEFINED;
-    wsrep->resume(wsrep);
+    DBUG_ASSERT(!provider_resumed());
```
kbauskar, line 1098: I like this than unknown suprise that can happen with if()

PR #39, opened 2015-10-14
- PXC#440: PXC is failing with handle_fatal_signal (sig=11) when we s…

kbauskar:
…et values

  for certain wsrep variables

  Default value setting for most of wsrep variables was done in faulty way
  causing server to crash. When user demands to set default value a separate
  buffer that caches configured default value should be used for checking
  and updating the value (vs buffer that is normally populated when user
  provides a value for a given varaible)

  Also, NULL setting semantics for most of these variable was faulty too.

  There were some in-consistency in code (wrt to documentation).
  [Well the documentation was right in this case so we fixed the code
   to make it consistent]

  default value for
- wsrep_provider is "none"
- wsrep_provider_option is ""
- wsrep_data_home_dir is "mysql data directory"
- wsrep_cluster_name is "my_wsrep_cluster"
- wsrep_cluster_address is ""
- wsrep_node_name is ""
- wsrep_node_address is ""
  [documentation says: Usually set up as primary network interface (eth0)]
- wsrep_node_incoming_address is ""
  [documentation says: wsrep_node_address:3306]
  
  Also, most of the wsrep variables sys_vars tcs were missing. Adding them
  so that we can enable sys_vars suite for daily testing.
  [Part of the code-fixes is ported from MariaDB tree]

### DIFF
--- a/mysql-test/suite/sys_vars/r/all_vars.result
+++ b/mysql-test/suite/sys_vars/r/all_vars.result
@@ -1,8 +1,8 @@
 create table t1 (test_name text);
 create table t2 (variable_name text);
 load data infile "MYSQLTEST_VARDIR/tmp/sys_vars.all_vars.txt" into table t1;
-insert into t2 select variable_name from information_schema.global_variables where variable_name not like 'wsrep_%' and variable_name not like 'innodb_disallow_writes';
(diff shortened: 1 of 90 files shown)


kbauskar:
Verified stability using Jenkins run
http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1743/
http://jenkins.percona.com/job/percona-xtradb-cluster-5.6-mtr/51/

PR #40, opened 2015-10-15
- PXC-449: Fixing test-cases failing in galera suite

kbauskar:
This patch aims at fixing failing tcs in galera suite.
  Most of these test-cases are failing due to newly added variables
  for example: keep_pages_count or difference in semantics on
  handling "FLUSH TABLE WITH READ LOCK", etc....

  (Note: This is phase-1 of the fix. There would more such phases
   to fix other issues that are being discovered)


kbauskar:
Verified with Jenkins run

http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1745/

kbauskar:
LGTM

PR #41, opened 2015-10-15
- PXC#450: Assertion `key_coun t > 0' failed in galera/src/certificat…

kbauskar:
…ion.cpp:590

  SEMANTICS:

---

  What is the issues.
- If wsrep_certify_nonPK is set to 1 (default = 1) it will allow certification
  of table w/o PK by appending a unique key to the write-set.
  (though it is not recommended to create table w/o PK as part of
   best practices)
- If user configures cluster to avoid certification of table w/o pk by setting
  wsrep_certify_nonPK=0 then attempt to perform any such action will result in
  DEADLOCK error.
  ## ISSUE:
  
  SQL TC mentioned in the bug report failed to generate the deadlock error and
  went ahead with certification which then failed (assert). Why ?
- In normal flow, attempt to create table w/o pk with append_key disabled
  skips setting of trx_id which is an indication for caller to avoid
  certification and instead generate an error.
- Test-Scenario involved CREATE...SELECT (which in turn is
  create->select->insert selected data). SELECT processing caused the trx_id
  to set but this trx_id was not resetted before INSERT execution as both the
  actions SELECT and INSERT are part of same statement.
  
  This eventually led to an inconsistency where-in append_key routine is not
  executed as configuration has disabled it but trx_id is set to valid value
  confusing caller to proceed with certification.
  ## SOLUTION:
  
  If append_key for table w/o pk is disabled then wsrep_ws_handle->trx_id
  should be initialized to UNDEFINED. If not, then make sure it is set
  accordingly.

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_certify_nonPK_off.result
+++ b/mysql-test/suite/galera/r/galera_var_certify_nonPK_off.result
@@ -4,6 +4,8 @@ CREATE TABLE t1 (f1 INTEGER) ENGINE=InnoDB /* Table has no primary key */;
 CREATE TABLE t2 (f1 INTEGER PRIMARY KEY) ENGINE=InnoDB;
 INSERT INTO t1 VALUES (1), (2);
 ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+CREATE TABLE T3 SELECT 1 c;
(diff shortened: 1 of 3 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1746/

LGTM

PR #42, opened 2015-10-16
- PXC#451: INSERT .... SELECT for MYISAM table with myisam replicate …

kbauskar:
…= on crashes.

  Best practice recommend not to replicate myisam using galera as myisam doesn't
  support rollback but if user demands it by setting the variable
  wsrep_replicate_myisam = on/1 (default = off/0) then framework will execute
  these actions using TOI.

  TOI initialization is suppose to be done only once per statement also the
  initialization takes care of all tables involved in the statement.

  Test-Scenario exercised a INSERT .... SELECT case that involved 2 tables
  (destination and source table can be same as done in bug test-case)
  which in turn-casued TOI to get invoked twice which is not needed
  as first invocation has taken care of all tables involved in statement.

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
+++ b/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
@@ -50,6 +50,8 @@ COUNT(*) = 1
 SELECT COUNT(*) = 1 FROM t2;
 COUNT(*) = 1
 1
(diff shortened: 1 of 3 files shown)


kbauskar:
Jenkins: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1747/

PR #44, opened 2015-10-16
- PXC#418: wsrep_cluster_name over 32 characters causes gmcast messag…

kbauskar:
…e max

  length overrun

  wsrep_cluster_name is used by joining node to join the cluster group started
  by the bootstrapping cluster node.

  This communication for handshake is done using gcomm protocol which enforces
  a static limit on buffer than holds group_name (technically very much needed
  to avoid memory bloat).

  This static limit for group_name is 32 bytes and so if cluster_name is
  greater than this static limit gcomm handshake will fail.

  In order to avoid this failure at handshake stage it is better to enforce
  such limit at wsrep_cluster_name validation step.

  From now on wsrep_cluster_name will have a hardlimit of 32 bytes.
  Trying to configure a cluster with cluster_name of size > 32 will result in
  server start failure. Trying to set cluster_name of size > 32 using SET
  command will return error without changing the cluster_name.

  [Note: We cautiously avoid auto-truncation as that would confuse user and
   may break a monitoring tool that may be using cluster-name for montioring]

### DIFF
--- a/mysql-test/suite/sys_vars/r/wsrep_cluster_name_basic.result
+++ b/mysql-test/suite/sys_vars/r/wsrep_cluster_name_basic.result
@@ -37,6 +37,22 @@ my_wsrep_cluster
 # invalid values
 SET @@global.wsrep_cluster_name=NULL;
 ERROR 42000: Variable 'wsrep_cluster_name' can't be set to the value of 'NULL'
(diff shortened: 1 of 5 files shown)


kbauskar:
Jenkins run: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1748/

LGTM

PR #45, opened 2015-10-19
- PXC#453: Hitting assert when myisam table is being loaded

kbauskar:
(with replicate-myisam) from schema based temporary table.

  If wsrep_replicate_myisam is turned ON it will try to replicate
  MyISAM table using TOI. But this replication is not applicable
  to temporary tables as it doesn't make sense to replicate session
  specific tables like temporary tables.

  Code snippet that processes prepare keys for isolation has check to
  identify and skip temporary tables.

  This check for temporary table retrieves table_cache key using
  "get_table_def_key" which in turns validates table key attributes
  by comparing db_name/table_name with mdl_request registered
  db_name/table_name.

  If table is temporary schema table (information_schema) then name of
  table is generally set to path of the table and original name is
  cached in alias.

  Validation check was only considering table_name which would not match
  with MDL key which is created before the table_name is set to path
  by open_and_process_table -> mysql_scheme_table function.

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
+++ b/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
@@ -83,6 +83,7 @@ use test;
 create table is1 (i int) engine=myisam;
 create table is2 (i int) engine=myisam;
 create table is3 (i int) engine=innodb;
+create table is4 (i int) engine=myisam;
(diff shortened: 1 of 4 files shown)


kbauskar:
Jenkins run: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/

LGTM

PR #47, opened 2015-10-21
- PXC#457: Enabling fake-change for PXC crashes server

kbauskar:
(fake-change should be offically blocked)

  Fake change is feature that allows slave to warm-up there-by avoiding
  further replication lag in case of async replication.

  PXC is meant for almost synchronous replication using galera plugin
  so fake-change feature doesn't apply.
  Semantics and Documentation has already clarified this but blocking
  of it was missed in codebase causing random failure as PXC flow
  was interrupted or modified by use of fake-changes.

  Fixed to disable setting of fake-change by getting rid of
  fake-change related variables.


kbauskar:
Jenkins run: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1751/

LGTM

PR #48, opened 2015-10-22
- PXC#460: wsrep_sst_auth don't work in Percona-XtraDB-Cluster-56-5.6…

kbauskar:
….25-25.12.1.el7

  Semantics:

---
- Generally end-user will create a separate user with needed privileges for
  performing DONOR action.
- This user credentials are specified using wsrep_sst_auth.
- Along with this user there could be other user(s) created on the server
  that sysadmin may use for normal or other operations
- Credentials for these user(s) can be specified in same cluster/server
  cnf file as part of [client] section
  
  When cluster act as DONOR and if wsrep_sst_auth is provided then it should
  strictly use it for performing SST based action.
  
  What if end-user has same credentials for performing both SST action and
  normal admin work ?
- Then end-user can simply specify these credentials as part of [client]
  section in cnf file and skip providing wsrep_sst_auth.
  ## Issue:
  
  MySQL client user/password parsing preference order is as follows:
- command line (through --user/--password)
- cnf file
- MYSQL_PWD enviornment variable.
  
  Recent change tried passing sst user password through MYSQL_PWD
  (and user though --user command line param as before).
  
  On the system where-in admin had another user for performing non-SST actions,
  credentials for such user were present in cnf file under [client] section.
  
  Due to mysql client preference order, SST user name was used (as it was
  passed through command line) but password of other user (meant for non-SST)
  action was being used as it was passed through cnf file.
  
  Password passed through MYSQL_PWD was completely ignored causing
  user-name/password mismatch.
  ## Solution:
- If user has specified credentials for SST then pass them through command
  line so that they are used in priority.
  
  (There could be security concern on passing things through command line but
  when I tried passing user-name and password through command line to mysql
  client and then did ps I saw this
  ./bin/mysql --user=sstuser --password=x xxxxxxxx -S /tmp/n1.sock
  so seems like password is not shown)

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -692,8 +692,7 @@ then
         fi
 
         if [ -n "${WSREP_SST_OPT_PSWD:-}" ]; then
-#           INNOEXTRA+=" --password=$WSREP_SST_OPT_PSWD"
-           export MYSQL_PWD="$WSREP_SST_OPT_PSWD"
+           INNOEXTRA+=" --password=$WSREP_SST_OPT_PSWD"
         elif [[ $usrst -eq 1 ]];then
            # Empty password, used for testing, debugging etc.
            INNOEXTRA+=" --password="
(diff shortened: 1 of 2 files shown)


kbauskar:
This is script level change and Jenkin build and mtr mtr test will not help to validate this change.

Verified the thing locally.

kbauskar:
LGTM

PR #49, opened 2015-10-22
5.6 - Multiple binary tarballs and disable TokuDB

tplavcic:
**PROBLEM:**
For PS we build multiple binary tarballs for release because of libssl linking so this is to make it the same for PXC (just added some variables into build-binary.sh the rest is in jenkins).
Also now that PXC is built with system readline we have the bug from below which will be resolved with this too.
The other change is to disable TokuDB build by default since it's not supported in PXC so I got some errors.

**BUG:**
https://bugs.launchpad.net/percona-xtradb-cluster/+bug/1506335

**TEST BUILD:**
http://jenkins.percona.com/job/percona-xtradb-cluster-5.6-binary-new/7/


tplavcic:
Merging as agreed with Krunal it's a simple change.

PR #53, opened 2015-10-28
Remove Makefile-pxc as it's not needed

tplavcic:
This falls under consolidation category.
We remove Makefile-pxc since it's currently used only for grepping the version from it - but that version is already available in the VERSION file (similar makefile was before in PS also but it was removed earlier).
The script build-binary-nogit.sh is moved to the "deprecated" folder since it's not used any more and seems out of date.

TEST BUILD:
http://jenkins.percona.com/job/percona-xtradb-cluster-RELEASE-5.6-new/10/


kbauskar:
Makefile-pxc is needed by build script that normal Jenkins Jobs use for validating bug/blueprint changes. I am yet to switch to a generic script that is build-binary.sh .. Will do it shortly. You can merge this PR post that. I will update the PR

tplavcic:
@kbauskar 
Which job would that be? Also if you wish I can fix it just not sure currently where is it...

PR #54, opened 2015-10-29
- PXC#461: Assertion `!provider_resumed()' failed in sql/lock.cc:1098:

kbauskar:
bool Global_read_lock::make_global_read_lock _block_commit(THD*)

  If FTWRL acquires lock and followup commands needs same lock it not
  re-acquired but if follow-up command fails then it shouldn't free
  the lock that is hasn't acquired.


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1761/

LGTM

PR #55, opened 2015-10-29
- PXC#458: DELAYED_INSERT on MYISAM table causes TOI state in-consist…

kbauskar:
…ency

```
       (TOI_END is not called)
```

  PXC#459: DELAYED_INSERT on MYISAM table causes WSREP-MDL conflict due even
           when wsrep is not enabled.

  INSERT DELAYED is executed in asynchronous fashion as in a separate handler
  thread is created and it executes the INSERT statement.

  This asynchronous replication doeesn't go well with galera's synchronous
  replication for following reasons:
  a. ID is supported by MyISAM (not by InnoDB) and galera doesn't by default
     support MyISAM replication
  b. Existing semantics is broken as ID is replicated using TOI/RSU.
     Code starts TOI but fail to close it. (Well this can be fixed but
     that is like allowing ID to execute using TOI model in asynchronous
     and till it is done it will cause all the other TOI enabled statement
     to wait even though client has reported success).

  In general DELAYED clause is ignored on asynchronous slave too to avoid
  in-consistency with master.

  Also, there are MDL level conflict that arises when RSU mode is used.
  ID handler trying to execute INSERT statement when parallely another
  DDL statement requesting the same MDL lock which again doesn't go
  well with galera MDL check semantics.

  With all those shortcoming we decided to ignore DELAYED clause to
  avoid in-consistency. (We could have fixed the pointed issues
  but semantics in-consistency would continue)

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
+++ b/mysql-test/suite/galera/r/galera_var_replicate_myisam_on.result
@@ -115,3 +115,53 @@ drop table is1;
 drop table is2;
 drop table is3;
 drop table is4;
+#node-1
+set @@global.wsrep_replicate_myisam = 1;
+set @@global.wsrep_OSU_method="TOI";
+use test;
+create table is1 (i int) engine=myisam;
+insert into is1 values (1), (2);
+#node-2
+select * from is1;
+i
+1
+2
+#node-1
+insert delayed is1 values (3), (4);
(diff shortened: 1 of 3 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1763/

LGTM

PR #56, opened 2015-10-29
- PXC#462: Turning off wsrep provider in middle of trx leaves trx in an

kbauskar:
inconsistent state.

  wsrep_on can be use to temporarily turn-off replication.

  Current scope of wsrep_on is global/local which means if some client
  connection turns it off then other client connections (connected to
  same node) will start experiencing the effect of it all of sudden
  unknowingly.
  If the value get toggled in middle of transaction then it would
  cause in-consistent state with part of transaction executed with
  replication on and other part without replication on.

  In order to solve all of these issues following solution has been
  implemented:
  a. wsrep_on will be session_only (local) variable moving forward.
  b. Trying to toggle its value in middle of transaction would
     result in error. Given the scope is local only other active
     transactions started by parallely connected client is not affected.

  Well sql_log_bin can help achieve the same effect.
  In theory yes but sql_log_bin stops bin-logging which means async
  replication slaves also doesn't see the events.
  wsrep_on effects only galera based setup by stopping replication
  to galera replicated cluster node only.

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_wsrep_on_off.result
+++ b/mysql-test/suite/galera/r/galera_var_wsrep_on_off.result
@@ -1,3 +1,4 @@
+call mtr.add_suppression("Cannot modify @@session.wsrep_on");
 CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) ENGINE=InnoDB;
 INSERT INTO t1 VALUES (1);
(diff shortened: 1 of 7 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1764/

LGTM

PR #57, opened 2015-10-30
- PXC#454: Stopping/Resetting wsrep_provider in middle of trx causes …

kbauskar:
…assert as

  cleanup is not done

  wsrep_provider points to the galera-plugin library. This variable is dynamic
  and can be set on the fly to either to point to a different version of
  galera-plugin or none.

  Generally user use it to point to none to get a specific node out of cluster
  and still continue using it as a normal standalone Percona Server node.

  Changing this variable will cause all connections (except the connection that
  triggered the change) to the given node to disconnect.

  If this change in setting is trigger while a transaction is active it
  can leave the server in an inconsistent state.

  To avoid this behavior changing this value in middle of transaction
  or as part of procedure or trigger is now blocked.


kbauskar:
Jenkins Build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1765/

LGTM

PR #59, opened 2015-11-05
- PXC#467: LDI on partitioned table fails

kbauskar:
Issue:

---

  LDI on partitioned table fails

  Reason:

---

  Partition table are handled using ha_partition handler which in turn creates
  "n" different ha_innodb handlers one for each partition.
  When data is inserted in partitioned table to avoid repeat bin-logging
  partition handler disables bin-logging while invoking ha_innodb interface and
  re-enables it on completion. bin-logging is then done as part of ha_partition
  handler interface.

  ha_innodb skips key-append if bin-logging is disabled (and especially with LDI
  involved). [This is looks like a design choice than bug]
  Without append_key, statement can't be be forward to galera plugin for
  certification which actually causes LDI to fail.

  Solution:

---

  Relax the append-key condition to consider involvement of LDI with partitioned
  table.


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1778/

LGTM

PR #60, opened 2015-11-05
- PXC#404: PXC node failure when drop table command share with tempor…

kbauskar:
…ary table

  DROP TABLE is ambiguous if there exist 2 objects with same name but
  different type TEMPORARY and NON_TEMPORARY.

  drop table t1;
  there could be 2 objects (temporary and non-temporary) with same name t1.

  It is like local and global scope in C programming language.

  MySQL semantics resolves such ambiguity by referening local (in this case
  temporary object).

---

  Galera replication doesn't replicate TEMPORARY object as their scope
  is localized to the thread that created it and that in case of replication
  would be applier thread making it in-accessible to end-user.

---

  DROP TABLE statement with mix of temporary and non-temporary object
  such that there exist a non-temporary object with same name as temporary
  object specified in DROP TABLE confuses replication.

  Replication algorithm detects presence of non-temporary object and replicate
  the event in form of query as is there-by carrying removal of temporary object
  too which is not part of replicated nodes.

  This means on replicated node drop table references non-temporary objects
  and get rid of it making it in-consistent with workload executor node.

  Algorithm now detect such case (mix of temporary and non-temporary)
  and creates a fake temporary object (as part of DROP workload) for
  replicated node to remove.

### DIFF
--- a/mysql-test/suite/galera/r/galera_create_table_like.result
+++ b/mysql-test/suite/galera/r/galera_create_table_like.result
@@ -46,3 +46,25 @@ DROP TABLE schema2.real_table2;
 DROP TABLE schema2.real_table3;
 DROP SCHEMA schema1;
 DROP SCHEMA schema2;
+#node-1
+use test;
+create table t1 (c1 int) engine=innodb;
+create table t2 (c1 int) engine=innodb;
+create temporary table t1 (c1 int) engine=innodb;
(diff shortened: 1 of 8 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1779/

LGTM

PR #61, opened 2015-11-10
- PXC#469: CTAS incompatibility and failures

kbauskar:
CTAS (CREATE TABLE ... AS SELECT) is executed in normal form
  (not as TOI statement like other CREATE due to involvement of SELECT)

  This effectively causes issue when source table is empty which
  in turns skip appending of keys and eventually result in a debug
  mode assert during certification.

  Debug mode assert is good but failed to consider all possible cases.
  Forcing a fake key for making assert happy is wrong idea so we commented
  the assert which anyway is debug mode assert and not so fatal.
  Alternative would be to suppress assert only if command = CTAS
  but galera-plugin doesn't have access to such command.
  (unless we make interface changes).

### DIFF
--- a/mysql-test/suite/galera/r/galera_create_table_like.result
+++ b/mysql-test/suite/galera/r/galera_create_table_like.result
@@ -68,3 +68,68 @@ select * from t1;
 c1
 11
 drop table t1;
+#node-1
+set wsrep_replicate_myisam = 1;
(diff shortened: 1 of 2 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/build-xtradb-cluster-binaries56-clone/1780/

LGTM

PR #64, opened 2015-11-27
Fix service-startup-timeout for systemd

Vlad1mir-D:
PXC won't start if found service-startup-timeout under [server] or [mysqld]
so parse_cnf should search for service-startup-timeout under [mysqld_safe].


tplavcic:
Thanks for contribution! We should also remove the "server" and "mysqld" sections from that search since the server doesn't know anything about "service-startup-timeout" variable.
`2016-09-15 03:27:13 11868 [ERROR] /usr/sbin/mysqld: unknown variable 'service-startup-timeout=100'`
So I'll close your PR, open a bug and propose another similar change where it will just search for that variable in mysqld_safe section of my.cnf.

PR #65, opened 2015-11-30
- PXC#486: Correcting the DDL/DML replication semantics for MyISAM ta…

kbauskar:
…ble with

  wsrep_replicate_myisam=off

  Existing or Updated Semantics:

  a. DDL statement: CREATE/DROP/TRUNCATE/.... for engine=myisam
     Will be replicated ir-respective of wsrep_replicate_myisam settings.
     This is done to ensure that name of the object is blocked so that user
     can't create another object with same name on replicated node (with or
     without different schema structure).
     [No change in innodb semantics]

  b. DML statement: INSERT/UPDATE/DELETE/SELECT .... for engine=myisam
     Will be replicated only if wsrep_replicate_myisam = 1.
     Expect node in-consistency if user try to run part of workload with
     wsrep_replicate_myisam = 0 and rest of the part with
     wsrep_replicate_myisam = 1

  c. SST xfer:
     A newly booted cluster node that get full SST will also get MyISAM table
     (schema + data) ir-respective of wsrep_replicate_myisam setting.

  d. enforce_storage_engine (options specific to PXC)
     If this option is set and sql_mode doesn't conflict with the setting
     (set sql_mode = "") then even if user request creation of MyISAM table it
     will be auto-converted to SE dictated by enforce_storage_engine option.
     Replication status is captured even before enforce storage engine variable
     modifies the table execution path and so for replication original statement
     as issued by user is used. If this option is set for #node-1 and not for
     #node-2 then expect in-consistency in cluster operation as #node-1 would
     end up auto-updating SE for the table to "InnoDB" but #node-2 may continue
     to create the same table with original user specified engine.
     (This limitation should be documented)

  e. CTAS statement: This is executed and handled differently as it involves
     DML + DDL. CTAS semantics are already confusing in upstream.
     PXC ensures that if the CTAS "select" has myisam table only
     (or no table select 1) such queries are not replicated.

```
 Why is this so ?
 - If there is involvement of innodb table then trx context has to be setup
   and wsrep-plugin is design to work only if trx context is setup.

 Then how come other myisam stmt replicated ?
 - Using TOI and CTAS doesn't use TOI.
```

### DIFF
--- a/mysql-test/suite/galera/r/galera_create_table_like.result
+++ b/mysql-test/suite/galera/r/galera_create_table_like.result
@@ -69,13 +69,13 @@ c1
 11
 drop table t1;
 #node-1
+use test;
+create table src_i (i int, primary key pk(i)) engine=innodb;
+create table src_m (j int, primary key pk(j)) engine=myisam;
 set wsrep_replicate_myisam = 1;
 select @@wsrep_replicate_myisam;
 @@wsrep_replicate_myisam
 1
-use test;
-create table src_i (i int, primary key pk(i)) engine=innodb;
-create table src_m (i int, primary key pk(i)) engine=myisam;
 create table t1 engine=innodb as select * from src_i;
 create table t2 engine=innodb as select * from src_m;
 create table t3 engine=innodb as select 1;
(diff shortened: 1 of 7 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1793/
Jenkins mtr: http://jenkins.percona.com/job/pxc56.test.mtr/

kbauskar:
LGTM

PR #67, opened 2015-12-03
- PXC#487: Nodes getting blocked with "flush tables" command

kbauskar:
Enabling replication of FLUSH TABLES causes hang on the node
  where the instruction gets replication with parallel DML
  workload active.

  Issue:

---

  Replicated event FLUSH TABLES is executed as an background
  action. Before clearing the table cache FLUSH TABLES
  marks existing table cache as invalid by incrementing
  refresh_version (which then != share->version()) and so
  new booting trx will detect share->has_old_version() and will
  wait for old version to die.
  (FLUSH TABLES MDL based co-ordination logic.)

  During this process if there are already active DML statements
  (before flush tables incremented refresh_version) then FLUSH TABLES
  thread will wait for ref-count of such shared version to drop to 0.

  Now with galera replication involved let's consider a use-case:

  a. DML statement actively executing.
  b. FLUSH TABLE starts execution as replication event.
  c. FLUSH TABLE gets galera-execution-token (seqno) assigned.
  d. DML statement now enter galera space for commit certification
     and get galera-execution-token assigned (that is > FLUSH TABLES
     token).
  e. FLUSH TABLES thread waits for ref-count to drop to 0.
  f. DML statement wait in galera commit-ordering as action with token
     less that it is pending execution. This blocks ref-cnt to drop to 0.

  DEADLOCK.

  This issue was reported in FTWRL context as bug#1280768.

  Proper fix would be to sort this double co-ordination.
  (MDL based on cordination in MySQL and
   Commit Ordering based cordination in Galera).

  Current fix (workaround) is to avoid FLUSH TABLES replication as it earlier.
  It was enabled to avoid GTID inconsistency as FLUSH TABLES command is written
  to bin-log but not replicated which then causes issues if async slave's master
  (that is using galera-node as master) is switched from galera-node-a
  to galera-node-b.

  Skipping FLUSH TABLES replication (if node is operating as galera node
  that is WSREP_ON) will have effect on async replication as FLUSH TABLES
  will not be replicated to async slave.

  Related bugs: lp:1520491, lp:1280768


kbauskar:
Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1797/
mtr: http://jenkins.percona.com/job/pxc56.test.mtr/92/

LGTM

PR #68, opened 2015-12-04
- PXC#485: handle_fatal_signal (sig=11) in get_quote_char_for_identif…

kbauskar:
…ier |

  sql/sql_show.cc:1209

  If prepare statement involves DDL galera replicate it using TOI.
  During this replication it creates a lex->definer that has needed
  details about host and user-name.

  This lex->definer should be created in persistent memory arena as
  it would be re-used when the prepare statement is re-executed.
  Galera TOI logic missed putting it to persistent memory arena
  which eventually lead to stale pointer pointing to freed memory
  if the same prepare statement is re-executed.

### DIFF
--- a/mysql-test/suite/galera/r/galera_prepared_statement.result
+++ b/mysql-test/suite/galera/r/galera_prepared_statement.result
@@ -31,3 +31,13 @@ DROP TABLE t1;
 DROP TABLE t2;
 DROP TABLE t3;
 DROP TABLE t4;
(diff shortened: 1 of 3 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1798/
mtr: http://jenkins.percona.com/job/pxc56.test.mtr/93/ + http://jenkins.percona.com/job/pxc56.test.mtr/94/

LGTM

PR #69, opened 2015-12-08
- PXC#480: xtrabackup-v2 SST fails with multiple log_bin directives i…

kbauskar:
…n my.cnf

  If any given variable the xtrabackup-v2 sst script looks for is specified
  multiple times in cnf file then it tend to pick both of them causing
  some of the follow-up command to fail.

  Avoid this programatic mistake by honoring only the last variable assigned
  setting as done by mysqld too.

  Check https://bugs.launchpad.net/percona-xtradb-cluster/+bug/1362830

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -240,7 +240,11 @@ parse_cnf()
 {
     local group=$1
     local var=$2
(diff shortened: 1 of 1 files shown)


kbauskar:
Jenkins run: http://jenkins.percona.com/job/pxc56.clone/1800/
MTR: http://jenkins.percona.com/job/pxc56.test.mtr/96/ http://jenkins.percona.com/job/pxc56.test.mtr/97/

kbauskar:
LGTM

PR #70, opened 2015-12-10
- PXC#479: Setting wsrep_desync=1 after FTWRL blocks a node

kbauskar:
FTWRL pauses cluster which is node specific action and wsrep_desync
  try to turn-off flow control for the said node which too is node specific
  action.

  FTWRL action is complete only after user executed UNLOCK TABLES.
  So if end-user try to toggle wsrep_desync while FTWRL is active (till unlock
  is executed) then it would cause galera plugin to halt/deadlock.

  Let's now understand the semantics a bit.

  a. FTWRL is mean to pause the cluster-node even though write-sets are recieved
     by cluster-node they are not applied.

  b. desync=1 suggest that this node will participate in flow-control.

  If node is paused then of-course the write-queue is bound to grow
  and will hit flow-control so if end-users can explicitly disable
  flow-control before triggerring FTWRL but once the cluster is paused
  then cluster node can't be desynced or re-synced till the FTWRL action is
  complete as both of the line would then act contraditory.

### DIFF
--- a/mysql-test/suite/galera/r/galera_flush.result
+++ b/mysql-test/suite/galera/r/galera_flush.result
@@ -1,4 +1,7 @@
 DROP TABLE IF EXISTS t1, t2;
+call mtr.add_suppression("InnoDB: Resizing redo log from .*");
+call mtr.add_suppression("InnoDB: Starting to delete and rewrite log files");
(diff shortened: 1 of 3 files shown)


kbauskar:
Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1801/
LGTM

PR #72, opened 2015-12-18
Merging PS-5.6.27 and codership-mysql-wsrep-5.6.27 changes to PXC-5.6.27

kbauskar:
This patch try to merge/refresh PXC-5.6.27 from PS-5.6.27 to get PS 5.6.27 specific changes including MySQL and same for codership maintain mysql-wsrep plugin changes.

No major code conflict observed.
Most of the conflicts are related to build/man/doc related.

Some key important changes:

[from codership refresh]
- The MySQL-wsrep packages are now built with OpenSSL rather than YaSSL (codership/mysql-wsrep#121)
- Galera error messages have been enhanced to contain the current schema name along with the query (codership/mysql-wsrep#202)
- IB atomic builtins are no longer used when compiling as they may cause the server to hang 
  (codership/mysql-wsrep#221)
- query cache is now compatible with wsrep_sync_wait (codership/mysql-wsrep#201)
- a deadlock could occur between the applier thread and an aborted transaction (codership/mysql-wsrep#184)
- DDL was not recorded in InnoDB header, causing InnoDB to recover to an earlier position (codership/mysql-wsrep#31)
  Rest of the details could be found at http://galeracluster.com/2015/11/announcing-galera-cluster-5-5-42-and-5-6-25-with-galera-3-12-2/

[from percona-server refresh]

Check for the fixes in https://www.percona.com/blog/2015/11/05/percona-server-5-6-27-75-0-is-now-available/ (Given this is PS refresh all fixes are PS specific nothing cluster related)


kbauskar:
First refresh from PS was successful. 
Here is the mtr and build jenkins job
http://jenkins.percona.com/job/pxc56.test.mtr/103/
http://jenkins.percona.com/job/pxc56.build/484/

PR #78, opened 2016-01-18
- PXC#490: crash-safe slaves: relay_log_recovery breaks replication (…

kbauskar:
…CID#62894)

  Issue:

---

  Replication info table like slave_master_info, slave_relay_log_info are
  created in InnoDB to support crash-safe replication that was introduce in
  MySQL-5.6

  Being a innodb tables they default qualify for galera replication but they
  shouldn't as per the protocol as they are meant to be kept on active
  ASYNC slave only.

  MySQL flow does block bin-logging of events performed on these tables but
  WSREP replication missed blocking its replication.
  This causes start of trx in galera cluster which was not needed.

  Infact, the condition to append_key that starts trx in the galera eco-system
  was not adequated to handle case where-in bin-logging is turned-off for
  a specific flow like update performed to replication info table.

  Solution:

---

  Corrected condition that trigger start of replication in galera cluster.
  If bin-logging is turned-off then we avoid starting trx in galera eco-system.

  Note: We also explicitly allows replication when binlog-format = STATEMENT.
  This is an existing semantics so nothing new here but we made is explict
  and there-by kept the hook open only for STATEMENT and not for MIXED.
  It is interesting to learn that this hook is needed only for pt-table-checksum
  to operate that unfortunately relied on so called non-supported functionality
  of PXC cluster (replication under SBR is not supported).

### DIFF
--- a/mysql-test/suite/galera/r/galera_as_slave_crash_safe.result
+++ b/mysql-test/suite/galera/r/galera_as_slave_crash_safe.result
@@ -1,0 +1,24 @@
+#node-3 (async slave)
+call mtr.add_suppression("Recovery from master pos");
+START SLAVE USER='root';
+Warnings:
+Note	1759	Sending passwords in plain text without SSL/TLS is extremely insecure.
+#node-1 (master)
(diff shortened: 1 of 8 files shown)


kbauskar:
LGTM

PR #82, opened 2016-01-29
Release Percona XtraDB Cluster 5.6.28-25.14 with Galera 3.14

tplavcic:
Release branch with tag


hrvojem:
There are issues with this merge with doc/ folder, because it contains PS docs (release notes, fixes to PS docs which are not part of PXC, tokudb docs) this should be cleaned up before merging to trunk otherwise it breaks the doc deployment

kbauskar:
Addressed PS inherited doc issue.

hrvojem:
LGTM for docs changes

kbauskar:
LGTM.
(Large merge that we see above is inherited from PS. PS inherited it by folding MS changes)

PR #85, opened 2016-02-04
- #PXC-474: Async replication + auto-increment + log-format (STMT) = inconsistent cluster node

sysprg:
This is patch for the PXC-474 issue: suppose we have a PXC 2+ node cluster which is replicating from an async master. Now if the binlog_format is STATEMENT and multi-row inserts are executed on a table with an auto_increment column such that values are automatically generated by MySQL, then the PXC node generates wrong auto_increment values which are different from what was generated on the async master.

Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1847/

### DIFF
--- a/mysql-test/suite/galera/t/galera_binlog_stmt_autoinc.test
+++ b/mysql-test/suite/galera/t/galera_binlog_stmt_autoinc.test
@@ -1,0 +1,114 @@
+##
+## Tests the auto-increment with binlog in STATEMENT mode.
+##
+
+--source include/galera_cluster.inc
+--source include/have_innodb.inc
+
+##
+## Verify the correct operation of the auto-increment when the binlog
+## format artificially set to the 'STATEMENT' (although this mode is
+## not recommended in the current version):
+##
+
+--connection node_2
+SET GLOBAL wsrep_forced_binlog_format='STATEMENT';
+
+--connection node_1
+SET GLOBAL wsrep_forced_binlog_format='STATEMENT';
+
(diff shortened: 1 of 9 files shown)


kbauskar:
Description which eventually will turn as commit message need to explain solution too.
Problem is well explained but solution implementation description is missing.

mysql-test/suite/galera/t/galera_binlog_stmt_autoinc.test
```diff
@@ -0,0 +1,84 @@
+##
+## Tests the auto-increment with binlog in STATEMENT mode.
+##
+
+--source include/galera_cluster.inc
```
kbauskar, line 5: Use-case involves async master with one of galera node from galera cluster acting as slave.

I see you have only galera cluster involved. How is this testing the said use-case.

sql/mysqld.cc
```diff
@@ -4172,6 +4172,13 @@ int init_common_variables()
   */
   global_system_variables.time_zone= my_tz_SYSTEM;
 
+#ifdef WITH_WSREP
```
kbauskar, line 4175: Please document all your code-snippet. It is difficult to reverse engineer each snippet + time consuming. Defeat purpose of multiple developers working.

kbauskar:
I think I can review the patch better once you have documented all the snippets explaining why you are saving auto_increment\* values. Also, commit message should clearly explain what was going wrong and how did you fixed it.

PR #89, opened 2016-02-15
- #PXC-468: Some of the galera mtr suite tcs are failing intermittently

sysprg:
Some of the galera mtr suite testcases are failing intermittently: galera_admin, galera_flush_local, galera_flush_gtid, galera_split_brain and galera_kill_nochanges.

This is partly due to the fact that most of them should be marked as "big test". Thus it is necessary to include "big_test.inc" to each of long-running test cases.

In addition, we need to adjust the output file for galera_flush_gtid.test. It does not reflect the recent changes in the galera_flush.test (which is indirectly included from the galera_flush_gtid.test).

In addition, we need to change one regexp in the main mtr script (mysql-test-run.pl) to ensure that it did not consider all the options that begin with the word "user" as the "user=<name>" and do not forget to add "--user=root" when it is required.

We also need to add a forgotten line (with notification that the server was killed) to the galera_kill_nochange results file.

Then we need to remove the corrected tests from the list of disabled tests.

Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1850/

### DIFF
--- a/mysql-test/mysql-test-run.pl
+++ b/mysql-test/mysql-test-run.pl
@@ -5225,7 +5225,7 @@ sub mysqld_arguments ($$$) {
   # to start unless we specify what user to run as, see BUG#30630
   my $euid= $>;
   if (!IS_WINDOWS and $euid == 0 and
-      (grep(/^--user/, @$extra_opts)) == 0) {
+      (grep(/^--user=/, @$extra_opts)) == 0) {
     mtr_add_arg($args, "--user=root");
   }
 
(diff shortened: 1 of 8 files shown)


kbauskar:
big-test should be used if overall execution time of test-case is higher.
TC that are marked with big-test in patch are failing due to timeout in wait-condition.
It has been observed that the default timeout these TCs uses is sufficient provided the machine is lightly loaded as the same TCs pass on 2nd/3rd automated try done by mtr if 1st try fails.
So may be increasing timeout so that 1st try too passes will help.

PR #90, opened 2016-02-18
- PXC#497: Warning: "BF applier failed to open_and_lock_tables"

kbauskar:
Issue:

---

  Let me explain it with an example. Would be simpler.

  Let's assume there are 2 pxc nodes and each node and following
  workload is being

  node-1: executes already prepare stmt and about to commit
  (update i > 2 and i < 5)

  node-2: executes some stmt that would conflict with node-1 (insert (3))
  inserting in gap.

  As per the protocol local trx is selected for rollback and so
  when node-1 get node-2 trx which is applier trx node-1 local trx that
  is prepare stmt execution is rolled back and re-tried using replay trx.

  Note: retried using replay trx effectively means only write-set
  is applied this makes it bit faster.
  Application of write-set uses the apply-log-event path which
  open and locks table so there is no need to keep track of
  table versioning change which is normally done (in normal path)
  by installing a reprepare observer.

  Solution:

---
- Avoid installing observer when prepare trx is being replayed
  as trx will open and lock the tables.

### DIFF
--- a/mysql-test/suite/galera/r/galera_transaction_replay.result
+++ b/mysql-test/suite/galera/r/galera_transaction_replay.result
@@ -30,3 +30,43 @@ SELECT COUNT(*) = 1 FROM t1 WHERE f2 = 'c';
 COUNT(*) = 1
 1
 DROP TABLE t1;
+#node_1
+create table t1 (i int primary key, c char(1)) engine=innodb;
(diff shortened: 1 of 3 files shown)


kbauskar:
LGTM.

http://jenkins.percona.com/job/pxc56.test.mtr/171/
http://jenkins.percona.com/job/pxc56.test.mtr/172/

PR #93, opened 2016-03-01
Remove code per PXC-500

kennt-percona:
Build completed

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -688,6 +688,7 @@ void wsrep_deinit()
 
 void wsrep_recover()
 {
+  #if 0
   if (!memcmp(&local_uuid, &WSREP_UUID_UNDEFINED, sizeof(wsrep_uuid_t)) &&
       local_seqno == -2)
   {
@@ -696,7 +697,7 @@ void wsrep_recover()
     WSREP_INFO("Position %s:%lld given at startup, skipping position recovery",
                uuid_str, (long long)local_seqno);
     return;
-  }
+  #endif
   wsrep_uuid_t uuid;
   wsrep_seqno_t seqno;
   wsrep_get_SE_checkpoint(uuid, seqno);


sql/wsrep_mysqld.cc
```diff
@@ -688,6 +688,7 @@ void wsrep_deinit()
 
 void wsrep_recover()
 {
+  #if 0
```
kbauskar, line 691: Please add comment explaining why this code is a dead-code.

kbauskar:
1. Commit message has to be descriptive explaining what changes are being done.
2. I assume you have tested some of the flow described in bug report. Please include them in commit message. (As in did you got the code to enter that if loop ? If yes how and then why is that ir-relevant)
3. Build done needs to be annotated with corresponding Jenkins link.
   (Also, I see the build scheduled for this change failed for some distro. You need to investigate what the is cause of the failure and if needed reschedule the build (only for failing distro unless there is a code change).
4. If build fails TCs will not be executed. Once build passes for all distro then TCs will be executed (In a way it is linked). So once you get build to pass for all distro then let the tc run and re-check what all things are failing.

kennt-percona:
To hit this code path, use "--wsrep-recover --wsrep-start-position=00000000-0000-0000-0000-000000000000:-2" from the command line.  mysql will start up, print out this uuid/seqno (in this case the uuid with all 0's and a seqno of -2), and then exits.

This code occurs in the wsrep_recover() in sql/wsrep_mysqld.cc.  This function only gets called when "--wsrep-recover" is used.  Since seq numbers start at -1 and are then incremented, calling it with -2 has to be done explicitly (via --wsrep-start-position).

The only other code I found that relies on the output from this specific case is in mysqld_safe.sh but that code only calls wsrep_recover() when the seqno is -1, so it should never see hit this codepath.

So, it seems hard to find a use for this, unless the act of starting and then shutting down the server itself is useful, but this is an undocumented feature.

PR #94, opened 2016-03-01
5.6 pxc 507

kennt-percona:
Build completed

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -688,6 +688,7 @@ void wsrep_deinit()
 
 void wsrep_recover()
 {
+  #if 0
   if (!memcmp(&local_uuid, &WSREP_UUID_UNDEFINED, sizeof(wsrep_uuid_t)) &&
       local_seqno == -2)
   {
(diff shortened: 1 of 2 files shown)


sql/wsrep_mysqld.cc
```diff
@@ -688,6 +688,7 @@ void wsrep_deinit()
 
 void wsrep_recover()
 {
+  #if 0
```
kbauskar, line 691: Seems like you cooked up this patch over and above the existing branch. Each JIRA issue should have its own separate branch created from trunk. Check the guideline page on wiki.

kbauskar:
1. Commit message missing. It should be descriptive to help understand reader without looking at the fix what is being done.
2. TCs is missing. Note: Any bug or issues found has to have a TCs (unless reproducing it is close to impossible or need huge infrastructure)
3. Also, don't limit your view to existing the TCs or issue pointed. Check what other things (other variables) can go wrong if cluster-address is set to invalid.
4. Please revise error message as to make it simple to understand what the issue is. Check sample PXC#507 error message update that I did.

PR #96, opened 2016-03-07
- PXC#508: LDI fails to load the data in multi-partitioned table.

kbauskar:
Issue:

---
- PXC/Galera logic re-uses the existing MySQL ALTER table hack that commit
  after every 10K rows. This hack was introduce to reduce the recovery time
  if server crashes while doing ALTER TABLE.
- MySQL hack is meant explicitly for ALTER TABLE and so it has alter-algorithm
  specific logic where-in there are 2 tables: src table (table that is being
  altered) and dest table (new table that is being populated that
  will replace original table once the operation is complete).
- "write_row" flow always have a reference to destination table but doesn't
  have a reference to source table. Hack traces source table by looking at the
  table-locks held by transaction as ALTER table transaction should only have
  references to these 2 tables. (Source table reference is needed so that it
  can be explicltly locked after commit as commit will release all the locks
  on the table. Destination table will be locked anyways as part of
  immediate insert).
- Imagine applying same theory for normal table that is partitioned.
  
  Let's say table is partitioned in 3 parts: p0, p1, p2
  (With existing partitioned infrastructure 3 different innodb-handlers and
   tables are created so from transaction perspective it is like user is
   trying to insert into 3 different tables.)
  
  Let's say we try to insert 60K data such that
  first 20K: p0
  second 20K: p1
  third 20K: p2
  - Insert to p0 .... trx holds lock on p0
  - intemediate commit (while inserting 10001st row)
    lock p0 released and reacquired.
  - Insert to p0 + p1 .... trx holds on p0 + p1
  - intermediate commit (while inserting 20001st row)
    (logic will flag p0: src table and p1: destination table and post commit
    lock will be held on src table and eventually insert will take it on
    destionation table.)
  - Insert continue to p0 + p1 (40K inserted)
  - Insert continue to p2 (50K inserted)
  - Insert of 50001 will invoke intermediate commit hook.
    
    At this point in time transaction will have reference to p0 + p1 + p2
    (logic will flag p0: src table and p2: destination as we are trying to
    insert to p2 table then it finds p1 and get confused which it flags
    as error and flow will skip commit.)
  
  This means transaction will grow in size even though user expect it
  to get intermediate committed.
  ## Solution:
  - PXC/GALERA use-case is damn simple and there is no concept of tracing
    source table that is causing this issue.
    There are also other checks that are not at-all needed for LDI use-case.
  - LDI case has been segregated to a different loop that is simpler
    and meant to check only LDI specific condition.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -7739,20 +7739,8 @@ ha_innobase::write_row(
 	if ((sql_command == SQLCOM_ALTER_TABLE
 	     || sql_command == SQLCOM_OPTIMIZE
 	     || sql_command == SQLCOM_CREATE_INDEX
-#ifdef WITH_WSREP
-	     || (wsrep_on(user_thd) && wsrep_load_data_splitting &&
-		 sql_command == SQLCOM_LOAD                      &&
-		 !thd_test_options(
-			user_thd, OPTION_NOT_AUTOCOMMIT | OPTION_BEGIN))
-#endif /* WITH_WSREP */
 	     || sql_command == SQLCOM_DROP_INDEX)
 	    && num_write_row >= 10000) {
-#ifdef WITH_WSREP
-		if (wsrep_on(user_thd) && sql_command == SQLCOM_LOAD) {
-			WSREP_DEBUG("forced trx split for LOAD: %s", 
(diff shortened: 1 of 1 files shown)


kbauskar:
LGTM.

Jenkins mtr: http://jenkins.percona.com/job/pxc56.test.mtr/183/

PR #97, opened 2016-03-11
Fix for centos7 systemd script for multiple same options occuring (#1490897)

tplavcic:
BUG:
mysql-systemd start-post never returns when pid_file specified in multiple sections of the my.cnf
https://bugs.launchpad.net/percona-xtradb-cluster/+bug/1490897

Small change just to add a "tail -n 1" so that only the last option is used. We already have this in some other places.

### DIFF
--- a/build-ps/rpm/mysql-systemd
+++ b/build-ps/rpm/mysql-systemd
@@ -9,9 +9,9 @@ parse_cnf()
     local def=$2
     shift
     local groups=$*
-    reval=$(my_print_defaults $groups | awk -F= '{if ($1 ~ /_/) { gsub(/_/,"-",$1); print $1"="$2 } else { print $0 }}' | grep -- "--$var=" | cut -d= -f2-)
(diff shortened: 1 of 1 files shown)


abychko:
tail -1 to have the value from conf.d/\* files?

tplavcic:
This specific my_print_defaults returns multiple values if you have them in the config (for example if you have multiple pid-file or socket in the config). So it's piped into "tail -n 1" there to select just the last one since the script can work with only one value.

abychko:
yes, I understand the logic. I asked about order - what if different pid files are specified in different configs - left default in main my.cnf and added new file also with pid file. which one will be picked? (it's all about tail command)

tplavcic:
MySQL uses config files in this order below, so my_print_defaults prints them in that order and that tail option will use the last one from those and that should be correct since "top files are read first, later files take precedence".
/etc/my.cnf Global options
/etc/mysql/my.cnf   Global options
SYSCONFDIR/my.cnf   Global options
$MYSQL_HOME/my.cnf  Server-specific options
defaults-extra-file The file specified with --defaults-extra-file=file_name, if any
~/.my.cnf   User-specific options
~/.mylogin.cnf  Login path options

abychko:
looks OK, thanks

PR #98, opened 2016-03-16
    PXC-507 - Trying to create slave thread end up in assert if

kennt-percona:
PXC#507: Trying to create slave thread end up in assert if cluster address is INVALID.
## Issue:

Trying to create extra applier thread(s) with cluster address set to INVALID (wsrep_connected = false) result in an assert. While the scenario is valid as it doesn't make sense to create applier thread(s) with wsrep disconnected asserting and getting server down is bit too strict.
## Solution:

Raise a warning (vs error before) in such scenario and avoid creation of thread(s).
Expanded test-case to cover this test-scenario.

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_cluster_address.result
+++ b/mysql-test/suite/galera/r/galera_var_cluster_address.result
@@ -49,6 +49,20 @@ VARIABLE_VALUE = 'Primary'
 SELECT VARIABLE_VALUE = 2 FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME = 'wsrep_cluster_size';
 VARIABLE_VALUE = 2
 1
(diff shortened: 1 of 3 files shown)


kbauskar:
Patch looks good. Some quick comments on commit-message.

Commit message format:

---

PXC#507: Trying to create slave thread end up in assert if cluster address is INVALID.

## Issue:

Trying to create extra applier thread(s) with cluster address set to INVALID (wsrep_connected = false) result in an assert. While the scenario is valid as it doesn't make sense to create applier thread(s) with wsrep disconnected asserting and getting server down is bit too strict.

## Solution:

Raise a warning (vs error before) in such scenario and avoid creation of thread(s).
Expanded test-case to cover this test-scenario.

kennt-percona:
build: http://jenkins.percona.com/job/pxc56.clone/1880/
mtr: http://jenkins.percona.com/view/XtraDB-Cluster/job/pxc56.test.mtr/189/

PR #99, opened 2016-03-17
PXC-500 : Possibly dead code in wsrep_recover()

kennt-percona:
Issue:
Some leftover special case code in the wsrep-recover code path. This
code is generally unreachable.  It requires a sequence number of -2, but
sequence numbers start at -1 and are incremented (as well as a all-zero
UUID). So the only way to reach this code is to explicitly set the
wsrep_start_position to this magic value.

Solution:
Removed special case code in wsrep_recover().
Add a test case to check for special case value


kennt-percona:
Build: http://jenkins.percona.com/job/pxc56.build/565/
MTR: http://jenkins.percona.com/job/pxc56.test.mtr/192/

kbauskar:
LGTM

PR #100, opened 2016-03-22
Fix for PXC-518 : galera_admin test case fails occasionally

kennt-percona:
Issue:
These tests were unreliable and would fail occasionally.
We were not checking to see that the tables replicated before
starting the tests. Depending on the timing, we would see wsrep_last_committed
incremented by 1 or 2.

Solution:
Before starting the test proper, check that all tables have been replicated.


kennt-percona:
http://jenkins.percona.com/job/pxc56.build/568/
http://jenkins.percona.com/view/XtraDB-Cluster/job/pxc56.test.mtr/195/

kbauskar:
LGTM

PR #101, opened 2016-03-23
PXC-519 : galera.galera_flush and galera.galera_flush_local test case…

kennt-percona:
…s use incorrect wait_condition

Issue:
The tests are not using wait_condition.inc properly. The tests they use do not change
and do not test for the watch variable. This may explain why this fails occasionally.
Also, some of the tests were not waiting for the test tables to be initialized/replicated.

Solution:
Change the wait to the proper test condition.  Also have the tests wait for the
test tables to be initialized/replicated.  Failure to do this will cause the counts
to be off (will cause the test to fail occasionally).


kennt-percona:
http://jenkins.percona.com/job/pxc56.build/569/
http://jenkins.percona.com/job/pxc56.test.mtr/196/

kbauskar:
LGTM

PR #102, opened 2016-03-24
PXC-523 : test case galera.mysql-wsrep#216 often fails on Jenkins

kennt-percona:
Issue:
There is a timing issue with how this test runs. The test looks at
the node_2 log, but due to timing that log may not have been created
yet.

Solution:
Modify the test so that we can check to see if node_2 has received
the data before checking its log.


kennt-percona:
http://jenkins.percona.com/job/pxc56.build/571/
http://jenkins.percona.com/job/pxc56.test.mtr/198/

Fixes the issue, haven't seen this test in any of the runs.

kbauskar:
LGTM

PR #103, opened 2016-03-24
Fix PXC-524 : test case galera.galera_var_cluster_address should wait…

kennt-percona:
… for cluster reconnect

Issue:
The test can sometimes fail on Jenkins.  The test will wait a second to reconnect
to the cluster, however this may not be enough time, especially on Jenkins.
This causes the test to occasionally fail on Jenkins while succeeding when
run locally.

Solution:
Change the sleep to an include/wait_until_connected_again.inc


kennt-percona:
http://jenkins.percona.com/job/pxc56.build/572/
http://jenkins.percona.com/job/pxc56.test.mtr/199/

Test case only changes.

kbauskar:
LGTM

PR #104, opened 2016-03-25
PXC-525 : galera_as_slave_nonprim fails due to wrong size in error me…

kennt-percona:
…ssage

Issue:
Occasional test failure, this error messge is sent because another node
has shutdown (expected shutdown).  Due to the timing, we may get this
error message, thus the occasional test failure.

Solution:
Correct the expected error message. The parameter that changed is
the size of the message in bytes.

### DIFF
--- a/mysql-test/suite/galera/r/galera_as_slave_nonprim.result
+++ b/mysql-test/suite/galera/r/galera_as_slave_nonprim.result
@@ -14,7 +14,7 @@ STOP SLAVE;
 RESET SLAVE ALL;
 CALL mtr.add_suppression("Slave SQL: Error 'Unknown command' on query");
 CALL mtr.add_suppression("Slave: Unknown command Error_code: 1047");
-CALL mtr.add_suppression("Send action {\\(nil\\), 328, TORDERED} returned -107 \\(Transport endpoint is not connected\\)");
+CALL mtr.add_suppression("Send action {\\(nil\\), 323, TORDERED} returned -107 \\(Transport endpoint is not connected\\)");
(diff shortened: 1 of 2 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc56.build/573/
http://jenkins.percona.com/job/pxc56.test.mtr/200/

mysql-test/suite/galera/r/galera_as_slave_nonprim.result
```diff
@@ -14,7 +14,7 @@ STOP SLAVE;
 RESET SLAVE ALL;
 CALL mtr.add_suppression("Slave SQL: Error 'Unknown command' on query");
 CALL mtr.add_suppression("Slave: Unknown command Error_code: 1047");
-CALL mtr.add_suppression("Send action {\\(nil\\), 328, TORDERED} returned -107 \\(Transport endpoint is not connected\\)");
+CALL mtr.add_suppression("Send action {\\(nil\\), 323, TORDERED} returned -107 \\(Transport endpoint is not connected\\)");
```
kbauskar, line 17: Why was the size 328 before and 323 now ?

kennt-percona:
This change is due to commit f3e541a (PXC-474).

That change in sql/sql_parse.cc meant that the thread's auto_increment_increment/offset were no longer
being set for that slave thread. This then caused the Query_log_event::write() to no longer write out the auto_increment_increment/offset (which translates to 5 less bytes for every Query_log_event being generated by the slave thread).

Depending on the timing, this message send would either succeed (no error message generated) or fail (error message generated). This explains the occasional failures.

kbauskar:
LGTM

PR #105, opened 2016-03-29
PXC-522 : wsrep_last_committed race conditions in several test cases

kennt-percona:
Issue:
These test cases were unreliable (occasionally failing) due to timing
issues when retreiving the value of wsrep_last_committed.

Solution:
Make sure the tests wait (if possible) for changes to be replicated
to the nodes before getting the value of wsrep_last_committed.
In some cases, this is not possible, since there is no change in value,
in those cases a sleep was added.


kennt-percona:
http://jenkins.percona.com/job/pxc56.build.mini/
http://jenkins.percona.com/view/XtraDB-Cluster/job/pxc56.test.mtr.mini/2/

kbauskar:
LGTM

PR #106, opened 2016-03-30
- #PXC-502: Silent abort (crash) at gcs/src/gcs_core.cpp:1152

sysprg:
The new node cannot join the cluster when the user reserves a very large amount of memory for the work buffers, for example, if she/he reserves 18GB for InnoDB buffer pool. In today's world, this is no surprise, but it is still a very large buffer. Because of this the user later receives "insufficient memory" (ENOMEM) error when calling fork() - when another node is trying to join the cluster and has requested the SST from this node. We do not need a lot of memory for SST, but the fork() function failed because the operating system tends to reserve in advance the such amount of memory that may be changed in future - after the fork() call (even if it uses copy-on-write semantics for memory operations).

The actual memory copying does not happen, but the operating system allocates memory pages in advance - because it fears that the whole memory of the process can be changed and all pages will be copied. It depends on the settings of the operating system and is especially important when swapping is off. Details here: https://www.kernel.org/doc/Documentation/vm/overcommit-accounting

Therefore on many configurations (especially when swapping is off) we cannot make a fork() if the amount of free memory is less than the amount of memory used by our process. And, in any case, when we run xtrabackup or mysqldump during the SST, a fork() function does a huge amount of unnecessary work related to bifurcation of the address space, which in a few microseconds is eliminated with the help of execvpe().

We can solve this issue (and win the performance) if we use posix_spawn[p] function instead of fork(). In addition, it will significantly increase the performance when we run SST in PXC on a heavily loaded system.

This patch replaces the creation of a new process (eg, SST) via fork() to creation of exactly the same process through posix_spawnp(), on all the systems where we have this feature.

Jenkins build: http://jenkins.percona.com/job/pxc56.clone/1896/

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -5302,12 +5302,18 @@ a file name for --log-bin-index option", opt_binlog_index_name);
     else // full wsrep initialization
     {
       // add basedir/bin to PATH to resolve wsrep script names
-      char* const tmp_path((char*)alloca(strlen(mysql_home) +
-                                           strlen("/bin") + 1));
+      int mysql_home_len= strlen(mysql_home);
+      char* const tmp_path((char*)alloca(mysql_home_len +
+                                         strlen("/bin") + 1));
       if (tmp_path)
       {
         strcpy(tmp_path, mysql_home);
-        strcat(tmp_path, "/bin");
+        /*
+          mysql_home may already contain a trailing slash:
+        */
+        strcat(tmp_path, (mysql_home_len == 0 ||
+                          mysql_home[mysql_home_len - 1] != '/') ? "/bin" :
+                                                                    "bin");
         wsrep_prepend_PATH(tmp_path);
       }
       else
--- a/sql/wsrep_utils.cc
+++ b/sql/wsrep_utils.cc
@@ -38,6 +38,10 @@
 #include <sys/prctl.h>
 #endif
 
+#ifdef _POSIX_SPAWN
+#include <spawn.h>
+#endif
+
 #ifdef HAVE_GETIFADDRS
 #include <net/if.h>
 #include <ifaddrs.h>
@@ -209,7 +213,7 @@ process::process (const char* cmd, const char* type, char** env)
 
     if (NULL == env) { env = environ; } // default to global environment
 
-    int pipe_fds[2] = { -1, };
+    int pipe_fds[2];
     if (::pipe(pipe_fds))
     {
         err_ = errno;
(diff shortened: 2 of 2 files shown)


kbauskar:
LGTM

kbauskar:
Here is the mtr http://jenkins.percona.com/job/pxc56.test.mtr.mini/5/

AndreasVerhoeven:
I've been working on identifying a similar issue related to fork() on FreeBSD, and doesn't this change revert the fixes done in https://code.launchpad.net/~akopytov/percona-xtradb-cluster/bug1382797-5.5 , especially the `prctl()` call? 

Shouldn't marking the buffer pool as `MADV_DONTFORK` work as well?

sysprg:
prctl (PR_SET_PDEATHSIG, SIGTERM) is not a universal solution, because:

1) This option do not inherited by child processes that can be run from the SST script, because PR_SET_PDEATHSIG flag is always reset at the fork. Here is a snippet of the Linux documentation (http://man7.org/linux/man-pages/man2/prctl.2.html):

“This value is cleared for the child of a fork(2) and (since Linux 2.4.36 / 2.6.23) when executing a set-user-ID or set-group-ID binary, or a binary that has associated capabilities”

2) This option do not inherited by child processes that have the set-user-ID or set-group-ID flag set. See the quote from Lunix documentation (shown above).

3) Scheme that based on the pctrl() is vulnerable to a race condition, when the parent process exited or had been killed before the child process will invoke the prctl() (just after the fork() was called).

4) It works only in the Linux (is not supported in the FreeBSD, Solaris or other operating systems).

5) Although currently it is impossible, but in the future, if we will launch new process from one thread but will wait for its completion in separate thread, then pctrl() will lead to premature sending of the SIGTERM signal. Because as stated in the Linux documentation:

“Warning: the "parent" in this case is considered to be the thread that created this process.  In other words, the signal will be sent when that thread terminates (via, for example, pthread_exit(3)), rather than after all of the threads in the parent process terminate.”

sysprg:
I can modify the patch to implement the behavior similar to prctl() by registering at-exit handler and sending SIGTERM signal to the process group, which launched as a result of wsp::process() call. This solution may to overcome a number of limitations of the original prctl() call. However, this approach leaves one problem - it does not guarantee the completion of all child processes with the kill -9, because the program cannot intercept SIGKILL and in this case at-exit handlers are not called.

sysprg:
Regarding madvise usage, we have two different problems:

1) In fact, the previous approach with madvise(..., MADV_DONOTFORK) does not work (as designed) because of an error in the Linux kernel, in the implementation of the fork function.

2) It is problematic to extend this approach to all buffers in the system, because it will require intervention to other storage engines. In addition, for example, Galera written in C++ and some of its parts, such as certification, use the built-in and standard types from the standard C++ library, which do not use mmap() for memory allocation, but use the standard heap. The madvise() call is inapplicable to the objects from default heap, while the certification structures can take a large amount of memory.

AndreasVerhoeven:
I'm not sure what's good to do here, just wanted to point out that this reverted the previous patches (which has been causing trouble on FreeBSD as well, due to the large memory shadowing on fork)

PR #111, opened 2016-04-04
PXC-509 : wsrep_node_name not defaulting to computer's node name

kennt-percona:
Issue: The system does not pick up the computer's node to use as
a default for WSREP_NODE_NAME (which it should according to the
documentation).

Solution:
The system is already trying to get the local node name, however
the value is getting overwritten when initialization the variables.
Change the settings for wsrep_node_name so that the default value
is not overwritten.

### DIFF
--- a/mysql-test/include/galera_actual_node_name.inc
+++ b/mysql-test/include/galera_actual_node_name.inc
@@ -1,0 +1,16 @@
+#
+# Gets the node name for the server
+# This will return the variable in $galera_actual_node_name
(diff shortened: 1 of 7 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc56.build.mini/10/
http://jenkins.percona.com/job/pxc56.test.mtr.mini/12/

kbauskar:
LGTM.

PR #112, opened 2016-04-07
- #PXC-541: WSREP_SST_AUTH_ENV variable is not transmitted into SST scripts after PXC-502

sysprg:
The code in sst_append_auth_env() and in the sst_prepare_other() functions violates the principle of the "black box". These functions know how the implementation of wsp::process() works and they transmit parameters for this implementation implicitly, through the environment of the parent process. But they continue to pass the cloned environment (without WSREP_SST_AUTH_ENV variable) to wsp::process() via explicit parameter – relying on the fact that the implementation of this function will ignore this parameter and the cloned environment will not be transferred to the child process. But they do so only for very old versions of Linux that do not have the execvpe() call (glibc prior to 2.11). Therefore, the PXC-502 patch works well on all systems except very old versions of Linux - we found this bug only on CentOS 5.

To correct this flaw, I removed implicit parameter passing from the wsrep_sst.cс code, and added emulation of execvpe() (made on the basis of the implementation of execvpe in glibc) to the code, which does not use posix_spawn().

I also made a small improvement in the treatment of errors in the new implementation of the wsp::process() function, which allow to avoid a small memory leak if SST failed. It is not essential for the current implementation, because SST failure stops node working, but it makes the code more clean.

Jenkins build: http://jenkins.percona.com/job/pxc56.test.mtr.mini/17/

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -466,18 +466,10 @@ static void* sst_joiner_thread (void* a)
         // Print error message if SST is not cancelled:
         if (!sst_cancelled)
         {
-          if (tmp)
-          {
-             WSREP_ERROR("Failed to read '%s <addr>' from: %s\n\tRead: '%s'",
-                         magic, arg->cmd, tmp);
-          }
-          else
-          {
-            // Null-pointer is not valid argument for %s formatting (even
-            // though it is supported by many compilers):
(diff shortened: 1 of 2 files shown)


kbauskar:
Code change stands approved. We just need a jenkins run updates. Are the failure genuine.

kbauskar:
LGTM. http://jenkins.percona.com/job/pxc56.test.mtr.mini/17/

PR #113, opened 2016-04-13
- #PXC-516: Unknown option causes cluster node to fail with assert in debug mode

sysprg:
Launching mysqld server with an unknown option triggers the assertion in the process of its termination. While it should shutdown with clean error message.

This is because analysis of the server options and transfer of the state (SST) are run simultaneously, but mysqld does not provide any way for emergency completion of the SST script (which running in separate process). When the server finds an unknown option, it calls unireg_abort() function that can detect incomplete thread and can trigger assertion in debug mode.

First, we have incomplete thread, which created in order to interact with the SST process. Second, until the SST process yet not terminated, we have an incomplete thread, which waits on the "sst_cond_" conditional variable in the Galera code. However, this variable will be set only at the end of SST process, in the sst_received() callback function. Therefore, if the SST is continued, this thread also freezes.

To solve these problems, I added code that emergency completes the SST process by sending a SIGTERM signal to it when unireg_abort() was called. This leads to completion of the thread, which interacts with the SST process. Also, sst_received() callback function will be called after termination of the SST process, and it unlocks the thread, which waiting on the "sst_cond_" conditional variable (in the Galera code).

Also, I corrected a race condition in the Galera code - calling of the sst_received() callback immediately after sending SST request should not be terminated with an error, even if we have not yet shifted to the S_JOINING state.

Related changes in the Galera code we can see here: https://github.com/percona/galera/pull/48


kbauskar:
Jenkins job: http://jenkins.percona.com/job/pxc56.test.mtr.mini/16/
LGTM

PR #114, opened 2016-04-21
- #PXC-546:  Seqno in the grastate.dat file is reset to "-1" and remains undefined until completion of the SST or IST

sysprg:
Seqno in the grastate.dat file is reset to "-1" (although the state is not marked as "unsafe") each time when the server is started (this done in Galera initialization, see the galera::ReplicatorSMM::update_state_uuid() function) and remains undefined until completion of the SST or IST.

To correct this error is sufficient to pass the current value of seqno into st_.set(...) function that is called from the galera::ReplicatorSMM::update_state_uuid(). Currently it called with the WSREP_SEQNO_UNDEFINED argument that causes this bug.

We also need to add a new logic into the recv_IST() function, to keep the state marked as having an undefined position after the start of the data changes during IST. To do this, if the current position is defined (for example, when there were no SST before IST), then we need to change it to an undefined position before applying the first transaction, since during the application of transactions (or after the IST) server may fail.

Related pull request for Galera is located here: https://github.com/percona/galera/pull/49

Jenkins build: http://jenkins.percona.com/job/pxc56.test.mtr.mini/20/


kbauskar:
Can we add more test-scenarios.
For example: scenario that revealed this issue. 3 node cluster all are in sync. Node-2 is shutdown and restarted with wrong option. node-2 should fail to start (of-course till PXC#516 is done it would be raise an warning) but then after the correction it should take start without SST.
This will take care of the case where-in node directly hunt for IST w/o SST.
Can we also test-scenario where-in SST len = 0 that will set seqno=UNDEFINED and then IST failure will result in complete SST on restart.

kbauskar:
LGTM
http://jenkins.percona.com/job/pxc56.test.mtr.mini/23/

PR #116, opened 2016-04-28
- PXC#547: Cannot set wsrep_desync to OFF dynamically when set in my.cnf

kbauskar:
Issue:

---
- wsrep_desync is used to indicate that node will not participate in
  flow control decision of the cluster.
- Setting it to ON will disable node participation in flow control.
  It can be toggled back to OFF.
  
  This action of desync can be done manually or even internally
  as part of RSU or SST.
- It being a configuration variable what if user try to set this
  as part of initial configuration (in my.cnf).
  It suggest user is requesting node to be desync even before it is
  synced with cluster which logically doesn't make any sense.
  
  With that background all such use-case should be blocked.
  ## Solution:
- Emit error if wsrep_desync is set during server initialization.
  User is free to desync a node after it has been successfully
  synced with the cluster.
  
  Node doesn't participate in flow control during initialization
  when it act as JOINER to recieve the state from DONOR.


kbauskar:
Jenkins job. http://jenkins.percona.com/job/pxc56.test.mtr.mini/21/

PR #118, opened 2016-05-07
Fix PXC-552 : mysql-wsrep#198 test case timing issue, fails on slower…

kennt-percona:
… nodes

Issue:
This test case will fail occasionally because the "CREATE TABLE t2" operation
will not have propagated to node 2 before the "LOCK TABLE t2".

Solution:
Have node 2 wait for the create table to propagate before continuing
with the test.


kennt-percona:
TC only change
http://jenkins.percona.com/job/pxc56.test.mtr.mini/24/
http://jenkins.percona.com/job/pxc56.build.mini/25/

kbauskar:
LGTM

PR #119, opened 2016-05-08
Fix PXC-556 : Restarting a cluster with innodb_page_size=4096 segfaults

kennt-percona:
Issue:
Using an innodb_page_size=4096 causes an assert, requires a start-stop-start
cycle.

Solution:
Galera takes a part of the page, if the size is too small then this will
start to overlap with some of the innodb structures.  For now, require
that we only use the 16K page size.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/204/
http://jenkins.percona.com/job/pxc56.clone/1918/

kbauskar:
LGTM. Also, we need to keep track of upstream issue.

PR #125, opened 2016-05-23
Fix dependencies due to galera naming changes

EvgeniyPatlan:


abychko:
duplicate galera dependencies? for -full and --server?

PR #134, opened 2016-05-31
- PXC#588: Avoid changing wsrep_provider when node is paused and/or d…

kbauskar:
…esycned.
- If node is paused then even though node may have recieved all the write-set
  it hasn't applied them. Switching off wsrep_provider during this time
  is not advisable to maintain data consistent state.
  Desync is another such operation which may have paused cluster.
  So it is not advisable to change wsrep_provider during this time too.


kbauskar:
LGTM

PR #136, opened 2016-06-01
- PXC#570: Asynchronous TOI action on non-primary node can cause node…

kbauskar:
… to crash

  Issue:

---

  CREATE event will be auto-removed on completion of tenure (tenure is defined
  by event end-time). This auto-removal is DDL action of DROP event that is
  replication in galera as expected.

  If event is active and node becomes non-primary then such auto-removal action
  are still triggered which eventually causes TOI error.

  Solution:

---
- Ensure that TOI check for node readiness to process DDL+DML before
  kickstarting the execution. Generally this is done by parent level
  for user-triggered command but for such auto-triggered command a
  local hook at TOI level is needed.


kbauskar:
LGTM

PR #138, opened 2016-06-01
Fixes for debian and centos according to BLD-436 and BLD-437

EvgeniyPatlan:


tplavcic:
OK, let's go with this for the next beta build

PR #140, opened 2016-06-02
- PXC#592: Tried closing fk-reference-table that was never opened.

kbauskar:
Function "wsrep_row_upd_check_foreign_constraints" tried to mark
  fk-reference-table opened without ensuring it table is really opened.


kbauskar:
LGTM

PR #142, opened 2016-06-02
Fix for centos7 systemd

tplavcic:
Running with only mysqld in systemd service was faulty if we used IP addresses to initialize new cluster.
So I have returned the mysqld_safe here and we will use it for now until better solution.
This is a change only for centos7.

I have tested and successfully created a two node cluster with xtrabackup sst:

```
| wsrep_evs_state                               | OPERATIONAL
| wsrep_cluster_size                            | 2
| wsrep_cluster_state_uuid                      | de8f79d8-28b8-11e6-a3e0-6f5158deb7b9
| wsrep_cluster_status                          | Primary
| wsrep_connected                               | ON
| wsrep_local_bf_aborts                         | 0
| wsrep_local_index                             | 0
| wsrep_provider_name                           | Galera
| wsrep_provider_vendor                         | Codership Oy <info@codership.com>
| wsrep_provider_version                        | 3.14.2(r53b88eb)
| wsrep_ready                                   | ON
```

### DIFF
--- a/build-ps/percona-xtradb-cluster.spec
+++ b/build-ps/percona-xtradb-cluster.spec
@@ -594,7 +594,7 @@ mkdir debug
            -DWITH_EMBEDDED_SERVER=OFF \
            -DWITH_INNODB_MEMCACHED=ON \
 %if 0%{?systemd}
-           -DWITH_SYSTEMD=ON \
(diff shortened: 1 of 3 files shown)


EvgeniyPatlan:
LGTM

PR #144, opened 2016-06-07
Second node can't join the cluster on CentOS (BLD-444)

EvgeniyPatlan:


tplavcic:
I think this part is needed `[ $ret -ne 0 ] && exit $ret`
because it checks the return status of chown line before so if you agree please don't remove it

EvgeniyPatlan:
No. this $ret is connected with the action^

action $"Installing validate password plugin: " /usr/sbin/mysqld --datadir="$datadir" --user=mysql --init-file="$initfile"
ret=$?

So as you can see we assign exit status value of acion command to $ret

tplavcic:
True.
LGTM

PR #151, opened 2016-06-10
Make changes for build of pxc-57

EvgeniyPatlan:
This commit contains ported changes from pxc-57-beta and ported changes from ps-57


tplavcic:
Since these are basically changes that were done for beta release I don't see anything wrong with it.

PR #152, opened 2016-06-14
Beta rel 5.7.12 25.16 packaging

EvgeniyPatlan:


kbauskar:
LGTM. These are basically PXC-5.7.11 changes that are now being merged to PXC-5.7.12 as we don't plan to fold PXC-5.7.11 (due to galera naming conflict dependency)

PR #153, opened 2016-06-14
PXC-333 : Invalid my_error() usage leading to incorrect/misleading er…

kennt-percona:
…ror messages

Issue: Incorrect usage of my_error()/my_message() results in some error
messages not appearing correctly (may not show the desired string or in
the wrong context).

Solution:
Correct the my_error()/my_message() usage to display more WSREP-specific
error messages.

### DIFF
--- a/mysql-test/suite/galera/r/galera_bf_abort.result
+++ b/mysql-test/suite/galera/r/galera_bf_abort.result
@@ -4,7 +4,7 @@ START TRANSACTION;
 INSERT INTO t1 VALUES (1);
 INSERT INTO t1 VALUES (1);
 INSERT INTO t1 VALUES (2);
-ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+ERROR 40001: WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction
 wsrep_local_aborts_increment
 1
 DROP TABLE t1;
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1437,7 +1437,7 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       else
       {
         mysql_reset_thd_for_next_command(thd);
-        my_error(ER_LOCK_DEADLOCK, MYF(0), "wsrep aborted transaction");
+        my_message(ER_LOCK_DEADLOCK, "WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction", MYF(0));
         WSREP_DEBUG("Deadlock error for: %s", WSREP_QUERY(thd));
         mysql_mutex_unlock(&thd->LOCK_wsrep_thd);
         thd->killed               = THD::NOT_KILLED;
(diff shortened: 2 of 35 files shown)


mysql-test/suite/galera/r/galera_bf_abort.result
```diff
@@ -4,7 +4,7 @@ START TRANSACTION;
 INSERT INTO t1 VALUES (1);
 INSERT INTO t1 VALUES (1);
 INSERT INTO t1 VALUES (2);
-ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+ERROR 40001: WSREP cancelled the transaction
```
kbauskar: We may need to capture the reason.

So something like  "WSREP detected deadlock/conflict. Transaction aborted. Trying restarting"
"dealock/conflict" -> reason that can be replaced if we have specific reason know while raising the error.

sql/sql_parse.cc
```diff
@@ -1437,7 +1437,7 @@ bool dispatch_command(THD *thd, const COM_DATA *com_data,
       else
       {
         mysql_reset_thd_for_next_command(thd);
-        my_error(ER_LOCK_DEADLOCK, MYF(0), "wsrep aborted transaction");
+        my_message(ER_LOCK_DEADLOCK, "WSREP cancelled the transaction", MYF(0));
```
kbauskar: I am wondering if we should be introducing a new error code to capture WSREP operation failure.
ER_WSREP_XXXX.

This will help easy classification of error that are raised by WSREP module.

sql/wsrep_mysqld.cc
```diff
@@ -1281,8 +1274,8 @@ static int wsrep_TOI_begin(THD *thd, const char *db_, const char *table_,
     WSREP_WARN("TO isolation failed for: %d, schema: %s, sql: %s. Check wsrep "
                "connection state and retry the query.",
                ret, (thd->db().length ? thd->db().str : "(null)"), WSREP_QUERY(thd));
-    my_error(ER_LOCK_DEADLOCK, MYF(0), "WSREP replication failed. Check "
-	     "your wsrep connection state and retry the query.");
+    my_message(ER_LOCK_DEADLOCK, "WSREP replication failed. Check "
```
kbauskar, line 1277: Is it safe to assume that all wsrep-xxxx error messages has been changed to use my_message (from my_error, etc..)

kennt-percona, line 1277: In some places, we are still using my_error(), since we are supplying a parameter to the error message, for example: ER_NOT_SUPPORTED_YET.

mysql-test/suite/galera/r/galera_bf_abort.result
kennt-percona: From what I can gather, this is possible in some places (may be able to detect BFA or LCF).

sql/sql_parse.cc
kennt-percona: We could do this, there are several types of errors that all use ER_LOCK_DEADLOCK, so it would be nice to split these off.  We would need to determine an appropriate range to avoid future conflicts (maybe start from 10000 or 20000).

mysql-test/suite/galera/r/galera_bf_abort.result
kbauskar: Well we can avoid to be specific (log should have the details) and keep it generic but with hint of what kind of failure it is about.

How about this message:

"WSREP detected deadlock or conflict. Transaction aborted. Trying restarting transaction"

sql/sql_parse.cc
kbauskar: Yeah. Right. I thought over this and looks like it would invite more mess than solving it. Not good time for us to solve it. Ignore this.

sql/wsrep_mysqld.cc
kbauskar, line 1277: Ok. Noted. Thanks.

PR #154, opened 2016-06-14
- wsrep_replication thread use to pre-set multi-statement transaction

kbauskar:
flag when ideally it should be set by the respective transaction
  statement like trans_begin (invoked by BEGIN statement).

  There is no specific reason quoted above why this was done but this
  is no more compatible with new semantics of MySQL-5.7 that avoids
  implicit commit with a transaction.
  begin
    begin .... this can cause implicit commit so should be avoided.
          .... when using gtid_mode = on.

  Unfortunately, with change above even the first begin use to get
  multi-statement transaction flag ON which is not true as transaction
  has not yet started.

  Solution: Avoid setting this flag. Let the followup flow handle it.


kbauskar:
LGTM http://jenkins.percona.com/job/pxc57.test.mtr/42/

PR #155, opened 2016-06-14
- #PXC-585: WSREP BF-BF conflicts causes server to crash.

sysprg:
MW-258 test sometimes fails (not at every startup). Currently we have only the error in the TC caused by the fact that before checking variable wsrep_desync_count we need to wait for "synced" callback, since the variable wsrep_desync_count shows the actual state of the desync counter, but synchronization takes some time.

Galera part of this patch is available here: https://github.com/percona/galera/pull/65


kbauskar:
This is not the real cause. Closing the PR.

PR #156, opened 2016-06-16
Beta rel 5.7.12 25.16 pxc strict mode

kbauskar:


kbauskar:
mtr test-case result: http://jenkins.percona.com/job/pxc57.test.mtr/44/

mysql-test/suite/galera/t/pxc_strict_mode.test
```diff
@@ -0,0 +1,1189 @@
...
+--mkdir $restartdir
+
+--connection node_1
+--echo #node-1
+--let $pxc_strict_mode_saved2 = `SELECT @@pxc_strict_mode`
+
+# to allow insertion in add_suppression table.
+set global pxc_strict_mode='DISABLED';
+
+call mtr.add_suppression("WSREP: Percona-XtraDB-Cluster doesn't recommends use of MyISAM table replication as it is an experimental feature");
```
kennt-percona: 'recommends' should be 'recommend'

mysql-test/suite/galera/t/pxc_strict_mode.test
```diff
@@ -0,0 +1,1189 @@
...
+# create test-bed
+#
+# let's cache the default value first.
+--connection node_1
+--echo #node-1
+--let $pxc_strict_mode_saved1 = `SELECT @@pxc_strict_mode`
+--let restartdir= $MYSQLTEST_VARDIR/tmp/restartdir
+--mkdir $restartdir
+
+--connection node_1
```
kennt-percona, line 20: I think that this should be for connection node_2 not node_1 (it gets restored on connection node_2).

mysql-test/suite/sys_vars/t/pxc_strict_mode_basic.test
```diff
@@ -0,0 +1,50 @@
+#
+# Test different setting option for pxc-strict-mode
+
+# let's cache the default value first.
+#--let $pxc_strict_mode_saved = `SELECT @@pxc_strict_mode`
+
+--echo # set default value
+--error ER_WRONG_VALUE_FOR_VAR
```
kennt-percona, line 8: I don't understand why changing the value of pxc_strict_mode fails for all cases (even for valid values).  Isn't pxc_strict_mode dynamic?

mysql-test/suite/galera/t/pxc_strict_mode.test
kbauskar: Not sure I understood this comment exactly as to which part it is referring too. BTW we no more need to set DISABLED mode as default for mtr is DISABLED.

kennt-percona, line 20: At the bottom of the test it restores pxc_strict_mode_saved2 for connection node_2, however it's saving the value of connection node_1 into pxc_strict_mode_saved2.

so, from line 21, shouldn't this be
--connection node_2
--echo #node-2
--let $pxc_strict_mode_saved2 = `SELECT @@pxc_strict_mode`

mysql-test/suite/sys_vars/t/pxc_strict_mode_basic.test
kbauskar, line 8: Right. This turned out bit funny.
We don't allow switching pxc-strict-mode if server is running as standalone.
(sys_vars run the server in non-wsrep mode).

Then why do we have this TCs and how do we check toggling of pxc-strict-mode.
(toggling is being check as part of pxc-strict-mode.test and we have 2 options with this tcs.
a. remove it  but that means adding exception as every variable should have test-case as per the existing enforced mysql semantics.
b. this tcs still help us make sure that value that is being passes is invalid and not garbage.
Garbage will produce SEGFAULT)

mysql-test/suite/galera/t/pxc_strict_mode.test
kbauskar: Done

mysql-test/suite/galera/t/pxc_strict_mode.test
kbauskar, line 20: Right. I got that and corrected in this new patch. Thanks for clarification.

kennt-percona:
LGTM

PR #157, opened 2016-06-17
- #PXC-585: WSREP BF-BF conflicts causes server to crash.

sysprg:
MW-258 test sometimes fails with MDL BF-BF confict (not at every startup).

The main reason for this problem is that we need to remove protection against repeated calls of the wsrep->pause() on the same node, since otherwise we lose protection against conflicts, which is provided by the local_monitor and commit_monitor in the Galera code. Previously, we used a similar defense (implemented via counter variable + mutex) against repeated calls to the wsrep->desync() and wsrep->pause() functions, but it is irrelevant to the new version of the Galera, because it have an internal desync counter. Therefore, we need to remove this extra protection.

Also MW-258 TC occasionally fails due to the fact that before final checking of the wsrep_desync_count variable we need to wait until node re-synched, because the wsrep_desync_count variable shows the actual state of the desync counter, but synchronization takes some time.

Galera part of this patch is available here: https://github.com/percona/galera/pull/66


mysql-test/suite/galera/r/MW-258.result
```diff
@@ -21,6 +22,8 @@ SHOW VARIABLES LIKE 'wsrep_desync';
 Variable_name	Value
 wsrep_desync	OFF
 UNLOCK TABLES;
+SET GLOBAL wsrep_provider_options = 'signal=wsrep_desync_left_local_monitor';
```
kbauskar: We should use script instead.
--source include/galera_clear_sync_point.inc
--source include/galera_signal_sync_point.inc

mysql-test/suite/galera/r/MW-258.result
```diff
@@ -9,6 +9,7 @@ Variable_name	Value
 wsrep_desync	OFF
 SET SESSION wsrep_sync_wait = 0;
 SET SESSION wsrep_osu_method = RSU;
+SET GLOBAL wsrep_provider_options = 'dbug=d,wsrep_desync_left_local_monitor';
```
kbauskar: We should use the script instead
--let $galera_sync_point = apply_monitor_slave_enter_sync
--source include/galera_set_sync_point.inc

mysql-test/suite/galera/t/MW-258.test
```diff
@@ -13,6 +15,7 @@ SHOW VARIABLES LIKE 'wsrep_desync';
 --connection node_1a
 SET SESSION wsrep_sync_wait = 0;
 SET SESSION wsrep_osu_method = RSU;
+SET GLOBAL wsrep_provider_options = 'dbug=d,wsrep_desync_left_local_monitor';
```
kbauskar: We should use the script instead
--let $galera_sync_point = apply_monitor_slave_enter_sync
--source include/galera_set_sync_point.inc

mysql-test/suite/galera/t/MW-258.test
```diff
@@ -28,12 +31,20 @@ SHOW STATUS LIKE 'wsrep_desync_count';
 SHOW VARIABLES LIKE 'wsrep_desync';
 UNLOCK TABLES;
 
+--connect node_1c, 127.0.0.1, root, , test, $NODE_MYPORT_1
+SET GLOBAL wsrep_provider_options = 'signal=wsrep_desync_left_local_monitor';
```
kbauskar: We should use script instead.
--source include/galera_clear_sync_point.inc
--source include/galera_signal_sync_point.inc

mysql-test/suite/galera/t/lp1376747-4.test
```diff
@@ -28,7 +28,7 @@ INSERT INTO t1 VALUES (2,3);
 SET session wsrep_sync_wait=0;
 SET session wsrep_causal_reads=OFF;
 SHOW CREATE TABLE t1;
-FLUSH TABLES t1 WITH READ LOCK;
+--send FLUSH TABLES t1 WITH READ LOCK;
```
kbauskar, line 31: Did you ran this test. I think it will generate a diff and result file will need re-recording. I don't see result for this test ?

sql/wsrep_var.cc
```diff
@@ -256,12 +256,12 @@ static int wsrep_provider_verify (const char* provider_str)
 @return false if no error encountered with check else return true. */
 bool wsrep_provider_check (sys_var *self, THD* thd, set_var* var)
 {
+  char wsrep_provider_buf[FN_REFLEN];
+
   if (WSREP(thd))
   {
-    mysql_mutex_lock(&LOCK_wsrep_pause_count);
-    bool node_paused = (wsrep_pause_count == 0) ? false : true;
+    bool node_paused = thd->global_read_lock.provider_resumed() ? false : true;
```
kbauskar: This will take care of pause if fired from same connection only.
(If this stmt is fired from another connection then it will not evaluate node_paused to true.)

mysql-test/suite/galera/t/lp1376747-4.test
sysprg, line 31: fixed

PR #158, opened 2016-06-20
- PXC#613: Toggling wsrep_provider leave behind stale/unreleased refe…

kbauskar:
…rence to

  wsrep_status

  show status result in fetching wsrep-status-variables by making a call
  to galera library function. This galera library function allocates
  a buffer, populate it with needed data and reference to it is passed
  that is then feeded in mysql-array to service the show status command.

  Who deallocates this buffer then ?
- buffer is deallocated only on next call to show status when it try to
  fetch latest values of wsrep-status-variables. (It could have re-used the
  same buffer but that. Not sure why it missed it.)
  
  If provider is unloaded, show status post unload will leave behind stale
  memory reference which turns into memory-leak when show status is fired with
  unloaded provider as unloaded provider can't fire wsrep-status api so it
  return dummy status array.
  
  Next time when provider is reloaded and show status is refired provider
  can call galera library wsrep-api but the caller passes dummy_stats
  reference to library (that is static allocation from stack).
  ## Fix:
- While toggling the wsrep_provider (this including unloading use case too)
  free the stats.
  
  [Note: THD::release_resources is not called on the client connecting
  running the wsrep_provider toggle command.]

### DIFF
--- a/mysql-test/suite/galera/r/galera_wsrep_provider_unset_set.result
+++ b/mysql-test/suite/galera/r/galera_wsrep_provider_unset_set.result
@@ -33,3 +33,18 @@ SELECT COUNT(*) = 4 FROM t1;
 COUNT(*) = 4
 1
 DROP TABLE t1;
+#node-2
+call mtr.add_suppression("WSREP: gcs_caused\\(\\) returned -103 \\(Software caused connection abort\\)");
+show status like 'wsrep_cluster_size';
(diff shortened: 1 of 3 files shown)


kbauskar:
http://jenkins.percona.com/job/pxc56.test.mtr/222/

http://jenkins.percona.com/job/pxc56.test.mtr/230/

PR #160, opened 2016-06-20
- pxc-strict-mode tried blocking DML/DDL on all Storage Engine including

kbauskar:
performance-schema. Performance-Schema DML and DDL should be allowed
  as that is part of normal workflow as documented by MySQL documentation.
  Performance_schema is more like special Storage Engine meant to monitor
  server and as expected it doesn't make sense to replicate updates
  to this schema.


kbauskar:
LGTM: http://jenkins.percona.com/job/pxc57.test.mtr/48/

PR #161, opened 2016-06-20
Fix PXC-612 : --wsrep_sst_method=xtrabackup-v2 fails if no datadir op…

kennt-percona:
…tion in cnf file

Issue:
When specifying the datadir via the command-line (and not in the cnf),
innobackupex (in the wsrep_sst_xtrabackup-v2.sh script) will fail,
because it does not accept a command-line datadir option.

Solution
The innobackupex used in xtrabackup 2.4 is a symlink to xtrabackup,
so we can rely on those versions to support the datadir option, on the
side.  So if we do not find the datadir in the config file, we add
it in via the command line (but only if the xtrabackup version >= 2.4.3)

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -72,6 +72,10 @@ xtmpdir=""
 scomp=""
 sdecomp=""
 
+# Use this to pass in the --datadir option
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/225/

kbauskar:
LGTM..patch changes looks good but all the tcs are failing ?

PR #163, opened 2016-06-22
PXC-612 : --wsrep_sst_method=xtrabackup-v2 fails if no datadir option…

kennt-percona:
… in cnf file

Issue:
Using only the command-line option fails during an SST with xtrabackup,
because innobackupex cannot accept command-line parameters for the
datadir.

Solution:
Force the use of a version of xtrabackup that accepts command-line
paramters. Thus we now require at least XB-2.4.3.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -607,6 +607,43 @@ send_donor()
 
 }
 
+# Returns the version string in a standardized format
+# Input "1.2.3" => echoes "010203"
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/227/

All of the nodes are failing in two tests:
galera.galera_defaults (missing wsrep_desync_count)
galera.galera_sst_xtrabackup_v2_options  (this is turned off later because it doesn't work with xb 2.4.2)

kbauskar:
LGTM

PR #164, opened 2016-06-22
PXC-612 : --wsrep_sst_method=xtrabackup-v2 fails if no datadir option in cnf file

kennt-percona:
(Null Merge for PXC 5.7, 5.6 only fix)

Issue:
Using only the command-line option fails during an SST with xtrabackup,
because innobackupex cannot accept command-line parameters for the
datadir.

Solution:
Force the use of a version of xtrabackup that accepts command-line
paramters. Thus we now require at least XB-2.4.3.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -607,6 +607,43 @@ send_donor()
 
 }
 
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/51/

kbauskar:
LGTM

PR #168, opened 2016-06-23
- #PXC-604: Setting wsrep_auto_increment_control to OFF doesn't restore the original user set value

sysprg:
If the user set values of auto_increment_increment and offset to x and y respectively, then setting wsrep_auto_increment_control to OFF should restore the original values.

Unfortunately, this works fine when SET operator used for setting the auto-increment options, but is not working, when auto-increment parameters passed via command line.

Even after setting auto-increment control to OFF, default values of 1 are being used. This is because of misplaced code snippet that caches these values to restore when wsrep_auto_increment_offset is set to OFF.

This fragment just need to be moved slightly lower - so that it executed after call to the get_options function (in the mysqld.cc file).

Jenkins build here: http://jenkins.percona.com/job/pxc56.build/606/

### DIFF
--- a/mysql-test/suite/galera/t/galera_auto_inc_off.test
+++ b/mysql-test/suite/galera/t/galera_auto_inc_off.test
@@ -1,0 +1,67 @@
+#
+# 1) Check that the values of auto-increment step and offset,
+#    which are set by the user, are saved and restored when user
+#    turning off the automatic control for these options.
+# 2) Check that these options are successfuly passed to new session.
+# 3) Check that the same logic works for the auto-increment parameters
+#    passed via command line.
+#
+--source include/galera_cluster.inc
+--source include/have_innodb.inc
+
+SET GLOBAL auto_increment_increment=10;
+SET GLOBAL auto_increment_offset=7;
+SET GLOBAL wsrep_auto_increment_control='OFF';
+
+SHOW GLOBAL VARIABLES LIKE '%auto_increment%';
+
+--let $galera_connection_name = node_1a
+--let $galera_server_number = 1
+--source include/galera_connect.inc
+
+--connection node_1a
+
+SHOW VARIABLES LIKE '%auto_increment%';
+
+SET GLOBAL wsrep_auto_increment_control='ON';
+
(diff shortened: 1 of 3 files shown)


mysql-test/suite/galera/t/galera_auto_inc_off.test
```diff
@@ -0,0 +1,51 @@
...
+--source include/galera_connect.inc
+
+--connection node_1a
+
+SHOW VARIABLES LIKE '%auto_increment%';
+
+--connection node_1
+
+SET GLOBAL wsrep_auto_increment_control='ON';
+
```
kbauskar, line 27: Show variables like %auto_increment% post turning it ON is missing.

sysprg, line 27: Let me explain my logic. In general, we do not know what the offset received by each of the nodes, and which step will be used when wsrep_auto_increment_control = 'ON'. We can put into the .result file some values, which given by the current algorithm, but we have no guarantee that they will be same tomorrow. We have occasionally saw TC, which failed on the Jenkins because the nodes "node_1" and "node_2" obtained auto_increment_offset = 2 (at the node #1) and auto_increment_offset = 1 (at the node #2), instead of auto_increment_offset = 1 (at the node #1) and auto_increment_offset = 2 (at the node #2) (this order is recorded in .result file). But if you think it is right to enter values relying on the current version of algorithm and the current order of the start of nodes (node #1 always first), then I can add it.

kbauskar, line 27: that swap can happen only if node-2 is started post node-1 and I am not sure how mtr can do that. MTR need to ensure node-1 is started before node-2.

I believe emission of those value is important to ensure correctness of the functionality so may be we should add it. We can always revisit if things start failing.
(Before adding just run the test-case 100 times to ensure there is no existing inconsistency reported).

kbauskar:
Also please upload 5.7 patch too.
Make the change run mtr and upload both patches for final review.

mysql-test/suite/galera/t/galera_auto_inc_off.test
sysprg, line 27: Added tests for wsrep_auto_increment_control = ON using wsrep variables to check values of the auto increment parameters (for the current automatic control algorithm).

kbauskar:
LGTM

PR #171, opened 2016-06-28
- PXC#585: Parallel RSU operation can cause "MDL BF-BF confict"

kbauskar:
MW-258 test sometimes fails with MDL BF-BF confict (not at every startup).

  The main reason for this problem is that we need to remove protection against
  repeated calls of the wsrep->pause() on the same node, since otherwise we lose
  protection against conflicts, which is provided by the local_monitor and
  commit_monitor in the Galera code. Previously, we used a similar defense
  (implemented via counter variable + mutex) against repeated calls to the
  wsrep->desync() and wsrep->pause() functions, but it is irrelevant to the new
  version of the Galera, because it have an internal desync counter.
  Therefore, we need to remove this extra protection.

  Also MW-258 TC occasionally fails due to the fact that before final checking
  of the wsrep_desync_count variable we need to wait until node re-synched,
  because the wsrep_desync_count variable shows the actual state of the
  desync counter, but synchronization takes some time.


kbauskar:
MTR run:
http://jenkins.percona.com/job/pxc56.test.mtr/238/
http://jenkins.percona.com/job/pxc56.test.mtr/239/

kbauskar:
LGTM

PR #172, opened 2016-06-29
PXC-623 : Fix galera_as_slave_autoinc fails occasionally (timing issue)

kennt-percona:
Issue:
The test expects a certain ordering of the auto_increment_offset, but due
to timing isses the order may sometimes be reversed. Which means that the
tests will occasionally fail.

Solution:
To make the tests more robust, verify that the nodes are being assigned
the values of [1,2] in an order-independent way.


kennt-percona:
http://jenkins.percona.com/job/pxc57.build/110/
http://jenkins.percona.com/job/pxc57.test.mtr/69/

PR #173, opened 2016-06-30
- #PXC-601: show status while updating group state can cause an assert (likely race)

sysprg:
The gcs_group_get_status function does not check that the group->my_idx can be equal to -1 (for example, when node receives self-leave message during non-primary state or technically due to implementation details of the gcs_group_handle_comp_msg function). Therefore, gcs_group_get_status function refers to memory beyond the array, which leads to SIGFAULT during execution of the to_string function.

To fix this error we should not read desync counter until the gcs_group_handle_comp_msg function returns the group->my_idx to definite state.

In addition, we should handle the situation where node is in the non-primary state and we got self-leave message. In this case node is desynchronized from the cluster, therefore I propose to return desync counter = 1.

Galera part of this patch is available here: https://github.com/percona/galera/pull/67

Jenkins job: http://jenkins.percona.com/job/pxc56.build/615/


kbauskar:
LGTM

PR #174, opened 2016-06-30
PXC-623 : [JIRA] (PXC-623) Fix galera_as_slave_autoinc fails occasion…

kennt-percona:
…ally (timing issue)

Issue:
The test assumes that the two nodes will receive the auto_increment_offset in a certain
order [1,2], however depending on the timing the values may appear as [2,1].

Solution:
Change the test assuming that each node can take on both values.


kennt-percona:
TC-only change

PR #175, opened 2016-07-01
- PXC#625: ROLLBACK to savepoint fail to register wsrep_ handler for

kbauskar:
cleanup/rollback

  Before a transaction is rollback in MySQL workflow if there it involves
  wsrep action then wsrep handler needs to be registered as one of the
  plugin to process the needed rollback action.
  This was done for normal rollback workflow but not for rollabck to
  savepoint workflow causing leaving behind the stale and open transaction
  in wsrep/galera sub-system.


kbauskar:
mtr testing: http://jenkins.percona.com/job/pxc56.test.mtr/246/

LGTM

PR #177, opened 2016-07-01
Fix PXC-624 : galera_bf_abort fails with "table t1 doesn't exist" occ…

kennt-percona:
…asionally

Issue:
This test case started to fail more often (in a recent run it failed on every
node).

Solution:
Need to ensure that the nodes form a cluster (not two independent nodes).
(1) Work with node 2 so that it forms up with node 1
(2) Check to ensure that we have a cluster of size 2 is formed


kbauskar:
LGTM

PR #179, opened 2016-07-01
- PXC#616: PXC is not starting with `--innodb_autoinc_lock_mode=1

kbauskar:
--pxc_strict_mode=DISABLED` combination.

  Check for innodb_autoinc_lock_mode was wrongly placed.
  innodb_autoinc_lock_mode is InnoDB specific variable and so the check
  should done as part of InnoDB startup validation routine.


kbauskar:
LGTM
http://jenkins.percona.com/job/pxc57.test.mtr/76/

PR #182, opened 2016-07-05
Fix PXC-571

kennt-percona:
Issue:
The use of THD_PROC_INFO is a little skewed as it doesn't always reflect what the thread
is doing at this moment (for debug purposes it records if it's perfomed an action).

Solution:
Change the code to reflect what the thread is currently doing.  Added some debug code
so that the tests will still continue to work.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/82/
http://jenkins.percona.com/job/pxc57.test.mtr/83/

Ran mtr 82, but there were some suspicious nodes (most of the nodes were ok or had 1-2 failures, but here 3-4 of the nodes had a lot of test failures >15, so I reran the tests on the failed nodes).

kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/84/
- added "wsrep: applier idle" stage
- added WSREP_DEBUG comments to print out trx seqno

kbauskar:
LGTM

PR #186, opened 2016-07-08
Fix PXC-564 : PXC changes to support the keyring option

kennt-percona:
Issue:
Original issue was to support re-encrypt because of the way 5.7.11 uses
keyring encryption, however 5.7.12 changes the way it uses the keyring
so re-encryption is no longer needed.  However, still need to add
support for sending the keyring over.

Solution:
Send the keyring over.  When using the keyring, have to ensure
that it is sent over an encrypted connection.
Also requires PXB 2.4.4


kbauskar:
LGTM

PR #190, opened 2016-07-08
- PXC#632: mysqldump sst doesn't work with pxc-strict-mode=ENFORCING.

kbauskar:
- ENFORCING mode prohibits direct update of MyISAM table using
  DML statement. SST method = mysqldump will try to dump and load
  the table as normal DML statement.
  
  With ENFORCING mode mysqldump method usage becomes restrictive.
  In order to address this issue pxc_strict_mode is set to
  DISABLED mode while running the load statements.
  It is fine to do so as mysqldump is used while during node
  bootup stage for recieving SST stream and statement used
  by MySQL are safe and fully supported from PXC perspective.


kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/87/

PR #191, opened 2016-07-11
- Fixed some test-cases to safe-guard against unpredictable values

kbauskar:
especially auto_increment_offset which depends on which node start
  first.


kbauskar:
LGTM

PR #192, opened 2016-07-11
- Fixed some test-cases to safe-guard against unpredictable values

kbauskar:
especially auto_increment_offset which depends on which node start
  first.

Conflicts:
    mysql-test/include/mtr_warnings.sql
    mysql-test/suite/galera/r/galera_wan_restart_ist.result
    mysql-test/suite/galera/t/galera_wan_restart_ist.test


kbauskar:
LGTM

PR #194, opened 2016-07-12
PXC-564 : PXC changes to support the keyring option

kennt-percona:
Issue: wsrep_sst_rsync.sh does not transmit the keyring (so any
use of the keyring option will fail). Also, some modifications
to the wsrep_sst_xtrabackup-v2.sh were needed.

Solution:
Modify wsrep_sst_rsync.sh to send/receive the keyring file.
Modify wsrep_sst_xtrabackup-v2.sh to use the directory the keyring
is in (rather than the datadir). We should not be copying the file
into other directories.  This also removes the need for a temporary
directory.

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -202,3 +202,26 @@ wsrep_check_programs()
 
     return $ret
 }
+
+#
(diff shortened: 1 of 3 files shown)


kbauskar:
LGTM.

PR #197, opened 2016-07-13
- PXC#635: Error message need to be updated for the condition when ru…

kbauskar:
…nning

  alter to switch from transaction to non-transactional engine

  Error message pointed to string that is relevant when ALTER is performed
  on non-transactional engine. Corrected error message.


kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/96/

kbauskar:
LGTM

PR #199, opened 2016-07-14
- PXC#634: Improve strict mode error message to be more customer frie…

kbauskar:
…ndly

  Improved error messages by including pxc_strict_mode value as part
  of error message.


kbauskar:
LGTM

PR #201, opened 2016-07-14
Fix PXC-636 : Doing an SST with encrypt=1 (symmetric encryption) fails

kennt-percona:
Issue:
When using symmetric encryption, the data stream was changed, starting
from PXB 2.3.  This broke wsrep_sst_xtrabackup-v2.sh.

Solution:
Change the code in wsrep_sst_xtrabackup-v2.sh to match PXB 2.3+.  Also,
add a requirement that at least PXB 2.4.3 is required.

Conflicts:
    mysql-test/suite/galera/t/disabled.def
    scripts/wsrep_sst_xtrabackup-v2.sh


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/99/

kbauskar:
LGTM

PR #202, opened 2016-07-15
- PXC#639: Exclude system tables (tables in mysql db) from pxc-strict…

kbauskar:
…-mode checks.

  pxc-strict-mode blocks insert to non-transactional tables which includes
  system table created with MyISAM.

  This causes inconvenience to end-user as there is no easy corrective action
  except changing the pxc-strict-mode to DISABLED or PERMISSIVE.

  With that thought-line in place we have decided to exclude system-tables
  (located in mysql database) from pxc-strict-mode gamut.


kbauskar:
LGTM

kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/Host=centos6-64,Ttype=debug/100/

PR #203, opened 2016-07-15
- PXC#628: Applier trx failure failed to rollback statement

kbauskar:
Ideally apply trx should never fail given that nodes are consistent.
  If for some un-foreseen reason it fails (generally due to human
  introduce incompatibilities) transaction is rollback.
  This rollback should ensure that statement open by apply_trx
  are also rollback before the real transaction is rollback.
  Check Relay_log_info::cleanup_context for detail comments.


kbauskar:
LGTM

mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/256/

PR #209, opened 2016-07-18
Fix 267 : segfault when "SET STATEMENT ... FOR OPTIMIZE TABLE" on tab…

kennt-percona:
…le that doesn't exist

Issue:
When running this command on a node, this will crash a replicated node.

Solution:
This happens within copy_system_variables() in sql_plugin.cc. The code
tries to do a deep copy and makes a copy of table_plugin and temp_table_plugin
that are set to NULL in WSREP threads. The fix is to check that these values are
non-NULL before copying.


kbauskar:
LGTM

kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/105/

PR #211, opened 2016-07-18
- PXC#641: CREATE TRIGGER TOI missed generating key.

kbauskar:
For TOI isolation to avoid conflict of it with normal operation
  isolation key is generated based on the table/object being modified.
  CREATE TRIGGER is done on a said table specified by user but TOI
  key was not being generated for it.


kbauskar:
LGTM

PR #214, opened 2016-07-19
- PXC#268: binlog_rows_query_log_events causes replicated node to crash.

kbauskar:
Setting binlog_rows_query_log_events to ON will generate ROW_QUERY_LOG_EVENT.
  This event logs an extra information while logging row information and so
  event should be kept infact till ROW_LOG_EVENT is processed and should be
  freed once ROW_LOG_EVENT is done.

  PXC/Galera logic freed this event immediately on handling it (like any other
  replicated event) this created 2 issues:
- ROW_LOG_EVENT is access a stale information.
- ROW_LOG_EVENT trying to clean this event as part of its clean up code
  result in double-free.


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/262/

PR #215, opened 2016-07-20
- PXC#647: PXC-5.7 to PXC-5.7 upgrade fails.

kbauskar:
PXC-5.7 upgrade needs to update INFORMATION_SCHEMA. Tables in
  INFORMATION_SCHEMA are maintained in-memory using MEMORY engine.
  Even though these are temporary table, MySQL is_temporary_table fail
  to classify them as they are system-temporary tables. Fixed it by
  treating information_schema as system db.


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/108/

PR #217, opened 2016-07-21
Documenting PXC-602: Strict mode

azhebel:


doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,296 @@
+.. _strict-mode:
+
+===========
+Strict Mode
+===========
+
+This mode is designed to avoid problems with improper configuration of |PXC|.
```
kbauskar: Mode is designed to avoid use of experimental or unsupported feature of PXC.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,296 @@
...
+It performs a number of validations against essential features of |PXC|.
+Some validations are performed only at startup,
+while others remain enabled during runtime.
+
+Depending on the actual mode you select,
```
kbauskar: Wondering why it is inserting newline for for such a short text.
May be it looks ok in html format but looks bit weird here.

azhebel: Just easier to edit compared to when each paragraph is on the same line or simply hard wrapping at 80 chars. Breaking lines at logical "blocks" of text, like when a new sentence is started or after a comma that defines two parts of a sentence. When I need to rephrase a part of the sentence, I can simply replace the whole line, because it is a separate logical part of the sentence. Sort of like treating text as code.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+except if the node is acting as a standalone server
+or the node is bootstrapping, then strict mode is disabled.
+
+To select the mode,
+use the :variable:`pxc_strict_mode` variable in the configuration file
+or the ``--pxc-strict-mode`` option during ``mysqld`` startup.
+
+.. note::
+
+   Although it is recommended to start the server with the necessary mode,
```
kbauskar: Percona highly recommend user to use pxc_strict_mode = ENFORCING only. If some part of workload fails you may want to relook at your workload as it may be exercising experimental or unsupported feature.

Also, it is important to have all the nodes running with same configuration (baring hardware limits) including pxc_strict_mode. (Changing pxc_strict_mode on selected node from cluster that is being used especially for WRITE workload is not a good idea and can lead to inconsistency.)

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+   you can dynamically change it during runtime.
+   For example, to enable permissive strict mode::
+
+      mysql> SET pxc_strict_mode=PERMISSIVE;
```
kbauskar: Before changing the mode from ENFORCING to anything else do study effect that associated with it.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+   :local:
+
+Storage engine
+--------------
+
+|PXC| currently supports replication only for tables
+that use a transactional storage engine (XtraDB or InnoDB).
+When you enable strict mode,
+non-system persistent tables are checked during runtime.
+If a write or alter operation is performed on an unsupported table,
```
kbauskar: write (DML), ALTER, TRUNCATE, ADMIN operation.

(Again I assume you are still using the document which is stale it was present for reference only
 and so I documented details notes in comment section.)

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+
+.. contents::
+   :local:
+
+Storage engine
+--------------
+
+|PXC| currently supports replication only for tables
+that use a transactional storage engine (XtraDB or InnoDB).
+When you enable strict mode,
```
kbauskar: s/When you enable strict mode/When you set pxc_strict_mode = ENFORCING/g

enable is bit confusing as variable can have multiple values.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+By default, this variable is set to ``OFF``.
+However, if MyISAM replication is enabled, one of the following happens:
+
+* In case of *enforcing* or *master* mode,
+  an error is thrown and server startup is halted.
+
+* In case of *permissive* mode,
+  a warning is logged and server startup continues.
+
+.. note:: This validation is not performed during runtime,
```
kbauskar: I see you are still using the document to create it which unfortunately has stale information.

"because the :variable:`wsrep_replicate_myisam` variable
-   cannot be set dynamically." ... is wrong.

From comment section:
DISABLED:
At startup, no check is enforced.
At runtime, setting wsrep_replicate_myisam to ON/OFF is allowed.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,263 @@
...
+-----------------
+
+The binary logging format must be set to ``ROW``.
+This is the default format.
+
+When you enable strict mode,
+it checks the value of the |binlog_format| variable
+both at server startup and during runtime.
+
+In case of *enforcing* or *master* mode:
```
kbauskar: For each section we should also list the DISABLED semantics like we are doing for ENFORCING AND PERMISSIVE.

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
+.. _strict-mode:
```
kbauskar: File name can be changed from 

doc/source/features/strict-mode.rst

doc/source/features/pxc-strict-mode.rst

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+
+===========
+Strict Mode
```
kbauskar: PXC Strict Mode

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+===========
+
+Strict mode is designed to avoid the use of
```
kbauskar: s/Strict mode/PXC Strict mode/gc

(please replace all the strict mode with pxc_strict_mode. There is innodb_strict_mode too that would confuse user)

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+without risking the consistency of your data.
+
+If you are planning to set strict mode to anything else than ``ENFORCING``,
+you should be aware of the limitations and effects
+that this may have on data integrity.
+For more information, see :ref:`validations`.
+
+To set the mode,
+use the :variable:`pxc_strict_mode` variable in the configuration file
+or the ``--pxc-strict-mode`` option during ``mysqld`` startup.
```
kbauskar: Mode can be set even dynamically using

set global pxc_strict_mode = ....

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+
+Binary log format
+-----------------
+
+|PXC| supports only the default row-based binary logging format.
+This validation checks the value of the |binlog_format|_ variable
+at session level, because it cannot be changed globally.
+
+Depending on the strict mode selected,
+the following happens:
```
kbauskar: Allowing setting of binlog_format at startup will set global variable something that is not allowed.

Check this

DISABLED:
At runtime, setting binlog_format to STATEMENT/MIXED/ROW is allowed.
PERMISSIVE:
At runtime, setting binlog_format to STATEMENT/MIXED logs a warning. New setting is accepted. (Setting it to ROW is welcome operation and doesn't result in a warning)
ENFORCING/MASTER:
At runtime, setting binlog_format to STATEMENT/MIXED logs an error. SET statement fails. (Setting it to ROW is welcome operation and doesn't result in an error).

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+Explicit table locking
+----------------------
+
+|PXC| has only experimental support for explicit table locking operations,
+The following undesirable operations lead to explicit table locking
+and are covered by this validation:
+
+* ``LOCK`` and ``UNLOCK``
+* ``GET_LOCK()`` and ``RELEASE_LOCK()``
+* ``FLUSH TABLE ... WITH READ LOCK``
```
kbauskar: there are 2 flush table.

flush table <table> with read read lock; (this is blocked)

flush table with read lock; (this is allowed)

So just to be extra careful we should probably mention it as "FLUSH TABLE <table> WITH READ LOCK"

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+.. _explicit-table-locking:
+
+Explicit table locking
+----------------------
+
+|PXC| has only experimental support for explicit table locking operations,
+The following undesirable operations lead to explicit table locking
+and are covered by this validation:
+
+* ``LOCK`` and ``UNLOCK``
```
kbauskar: LOCK Table. (Unlock we support based on context so we can remove UNLOCK from here)

doc/source/features/strict-mode.rst
```diff
@@ -0,0 +1,384 @@
...
+
+``DISABLED`` or ``MASTER``
+
+ At startup, no validation is performed.
+
+ At runtime, all operations are permitted.
+
+``PERMISSIVE``
+
+ At startup, no validation is perfromed.
```
kbauskar: s/perfromed/performed/gc

doc/source/index.rst
```diff
@@ -65,6 +65,7 @@ Features
 
    features/highavailability
    features/multimaster-replication
+   features/strict-mode
```
kbauskar: pxc-strict-mode

doc/source/wsrep-system-index.rst
```diff
@@ -309,6 +309,47 @@ This variable should contain the path to the Galera library (like :file:`/usr/li
...
+  :ref:`explicit table locking <explicit-table-locking>` is not performed.
+  This mode can be used with clusters
+  in which write operations are isolated to a single node.
+
+By default, strict mode is set to ``ENFORCING``,
+except if the node is acting as a standalone server
+or the node is bootstrapping, then strict mode defaults to ``DISABLED``.
+
+For more information, see :ref:`strict-mode`.
+
```
kbauskar, line 361: Another important point:

Note: Setting pxc_strict_mode from DISABLED/PERMISSIVE to ENFORCING/MASTER needs to ensure following conditions are met. wsrep_replicate_myisam=OFF, binlog_format=ROW, log_output=FILE/NONE, transaction isolation level != SERIALIZABLE.

PR #219, opened 2016-07-25
Fix PXC-648 : Remove unused option (sst_special_dirs) from code and docs

kennt-percona:
Issue:
sst_special_dirs is an unused variable (it was in use previously but those
have been removed).  This variable serves no purpose anymore.

Solution:
Remove the variable from the code and the docs.

Conflicts:
    scripts/wsrep_sst_xtrabackup-v2.sh


kbauskar:
LGTM

PR #220, opened 2016-07-26
Fix PXC-564 : PXC changes to support the keyring option

kennt-percona:
Issue:
Code required PXB 2.4.3, needs to require 2.4.4

Solution:
Change PXB requirement to 2.4.4


kbauskar:
LGTM

PR #221, opened 2016-07-26
- Corrected error handling message to use proper enum. Before everyth…

kbauskar:
…ing was wrong using TABLE enum value


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/265/

PR #224, opened 2016-07-27
- PXC#651: Followup from PXC#628 that is already fixed but in limited…

kbauskar:
… way.


kbauskar:
LGTM

mtr: http://jenkins.percona.com/job/pxc56.test.mtr/266/

PR #225, opened 2016-07-27
Added a my_setwd call for SST remounting the datadir

y-trudeau:
This is a fix to allow SST operations that mount the datadir to work correctly.  It adds a my_setwd call right after the SST so that MySQL opens its files in the mounted device instead of in the empty mount point.


kbauskar:
LGTM

PR #227, opened 2016-07-27
- PXC#558: sst fails with mysqldump + gtid_on

kbauskar:
When executing mysqldump with partial backup on node with GTID enable
  then use --set-gtid-purged=OFF to avoid emission of GTID of all transactions,
  even those that changed suppressed parts of the the databases.


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/114/

PR #231, opened 2016-07-28
Fix PXC-636 : Doing an SST with encrypt=1 (symmetric encryption) fails

kennt-percona:
Issue:
Previously, this didn't work so the code limited the required encryption
to modes 2 and 3.  Now that this is fixed, encrypt=1 should be allowed also.

Solution:
Allow encrypt mode = 1


kbauskar:
LGTM

kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/115/

PR #232, opened 2016-07-29
Fix PXC-655 : when using encrypt=1, performing an SST will display th…

kennt-percona:
…e encryption key in the logs

Issue:
When using "encryt=1" in the [sst] section, the encryption key will show in
the logs when an SST is done.

Solution:
Have the script late-evaluate the variables, thus they won't show up when
we print out the commands in the script, it'll show up as "$ecmd".


kbauskar:
LGTM

PR #234, opened 2016-08-01
- PXC#658: INSERT .. SELECT * from table fails with pxc-strict-mode

kbauskar:
INSERT .... SELECT \* from table involves 2 different tables.
  source table (copy data from) and destination table (copy data to)

  We just need to check that destination table is not a non-transactional
  table. Source table could be non-transactional table and still it should
  work. pxc-strict-mode check wrongly enforced it on all tables causing
  insert .... select \* to fail.


kbauskar:
LGTM
mtr: http://jenkins.percona.com/job/pxc57.test.mtr/117/
http://jenkins.percona.com/job/pxc57.test.mtr/119/

PR #235, opened 2016-08-01
- PXC#659: Disable switching slave_preserve_commit_order to ON

kbauskar:
When slave is configured to use multi-threads for applying master
  replicated action, slave_preserve_commit_order can help enforce
  application of these action in same sequence like master.

  In short, this is nothing but a commit monitor.

  If slave is pxc-node (galera-node) then galera already has commit
  monitor (being multi-master setup) but with different semantics
  to cater to multi-master setup. These 2 semantics conflicts
  and PXC need to opt for one of the semantics and naturally PXC
  opts for galera semantics.

### DIFF
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -3736,6 +3736,22 @@ static bool check_not_null_not_empty(sys_var *self, THD *thd, set_var *var)
   if (res && res->is_empty())
     return true;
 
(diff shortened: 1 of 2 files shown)


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/120/

PR #241, opened 2016-08-04
ProxySQL with PXC how-to (DOC-278) (5.7)

hrvojem:


kbauskar:
At top level looks okay. Just run it through Ramesh and Marcos who are actually using the Proxy-SQL.

PR #247, opened 2016-08-08
Fix PXC-654 : Better support in the wsrep_sst_* scripts for the defau…

kennt-percona:
…lts-group-suffix option

Issue:
When looking up the value of options in the config files, the group suffix is often
ignored.  This makes it difficult to use a single config file for a test, which
would make it easier to describe a particular test configuration.

Solution:
Have the parse_cnf() function also check and use the group suffix when
looking for a particular configuration value.

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -207,3 +207,38 @@ wsrep_check_programs()
 
     return $ret
 }
+
+#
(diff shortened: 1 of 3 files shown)


kbauskar:
LGTM.

kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/281/

PR #254, opened 2016-08-10
- Some test-cases failed due to underlying infrastructure differences.

kbauskar:
For example: Depending on which node boot first based on machine load
  value of auto-increment column differs.
  This patch aims at making these test-cases infrastructure independent.


kbauskar:
LGTM.
mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/280/

PR #259, opened 2016-08-12
- #PXC-665: SST donor node has status "JOINED" after successful resync

sysprg:
Operations that use the global lock (such as FLUSH TABLES FOR READ LOCK) and DDL operations that are protected via the RSU can be performed during SST. In this case, they come into interference with the SST process, which leads to a variety of bugs, including hang or crash of the donor node after SST donation. Alternatively, the SST donor node stays in the "JOINED" status, even after the completion of the SST.

This patch solves the problem with the incorrect classification of the "JOIN" message (which is generated as response to the wsrep->resync() call) at the GCS layer. Improper handling of the JOIN message leads to ignoring of the subsequent SYNC message and generates the above-described problems.

Pausing of the RSU operations during the SST will be solved as a separate issue in another patch.

Galera part of this patch is located here: https://github.com/percona/galera/pull/80


kbauskar:
LGTM

kbauskar:
mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/282/

PR #261, opened 2016-08-15
- #PXC-608: Assertion `node->desync_count > 0' failed in void gcs_node_update_status(gcs_node_t*, const gcs_state_quorum_t*)

sysprg:
This error occurs because although disconnection node from the cluster involves stop of replication, closing of all client connections and consequently the cancellation of all transactions (which are tied to client connections), but the desynchronization counter associated with the current node, as well as the last state of the node, which is stored in the prim_state field of the gcs_group_t structure, are not cleared even after disconnecting the node from the cluster.

Galera part of this patch is located here: https://github.com/percona/galera/pull/81

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -1204,6 +1204,16 @@ void wsrep_stop_replication(THD *thd)
   /* wait until appliers have stopped */
   wsrep_wait_appliers_close(thd);
 
+  /* After disconnecting the node from the cluster and closing all */
+  /* client connections we need to reset the desynchronization counter: */
+
+  wsrep_desync = 0;
+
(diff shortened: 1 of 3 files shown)


sql/wsrep_mysqld.cc
```diff
@@ -1204,6 +1204,11 @@ void wsrep_stop_replication(THD *thd)
   /* wait until appliers have stopped */
   wsrep_wait_appliers_close(thd);
 
+  /* After disconnecting the node from the cluster and closing all */
+  /* client connections we need to reset the desynchronization counter: */
+
+  wsrep_desync = 0;
```
kbauskar, line 1210: Do we also need to reset local_status ?

PR #265, opened 2016-08-22
- PXC#671: Node transitioning to non-primary with active ALTER TABLE …

kbauskar:
…result in crash.

  If ALTER (or any DDL command that needs TOI) is active before it execute
  TOI action if node is transitioned to non-primary state then ALTER command
  should fail. This failure is handled and generated by check in
  wsrep_to_isolation_begin. As part of PXC#570 fix this check got suppressed
  and instead of returning error it started to return success.
  Corrected it now to return error.


kbauskar:
LGTM

PR #266, opened 2016-08-22
- PXC#672: ALTERing non-existing table emitted wrong message from

kbauskar:
pxc-strict-mode module.

  If an non-existing table is altered then pxc-strict-mode claimed it to be
  a table with non-transaction storage engine. Corrected the table existence
  check.


kbauskar:
LGTM

PR #267, opened 2016-08-22
- PXC#561: InnoDB: Failing assertion: lock_get_wait(other_lock) in fi…

kbauskar:
…le lock0lock.cc line 5944

  Original issue was due to use of table without explict primary key but we
  re-observed this issue post 5.7 merge due to accidentally missed merge
  that tried to mix semantics of galera + mysql group replication.


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/140/

PR #270, opened 2016-08-23
- PXC#674: Skip PFS transaction registration for WSREP engine.

kbauskar:
WSREP has dependent action as in WSREP is not a standalone storage
  engine. It replicate the action performed by the mainline storage
  engine (for example: InnoDB) and so the performance schema registration
  of transaction should be done mainline storage engine and not by
  WSREP storage engine. (Same as binlog storage engine).


kbauskar:
LGTM

mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/143/

PR #274, opened 2016-08-30
- PXC#680: If TOI action fail for DROP event it can leave behind stal…

kbauskar:
…e reference.

  If a node with an active event turns non-primary, and event is due for
  automated drop due to completion of its tenure TOI replication action will
  fail leaving behind a stale reference to lex that is temporarily updated
  for TOI replication action.
  Failure of TOI should restore the original lex reference.


kbauskar:
LGTM.
mtr: http://jenkins.percona.com/job/pxc56.test.mtr/296/

PR #275, opened 2016-08-31
- PXC#681: Avoid generating audit-log event for closing connection

kbauskar:
While closing connection audit-log event should not be generated.
  Fix for this was implemented in upstream but there are different
  classes for closing wsrep connection that failed to inherit this
  fix. Corrected it now.


kbauskar:
LGTM.
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/157/

PR #279, opened 2016-09-09
- #PXC-618: Assertion `length > 0' failed in virtual type_conversion_status Field_geom::store_internal(const char*, size_t, const CHARSET_INFO*)

sysprg:
For debug builds, Oracle fixed Bug #22195364: when an indexed nonnullable generated column with an empty string generated expression was updated during an insert for a duplicated key, there was an optimization that resulted in the server failing to find the duplicated index, causing an assertion to be raised.

However, they forgot to fix the Field_geom::store_internal() method, which leads to the new assertion failure.

To avoid the assertion failure we need to bypass the extra testing inside of this function, by analogy with the changes that were made in the Field_blob::store_internal() method (which were made by the Oracle).

### DIFF
--- a/sql/field.cc
+++ b/sql/field.cc
@@ -8726,24 +8726,25 @@ type_conversion_status
 Field_geom::store_internal(const char *from, size_t length,
                            const CHARSET_INFO *cs)
 {
-  DBUG_ASSERT(length > 0);
-
-  // Check that the given WKB
-  // 1. isn't marked as bad geometry data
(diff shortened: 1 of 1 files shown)


sysprg:
Jenkins job: http://jenkins.percona.com/job/pxc57.test.mtr/163/

dutow:
Hello!

I'm closing this as this was an upstream bug and was already fixed differently by Oracle.

PR #302, opened 2016-09-23
DOC-290: wsrep_cert_interval missing document in percona

azhebel:


kbauskar:
Ok for now but we need to expand the description in due-course.

Is this needed even for 5.6. If yes, do create a PR for it too.

PR #306, opened 2016-09-26
- #PXC-252: [pz@percona.com: Testing Auto Recover]

sysprg:
In fact, this hotfix covers three different issues (PXC-426, PXC-252 and it extension, which requires API change). Unfortunately, these three patches partially overlap in fragments of code and change the same files repeatedly. Enough difficult to separate them into independent pieces and in this case we need to accept them only one-by-one in fixed sequence - because of overlapping changes in the code.

Since we have already debugged them individually in the past, I combined them into one big patch.
## PART I: #PXC-426: Race condition during IST

The node lost connectivity with other nodes in the cluster for some time due to network problems. After this, it need to run IST to synchronize their state with the cluster. However, the IST request has failed.

If we already have large network delays and packet losses, then during IST we can encounter situation when seqno on the donor node has moved forward so much that IST is not possible and we need to run "full" SST request.

To avoid unsuccessful IST attempts, the Galera introduces the concept of "safety gap" - the window for seqno values in which the heuristic algorithm for selecting the donor prefers to switch to SST.

The immediate cause of the PXC-426 error is the inability of the PXC to start new SST request during node operation (using methods other than mysqldump). PXC and Galera written in such way that normally we need a full restart before launching the new SST.

To resolve this contradiction, we need to change the heuristic algorithm that selecting a donor in such way, that it took into account the impossibility of SST. If a joiner node sends a request without SST part, then after applying this patch the donor node must ignore the safety gap window and still tries to execute IST.

It is much better to have a small probability of failure in the IST, than immediately suffer from failure with 100% guarantee (due to heuristics associated with safety gap window). Moreover, if the request contains the SST-related part (not only IST), then we can leave current logic with the additional heuristics related to safety gap window.

However, the algorithm of the donor selection, which implemented in the GCS, contains the potential flaw, and in rare situation it still may choose as a donor such node, where seqno can gone forward before the IST request will be received.

Therefore, since heuristic that selects the donor node does not give us a 100% guarantee that seqno does not moved forward while the new node joining the cluster, and during node operation we have only a requests for the IST without SST part, then we need a way to informing joiner node, that it should prepare to receive SST. This return path also added to the patch code.

Especially we need this signal to notify the user that server unable to perform the SST instead of IST when xtrabackup or rsynch selected as the state transfer method.

In these cases, we need to show the user an explicit message that indicates that we need to perform SST after node restart. However, current version of the server continues to work, even when the synchronization of the node is no longer possible and it cannot continue normal operation in a cluster. Or vice versa, it falls for one of the assertions in the Galera code. This corrected in the patch.

In addition, I changed the Galera code to allow restart SST after failing to perform IST - to be protected against failures in the future, when implementation of the wsrep_sst_prepare function may change.

In addition, if the IST request canceled due to an error at the GCS level or by initiative of the server, and if the node remain in the S_JOINING state, then we must return it to the S_CONNECTED state.
## PART II: #PXC-252: [pz@percona.com: Testing Auto Recover]

Usualy, we cannot restart the SST automatically whenever SST fails because the server usually does not have enough information to understand why this was happened. For example, the analysis of the log indicates that the connection has been restored and the 22 error was not fatal:

```
>2014-12-05 17:16:08 2596 [Warning] WSREP: Could not find peer:
>2014-12-05 17:16:08 2596 [Warning] WSREP: 1.0 (nuc3): State transfer to
>-1.-1 (left the group) failed: -22 (Invalid argument)
...
>2014-12-05 17:16:08 2596 [Note] WSREP: Member 1.0 (nuc3) synced with group
```

At the same time log analysis shows that, in fact, the SST process was completed with another error = 1 ("Operation not permitted"). Perhaps it was a mistake due to insufficient access rights or incorrect password, but not a network error.

If this diagnosis is correct, then restart of the SST did not help us to resolve the problem. If it is wrong, it turns out that at this point we do not have enough information to make the right decision. A simple patch cannot eliminate this (in this case, I think here we need to talk about architecture refining as a separate development task, not just a patch on the product support level).

However, we definitely has a serious problem that we should eliminate. Namely - after detecting a failure due to permissions, the server process is completed improperly, leaving the unfinished thread:

```
Error in my_thread_global_end(): 1 threads didn't exit
>141205 17:16:28 mysqld_safe mysqld from pid file /var/lib/mysql/mysqld.pid
>ended
```

We have already eliminated this error in most scenarios by a previous patches, but I found new possible branches, which similar error arise (for example, it is possible when we repeat the connection attempts after the failure of the SST - although I was not able to reproduce this situation manually or by automatic script since it is rare race condition).
## PART III: Changes in the wsrep API (continuation of the #PXC-252):

In addition, I found that the failure during the SST, which is diagnosed on the Galera level, leads to the completion of the server using abort() call, but in this case the child processes that have been launched for the SST continues to run after completion of the server.

This makes it impossible to re-start the server before all timeouts expired (in these SST-related processes). Otherwise we failed due to the busy sockets or it leads to other fatal errors due to interference between new and old instances of the SST scripts.

In this patch, I added new checks to safely shutdown the server, for correct processing of a failed attempts to make a new connection and the SST, and for the destruction of all child processes when the server terminated abnormally by initiative of the Galera.

To do this, I needed to add a new callback to wsrep API, because the Galera intercepts the SIGABRT signal. Therefore we cannot intercept abnormal termination of the server process (after it calls abort() function from the Galera side) without expanding the Galera API by adding to it new callback.

Galera part of this patch is located here: https://github.com/percona/galera/pull/84


mysql-test/suite/galera/t/galera_seqno_gone_forward.test
```diff
@@ -0,0 +1,105 @@
...
+
+--connection node_1
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 1 FROM INFORMATION_SCHEMA.GLOBAL_STATUS WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+# Do some changes on node 1:
+
+CREATE TABLE t2 (f1 INTEGER);
+INSERT INTO t2 VALUES (1), (2), (3), (4), (5), (6), (7), (8), (9), (10);
```
kbauskar: With such a small seqno change I assume either way it will use SST.
safety_gap logic is such that a small change in IST causes it to select SST.

Can we have some significant change while node is down.

kbauskar (changes requested):

PR #307, opened 2016-09-28
Fix conflict between xtrabackup versions

EvgeniyPatlan:


tplavcic:
I think this is NOT something you want to do as it means that PXC in the repository will replace percona-xtrabackup for EVERYBODY even if they don't want PXC installed or if they want to have PXB 2.3 and PT without PS/PXC. But I can always be wrong.

kbauskar:
@tplavcic  I assume the check will triggered only when user try to install PXC-5.6. If user is not trying to install PXC and just want PXB then there is no such use-case right ?

EvgeniyPatlan:
@tplavcic It will just obsoletes old version of percona-xtrabackup during percona-xtradb-cluster-server installation.
No you are right! I would try to find another solution

tplavcic:
It will obsolete it also when doing "yum update", so if you have percona-xtrabackup installed from our repo and do "yum update" percona-xtrabackup will be replaced with Percona-XtraDB-Cluster-server package (we have the same on centos7 where our -shared packages Obsoletes mariadb-libs).
Make a test packages and put them into experimental repository and try it out, but this is what I expect.

tplavcic:
When some package in repository has "Obsoletes" clause the package manager always looks at that regardless if you want to install that package or not.

PR #311, opened 2016-09-29
Fix percona-xtrabackup conflict

EvgeniyPatlan:


EvgeniyPatlan:
I have found a bit tricky solution which could help. Please check if it is possible to use it

build-ps/percona-xtradb-cluster.spec
```diff
@@ -906,11 +907,16 @@ if [ -x %{_sysconfdir}/init.d/mysql ] ; then
 fi
 
 %post -n Percona-XtraDB-Cluster-server%{product_suffix}
-
 if [ X${PERCONA_DEBUG} == X1 ]; then
         set -x
 fi
 
+ver=`/bin/rpm -qi percona-xtrabackup | grep Version | awk '{print $3}'`
+if [ $ver = '2.3.5' ]; then
```
tplavcic, line 915: Maybe it would be better to check only for "2.3" and not the whole number, users don't have to upgrade right away so the version in repository can change and although not likely but possible we will push "2.3.6" and it will not work in this case.

EvgeniyPatlan: @tplavcic yes sure I would update the version check. But what about the solution in general?
Is it acceptable or not?

tplavcic: Well it's definitely not something you would normally do, but if we're not planning to replace default percona-xtrabackup with version 2.4 and just telling the users to remove packages and replace PXB version with the new one is not the option then I think this is the only solution left.
@abychko needs to check this as well and agree on it.

abychko:
in general calling rpm (or yum & rpm) from inside the another rpm process is very bad idea. do you have some another solution?

PR #312, opened 2016-09-29
- PXC#695: Reduce PXC 5.6 dependency from 2.4 to 2.3 due to package u…

kbauskar:
…pgrade issue

  PXC-5.6 was made dependent with latest XB-2.4 but creates a problem for inline
  minor upgrade as name of the XB-2.4 and XB-2.3 packages are different.
  Upgrade installer think it is different packages and inline upgrade
  can't be done.

  Resolve it by downgrading PXC-5.6 support to XB-2.3


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc56.test.mtr/306/

PR #326, opened 2016-10-19
- PXC#715: Ensure that super_read_only doesn't apply to galera replic…

kbauskar:
…ation applier
- read_only mode is not applied to galera applier.
- super_read_only mode shouldn't apply to galera applier too.
  (this also makes it compatible with MySQL async master replication semantics
   introduced in MySQL-5.7)
  
  (MySQL did a good job of unifying check in single function that avoid
  spread out implemention and so the conflicts with 5.6 vs 5.7)

Conflicts:
    sql/handler.cc
    sql/lock.cc
    sql/sql_parse.cc
    sql/sql_trigger.cc


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc57.test.mtr/172/

PR #327, opened 2016-10-19
- PXC#715: Ensure that super_read_only doesn't apply to galera replic…

kbauskar:
…ation applier
- read_only mode is not applied to galera applier.
- super_read_only mode shouldn't apply to galera applier too.
  (this also makes it compatible with MySQL async master replication semantics
   introduced in MySQL-5.7)


kbauskar:
LGTM

kbauskar:
http://jenkins.percona.com/job/pxc56.test.mtr/310/

PR #330, opened 2016-10-25
- PXC#717: ALTER USER is not replicated

kbauskar:
CREATE/DROP USER are replicated but as listed here
  https://bugs.launchpad.net/codership-mysql/+bug/1376269 ALTER USER
  is not replicated.


kbauskar:
LGTM

mtr run: http://jenkins.percona.com/job/pxc56.test.mtr/315/

PR #332, opened 2016-10-26
PXC-701 : docker image relies on init.ok, that is deleted when SST is…

kennt-percona:
… triggered

Issue:
The startup script creates an init.ok file in the datadir to indicate that the
node has been configured. On joining a cluster, the joiner side will clean the
datadir in preparation for receiving an SST.  This cleaning will remove the
init.ok file.

Solution:
Add init.ok to the list of files that won't get removed by the SST.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/174/

There is a crash in the MTR run, will be filing a separate issue for that (it's an assert that's being hit in the instrumentation code).

kbauskar:
LGTM

PR #334, opened 2016-10-28
- PXC#252/PXC#426: Race condition in IST

kbauskar:
(patch from Julius)

  PART I: PXC#426: Race condition during IST

---

  The node lost connectivity with other nodes in the cluster for some time due
  to network problems. After this, it need to run IST to synchronize the state
  with the cluster. However, the IST request still fails.

  (Probable cause: donor has made progress and no more holds a needed
   seqno needed for IST to complete successfully)

  Galera has concept of safety_gap to handle this problem (though not 100%
  fool-proof) and if the check fails then it suggest switching to SST.
  If server is configured to use rsync or XB method online SST during
  node operation is not possible.

  It is important that if server is forced to switch to SST
  and SST is not possible then node indicate so and shutdown properly.
  (State of node should be restored too).

  PART II: PXC#252: SST operation failure should ensure cleanup

---

  If the SST fails for the reason mentioned above the node should successully
  clean the resources before shutdown.

  PART III: Changes in the wsrep API (continuation of the #PXC-252):

---

  In addition, I found that the failure during the SST, which is diagnosed on
  the Galera level, leads to the completion of the server using abort() call,
  but in this case the child processes that have been launched for the SST
  continues to run after completion of the server.

  This makes it impossible to re-start the server before all timeouts expired
  (in these SST-related processes). Otherwise we failed due to the busy sockets
  or it leads to other fatal errors due to interference between new and old
  instances of the SST scripts.

  In this patch, I added new checks to safely shutdown the server, for correct
  processing of a failed attempts to make a new connection and the SST, and for
  the destruction of all child processes when the server terminated abnormally
  by initiative of the Galera.

  To do this, I needed to add a new callback to wsrep API, because the Galera
  intercepts the SIGABRT signal. Therefore we cannot intercept abnormal
  termination of the server process (after it calls abort() function from the
  Galera side) without expanding the Galera API by adding to it new callback.

Conflicts:
    sql/signal_handler.cc
    sql/wsrep_mysqld.cc
    wsrep/wsrep_api.h

### DIFF
--- a/mysql-test/suite/galera/r/galera_seqno_gone_forward.result
+++ b/mysql-test/suite/galera/r/galera_seqno_gone_forward.result
@@ -1,0 +1,18 @@
+CREATE TABLE t1 (f1 INTEGER);
+INSERT INTO t1 VALUES (1), (2), (3), (4), (5), (6), (7), (8), (9), (10);
+Shutting down server ...
+CREATE TABLE t2 (f1 INTEGER);
+INSERT INTO t2 VALUES (3);
+INSERT INTO t2 VALUES (2);
+INSERT INTO t2 VALUES (1);
+SET GLOBAL wsrep_provider_options = 'dbug=d,simulate_seqno_shift';
+Starting server ...
+# restart
+SELECT COUNT(*) = 10 FROM t1;
+COUNT(*) = 10
+1
+SELECT COUNT(*) = 384 FROM t2;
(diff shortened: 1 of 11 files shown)


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc57.test.mtr/177/

PR #335, opened 2016-10-28
mysql-systemd: Use softcoded mysqld path

alvaroaleman:


EvgeniyPatlan:
Why do we need this. As for me the best way is to use absolute path.

alvaroaleman:
My usecase is an extradb that is not installed in the default location.

Since all other paths in that script are not absolute and there is a `PATH="/sbin:/usr/sbin:/bin:/usr/bin:$bindir:$sbindir"`, this would make things easier because then I would only have to adjust that `PATH``setting

EvgeniyPatlan:
Thank you for your contribution - we really appreciate your input! Unfortunately we cannot merge this PR as it does not fulfill our requirements for a successful merge. In this case, our packages are not relocatable and install all binaries into predefined paths. In such case we try to be sure that we would use the binary which was installed from our package and your database wouldn't be broken by the incorrect call. PATH in mysql-systemd script is used for better parsing of input options. 

If you would like to make your own build including your contribution, please refer to the following instructions:  https://www.percona.com/doc/percona-xtradb-cluster/5.7/install/compile.html#compile

PR #338, opened 2016-11-02
- PXC#720: wsrep_max_ws_rows check should skip non-replicated write-sets

kbauskar:
* wsrep_max_ws_rows should skip non-replicated write-sets.
    non-replicated write-sets are generated by DML action on temporary
    table (explict or implicit).


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc56.test.mtr/321/

PR #339, opened 2016-11-16
- PXC#710: ProxySQL assisted PXC maintenance mode.

kbauskar:
Often PXC nodes needs to be taken down for maintenance
  and this generally involve co-ordination with load balance
  and the real node.

  To simply this flow we now introduce a proxysql assisted
  pxc maintenance mode. User will shutdown or set the maintenance
  mode on real node as usual without caring if load balancer
  has active traffic diverted to this node.
  Load Balance (currently only ProxySQL) will be tuned to
  understand this change and divert the traffic to other nodes.

  This of-course means node action will be delayed so that
  active transaction are given time to complete.
  (long running transaction will be killed as before
   but user can still increase this transition time by setting
   pxc_maint_transition_period)

### DIFF
--- a/mysql-test/include/default_mysqld.cnf
+++ b/mysql-test/include/default_mysqld.cnf
@@ -32,6 +32,7 @@ log-bin=mysqld-bin
 # It is maintainance challange to keep on modifying n different
 # galera.cnf file for this setting
 pxc_strict_mode=DISABLED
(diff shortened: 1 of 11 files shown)


kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/187/

PR #342, opened 2016-11-16
Fix PXC-723 : "dh key too small" error when using SSL connections for…

kennt-percona:
… SST

Issue:
With socat versions < 1.7.3, they have not updated the DH parameter
length (due to the LogJam vulnerability, SSL client implementations now
require DH parameters > 1024 bits). On the donor, this error would appear
  socat[35371] E SSL_connect(): error:14082174:SSL routines:SSL3_CHECK_CERT_AND_ALGORITHM:dh key too small
This affects PXC SST encrypt=2 and encrypt=3

Solution:
When needed (socat < 1.7.3 and running on the joiner during an SST), look for
an existing dhparams.pem file in the datadir, if there isn't one create
it with 2048 bits.  Then pass this as a parameter to socat (on the joiner side,
since the joiner is acting as a server).

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -70,6 +70,7 @@ xtmpdir=""
 
 scomp=""
 sdecomp=""
+ssl_dhparams=""
 
 # Required for backup locks
 # For backup locks it is 1 sent by joiner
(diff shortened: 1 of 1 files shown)


kennt-percona:
Modified the Jenkins script to set MTR_START_TIMEOUT from 180 to 300 seconds,  sometimes the code to generate the 2048-bit DH params can take a long time (several minutes)

kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/329/
http://jenkins.percona.com/job/pxc56.test.mtr/330/

kbauskar:
LGTM

PR #346, opened 2016-11-17
PXC-713 : Add SST encrypt=4 mode, support MySQL-generated SSL files

kennt-percona:
Issue:
Current methods cannot use the MySQL-generated files.
They use SSL files (ca,cert,and key) that are not generated in the same way.

Solution:
We are adding a new mode (encrypt=4) that requires the cert, ca, and key files.
This will do peer cert verification, but will not do hostname/commonname verification.
Modes encrypt=1,2,3 are being deprecated (for now, warnings will be issued).


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/191/
http://jenkins.percona.com/job/pxc57.test.mtr/192/

kbauskar:
LGTM

PR #347, opened 2016-11-18
- PXC#729: Switching to maintenance mode doesn't for x secs.

kbauskar:
Snippet to introduce wait was missing.
  Added it back.


kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/193/
LGTM

PR #348, opened 2016-11-21
PXC-724: pxc-encrypt-cluster-traffic option

kennt-percona:
Issue: SSL configuration of internal Galera communication and SST traffic
is complicated (especially since the expected files differ).

Solution: With PXC-713 (encrypt=4 option), this checkin adds the
pxc-encrypt-cluster-traffic option that enables PXC SSL auto-configuration.
PXC now uses the MySQL-generated SSL files to configure both
internal Galera communication and SST traffic.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/194/
http://jenkins.percona.com/job/pxc57.test.mtr/195/

kennt-percona:
Modified the code to only generate new certs if pxc-encrypt-cluster-traffic is set and we are bootstrapping.
http://jenkins.percona.com/job/pxc57.test.mtr/196/

kbauskar:
LGTM

PR #349, opened 2016-11-22
Fix PXC-722 : SST falls back to unencrypted mode if socat is not SSL …

kennt-percona:
…enabled, should just fail

Issue:
If socat is not OpenSSL enabled, then the code falls back to using socat as
unencrypted transport.  If the user requests encryption but we cannot provide
it, we should error out rather than use an unencrypted connection.

Solution:
If socat is not OpenSSL enabled, fail with a fatal error during an SST.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/332/

kbauskar:
LGTM.

PR #354, opened 2016-11-23
Fix PXC-716 : netcat help output in error log when running xtrabackup…

kennt-percona:
… SST

Issue:
Some versions of netcat write the info to stderr rather than just to stdout.

Solution:
Redirect stderr to stdout also (use '2>&1')


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/199/

kbauskar:
LGTM

PR #355, opened 2016-11-24
Fix PXC-732: Improve SST SSL encryption error message

kennt-percona:
Issue:
If an SSL file is missing (CA, key, or cert), socat/OpenSSL returns
a cryptic message.

Solution:
Before calling into socat/OpenSSL, verify that the file exists and that
we can read it.  If our check fails, put out an error message and
exit the SST.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/336/

kbauskar:
LGTM

PR #358, opened 2016-11-24
Fix PXC-735: SST with netcat ignores encrypt option

kennt-percona:
Issue:
If nc(netcat) is used as the transfer mechanism (with the 'transferfmt' option)
and SSL encryption is specifed (encrypt=2,3,4), the the encryption options are
ignored.

Solution:
If encryption is requested, then if we can't provide it, we will fail the
SST rather than silently defaulting to unencrypted transport.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/202/

kbauskar:
LGTM

PR #359, opened 2016-11-28
Fix PXC-690: Display Flow control interval as a status variable

kennt-percona:
Issue:
The values for the flow control interval can only be seen in the
logs when the values are changed.

Solution:
Add a status variable ('wsrep_flow_control_interval') to expose
the lower and upper limits of the flow control mechanism.
This is a read-only variable.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/203/

kbauskar:
LGTM

PR #360, opened 2016-11-29
PXC-594 PXC-608: Assertion `sender->count_last_applied' failed in gcs_group_handle_join_msg(gcs_group_t*, const gcs_recv_msg_t*)

sysprg:
This patch fixes two bugs:

PART I, PXC-594:
----------------

Assertion `sender->count_last_applied' failed in the
gcs_group_handle_join_msg() function.

This error stems from the fact that we have added
a direct transition from JOINED to DONOR state bypassing
SYNCED state, but does not take into account that in this
case the count_last_applied flag is lost. It is necessary
to change it in such a direct transition.

PART II, PXC-608:
-----------------

Assertion `node->desync_count > 0' failed in the
the gcs_node_update_status() function.

This error occurs because although disconnection node
from the cluster involves stop of replication, closing
of all client connections and consequently the cancellation
of all transactions (which are tied to client connections),
but the desynchronization counter associated with the current
node, as well as the last state of the node, which is stored
in the prim_state field of the gcs_group_t structure, are not
cleared even after disconnecting the node from the cluster.

Related Galera patch is located here: https://github.com/percona/galera/pull/95

### DIFF
--- a/mysql-test/suite/galera/r/galera_non_prim_after_desync.result
+++ b/mysql-test/suite/galera/r/galera_non_prim_after_desync.result
@@ -1,0 +1,36 @@
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY, f2 CHAR(1));
+Shutting down server ...
+Starting server ...
+SET SESSION wsrep_sync_wait=0;
+SET GLOBAL wsrep_OSU_method='RSU';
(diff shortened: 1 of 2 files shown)


dutow:
Hello

Closing this as PXC 5.6 is EOL, and one of the referenced issues was also fixed by Codership.

PR #387, opened 2017-01-26
Fix PXC-760 : gcache.page files not being deleted

kennt-percona:
Fix PXC-686 : Review semantics of gcache keep_pages_size and keep_pages_count

Issue
(760) When the limits on the gcache were exceeded (using gcache.keep_pages_size or
gcache.keep_pages_count), the files were not being deleted on a timely basis.
The decision on discard the seqnos was decided with static constants within
index_purge_required()

(686) The semantics of the gcache.keep_pages_size and gcache.keep_pages_count is
confusing because we are doing an AND condition (most people expect an OR).

Solution
Adding an extra call (GCache::cleanup_required) that checks the page store
within the gcache to see if the limits have been reached.  This is called
along index_purge_required().

Also, change the limits check to an OR.

These are the test cases.  The code change is in galera.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/213/

kbauskar (approved):

kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/224/

PR #388, opened 2017-02-24
- PXC#768: Allow CTAS with temporary table in PXC strict mode

kbauskar:
CTAS is blocked in pxc_strict_mode = ENFORCING as it involves
  DDL + DML which and can lead to GTID consistency issue.
  This is now applicable if table is temporary as it is local
  to the node and statement is not replicated.
  Allow CTAS on temporary table.


kbauskar:
mtr-run: http://jenkins.percona.com/job/pxc57.test.mtr/225/

PR #390, opened 2017-02-28
Fix PXC-755 : Passing use-memory to XB

kennt-percona:
Issue
The use-memory param to xtrabackup for --apply-log was only being
supplied through setting the inno-apply-opts options.

Solution
Add support for using the options set in [xtrabackup] "use-memory"
and [mysqld] "innodb_buffer_pool_size.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/229/

PR #392, opened 2017-03-01
Fix PXC-685 : Stablize 5.6/5.7 mtr suite for our Jenkins environment

kennt-percona:
Issue:
Additional warnings pop up when system is under load

Solution:
Suppress the warnings


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/376/

PR #395, opened 2017-03-02
Fix PXC-769: galera_gcache_recover fails intermittently

kennt-percona:
Issue
This test case fails intermittently.

Solution
It appears that we kill PXC too quickly, we don't give InnoDB
enough time to commit (especially under load). This causes InnoDB
to rollback the transction during the --wsrep-recover stage (this
output goes to a different log file so we don't see this output).
Thus the IST becomes an SST since a transaction is missing.

So the solution is to give the operation a little more time before
killing the node.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/377/

PR #397, opened 2017-03-06
Fix PXC-759 : Explictly show variable to indicate if node is in flow …

kennt-percona:
…control (paused)

Issue
When a PXC node enters flow control, there is no way of knowing that it
has done so.  This would be useful for debugging, especially if a
node is causing other nodes to block.

Solution
Add a status variable, 'wsrep_flow_control_status' that has the
values of 'ON' or 'OFF'.  'ON' indicates that the node is in
flow control.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/236/

PR #399, opened 2017-03-09
Fix PXC-766: Show progress during an IST

kennt-percona:
Issue
During a slow IST, there's no way of knowing if the system is making progress
(or how far along it is).

Solution
Add a status variable that can be queried for the IST status.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/243/

PR #405, opened 2017-03-10
- PXC#781: Failing assertion: lock_get_wait(other_lock) in file lock0…

kbauskar:
…lock.cc line 5884

  Said assertion doesn't hold true with PXC/Galera BF locks.
  Suppress them but do print details if that path is hit.


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/249/

PR #407, opened 2017-03-13
5.7 pxc 754

kbauskar:


kbauskar:
LGTM
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/250/

PR #408, opened 2017-03-13
Fix PXC-694 : garbd shouldn't recompute flow control limits

kennt-percona:
Issue: The current algorithm includes the garbd node in the flow
control calculations.

Solution: When calculating the flow control limits, do not include
the garbd node.

Also includes:
Additional fix to suppress the "Difficult to find free blocks" warning
Fixes to galera_load_data


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/251/

PR #410, opened 2017-03-13
Set correct option in wsrep.cnf

EvgeniyPatlan:


abychko:
merging to rollback mistakingly commented option which doesn't work under selinux

PR #413, opened 2017-03-14
Fix PXC-725 : check that SSL files work with each other

kennt-percona:
Issue:
We do not do any checking of the SSL files before using them.  If the
files do not work with each other, then the result is a cryptic
SSL error message coming from socat.

Solution:
Do some validation before using the SSL files (verify that the
cert and key match or the cert and CA match).


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/255/

PR #417, opened 2017-03-15
[BLD-668] Mysql common shouldn't be installed togeather with pxc-common

EvgeniyPatlan:


delgod:
LGTM, may be it is needed to reopen BLD-663

delgod (approved):

PR #419, opened 2017-03-17
Fix PXC-782: PXC xtrabackup-v2 option should use tmpdir option

kennt-percona:
Issue:
wsrep_sst_xtrabackup-v2.sh does not use the tmpdir option (either [xtrabackup] or [mysqld]).
It creates any needed temporary directories via "mktemp -d").

Solution:
Change the script so that it uses the tmpdir (if in [sst], [xtrabackup] or [mysqld], in
that order).

Additionally:
Fixes to the MW-328A test case.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/382/

kbauskar:
LGTM

PR #426, opened 2017-03-23
Fix PXC-758: Debug status output is misleading

kennt-percona:
PXC-794: PXC SST: usability, prepend ',' to sockopt if needed
    PXC-795: PXC SST: set "--parallel=4" as default option to wsrep_sst_xtrabackup-v2

Issue:
PXC-758: status output does not check the result code before reporting result.
PXC-794: make the "sockopt" less error-prone
PXC-795: perf, increase use of threads when backing up on the donor

Solution:
PXC-758: check the result code and output the according output
PXC-794: prepend a comma (",") to the sockopt before using
PXC-795: add the "--parallel=4" to socat as a default

### DIFF
--- a/mysql-test/suite/galera/r/MW-328A.result
+++ b/mysql-test/suite/galera/r/MW-328A.result
@@ -18,4 +18,4 @@ have_deadlocks
 Got one of the listed errors
 DROP PROCEDURE proc_update;
 DROP TABLE t1, t2;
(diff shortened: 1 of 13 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/263/

kbauskar:
LGTM

PR #429, opened 2017-03-24
Fix PXC-794 : PXC SST: usability, prepend ',' to sockopt if needed

kennt-percona:
Issue
The sockopt option requires the option to start with a comma(",").
This is easily missed and makes it harder to set than necessary.

Solution
If there is no comma, prepend it to the option before using it.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/384/

PR #446, opened 2017-04-25
Fix PXC-749: Assertion `pool_.size() == allocd_' failed in gu::MemPoo…

kennt-percona:
…l<thread_safe>::~MemPool() [with bool thread_safe = false]

Issue:
Code asserts when wsrep-certify-nonPK=0 is set and an INSERT is tried on a no-PK table.
The code results in a memory leak (which causes the assert)

Solution:
If the table does not have a PK, and wsrep-certify-nonPK is set to 0, do not
append non-unique keys to the row.

Also: fixes to the garbd test


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/281/

PR #448, opened 2017-05-02
Fix PXC-813: correct timezone for sst script

kennt-percona:
Issue:
The SST script (when printing output) does not show the TZ in the same
format as the rest of the error log.

Solution:
Change the format to match. We still use the SYSTEM timezone (as this
is the default for MySQL)

Also, some additional test case fixes.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/390/Host=ubuntu-yakkety-32bit,Ttype=debug/console

PR #449, opened 2017-05-02
Fix PXC-813: correct timezone for sst script

kennt-percona:
Issue:
The SST script (when printing output) does not show the TZ in the same
format as the rest of the error log.

Solution:
Change the format to match what MySQL is doing. Change the script
to use the same format (see --log-timestamps for supported TZs).

Also, some additional test case fixes.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/285/

PR #451, opened 2017-05-04
- PXC#816: Regression in GTID consistency when parallel applying used…

kbauskar:
… for async replication

  - GTID event generated from MASTER is processed by SLAVE.
  - SLAVE can process this event using master-thread if slave-parallel-workers=0
    or assign it to slave worker is slave-parallel-workers > 0.
  - Starting 5.7, MySQL has stopped writing this event to bin-log but PXC need
    this event for proper creation of write-set so PXC has logic to cache this event.
  - Hook to cache this event was wrongly placed that was triggered only if
    GTID event is being processed by master-thread (that is slave-parallel-workers=0)

  Fix:
  - Corrected hook such that the thread that is processing this event will
    cache the event.

------------------------

- PXC#820: unable to query pxc_maint_mode while switching to maintenance

  - After updating pxc_maint_mode mode, flow use to trigger a sleep for
    pxc_maint_transition_period so that ProxySQL can detect the change
    and update the state of the node while user still wait for node change.

  - Unfortunately, MySQL holds global-variable locks longer than expected.
    on_update routine is called after the variable is updated and theortically
    global-variable lock should not be held during on_update as update action
    is already complete by MySQL continue to hold locks and correcting it
    looks difficult now as legacy variable's on_update routine continue to
    update the variable even though semantically it should use on_update
    function for follow-up action.

  To get around this we now handles pxc_maint_mode case specially by checking
  if there is change in pxc_maint_mode post update. If the new mode is
  maintenance then trigger sleep.

### DIFF
--- a/build-ps/build-binary.sh
+++ b/build-ps/build-binary.sh
@@ -302,7 +302,7 @@ export MAKE_JFLAG="${MAKE_JFLAG:--j$PROCESSORS}"
 
 export DEBIAN_VERSION="$(lsb_release -sc)"
 echo $DEBIAN_VERSION
-if [[ "$CMAKE_BUILD_TYPE" == "Debug" ]] && [[ "$DEBIAN_VERSION" == "yakkety" ]]; then
+if [[ "$CMAKE_BUILD_TYPE" == "Debug" ]] && ([[ "$DEBIAN_VERSION" == "yakkety" ]] || [[ "$DEBIAN_VERSION" == "zesty" ]]); then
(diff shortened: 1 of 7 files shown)


kbauskar:
mtr run:
http://jenkins.percona.com/job/pxc57.test.mtr/288/ &
http://jenkins.percona.com/job/pxc57.test.mtr/287/

kennt-percona:
LGTM

PR #452, opened 2017-05-05
Fix PXC-812: PXC SST: joiner should not remove the donor-keyring

kennt-percona:
Issue:
If the keyring file is located in the datadir (or in a subdirectory
of the datadir), the SST will fail.  The reason is that the donor's
keyring file will be copied to the directory indicated in the
keyring-file-data option and will be deleted by the joiner (when
the joiner clears its datadir).

Solution:
Do not delete the keyring file that is being transferred ('donor-keyring') and
do not delete the directory of the keyring file.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/289/

kbauskar:
LGTM

PR #455, opened 2017-05-11
- PXC#824: shutdown of pxc doesn't wait for applier thread to finish.

kbauskar:
When the node is gracefully shutdown it should wait for applier
  to finish the assigned work (this is how 5.6 operates).

  In recent ported 5.7 the flow didn't wait for applier to finish,
  instead notified applier about server shutdown.

  Applier on getting notification fails to apply the write-set.
  Failure to apply write-set is treated as catastrophic error
  there-by causing applier thread to abort with fatal error
  about node-consistency compromised.

  Fix:
  ---
  * shutdown flow should wait for applier to finish.
  * applier will not see new workload as the group communication
    channel has been disconnected but applier should finish up
    existing work. (this is now in sync with 5.6 logic).


kbauskar:
mtr run: http://jenkins.percona.com/job/pxc57.test.mtr/294/

PR #474, opened 2017-06-16
Fix PXC-827: MySQL GTID metadata not synced upon full SST

kennt-percona:
Issue:
The SST did not handle correctly the case when binlog names on the
donor/joiner are different (it was assumed they are the same).
On 5.6, with GTIDs enabled, this would lead to differences in the
gtid_executed table.  Note that 5.7 does not rely on the binlog,
but just restores the gtid_executed table.

Solution:
XtraBackup only copies the last binlog (to recover the GTIDs). So
we ensure that the binlog is renamed and copied correctly.  This
requires that we send over the name of the binlog on the donor. To
enable this, a change was made to how this data is sent over (instead
of a pure one-line data file, we send a cnf-like file that is parsed
on the joiner to recover the values).


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/

kbauskar:
LGTM

PR #485, opened 2017-07-14
Fixes for ProxySQL doc, PXC var descriptions and outdated limitation

azhebel:


rameshvs02 (approved):
proxysql-admin changes looks good.

michaelcoburn:
- [ ] DOC-531 please derive the example from a ProxySQL instance with proxysql-admin running because then COMMENT field will be populated
Thanks @azhebel

azhebel:
@michaelcoburn this example is in the Manual Configuration section. When I added nodes manually (according to the described procedure), the column was not populated.

michaelcoburn:
Hi @azhebel , not sure what you're saying? There is a column named COMMENT and I think we should put something in there.. proxysql-admin already does write READWRITE or some such information, for completeness we should fill this in even in manual configuration.

kbauskar (approved):

azhebel:
@michaelcoburn pls take a look at the page. There are two big sections:
**Automatic Configuration** - this is OUR recommended way, which uses our own `proxysql-admin` script to add all nodes (it is available only in Percona's fork of ProxySQL). The script, among other things, populates the `comment` column in `mysql_servers` table.
**Manual Configuration** - this is alternative. For example, if the user wants to run vanilla ProxySQL (without the `proxysql-admin` script), or whatever other reason they might have to do everything manually, maybe to understand each step in detail rather than rely on some script. Here the procedure tells to add nodes as follows:
> INSERT INTO mysql_servers(hostgroup_id, hostname, port) VALUES (0,'192.168.70.61',3306);

In this case, the `comment` column remains empty.

PR #488, opened 2017-07-21
- PXC#839: GTID inconsistency if user set gtid_next="UUID:seqno"

kbauskar:
Problem:
  -------

  - If user assign an event with explict gtid using gtid_next="UUID:seqno"
    then originating node will have event recorded with said gtid
    but replicating node will continue to use automatic gtid.
  - This gap is because replicating write-set doesn't have explict
    GTID event appended to it.

  Solution:
  ---------

  Starting MySQL-5.7, MySQL appends GTID directly to binlog during commit
  time. PXC/Galera create replication write-set during pre-commit time.
  During this pre-commit time GTID event is not present in binlog and so
  GTID event needs to be explictly appended to replication write-set.

  If gtid_next = AUTOMATIC this is not needed as it is auto-managed
  by each node local counter but if user set gtid explictly then
  appending gtid event is must.

  (Fix for this was present in PXC-5.7.16 but got accidentally removed
   in latter release during upstream merge).

### DIFF
--- a/sql/wsrep_binlog.cc
+++ b/sql/wsrep_binlog.cc
@@ -167,20 +167,21 @@ static int wsrep_write_cache_once(wsrep_t*  const wsrep,
 
     uint length(my_b_bytes_in_cache(cache));
 
-    /* Start 5.7 MySQL server will write GTID directly to binlog file.
-    This means GTID event is not present in the case if GTID is set explictly.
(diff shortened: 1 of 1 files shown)


kbauskar:
http://jenkins.percona.com/job/pxc57.test.mtr/313/

PR #494, opened 2017-08-02
- PXC#842: Deadlocks and duplicate Entries Load Data Infile

kbauskar:
Problem:
  -------

  * Load Data Infile (LDI) with log-bin=0 fails.
  * LDI with > 10K rows is split into mini transactions, each having
    10K rows (except the last bucket). Each mini transaction is committed
    independently once the bucket is full.
    A new transaction is started before processing next set of bucket.
  * New transaction should be started after the transaction is committed
    in wsrep and innodb world but the exsiting flow tried start new
    transaction before innodb commit was complete.
  * Recent change for performance optimization has tweaked innodb
    code to invoke post-commit that committed newly started transaction.
    In turn followup bucket is processed without active transaction.

  Solution:
  ---------

  * Ensure that new transaction starts only after the existing mini
    transaction is committed in wsrep and innodb world.


kbauskar:
http://jenkins.percona.com/job/pxc57.test.mtr/316/

PR #497, opened 2017-08-03
- PXC#841: DDL is replicated even when sql_log_bin=0, and GTID is

kbauskar:
incremented in remote node

  sql_log_bin=0 will stop generation of binary logs data.
  In turn it will also stop replication and GTID increment.

  PXC executes DDL through TOI and DML through normal replication.
  Since binary log is not-generated DML replication was already blocked.
  DDL path didn't had a check for sql_log_bin.
  Added the needed check.


kbauskar:
LGTM
http://jenkins.percona.com/job/pxc56.test.mtr/405/

PR #506, opened 2017-08-14
PXC-825 : using the syslog option (for an SST) does not specify defau…

kennt-percona:
…lts-group suffix

Issue:
When using syslog (for an SST), the defaults-group-suffix was not included, thus
making it difficult to use one configuration file (for multiple servers).

Solution:
Include the defaults-group-suffix when calling xtrabackup.

Also, added some missing logging functions when using syslog.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/414/

kbauskar (approved):

PR #508, opened 2017-08-18
PXC-851: Avoid doing SST from lower version (5.7 node trying to get S…

kennt-percona:
…ST from 5.6)

Issue:
Currently, PXC does not check the versions between the donor and the joiner.
Thus, we will clean the datadir and then receive the donor's datadir which will
probably be incompatible (thus erasing the datadir for no good reason).

Solution:
Pass the version from the donor to the joiner and let the joiner reject
the SST if the version is a higher version.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/415/

kbauskar (approved):

PR #509, opened 2017-08-19
PXC-851: Avoid doing SST from lower version (5.7 node trying to get S…

kennt-percona:
…ST from 5.6)

Issue:
Currently, PXC does not check the versions between the donor and the joiner.
Thus, we will clean the datadir and then receive the donor's datadir which will
probably be incompatible (thus erasing the datadir for no good reason).

Solution:
Pass the version from the donor to the joiner and let the joiner reject
the SST if the version is a higher version.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1060,7 +1060,7 @@ normalize_version()
 # Compares two version strings
 # The first parameter is the version to be checked
 # The second parameter is the minimum version required
-# Returns 1 (failure) if $1 >= $2, 0 (success) otherwise
+# Returns 0 (success) if $1 >= $2, 1 (failure) otherwise
 check_for_version()
 {
     local local_version_str="$( normalize_version $1 )"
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/337/

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1546,6 +1550,33 @@ then
...
+            wsrep_log_error "Upgrade this node before joining the cluster."
+            wsrep_log_error "****************************************************** "
+            exit 2
+        fi
+
+        # Is the donor's pxc version < this node's pxc version?
+        if ! check_for_version $donor_version_str $local_version_str; then
+            wsrep_log_warning "WARNING: PXC is receiving an SST from a node with a lower version."
+            wsrep_log_warning "This node's PXC version is $local_version_str. The donor's PXC version is $donor_version_str."
+            wsrep_log_warning "Run mysql_upgrade to upgrade the data."
```
kbauskar: Run mysql_upgrade in non-cluster (standalone mode) to upgrade. Check the upgrade process here https://www.percona.com/doc/percona-xtradb-cluster/LATEST/howtos/upgrade_guide.html

kbauskar: I assume you have tried the upgrade process manually and ensured that post SST data-dir copy-over mysql_upgrade works ?

kbauskar (commented):
Please check the said comments.

kbauskar (approved):

PR #510, opened 2017-08-23
PXC-863: PXC 5.6, update PXB dependency to >= 2.4.3

dannyk81:
Update PXC 5.6 dependency of PXB to use `percona-xtrabackup-24` instead of `percona-xtrabackup`

Without this, it is impossible to install Percona XtraBackup 2.4, as it breaks dependcy of PXC 5.6 Server which currently depends on `percona-xtrabackup`

Related to https://github.com/percona/percona-xtradb-cluster/commit/cb14df8ebfaaeb2b2450f5825c189f717b918356 & https://bugs.launchpad.net/percona-xtradb-cluster/+bug/1596993


dannyk81:
Ping :) any chance to get a quick review on this?

kbauskar:
Thanks for the patch. Sorry for delay. Reviewer will look into this and will get back to you.

EvgeniyPatlan:
Thanks for the patch @dannyk81. Such change is possible but it requires proper testing. Right now we have already made changes for centos/rhel systems and from the upcoming release it would be possible to install  pxb-23 and pxb-24 together on the same system. Update-alternatives should help in configuring the needed version.
Right now I'm working on the same changes for deb packages.
And once again thank you for your help and for your patch.

dannyk81:
Thanks @EvgeniyPatlan, looking forward for this :)

Any ETA you can share?

EvgeniyPatlan:
@dannyk81 I hope it will be available in the upcoming release

PR #511, opened 2017-08-25
PXC-826: Multi-source replication + galera fails when using the same …

kennt-percona:
…user

Issue:
With the PXC node being a slave to 2 async masters, upon restarting the
PXC node, one of the slave replication threads will be locked.

Solution:
The problem was that we were using mysql_cond_signal() when changing
wsrep_ready to TRUE.  This would only wake up one of the threads. This
should be using mysql_cond_broadcast() to wakeup all threads waiting
on the condition variable when changing wsrep_ready.

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -788,12 +788,12 @@ out:
 
 void wsrep_ready_set (my_bool x)
 {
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/344/

kbauskar (approved):

PR #512, opened 2017-08-28
PXC-853: When nodes all go non-primary, PXC can recover, but does not…

kennt-percona:
… set wsrep_ready

Issue:
There are times, during the MTR run, when all the nodes can go NON-PRIMARY
due to timing issues.  The cluster can recover, but in this case the cluster
goes straight from OPEN -> DONOR/DESYNCED, and thus does not re-enable
wsrep_ready.

Solution:
Enable wsrep_ready whenever we go PRIMARY.

Also, test case fixes


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/345/

kbauskar (approved):

PR #520, opened 2017-09-05
Add systemd unitfile for debian systems

EvgeniyPatlan:


kbauskar:
I am not expert so just took a cursory look and found ok to proceed.
Final approval will come from @tplavcic.

build-ps/percona-xtradb-cluster.spec
```diff
@@ -1539,6 +1545,8 @@ fi
 %attr(755, root, root) %{_bindir}/wsrep_sst_mysqldump
 %attr(755, root, root) %{_bindir}/wsrep_sst_xtrabackup-v2
 %attr(755, root, root) %{_bindir}/wsrep_sst_rsync
+%attr(755, root, root) %{_bindir}/ps-admin
```
tplavcic: Why ps-admin?
1. If it's going to be shipped with PXC it needs changes there because it includes options to install TokuDB and RocksDB which are not part of PXC - and also other plugins need to be checked.
2. If it will be added to PXC currently I see it added only in `rpm` and not `deb` - why difference?

build-ps/debian/percona-xtradb-cluster-server-5.7.mysql.service
```diff
@@ -0,0 +1,23 @@
...
+User=mysql
+Group=mysql
+Type=forking
+TimeoutSec=0
+PermissionsStartOnly=true
+ExecStartPre=/usr/share/mysql/mysql-systemd-start pre
+ExecStart=/usr/sbin/mysqld --daemonize --pid-file=/var/run/mysqld/mysqld.pid $MYSQLD_OPTS $MYSQLD_RECOVER_START
+ExecStartPost=/usr/share/mysql/mysql-systemd-start post
+EnvironmentFile=-/etc/sysconfig/mysql
+LimitNOFILE = 5000
```
tplavcic, line 19: Is this OK for PXC? If not sure check with Krunal.

build-ps/debian/percona-xtradb-cluster-server-5.7.mysql.service
```diff
@@ -0,0 +1,23 @@
...
+Restart=on-failure
```
tplavcic, line 20: I'm not sure if this is ok for PXC. When talking with Raghu long before he mentioned that for PXC automatic restarts are not good.
Also for `RestartPreventExitStatus=1` I have reported an upstream bug and it seems that the server is restarted at least several times even if this option is set, check this: https://bugs.mysql.com/bug.php?id=87210
So also check what @kbauskar  thinks about this.

build-ps/ubuntu/percona-xtradb-cluster-server-5.7.mysql.service
```diff
@@ -0,0 +1,23 @@
...
+Group=mysql
+Type=forking
+TimeoutSec=0
+PermissionsStartOnly=true
+ExecStartPre=/usr/share/mysql/mysql-systemd-start pre
+ExecStart=/usr/sbin/mysqld --daemonize --pid-file=/var/run/mysqld/mysqld.pid $MYSQLD_OPTS $MYSQLD_RECOVER_START
+ExecStartPost=/usr/share/mysql/mysql-systemd-start post
+EnvironmentFile=-/etc/sysconfig/mysql
+LimitNOFILE = 5000
+Restart=on-failure
```
tplavcic, line 20: Same comment as for Debian.

tplavcic (changes requested):

build-ps/percona-xtradb-cluster.spec
EvgeniyPatlan: fixed

tplavcic (approved):

PR #525, opened 2017-09-15
Add missed wsrep_pos

EvgeniyPatlan:


kbauskar (approved):
looks good to me. if needed we can consult RE member or additional review.

PR #526, opened 2017-09-15
BLD-820: [PXC-5.7] SST fails when adding a 2nd node to the cluster

kennt-percona:
Issue:
We introduced an SST compatibility issue when making a change to the way
the donor and joiner nodes are communicating.

Solution:
The joiner-side code is now backwards-compatible, it will accept the SST
from an older donor version (with the bugs that were fixed by using the
new protocol).


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/433/

PR #531, opened 2017-09-20
[BLD-827]Fix unitfile

EvgeniyPatlan:


abychko (approved):
approved as temp solution, will be changed in next release

PR #533, opened 2017-09-21
[BLD-830] Added usage of mysql-helper on service start

EvgeniyPatlan:


abychko (approved):
looks good, it solves missing /var/run/mysqld on reboot

PR #535, opened 2017-09-21
[PXC-869] Add pxc_cluster_view.sql and fix path on Centos systems

EvgeniyPatlan:


dbadapt (changes requested):
we don't need to check-in 3 copies of the same pxc_cluster_view.sql file.   Please place this file in the /scripts directory

EvgeniyPatlan:
fixed

dbadapt (approved):
LGTM

PR #547, opened 2017-10-06
- PXC#874: PXC with async slave reporting Rolling back unfinished tra…

kbauskar:
…nsaction

  Issue:
  ------

  PXC acting as async slave will replicate an event that is not handled
  by wsrep-applier causing erroneous messages in the replicated node.

  setup: async-master (could be pxc node) replicating to
         async-slave (pxc node that further replicates the said traffic in
                      its cluster group channel).

  If async master has gtid_mode=OFF it will emit GTID_NEXT="ANONYMOUS" event.
  PXC node acting as async slave can handle this event as it is done by
  MySQL async replication applier/worker thread but when the same event is
  further replicated on group channel PXC applier thread doesn't have logic
  to handle this event causing erroneous messages.

  Solution:
  --------

  Add logic in wsrep-applier to handle said event.

### DIFF
--- a/mysql-test/suite/galera/r/galera_as_slave_without_gtid.result
+++ b/mysql-test/suite/galera/r/galera_as_slave_without_gtid.result
@@ -1,0 +1,22 @@
+# pxc-slave
+START SLAVE USER='root';
+Warnings:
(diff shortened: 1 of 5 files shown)


kbauskar:
http://jenkins.percona.com/job/pxc57.test.mtr/385/

PR #549, opened 2017-10-10
PXC-875:  [Warning] WSREP: Unsupported protocol downgrade: incrementa…

kennt-percona:
…l data collection disabled. Expect abort.

Issue:
Some tests would fail with the warning message: Unsupported protocol downgrade: incremental...
This was happening because the servers start with version -1 and then get established
at version 7.  During some tests, wsrep_provider is turned off which resets the version to -1,
but the internal variable for incremental data collection is still set to 1 and is enabled.
So upon wsrep restart, we get the error, since we see the version going from 7 to -1.

Solution:
Clear the use of the incremental data collection settings upon wsrep initialization.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/387/

kbauskar (approved):

PR #552, opened 2017-10-11
PXC-873: gtid/keyring stage: No such file or directory

kennt-percona:
Issue:
Text was cut-off, leaving to a partial error message

Solution:
Make the error text appear on the same line


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/390/

kbauskar (approved):

PR #553, opened 2017-10-12
- PXC#878: Threads stuck in permanent statistics state with innodb_th…

kbauskar:
…read_concurrency>0

  * Upstream merge accidentally got a wrong code that looks like a copy-paste
    mistake.

  * When the threads leaves InnoDB it should mark exit from innodb server
    concurrency. It was doing exactly opposite.


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc57.test.mtr/389/

PR #555, opened 2017-10-18
PXC-870: User without Create View privileges get "ERROR 1142" but VIE…

kennt-percona:
…W definition

         still gets replicated to other attached nodes!

Issue:
A user that doesn't have permission to perform certain DDL actions, will
see the command fail on their local node, but the command will be
replicated to other nodes where it will succeed.

Solution:
For some DDL actions, we were replicating than doing the access checks.
The fix is to do it in the proper order, access checks then replicate.

Note:
CVE-2017-15365

Also fixed the galera_toi_ddl_locking test case.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/397/

sql/sql_view.cc
```diff
@@ -432,6 +432,9 @@ bool mysql_create_view(THD *thd, TABLE_LIST *views,
     goto err;
 
   lex->link_first_table_back(view, link_to_local);
+
+  WSREP_TO_ISOLATION_BEGIN(WSREP_MYSQL_DB, NULL, NULL)
```
kbauskar, line 436: Just cross-check that if the existing code is invoked with last params as NULL then this is okay but if existing code is invoked with last 2 param or 1 param as table value then we need to get same string here.

WSREP_TO_ISOLATION_BEGIN(WSREP_MYSQL_DB, NULL, NULL)

* Not that this is wrong but a safe check as it can mess lot of things.
* Validate this check or point for each TO ISOLATION call added.

kbauskar (approved):

sql/sql_view.cc
kennt-percona: Most of the time it is called with (WSREP_MYSQL_DB, NULL, NULL).  The only time a parameter is passed is for CREATE_TRIGGER.

PR #556, opened 2017-10-18
- PXC#877: PXC hangs on what looks like an internal deadlock

kbauskar:
- PXC protects wsrep_xxxx state variables through a native mutex
    name LOCK_wsrep_thd.

  - This mutex should be held only while checking the state
    and should be released immediately.

  - In the said bug, this mutex was being held during complete
    cleanup. Other thread with conflicting lock, booted up
    at same time and tried and invoked kill action on the said
    thread. For invoking kill action it acquired transaction mutex
    and LOCK_wsrep_thd in respective order. Said thread already
    had LOCK_wsrep_thd and tried acquring transaction mutex
    leading deadlock.


kbauskar:
LGTM

http://jenkins.percona.com/job/pxc57.test.mtr/398/

PR #557, opened 2017-10-18
PXC-870: User without Create View privileges get "ERROR 1142" but VIE…

kennt-percona:
…W definition

         still gets replicated to other attached nodes!

Issue:
A user that doesn't have permission to perform certain DDL actions, will
see the command fail on their local node, but the command will be
replicated to other nodes where it will succeed.

Solution:
For some DDL actions, we were replicating than doing the access checks.
The fix is to do it in the proper order, access checks then replicate.

Note:
CVE-2017-15365

Also fixed the galera_toi_ddl_locking test case.


kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/447/

kbauskar:
LGTM

PR #558, opened 2017-10-19
- PXC#879: Wrong count if LOAD DATA used with GTIDs

kbauskar:
* wsrep splits LOAD DATA INFILE into 10K bucket (except last)
    and commits each bucket independently.

  * While committing such bucket if the bin-log
    reaches its max size then bin-log is switched over.
    Existing binlog is closed and GTIDs are saved.

    As part of this process commit action is invoked
    that causes clearing of registered storage engine
    plugin given that commit is last step.

  * But in this use-case it is not the last step but an intermediate
    commit of 10K bucket and so degistration of storage engine plugin
    shouldn't happen.

  * Also, LDI being a logical single command or big transaction
    all binlogs related information for it should be logged
    to same bin-log file. wsrep optimization for 10K commit
    shouldn't affect the split of binlog information.

  So we worked it out by skipping switch-over (or rotate as code
  name it) of binlogs if the wsrep is processing intermediate commit.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -9888,7 +9888,11 @@ commit_stage:
    */
   if (DBUG_EVALUATE_IF("force_rotate", 1, 0) ||
       (do_rotate && thd->commit_error == THD::CE_NONE &&
-       !is_rotating_caused_by_incident))
+       !is_rotating_caused_by_incident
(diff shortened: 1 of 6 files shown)


kbauskar:
http://jenkins.percona.com/job/pxc57.test.mtr/399/

PR #560, opened 2017-10-25
- PXC#883: ROLLBACK to SAVEPOINT does not rollback correctly on slaves

kbauskar:
* ROLLBACK to savepoint involves 2 main actions:
    - truncate binlog to the said position
    - rollback at innodb storage engine level.

    wsrep plugin has no role to play in savepoint rollback and
    call to wsrep plugin should be avoided in such case.

  * What would happen if we call wsrep plugin rollback action
    for savepoint rollback ?
    * wsrep plugin doesn't differentiate between normal and savepoint
      rollback and so will cause complete transaction rollback there-by
      discarding valid action even befor the savepoint.
    * this drawback currently is limited to log-bin=0 but ideal solution
      is to avoid calling wsrep plugin for savepoint rollback as it doesn't
      have role to play.

(cherry picked from commit 66b8ade02a2fcfa3c107e30355b2e8349287fcf9)

- Also pushing a left out piece of code for PXC#879
  that got removed during PXC#879 push.

### DIFF
--- a/mysql-test/suite/galera/r/galera_rollback_to_savepoint.result
+++ b/mysql-test/suite/galera/r/galera_rollback_to_savepoint.result
@@ -1,0 +1,41 @@
+#node-1
+use test;
+CREATE TABLE example (id INT NOT NULL PRIMARY KEY, name VARCHAR(64));
+INSERT INTO example VALUES(0, 'zero');
+BEGIN;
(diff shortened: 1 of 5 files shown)


kbauskar:
LGTM

kbauskar:
http://jenkins.percona.com/job/pxc57.test.mtr/404/
LGTM

PR #566, opened 2017-10-31
PXC-870, PXC-880

kennt-percona:
PXC-870: User without Create View privileges get "ERROR 1142" but VIEW definition still gets replicated to other attached nodes!
PXC-880: INSERT to table without PK possible with insertable VIEW when pxc_strict_mode=ENFORCING Edit

Additional tests added for PXC-870 to ensure that allowed actions are replicated.
Views with an underlying table that does not have a primary key, do not allow inserts if pxc_strict_mode is enabled.


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/408/

kbauskar (approved):

PR #567, opened 2017-11-01
- PXC#885: IST hangs when keyring_file_data is set

kbauskar:
* When a node joins the cluster depending on existing
    state it is synced using SST or IST.

  * If SST is done then keyring needs is sent from DONOR
    to JOINER.

  * If IST is done only write-sets are send. keyring
    sync is not needed.

  bug was caused as the JOINER flow expected keyring
  even in IST flow.


kbauskar:
LGTM
http://jenkins.percona.com/job/pxc57.test.mtr/411/
http://jenkins.percona.com/job/pxc57.test.mtr/412/

PR #568, opened 2017-11-09
PXC-889: Fallback when wsrep_provider cannot load the library allows …

kennt-percona:
…writes

Issue:
When we fail to load the WSREP provider (wsrep-provider option), due to a
problem with the shared library (either a code problem or the location is
not entered correctly), an error is logged, but the server is allowed to
start as a non-cluster node (and still allows writes).
This can lead to a problem if this is not detected, since this standalone
server will not represent the true cluster state.

Solution:
Have the node fail if the wsrep_provider fails to load.  Note that
the special case of an empty wsrep_provider will still continue to start.

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -461,12 +461,21 @@ int wsrep_init()
 
   if ((rcode= wsrep_load(wsrep_provider, &wsrep, wsrep_log_cb)) != WSREP_OK)
   {
-    if (strcasecmp(wsrep_provider, WSREP_NONE))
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/418/

kbauskar:
Approving based on 5.6-mtr run http://jenkins.percona.com/job/pxc56.clone/2169/

kbauskar (approved):

kennt-percona:
http://jenkins.percona.com/job/pxc56.test.mtr/456/

PR #569, opened 2017-11-09
PXC-889: Fallback when wsrep_provider cannot load the library allows …

kennt-percona:
…writes

Issue:
When we fail to load the WSREP provider (wsrep-provider option), due to a
problem with the shared library (either a code problem or the location is
not entered correctly), an error is logged, but the server is allowed to
start as a non-cluster node (and still allows writes).
This can lead to a problem if this is not detected, since this standalone
server will not represent the true cluster state.

Solution:
Have the node fail if the wsrep_provider fails to load.  Note that
the special case of an empty wsrep_provider will still continue to start.

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -904,13 +904,21 @@ int wsrep_init()
 
   if ((rcode= wsrep_load(wsrep_provider, &wsrep, wsrep_log_cb)) != WSREP_OK)
   {
-    if (strcasecmp(wsrep_provider, WSREP_NONE))
+    if (strlen(wsrep_provider) == 0)
     {
-      WSREP_ERROR("Failed to load wsrep_provider (%s). Error: %s (code: %d)."
-                  " Reverting to no provider.",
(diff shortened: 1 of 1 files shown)


kennt-percona:
http://jenkins.percona.com/job/pxc57.test.mtr/418/

sql/wsrep_mysqld.cc
```diff
@@ -904,14 +904,22 @@ int wsrep_init()
...
+      WSREP_ERROR("wsrep_load() failed to load the provider('%s'): %s (%d). Reverting to no provider.",
                   wsrep_provider, strerror(rcode), rcode);
-      strcpy((char*)wsrep_provider, WSREP_NONE); // damn it's a dirty hack
+
+      if (wsrep_provider) my_free((void *)wsrep_provider);
+      wsrep_provider = my_strdup(key_memory_wsrep, WSREP_NONE, MYF(0)); // damn it's a dirty hack
+
       return wsrep_init();
     }
+    else if (strcasecmp(wsrep_provider, WSREP_NONE))
```
kbauskar, line 917: Please check that unireg_abort doesn't affect data-dir such that next restart with valid configuration causes SST.

kbauskar (approved):

PR #570, opened 2017-11-28
PXC-890: PXC 5.7 MTR: ASAN failure for galera.mysql-wsrep#31

kennt-percona:
Issue:
ASAN reported an error accessing a variable in an object that had
been deleted.  This was due to a BF abort which caused the
rollbacker thread to maintain a reference to a THD object which
was deleted on another thread.

Note that this can occur with any test that causes a BF abort and
is then followed by a test that restarts the server.

Solution:
Have the rollbacker thread clear the reference to the THD object.


kennt-percona:
http://jenkins.percona.com/view/XtraDB-Cluster/job/pxc56.test.mtr/

kbauskar (approved):

PR #573, opened 2017-11-30
- PXC#895: Transaction lost after recovery

kbauskar:
If binlog is enabled then SE will persist redo at following stages:
    - during prepare
    - during binlog write
    - during innodb-commit
    3rd stage (fsync) can be skipped as transaction can be recovered
    from binlog.

    During 3rd stage, commit co-ordinates are also recorded in sysheader
    under wsrep placeholder. These co-ordinates are then used to stop
    binlog scan (in addition to get recovery position in case of node-crash).

    Since fsync of 3rd stage is delayed (as transaction can be recovered
    using binlog) it is possible that even though transaction is reported
    as success to end-user said co-ordinates are not persisted to disk.

    On restart, a prepare state transaction could be recovered and committed
    but since wsrep co-ordinates are not persisted a successfully committed
    transaction recovery is skipped causing data inconsistency from end-user
    application perspective.

    This behavior is now changed to let recovery proceed independent of
    wsrep co-ordinates and wsrep co-ordinates are updated to reflect
    the recovery committed transaction.

- PXC#897: gitd_executed is empty post wsrep_recover

  Don't spawn a new binlog file during wsrep-recovery. Why ?
  - Recovery flow is only going to read existing wsrep saved co-ordinate
    from sys_header. No other action is performed that needs binlogging.

  - Existing server flow looks at last binlog to set gtid_executed if server
    was shutdown abruptly. Newly created binlog will not have any such
    information and so restart post wsrep_recover will result in gtid_executed
    to be empty.

### DIFF
--- a/mysql-test/suite/galera/r/galera_trx_lost.result
+++ b/mysql-test/suite/galera/r/galera_trx_lost.result
@@ -1,0 +1,25 @@
+#node-2
+#shutdown node-2.
+#node-1
+select @@sync_binlog;
+@@sync_binlog
+1
+#node-1
+create table t1 (i int, primary key pk(i)) engine=innodb;
+insert into t1 values (1);
+insert into t1 values (2);
+insert into t1 values (3);
+set debug="+d,crash_before_trx_commit_in_memory";
+insert into t1 values (4);
+ERROR HY000: Lost connection to MySQL server during query
+#node-1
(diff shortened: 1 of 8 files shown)


kbauskar:
LGTM
http://jenkins.percona.com/job/pxc56.test.mtr/461/

PR #578, opened 2017-12-29
Do not alter gtid on startup

baloo:
When a node is joining a cluster and if root is not using 'auth_socket' plugin, the gtidset is locally changed on the local data (just out of xtrabackup) without running in the cluster (and thus not altering data in other members of the cluster).

This is effectively dangerous, because it alters the local gtid. If a slave comes to this master via async replication, it will have an extraneous component in the gtid set and won't be able to replicate from any other member of the cluster anymore.

I believe it is safer to remove any instruction that alters local data while not in the cluster altogether.

Please let me know what you think

### DIFF
--- a/build-ps/debian/extra/mysql-systemd-start
+++ b/build-ps/debian/extra/mysql-systemd-start
@@ -30,16 +30,6 @@ sanity () {
                 install -d -m 0750 -o mysql -g mysql ${MYSQLDATA}/mysql
 	fi
 
-	if [ ! "$(ls -A ${MYSQLDATA}/mysql)" ];
-	then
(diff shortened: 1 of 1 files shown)


baloo:
As a workaround, we just commented out `ExecStartPre=/usr/share/mysql/mysql-systemd-start pre` in the systemd service unit for now.

baloo:
My fix could be incorrect and it could be caused by this instead:
https://github.com/percona/percona-xtradb-cluster/blob/0b17da8262c024ff17e316625886dcb73dd8c815/packaging/deb-in/extra/mysql-helpers#L184-L190

I'll have a look tomorrow morning

PR #581, opened 2018-01-10
* PXC#915: InnoDB: Failing assertion: trx_is_registered_for_2pc(trx) in

kbauskar:
file ha_innodb.cc line 19502

  As part of PXC#841 we tried suppressing DDL/TOI replication
  if sql_log_bin = 0. While the approach worked it missed an important
  condition where-in the if the DDL statement modifies existing table
  say using ALTER TABLE, OPTIMIZE TABLE, etc.... then transaction
  for this operation is registered with MySQL 2PC co-ordinator
  that will eventually invoke sequence to prepare -> commit the given
  transaction.

  prepare stage will try to register wsrep-hton handler which is not
  expected for any DDL/TOI statement. This erroneous action eventually
  resulted in the said error/assert.


kbauskar:
LGTM
https://jenkins.percona.com/job/pxc56.test.mtr/482/

PR #588, opened 2018-01-16
Fix for DOC-612 by adding wsrep_last_applied variable

fiowro:


kbauskar:
Should this request go to release branch or trunk?
In past request specific to release goes to release branch and then release branch is folded to trunk.
I presume Alexey had to adjust some job to point to release branch so that docs are re-generated but we can re-check with @hrvojem

hrvojem:
If this is tied to a specific release item, like documenting a bug fix or a new feature that's coming with this release it should go to the release branch, otherwise trunk should/can be used (no need to wait for a DOC bug fix to be rolled out with a release).

kbauskar:
Thanks @hrvojem 
@fiowro This issue is part of newly introduced variable in an upcoming release. So can you port this patch to the release branch

PR #591, opened 2018-01-21
Fix for doc-613, a note mentioning the use of tmpdir variable on join…

fiowro:
…er node

### DIFF
--- a/doc/source/manual/xtrabackup_sst.rst
+++ b/doc/source/manual/xtrabackup_sst.rst
@@ -338,8 +338,14 @@ To disable initial SST timeout, set ``sst-initial-timeout=0``.
    :Example: /path/to/tmp/dir
 
 This option specifies the location for storing the temporary file
-where the transaction log is stored
+on a donor node where the transaction log is stored
 before streaming or copying it to a remote host.
+
+.. Note:: Starting from |PXC| :rn:`5.7.20-29.24` this option can be used on
+   joiner node also, to specify non-default location to receive
+   temporary SST files. This location must be large enough to hold the
+   contents of the entire database. If tmpdir is empty then default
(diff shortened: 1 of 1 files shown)


doc/source/manual/xtrabackup_sst.rst
```diff
@@ -338,9 +338,13 @@ To disable initial SST timeout, set ``sst-initial-timeout=0``.
...
 
 This option specifies the location for storing the temporary file
-where the transaction log is stored
+on a donor node where the transaction log is stored
 before streaming or copying it to a remote host.
 
+.. Note:: Starting from |PXC| 5.7.20-29.24 this option can be used on
+   joiner node also, to specify location for the received SST file other
+   than *datadir/.sst*.
+
```
kennt-percona: Maybe something like this would be better (this is temporary storage and must be large enough)
=====

to specify a location to receive temporary SST files other than *datadir/.sst*.  This location must be large enough to hold the contents of the entire database.

Also, if tmpdir is empty then *datadir/.sst* will be used.

doc/source/manual/xtrabackup_sst.rst
```diff
@@ -338,9 +338,15 @@ To disable initial SST timeout, set ``sst-initial-timeout=0``.
    :Example: /path/to/tmp/dir
 
 This option specifies the location for storing the temporary file
-where the transaction log is stored
+on a donor node where the transaction log is stored
 before streaming or copying it to a remote host.
 
+.. Note:: Starting from |PXC| 5.7.20-29.24 this option can be used on
```
hrvojem: you can use :rn:`5.7.20-29.24` so it links to the release notes

hrvojem:
LGTM

PR #592, opened 2018-01-22
fix for DOC-682: use mysql_upgrade in the upgrade procedure desctipti…

fiowro:
…on (in script messages - I think it is the only place saying about the previous upgrade procedure)

### DIFF
--- a/build-ps/debian/percona-xtradb-cluster-server-5.7.preinst
+++ b/build-ps/debian/percona-xtradb-cluster-server-5.7.preinst
@@ -157,7 +157,7 @@ case "$1" in
 	if [[ $currentver < $requiredver ]]; then
 	  echo -e "\n\n * Please execute the following operations to get the full cluster updated:"
 	  echo -e "\t - Upgrade each node to current PXC version (5.7.19+)"
-	  echo -e "\t\t mysql -uroot -p < /usr/share/mysql/pxc_cluster_view.sql"
+	  echo -e "\t\t mysql_upgrade -uroot -p --force"
 	  echo -e "\t - Restart each cluster node, one node at a time:"
 	  echo -e "\t\t service mysql restart\n\n"
         fi
--- a/build-ps/percona-xtradb-cluster.spec
+++ b/build-ps/percona-xtradb-cluster.spec
@@ -1324,7 +1324,7 @@ fi
 if [ "$UPDATE_REQUIRES" = "true" ] ; then
 echo " * Please execute the following operations to get the full cluster updated:"
 echo " - Upgrade each node to current PXC version (5.7.19+)"
-echo "       mysql -uroot -p < /usr/share/percona-xtradb-cluster/pxc_cluster_view.sql"
+echo "       mysql_upgrade -uroot -p --force"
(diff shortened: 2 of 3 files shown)


EvgeniyPatlan (approved):
lgtm if the instruction is correct

RoelVdP:
The `-s` option (https://dev.mysql.com/doc/refman/5.7/en/mysql-upgrade.html#option_mysql_upgrade_upgrade-system-tables) should not be used for `mysql_upgrade`; we want to check all tables not just system tables. However, this raises a different question; if no-one ever ran `mysql_upgrade` on PXC servers without `-s` (i.e. on all tables) in the past, then running it now 1) may be incorrect; i.e. with version updates one normally has to go to "latest supported of series > upgrade" etc. 2) may cause a major upstart delay on the server 3) any other unforeseen issues. Also, `--force` is missing etc. This PR needs 1) changes as described (IMHO) 2) more review by @kbauskar (this is the main point - developers have to sign off on "known good functionality"), 3) extensive testing by @rameshvs02 - though I am not sure we can cover all use case scenario's in QA for this, making #2 (proper dev review & signoff) all the more important.

`$MYSQLD_UPGRADE --force --socket=$UPGRADE_SOCKET --user=$WSREP_SST_OPT_USER --password=$WSREP_SST_OPT_PSWD ` is what Krunal had in the script.

RoelVdP:
Thank you for the updated fix. One after thought is this; does removing the `-s` somehow replicate the changes of `mysql_upgrade` to the tables to the cluster? Or, wording it another way; I **assume** that previously when we had `-s` in place, only system tables would be updated and that such system table changes would not replicate/sync out to other nodes. Whereas now we are removing `-s` (which would seem to be the correct action as all tables need `mysql_upgrade` run on them), but would the resulting table changes be replicated, thereby creating havoc on other nodes (not upgraded yet). This is all still besides the thought of "will it work in the first place?" given that `-s` has apparently never been used in cluster upgrades (why?) (and should it have been as we think it should)? This is a complex issue.

RoelVdP:
Backtracking with @rameshvs02 I found that this issue started at https://jira.percona.com/browse/PXC-869. Now, I am not sure how we came to the conclusion that the `/usr/share/mysql/pxc_cluster_view.sql` import needed to be changed to `mysql_upgrade`. This must have been some misunderstanding. Even in that ticket, both @hrvojem and @kbauskar execute mysql_upgrade (more on this in next comment).

RoelVdP:
https://www.percona.com/doc/percona-xtradb-cluster/LATEST/howtos/upgrade_guide.html mentions running `mysql_upgrade`, and it does not include `-s` so it seems we're fine from that perspective.That clears a few question of the table (all tables **are already upgraded**). It only leaves the question about the script (i.e. does it need to be run?) and `mysql_upgrade` which comes back to https://jira.percona.com/browse/BLD-835.

RoelVdP:
Updated https://jira.percona.com/browse/BLD-835 with final thoughts.

RoelVdP:
This PR is **not** to be merged, ref DOC-682 and BLD-835 for info.

PR #600, opened 2018-01-31
* PXC#2039: Node consistency compromized for simple

kbauskar:
INSERT INTO ... ON DUPLICATE KEY UPDATE workload

  - A high priority (read replicated) transaction can cause
    existing local transaction to get aborted.

  - Generally if a local transaction is aborted due to lock-conflict
    then the said transaction needs to be evaluated for certification
    conflict and accordingly decision to replay or abort needs to be
    called out.

    This logic went missing (from pre-commit hook) when replicate_pre_commit
    hook was split into 2 different actions (as part of performance).

  - Corrected the logic to handle the said condition in pre-commit hook.


kbauskar:
LGTM

https://jenkins.percona.com/job/pxc57.test.mtr/444/
https://jenkins.percona.com/job/pxc57.test.mtr/443/

PR #602, opened 2018-02-04
PXC-904 : Replication filters not working with account management sta…

kennt-percona:
…tements (CREATE USER)

Issue:
There is a different code path for account management statements
(such as CREATE USER, DROP USER, etc..).  For these cases, the replication
filters are checked only for the async slave thread.  The galera applier
threads, therefore, do not check the replication filters.  In addition,
the replication filter check occurs AFTER the statement is replicated
by galera.  This leads to the CREATE USER being galera replicated and
applied on the other galera nodes, but being blocked by the replication
filter on the async slave thread on the async slave node.

Solution
Move the galera replication code to occur AFTER the replication filter
check.  This will apply only to the account management statements to
avoid USER account inconsistency between galera nodes.

### DIFF
--- a/mysql-test/suite/galera/r/galera_repl_filtersA.result
+++ b/mysql-test/suite/galera/r/galera_repl_filtersA.result
@@ -1,0 +1,1753 @@
+# connection node_2 : starting async slave
+START SLAVE USER='root';
+Warnings:
+Note	1759	Sending passwords in plain text without SSL/TLS is extremely insecure.
+#
+# Test preparation
+#
+# connection node_1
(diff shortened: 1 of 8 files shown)


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/491/

kbauskar:
@kennt-percona 
Overall changes looks good. let's run this through QA. @rameshvs02 can you test this change?

rameshvs02:
@kbauskar sure!

kbauskar:
lgtm

PR #604, opened 2018-02-05
- PXC#2051: General code improvement in GTID event handling in PXC

kbauskar:
* GTID event captured as part of slave replication is appeneded
    to galera replicated write-set.
    Improved this part of the code to make all the internal flow
    consistent.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc57.test.mtr/448/

PR #607, opened 2018-02-06
Attempt to provide fix for DOC-686

fiowro:
describing the SST problem on rejoin on major and minor upgrades

### DIFF
--- a/doc/source/howtos/upgrade_guide.rst
+++ b/doc/source/howtos/upgrade_guide.rst
@@ -4,14 +4,16 @@
 Upgrading Percona XtraDB Cluster
 ================================
 
-This guide describes the procedure for upgrading |PXC| version 5.6
-to the latest version 5.7 without downtime (*rolling upgrade*).
-This means that replication between nodes will continue during the upgrade.
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. A "rolling upgrade" means
+there is no need to take down the complete cluster during the upgrade.
 
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+Both major upgrades (from 5.6 to 5.7 version) and minor ones (from 5.7.x
+to 5.7.y) can be done in this way. Rolling upgrades to 5.7 from versions
+older than 5.6 are not supported. Therefore if you are running |PXC| version
+5.5, it is recommended to shut down all nodes, then remove and re-create
+the cluster from scratch. Alternatively, you can perform a
+`rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
 and then follow the current procedure to upgrade from 5.6 to 5.7.
 
 The following documents contain details about relevant changes
@@ -24,6 +26,12 @@ mentioned in these documents when upgrading to |PXC| 5.7.
 * `Upgrading MySQL <http://dev.mysql.com/doc/refman/5.7/en/upgrading.html>`_
 
 * `Upgrading from MySQL 5.6 to 5.7 <http://dev.mysql.com/doc/refman/5.7/en/upgrading-from-previous-series.html>`_
+
+.. contents::
+   :local:
+
+Major upgrade
+-------------
 
 To upgrade the cluster, follow these steps for each node:
 
@@ -67,7 +75,7 @@ To upgrade the cluster, follow these steps for each node:
 #. Back up :file:`grastate.dat`, so that you can restore it
    if it is corrupted or zeroed out due to network issue.
 
-#. Start the node outside the cluster
+#. Start the node outside the cluster (in standalone mode)
    by setting the :variable:`wsrep_provider` variable to ``none``.
    For example:
 
@@ -77,6 +85,12 @@ To upgrade the cluster, follow these steps for each node:
 
    .. note:: As of |PXC| 5.7.6, the ``--skip-grant-tables`` option
       is not required.
+
+   .. note:: To prevent any users from accessing this node while performing
+      work on it, you may add `--skip-networking <https://dev.mysql.com/doc/refman/5.7/en/server-options.html#option_mysqld_skip-networking>`_
+      to the startup options and use a local socket to connect, or
+      alternatively you may want to divert any incoming traffic from your
+      application to other operational nodes.
 
 #. Open another session and run ``mysql_upgrade``.
 
@@ -104,9 +118,8 @@ To upgrade the cluster, follow these steps for each node:
       upon encountering a failed validation.
 
       If you are not sure, it is recommended to first start the node
-      with the :variable:`pxc_strict_mode` variable set to ``PERMISSIVE``::
-
-       $ sudo mysqld --pxc-strict-mode=PERMISSIVE
+      with the :variable:`pxc_strict_mode` variable set to ``PERMISSIVE``
+      in the in the |MySQL| configuration file, :file:`my.cnf`.
 
       After you check the log for any experimental or unsupported features
       and fix any encountered incompatibilities,
@@ -114,6 +127,169 @@ To upgrade the cluster, follow these steps for each node:
 
        mysql> SET pxc_strict_mode=ENFORCING;
 
+      Also switch back to ``ENFORCING`` may be done by restarting the node
+      with updated :file:`my.cnf`.
+
 #. Repeat this procedure for the next node in the cluster
    until you upgrade all nodes.
 
+It is important that on rejoining, the node should synchronize using
+:term:`IST`. For this, it is best not to leave the cluster node being
+upgraded offline for an extended period. More on this below.
+
+When performing any upgrade (major or minor), :term:`SST` could
+be initiated by the joiner node after the upgrade if the server
+was offline for some time. After :term:`SST` completes, the data
+directory structure needs to be upgraded (using mysql_upgrade)
+once more time to ensure compatibility with the newer version
+of binaries.
+
+   .. note:: In case of :term:`SST` synchronization, the error log
+             contains statements like "Check if state gap can be
+             serviced using IST ... State gap can't be serviced
+             using IST. Switching to SST" instead of "Receiving
+             IST: ..." lines appropriate to  :term:`IST`
+             synchronization.
+
+Minor upgrade
+-------------
+
+To upgrade the cluster, follow these steps for each node:
+
+1. Make sure that all nodes are synchronized.
+
+#. Stop the ``mysql`` service:
+
+   .. code-block:: bash
+
+      $ sudo service mysql stop
+
+#. Upgrade |PXC| and Percona XtraBackup packages.
+   For more information, see :ref:`install`.
+
+   For example, if you have Percona software repositories configured,
+   you might use the following commands:
+
+   * On CentOS or RHEL:
+
+     .. code-block:: bash
+
+        $ sudo yum update Percona-XtraDB-Cluster-57
+
+   * On Debian or Ubuntu:
+
+     .. code-block:: bash
+
+        $ sudo apt-get install --only-upgrade percona-xtradb-cluster-57
+
+#. In case of Debian or Ubuntu,
+   the ``mysql`` service starts automatically after install.
+   Stop the service:
+
+   .. code-block:: bash
+
+      $ sudo service mysql stop
+
+#. Back up :file:`grastate.dat`, so that you can restore it
+   if it is corrupted or zeroed out due to network issue.
+
+#. Start the node outside the cluster (in standalone mode)
+   by setting the :variable:`wsrep_provider` variable to ``none``.
+   For example:
+
+   .. code-block:: bash
+
+      sudo mysqld --skip-grant-tables --user=mysql --wsrep-provider='none'
+
+   .. note:: As of |PXC| 5.7.6, the ``--skip-grant-tables`` option
+      is not required.
+
+   .. note:: To prevent any users from accessing this node while performing
+      work on it, you may add `--skip-networking <https://dev.mysql.com/doc/refman/5.7/en/server-options.html#option_mysqld_skip-networking>`_
+      to the startup options and use a local socket to connect, or
+      alternatively you may want to divert any incoming traffic from your
+      application to other operational nodes.
+
+#. Open another session and run ``mysql_upgrade``.
+
+#. When the upgrade is done, stop the ``mysqld`` process.
+   You can either run ``sudo kill`` on the ``mysqld`` process ID,
+   or ``sudo mysqladmin shutdown`` with the MySQL root user credentials.
+
+   .. note:: On CentOS, the :file:`my.cnf` configuration file
+      is renamed to :file:`my.cnf.rpmsave`.
+      Make sure to rename it back
+      before joining the upgraded node back to the cluster.
+
+#. Now you can join the upgraded node back to the cluster.
+
+   In most cases, starting the ``mysql`` service
+   should run the node with your previous configuration::
+
+    $ sudo service mysql start
+
+   For more information, see :ref:`add-node`.
+
+   .. note:: As of version 5.7,
+      |PXC| runs with :ref:`pxc-strict-mode` enabled by default.
+      This will deny any unsupported operations and may halt the server
+      upon encountering a failed validation.
+
+      If you are not sure, it is recommended to first start the node
+      with the :variable:`pxc_strict_mode` variable set to ``PERMISSIVE``
+      in the in the |MySQL| configuration file, :file:`my.cnf`.
+
+      After you check the log for any experimental or unsupported features
+      and fix any encountered incompatibilities,
+      you can set the variable back to ``ENFORCING`` at run time::
+
+       mysql> SET pxc_strict_mode=ENFORCING;
+
+      Also switch back to ``ENFORCING`` may be done by restarting the node
+      with updated :file:`my.cnf`.
+
+#. Repeat this procedure for the next node in the cluster
+   until you upgrade all nodes.
+
+Dealing with IST/SST synchronization while upgrading
+----------------------------------------------------
+
+It is important that on rejoining, the node should synchronize using
+:term:`IST`. For this, it is best not to leave the cluster node being
+upgraded offline for an extended period. More on this below.
+
+When performing any upgrade (major or minor), :term:`SST` could
+be initiated by the joiner node after the upgrade if the server
+was offline for some time. After :term:`SST` completes, the data
+directory structure needs to be upgraded (using mysql_upgrade)
+once more time to ensure compatibility with the newer version
+of binaries.
+
+   .. note:: In case of :term:`SST` synchronization, the error log
+             contains statements like "Check if state gap can be
+             serviced using IST ... State gap can't be serviced
+             using IST. Switching to SST" instead of "Receiving
+             IST: ..." lines appropriate to  :term:`IST`
+             synchronization.
+
+The following additional steps should be made to upgrade the data
+directory structure after :term:`SST` (after the normal major or
+minor upgrade steps):
+
+1. shutdown the node that rejoined the cluster using :term:`SST`:
+
+   .. code-block:: bash
+
+      $ sudo service mysql stop
+
+#. restart the node in standalone mode by setting the
+   :variable:`wsrep_provider` variable to ``none``, e.g.:
+
+   .. code-block:: bash
+
+      sudo mysqld --skip-grant-tables --user=mysql --wsrep-provider='none'
+
+#. run ``mysql-upgrade``
+
+#. restart the node in cluster mode (e.g by executing ``sudo service mysql
+   start`` and make sure the cluster joins back using :term:`IST`.


doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
-to the latest version 5.7 without downtime (*rolling upgrade*).
-This means that replication between nodes will continue during the upgrade.
-
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. Rolling upgrade means that
+replication between nodes will continue during the upgrade.
```
kbauskar: Rolling upgrade means no need to take down complete cluster during the upgrade.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,17 @@ To upgrade the cluster, follow these steps for each node:
 #. Repeat this procedure for the next node in the cluster
    until you upgrade all nodes.
 
+.. note:: It is important that when rejoining, the node should synchronize using
+   :term:`IST`. If for some reason :term:`SST` syncronization is used
+   instead, there is a problem, as the node will use data-directory from
```
kbauskar: Can you re-word to avoid highlighting this as a problem?
Instead, we need to indicate that if node rejoins please follow the extra step to ensure that the data-directory of the JOINER is compatible with the installed binaries.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,17 @@ To upgrade the cluster, follow these steps for each node:
...
+   the previous version, and newer version binaries would throw a lot of
+   errors about wrong structure. This :term:`SST`-related problem is
+   garanteed in case of major upgrade, and is possible in case of minor
+   upgrade also.
+
+   If this problem occurs, it can be solved with the following steps:
```
kbauskar: Rather than saying problem occurs let's simply put that as the user need to do it whenever the node joins using SST. (Always needed step).

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,17 @@ To upgrade the cluster, follow these steps for each node:
...
+.. note:: It is important that when rejoining, the node should synchronize using
+   :term:`IST`. If for some reason :term:`SST` syncronization is used
+   instead, there is a problem, as the node will use data-directory from
+   the previous version, and newer version binaries would throw a lot of
+   errors about wrong structure. This :term:`SST`-related problem is
+   garanteed in case of major upgrade, and is possible in case of minor
+   upgrade also.
+
+   If this problem occurs, it can be solved with the following steps: 
+   * shutdown node that rejoined cluster using SST
```
kbauskar: User will need real command just like up normal upgrade sequence. Hrvoje or Ramesh can help with them as they have tried this sequences.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,17 @@ To upgrade the cluster, follow these steps for each node:
...
+   * restart in standalone
+   * run mysql-upgrade
+   * restart in cluster mode. (ensure cluster joins back using IST).
+
```
kbauskar: Also, we should introduce this section as part of relevant steps. 
Currently, upgrade enlist 9 steps. These steps are post 9 where-in user check and run extra steps (10-13).

kbauskar (changes requested):

doc/source/howtos/upgrade_guide.rst
RoelVdP: With a rolling upgrade, the cluster remains active during the upgrade.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. Rolling upgrade means that
+replication between nodes will continue during the upgrade.
+
+As major upgrade (from 5.6 to 5.7 version), so minor one (from 5.7.x
+to 5.7.y)  can be done in this way. Rolling upgrade to 5.7 from versions
```
RoelVdP: The first sentence seems to have grammar issues - "As major upgrade (from 5.6 to 5.7 version)," - the meaning is not defined/stated?

doc/source/howtos/upgrade_guide.rst
RoelVdP: Idem. Always try and use positive language and to state things in a way that is positive. Avoid things like "throw a lot of errors" "wrong structure", "SST-related problem", etc.

For example;
"When performing any upgrade (major or minor), SST could be initiated by the JOINER node after the upgrade if the server was offline for some time. After SST completes, the data directory structure needs to be upgraded (using mysql_upgrade) once more to ensure compatibility with the higher version DONOR. For this, ..."

doc/source/howtos/upgrade_guide.rst
RoelVdP: As above

RoelVdP (changes requested):
Please request another review after changes are made. Thank you

doc/source/howtos/upgrade_guide.rst
rameshvs02: @kbauskar, Since server started with SST we need to mention only following steps.
run  mysql_upgrade
sudo service mysql restart

doc/source/howtos/upgrade_guide.rst
fiowro: Ok, that's why this is an "attempt to fix DOC-686" :)
There is a question about mentioning errors. I think we need to specify some indicator - a clear sign that additional steps are necessary. As far as I got into the topic, messages about SST appear in both variants (SST and IST)... they are different but still may confuse... so currently the most obvious indicator I know about is "a lot of errors". Is there any specific sign we can pay user's attention on?

doc/source/howtos/upgrade_guide.rst
fiowro: Hm... It's a typical construction, "as ... so", e.g. "as IST, so SST may occur". I'll change it to "both ... and ..." for better readability.

doc/source/howtos/upgrade_guide.rst
fiowro: Also, it seems like there is still a possibility, that node would join with SST for the second time. I've searched for instruction, which makes node for sure join with IST, and got one, but it it can spoil the whole howto (low-level things done by hands). So changing the phrase "ensure cluster joins back using IST" with some clear indicator (and, may be, external link?) sounds reasonable.

kbauskar: 1. lot of error is not a good sign. Especially with the minor upgrade. A user can look out for the messages log file that clearly indicates IST or SST. If needed I can pull out those messages

For SST:

2018-02-06T09:14:48.958792Z 2 [Note] WSREP: Check if state gap can be serviced using IST
2018-02-06T09:14:48.958804Z 2 [Note] WSREP: Local UUID: 00000000-0000-0000-0000-000000000000 != Group UUID: 2b1d7ac7-0b1e-11e8-bd3a-e3f0e97b7f5f
2018-02-06T09:14:48.958857Z 2 [Note] WSREP: State gap can't be serviced using IST. Switching to SST

For IST:

2018-02-06T09:16:57.581259Z 2 [Note] WSREP: Receiving IST: 9120 writesets, seqnos 72-9192
2018-02-06T09:16:57.581345Z 0 [Note] WSREP: Receiving IST...  0.0% (   0/9120 events) complete.

kbauskar (changes requested):

doc/source/howtos/upgrade_guide.rst
```diff
@@ -25,6 +27,7 @@ mentioned in these documents when upgrading to |PXC| 5.7.
 
 * `Upgrading from MySQL 5.6 to 5.7 <http://dev.mysql.com/doc/refman/5.7/en/upgrading-from-previous-series.html>`_
 
+
 To upgrade the cluster, follow these steps for each node:
 
 1. Make sure that all nodes are synchronized.
```
kbauskar: Make sure that all nodes are synchronized.
Select a node to upgrade. Stop the mysql service on the selected node.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,43 @@ To upgrade the cluster, follow these steps for each node:
...
+To upgrade data directory structure after :term:`SST`, following
+additional steps should be done: 
+
+11. shutdown node that rejoined cluster using :term:`SST`:
+   
+   .. code-block:: bash
+
+      $ sudo service mysql stop
+
+#. restart the node in standalone mode by setting the
```
kbauskar, line 285: bullet numbering went off.

kbauskar (approved):
Looks good. Suggest 2 small changes.
Please check with Roel.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
-
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. Rolling upgrade means no
+need to take down complete cluster during the upgrade.
+
+Both major upgrade (from 5.6 to 5.7 version) and minor one (from 5.7.x
```
RoelVdP: s/upgrade/upgrades/ s/one/ones/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
-This guide describes the procedure for upgrading |PXC| version 5.6
-to the latest version 5.7 without downtime (*rolling upgrade*).
-This means that replication between nodes will continue during the upgrade.
-
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. Rolling upgrade means no
```
RoelVdP: A "rolling upgrade" means there is no

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
+need to take down complete cluster during the upgrade.
```
RoelVdP: s/complete/the complete/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
+
+Both major upgrade (from 5.6 to 5.7 version) and minor one (from 5.7.x
+to 5.7.y)  can be done in this way. Rolling upgrade to 5.7 from versions
```
RoelVdP: s/)  c/) c/ (double space)

RoelVdP: s/Rolling upgrade/Rolling upgrades/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
+older than 5.6 is not supported. Therefore if you are running |PXC| version
```
RoelVdP: s/is not/are not/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
 #. Repeat this procedure for the next node in the cluster
    until you upgrade all nodes.
 
+It is important that when rejoining, the node should synchronize using
```
RoelVdP: s/when rejoining/on rejoining/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+:term:`IST`.
```
RoelVdP: s/./. For this, it is best not to leave the cluster node being upgraded offline for an extended period. More on this below./

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+
+When performing any upgrade (major or minor), :term:`SST` could
+be initiated by the joiner node after the upgrade if the server
+was offline for some time. After :term:`SST` completes, the data
+directory structure needs to be upgraded (using mysql_upgrade)
+once more time to ensure compatibility with the newer version
+of binaries.
+
+   .. note:: In case of :term:`SST` synchronization,  messages log
```
RoelVdP: s/synchronization,  messages/synchronization, the messages/

RoelVdP: Is the "messages log" the "error log"? If so, should it say "error log" instead? @kbauskar ?

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+             file contains statements like "Check if state gap can
+             be serviced using IST ... State gap can't be serviced
+             using IST. Switching to SST" instead of "Receiving
+             IST: ..." lines appropriate to  :term:`IST`
+             synchronization.
+
+
+To upgrade data directory structure after :term:`SST`, following
```
RoelVdP: s/upgrade data/upgrade the data/

RoelVdP: s/, following/, the following/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+additional steps should be done:
```
RoelVdP: s/done/made/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+
+11. shutdown node that rejoined cluster using :term:`SST`:
```
RoelVdP: s/shutdown node/shutdown the node/

RoelVdP: s/cluster/the cluster/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+    .. code-block:: bash
+
+       $ sudo service mysql stop
+
+#. restart the node in standalone mode by setting the
+   :variable:`wsrep_provider` variable to ``none``, e.g.:
+
+   .. code-block:: bash
+
+      sudo mysqld --skip-grant-tables --user=mysql --wsrep-provider='none'
```
RoelVdP, line 290: If there are customer client instances trying to access the IP of this server, will they be able to connect when the seerver is in this mode? That could be a security risk? @kbauskar thoughts?

doc/source/howtos/upgrade_guide.rst
```diff
@@ -117,3 +120,41 @@ To upgrade the cluster, follow these steps for each node:
...
+
+#. run ``mysql-upgrade``
+
+#. restart the node in cluster mode (e.g by executing ``sudo service mysql
+   start`` and make sure the cluster joins back using :term:`IST`.
```
RoelVdP, line 295: s/and make sure the cluster joins back using/ensuring that the node rejoins the cluster using/

RoelVdP (changes requested):

doc/source/howtos/upgrade_guide.rst
kbauskar: @RoelVdP "it is error log."

doc/source/howtos/upgrade_guide.rst
kbauskar: I would call it more like an inconsistency risk. If the server comes for few minutes and existing application connects to it and start firing workload that is risky.
(Of-course this problem is more generic and I recall a bug that was floating around it wherein a node of the cluster boots in standalone mode and application start firing workload w/o probing for cluster status).
While the generic problem is something for upstream to fix we can propose to use --skip-networking so that user/admin is forced to connect only through local socket (less likely option for remote application)

doc/source/howtos/upgrade_guide.rst
RoelVdP: Okay, let's use "error log" then as that is what people are used to. (@fiowro this applies to all documentation.)

doc/source/howtos/upgrade_guide.rst
RoelVdP: "To prevent any users from accessing this node while performing work on it, you may add [--skip-networking:LINK IT TO MYSQL UPSTREAM MANUAL?] to the startup options and use a local socket to connect, or alternatively you may want to divert any incoming traffic from your application to other operational nodes."

kbauskar:
@fiowro 
Seems like I can't add a comment on existing test. So adding it here
"$ sudo mysqld --pxc-strict-mode=PERMISSIVE" ... most of the user doesn't start the server this way. So we should suggest user to start the server with pxc-strict-mode by adding the set pxc_strict_mode=PERMISSIVE in my.cnf and after validation is done user can switch it back to ENFORCING through the command line or by restarting the node with updated my.cnf.

doc/source/howtos/upgrade_guide.rst
hrvojem: "Both major upgrade (from 5.6 to 5.7 version) and minor one (from 5.7.x
+to 5.7.y)"

Isn't this a bit complicated procedure to follow for minor version upgrade, yum/apt upgrade should work instead of manually stopping/upgrading each node.

kbauskar: @hrvojem 

If I understand this correctly then only following 2 commands can be replaced
$ sudo apt-get remove percona-xtrabackup* percona-xtradb-cluster*
$ sudo apt-get install percona-xtradb-cluster-57

with
sudo apt-get upgrade percona-xtradb-cluster-57

(Given your expertise in this field we can simply go ahead with your suggestion).

But need to stop the server and need to run mysql-upgrade in standalone is needed. Right?

doc/source/howtos/upgrade_guide.rst
fiowro: Hm... as far as I understand, it is still possible that node will continue to attempt SST rejoins.  "ensuring" may be understood as related to "sudo service mysql start", which looks to be not correct... This "make sure" appeals to previous comment (lines in the error log). If speaking about real instruction to "ensure", I've found only low-level manual operations (file copy, etc) to avoid SST, so perhaps it's not a good place for such details...

doc/source/howtos/upgrade_guide.rst
hrvojem: @kbauskar 

IMO, to avoid the confusion we should split the minor from major upgrade docs, either in separate page or just a separate section. 

> But need to stop the server and need to run mysql-upgrade in standalone is needed. Right?

We can advise on stopping the service (but upgrade process should take care of it automatically), yes running mysql_upgrade should be mentioned.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -4,14 +4,16 @@
...
-Rolling upgrade from earlier versions to 5.7 is not supported.
-If you are running |PXC| version 5.5, it is recommended to shut down all nodes,
-then remove and re-create the cluster from scratch.
-Alternatively, you can perform a `rolling upgrade from PXC 5.5 to 5.6 <https://www.percona.com/doc/percona-xtradb-cluster/5.6/upgrading_guide_55_56.html>`_,
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the latest 5.7 version. A "rolling upgrade" means
+there is no need to take down the complete cluster during the upgrade.
+
+Both major upgrades (from 5.6 to 5.7 version) and minor ones (from 5.7.x
+to 5.7.y) can be done in this way. Rolling upgrades to 5.7 from versions
```
hrvojem, line 12: Add link from "major" and "minor" to specific content, so users don't have to scroll entire text to reach what they're interested in. Another option is to use 
>.. contents::
>  :local:

kbauskar:
Re-reviewed and look good now.
@RoelVdP Please take a final look.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -104,16 +118,178 @@ To upgrade the cluster, follow these steps for each node:
...
 
        mysql> SET pxc_strict_mode=ENFORCING;
 
+      Also switch back to ``ENFORCING`` may be done by restarting the node
+      with updated :file:`my.cnf`.
+
 #. Repeat this procedure for the next node in the cluster
    until you upgrade all nodes.
 
+Dealing with IST/SST synchronization while upgrade
```
RoelVdP: s/upgrade/upgrading/

doc/source/howtos/upgrade_guide.rst
```diff
@@ -104,16 +118,178 @@ To upgrade the cluster, follow these steps for each node:
...
+--------------------------------------------------
+
+It is important that on rejoining, the node should synchronize using
```
RoelVdP: Is this a repeat/duplication of line 136 and onwards?

doc/source/howtos/upgrade_guide.rst
RoelVdP: Perhaps something like 
s/and make sure the cluster joins back using :term:`IST`./and check if the node rejoins the cluster using :term:`IST`. If the node rejoins the cluster using SST, please refer to the SST related instructions above./

RoelVdP (approved):
Few outstanding/new comments, but with those addressed ok for me to merge. Assumptions made; that @kbauskar checked PXC functionality logic. And @rameshvs02 could you please test the reworded procedure in low-skill-user-mode to ensure any of us did not miss anything.

PR #608, opened 2018-02-14
- PXC#2054: Redo-log optimized DDL operation causing SST to fail

kbauskar:
* MySQL-5.7 introduced REDO-log optimized DDL (CREATE INDEX using
    sorted build) where-in all modified pages are not flush to the disk.

  * To protect against this change XB advise invoking backup with
    lock-ddl there-by blocking DDL during SST.

  * PXC currently has blocking DDL problem where-in a blocked DDL
    can cause DML to block too. (This is an existing PXC issue).


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/452/

PR #609, opened 2018-02-15
PXC-2043 : if pv not installed, sst script still tries to use it and …

kennt-percona:
…fails

Issue
When using "progress" or "rlimit" options in the [sst] section, the wsrep_sst_xtrabackup-v2 script
will fail because it tries to use the pcmd variable (which has been set to "").

Solution:
If pv cannot be found in the path, clear the progress and rlimit options.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/495/

PR #619, opened 2018-03-01
PXC-2059 : Create trigger is failing on PXC 5.6.29 node with error th…

kennt-percona:
…at binary logging is enabled but log_bin is OFF

Issue:
WSREP emulates the binlog when the real binlog is disabled.  Thus the requirement
for CREATE TRIGGER requires the SUPER privilege (or trust_function_creators) all
the time when WSREP is used.

Solution:
Change the error message to reflect the fact that the SUPER privilege is required.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/464/

PR #622, opened 2018-03-05
5.7 doc 691

fiowro:
Changed structure to contain SSL Automatic Configuration first, following by Generating Keys and Certificates, and Enabling Encryption [manually]. With some additional fixes to make restructured doc more understandable


doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+
+|PXC| includes the :variable:`pxc-encrypt-cluster-traffic` variable
+that enables automatic configuration of SSL encryption.
+By default, it is disabled, meaning that you need to configure SSL manually
+if you want to encrypt SST and IST traffic, and internal communication.
+
+This variable cannot be changed on the command line during runtime.
+To enable automatic configuration of SSL encryption,
+set ``pxc-encrypt-cluster-traffic=ON`` in the :file:`my.cnf` file
+and restart the cluster.
```
kennt-percona: Should mention that the pxc-encrypt-cluster-traffic option is in the [mysqld] section.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+There are two approaches to configure encryption: automatic and manual.
+
+.. _ssl-auto-conf:
+
+SSL Automatic Configuration
+===========================
+
+|PXC| includes the :variable:`pxc-encrypt-cluster-traffic` variable
+that enables automatic configuration of SSL encryption.
+By default, it is disabled, meaning that you need to configure SSL manually
```
kbauskar: "By default ...." statement contradict presence of the said variable.
So this statement (though existing) in new context should be changed.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+if you want to encrypt SST and IST traffic, and internal communication.
+
+This variable cannot be changed on the command line during runtime.
+To enable automatic configuration of SSL encryption,
+set ``pxc-encrypt-cluster-traffic=ON`` in the the ``[mysqld]`` section of
+the :file:`my.cnf` file, and restart the cluster.
+
+Automatic configuration of the SSL encryption needs key and certificate files.
```
kbauskar: Starting 5.7 key and cert are auto-generated by MySQL so this mode is like single step option. Let's keep it that simple.

It is like by default it would use the auto-generate cert but if user want user can override them by specifying the said options under mysqld section.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+Creation of these files is covered by the :ref:`generate-keys-certs` section.
+Necessary key and certificate files are first searched at the ``ssl-ca``,
+``ssl-cert``, and ``ssl-key`` options under ``[mysqld]``.
+If these options are not set, it then looks in the data directory for
+:file:`ca.pem`, :file:`server-cert.pem`, and :file:`server-key.pem` files.
+
+.. note:: The ``[sst]`` section is not searched.
+
+If all three files are found, they are used to configure encryption for SST
```
kbauskar: Again, since encrpyt=4 is not known to user (it now appears) in text below we should simply explain that enabling this option will 
* encrypt SST traffic
* encrypt IST traffic
* encrypt replication traffic.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+using XtraBackup (with ``encrypt=4`` mode, as explained in :ref:`xtrabackup`
+subsection).
+If any of the files are missing, a fatal error is generated.
+
+The following settings are applied (and overridden if necessary)::
```
kbauskar: "The following settings are ..." could be reworded something on this line

"Setting pxc-encrypt-cluster-traffic=ON is like setting following things in my.cnf"

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,6 +12,52 @@ This includes traffic between client applications and cluster nodes,
...
+ ssl-ca=ca.pem
+ ssl-cert=server-cert.pem
+
+For :variable:`wsrep_provider_options`, only the mentioned options are affected
+(``socket.ssl_key``, ``socket,ssl_cert``, and ``socket.ssl_ca``),
+the rest are not modified.
+
 .. _generate-keys-certs:
 
 Generating Keys and Certificates
```
kbauskar: "To encrypt |PXC| traffic, you must generate the following sets of files:" ....

It is not must to generate these key and certs. 

A story is like if the user
"Starting mysql-5.7, mysql will generate default key and cert and place them in data-directory. If user want to over-ride these certificates and place generate the new certificates then user can do so using the process laid below."

kbauskar: In general please reduce the spacing of this "Generating Keys and Certificates" section. It seems like lot of things are spaced with 1.5 lines para or even more (especially commands + The following files are required:.... section)

doc/source/security/encrypt-traffic.rst
```diff
@@ -134,12 +180,15 @@ The following files are required:
...
+
 .. _enable-encryption:
 
-Enabling Encryption
-===================
+Enabling Encryption Manually
+============================
 
-To enable encryption, you need to specify the location
+To enable encryption manually, you need to specify the location
```
kbauskar: In this section we are introducing a new term name ":ref:`encrypt-client-server`" ... but then following text re-name it as "Encrypting Database Traffic"...

Also, encrypting database traffic is more generic ... the said setting is only encrypting client-server communication.

kbauskar: Move the SST section below IST section.

kbauskar (changes requested):

doc/source/security/encrypt-traffic.rst
kennt-percona: I agree with Krunal here. We need to make it clear what the automatic configuration (pxc-encrypt-cluster-traffic) does.  It reuses the client SSL settings, but does not force the use of client SSL.  So it would be good to make that difference clear.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,19 +12,67 @@ This includes traffic between client applications and cluster nodes,
...
+There are two approaches to configure encryption: automatic and manual.
+
+.. _ssl-auto-conf:
+
+SSL Automatic Configuration
+===========================
+
+|PXC| includes the :variable:`pxc-encrypt-cluster-traffic` variable that
+enables automatic configuration of SSL encryption.
+
```
kennt-percona: Maybe we can make this clearer by making this into a numbered list of things to do
(1) create certs
(2) enable pxc-encrypt-cluster-traffic

doc/source/security/encrypt-traffic.rst
```diff
@@ -281,7 +334,52 @@ Make sure that the following options are set in :file:`my.cnf` on all nodes:
    if keyring configuration across nodes is different.
 
 The following SST methods are available:
-``rsync``, ``mysqldump``, and ``xtrabackup``.
+``xtrabackup``, ``rsync``, and ``mysqldump``.
```
kennt-percona: This is a comment about the keyring_file information above.   It's a little misleading.

If the keyring is used, then SST encryption is required.
If SST encryption is used, there is no keyring requirement.

It sounds like using the keyring is required for SST encryption, which is not true.
So the section should say something like, "if the keyring_file plugin is being used then SST encryption is required"

doc/source/security/encrypt-traffic.rst
fiowro: Nice idea!

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,19 +12,84 @@ This includes traffic between client applications and cluster nodes,
...
+
+.. _creating-certs:
+
+Creating Certificates
+---------------------
+
+Automatic configuration of the SSL encryption needs key and certificate files.
+These files are auto-generated by |MySQL| by default, and no manual operations
+are needed to use this default variant. 
+
```
rameshvs02: If we use auto-generate certificates from mysql, pxc node will terminate with following error
```
2018-03-22T05:33:30.274087Z 0 [Note] WSREP: gcomm: connecting to group 'cluster_one', peer '192.168.100.10:,192.168.100.20:,192.168.100.30:'
2018-03-22T05:33:30.276357Z 0 [ERROR] WSREP: handshake with remote endpoint ssl://192.168.100.20:4567 failed: asio.ssl:67567722: 'block type is not 01' ( 67567722: 'error:0407006A:rsa routines:RSA_padding_check_PKCS1_type_1:block type is not 01')
2018-03-22T05:33:30.276576Z 0 [ERROR] WSREP: handshake with remote endpoint ssl://192.168.100.10:4567 failed: asio.ssl:67567722: 'block type is not 01' ( 67567722: 'error:0407006A:rsa routines:RSA_padding_check_PKCS1_type_1:block type is not 01')
2018-03-22T05:33:30.279618Z 0 [Note] WSREP: SSL handshake successful, remote endpoint ssl://192.168.100.30:33532 local endpoint ssl://192.168.100.30:4567 cipher: AES128-SHA compression: none
2018-03-22T05:33:30.279748Z 0 [Note] WSREP: SSL handshake successful, remote endpoint ssl://192.168.100.30:4567 local endpoint ssl://192.168.100.30:33532 cipher: AES128-SHA compression: none
2018-03-22T05:33:30.279943Z 0 [Note] WSREP: (8dd3a4df, 'ssl://0.0.0.0:4567') connection established to 8dd3a4df ssl://192.168.100.30:4567
```
Please add one note as;
```
Note

You must use the same key and certificate files on all nodes.
```

doc/source/security/encrypt-traffic.rst
fiowro: So in this case it's better to make such a note, and remove the "manually" word from the "Generating Keys and Certificates Manually" title (user may need information from this section to populate auto-generated certs among all nodes).

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,19 +12,86 @@ This includes traffic between client applications and cluster nodes,
...
+enables encryption of the SST, IST, and replication traffic). If any of the
+files are missing, a fatal error is generated.
+
+.. _enabling_encrypt-cluster-traffic:
+
+Enabling :variable:`pxc-encrypt-cluster-traffic`
+------------------------------------------------
+
+|PXC| includes the :variable:`pxc-encrypt-cluster-traffic` variable that
+enables automatic configuration of SSL encryption.
```
kbauskar: SSL encryption there-by encrypting SST, IST and replication traffic.

doc/source/security/encrypt-traffic.rst
```diff
@@ -12,19 +12,86 @@ This includes traffic between client applications and cluster nodes,
    :local:
    :depth: 1
 
+There are two approaches to configure encryption: automatic and manual.
```
kbauskar: There are 4 kind of traffic
- client-server
- sst
- ist
- replication traffic

automatic and manaul is to faciliate last 3 category.
automatic configuration (pxc-encrypt-cluster-traffic) doesn't automate client-server traffic.

kbauskar (changes requested):

doc/source/security/encrypt-traffic.rst
fiowro: I've looked through this section once again and this comment looks unclear: the section contains no mentions of database traffic encryption anymore, and SST subsection is placed after encrypting replication traffic / IST, so I don't see what else can be fixed here.

doc/source/security/encrypt-traffic.rst
```diff
@@ -4,291 +4,231 @@
...
-* :variable:`socket.ssl_key`
-
-To set these options, use the :variable:`wsrep_provider_options` variable
-in the configuration file::
-
- wsrep_provider_options="socket.ssl=yes;socket.ssl_ca=/etc/mysql/certs/ca.pem;socket.ssl_cert=/etc/mysql/certs/server-cert.pem;socket.ssl_key=/etc/mysql/certs/server-key.pem"
-
-.. note:: You must use the same key and certificate files on all nodes,
-   preferably those used for :ref:`encrypt-client-server`.
+* :ref:`encrypt-replication`
```
kbauskar: * :ref:`encrypt-replication`
* :ref:`encrypt-ist`
(We are explicitly pointing out IST as a different mode of traffic so we need to ensure that both the heading are mentioned).

doc/source/security/encrypt-traffic.rst
```diff
@@ -362,92 +302,202 @@ If you do everything correctly,
...
+  (replicating transactions that execute on one node to all other nodes).
+* Incremental State Transfer (:term:`IST`) which
+  is copying only missing transactions from DONOR to JOINER node.
+* Service messages which ensure that all nodes are synchronized.
 
-* ``encrypt=1``, ``encrypt=2``, and ``encrypt=3`` have been deprecated.
+All this traffic is transferred via the same underlying communication channel
+(``gcomm``). Securing this channel will ensure that :term:`IST` traffic,
+write-set replication, and service messages are encrypted.
+(For IST, a separate channel is to be configured using the same configuration
```
kbauskar: s/For IST, a separate channel is to be configured using the same configuration/For IST, a separate channel is configured using the same configuration/g

kbauskar (approved):

PR #626, opened 2018-03-09
- PXC#2072: "flush table <table> for export" should be blocked with m…

kbauskar:
…ode=ENFORCING

  - FLUSH TABLE FOR EXPORT is meant to lock table and flush all related
    logs to prepare table for export.

  - This command is local to the node and is not replicated.

  - Just like FLUSH TABLE <table> for READ LOCK this command too causes
    table to get locked locally and this lock is non-preemptable.

  With that background is make sense to block this command with
  pxc-strict-mode=ENFORCING.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/467/

PR #627, opened 2018-03-12
PXC-2073 : PXC 5.7 : CREATE TABLE AS SELECT doesn't replicate if resu…

kennt-percona:
…lt set is empty

Issue
Running CTAS (Create Table As Select) with an empty source table on the last
node added to a cluster will fail (the command does not get replicated).

Solution
The wsrep_incremental_data_collection was not being set correctly.  This gets
set as part of a view callback and is dependent on the protocol version.
But the protocol version gets set AFTER the view callback is performed.
So if the node only receives one process conf change message, then it will
not set wsrep_incremental_data_collection correctly.

The solution is to set the protocol version before the view callback, so that
any version-dependent code in the view callback executes with the correct
version and not the default value (-1).


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/468/

kbauskar:
LGTM

PR #629, opened 2018-03-29
- PXC#2087: WSREP: BF lock wait long - deadlock

kbauskar:
- Said test-case is invoking COM_STMT_EXECUTE direct API level
    call to execute prepare statement.

  - If the local transaction is aborted by an high priority background
    transaction and there is no certification failure then such local
    transaction is replayed.

  - Existing flow, use to set the conflict_state=REPLAYED and on successful
    completion of the command, conflict_state was reset back to NO_CONFLICT.

  - This reset command block was not present if the prepare-statement
    is invoked through direct API call listed above.

  Of-course natural fix is to add the missing comamnd block but the reason
  why conflict_state was set to REPLAYED (after it is cleared to NO_CONFLICT)
  currently doesn't hold. Mentioned regression in commit message is not observed.

  Anyway, the said REPLAYED setting fix was a workaround fix.

### DIFF
--- a/sql/sql_prepare.cc
+++ b/sql/sql_prepare.cc
@@ -3910,7 +3910,21 @@ reexecute:
     thd->pop_reprepare_observer();
     observer_popped= true;
     (void)wsrep_replay_transaction(thd);
-    thd->wsrep_conflict_state= REPLAYED;
+    // This extra state was added as a workaround fix to help skip
+    // lex->unit->cleanup. While writing this commit-message, commenting
+    // the said conflict_state doesn't cause galera_transaction_replay
+    // test-case to fail which was suppose to fail and so the fix was
+    // added.
+    // Why it is wrong to set the conflict_state at this point ?
+    // As part of replay transaction flow, conflict_state is already reset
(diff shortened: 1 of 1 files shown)


kbauskar:
LGTM
mtr testing: https://jenkins.percona.com/job/pxc57.test.mtr/473/

s4sukan:
@kbauskar should we consider this was pushed to latest MySQL 5.7? We are facing this BF wait too long in our 5.6.39 version.

kbauskar:
@s4sukan 
While the error message looks same we will have to confirm if the original bug that you are referring too is the same too. But if you have reproducible test-case you can test the said fix and let us know if this solves the issue.

s4sukan:
@kbauskar , 
This bug is the what we are taking about, [MDEV-12837](https://jira.mariadb.org/browse/MDEV-12837). This was fixed in MariaDB 10.2.12, I wanted to see if this was fixed in PXC MySQL 5.7, I can confirm that issue is exists on PXC MySQL 5.6.39

PR #632, opened 2018-04-10
- PXC#2091: LOAD DATA INFILE fails with ERROR 1180 (HY000):

kbauskar:
wsrep_max_ws_rows exceeded even when wsrep_on is OFF

  - Galera limits number of rows that can be replicated as part of
    single transaction. This check was accidentlly enforced even
    when wsrep_on=off.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc56.test.mtr/510/

PR #634, opened 2018-04-12
- PXC#2092: RSU failure should be reported and command should be aborted

kbauskar:
- If RSU operation fails (say due to pending transaction or desync failure)
    then attached DDL statement should also fail.

  - RSU executes statement at said node in local fashion
    (not replicated)

  - If RSU fails but attached DDL statement is allowed to execute then
    it is like execution of DDL statement with replication turn-off for
    the given statement and it in turn can cause other replicating
    statement to stall and there-by halting complete cluster.
    [This is not what end-user expect].


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc56.test.mtr/511/

PR #637, opened 2018-05-07
- PXC#907: RSU wait for active commit connection timeout should be co…

kbauskar:
…nfigurable

  - While running RSU it is expected that user has isolated the node
    and there is no active traffic executing on the node.

  - RSU has check to ensure this. This check will cause RSU to wait for any
    active connection in commit state before starting RSU.

  - Check had timeout of 5 milli-seconds. This timeout may not be sufficient
    in some cases where-in COMMIT is taking longer.

  - In order to handle such scenarios this timeout should be made configurable.
    Said patch makes it configurable by introducting wsrep_RSU_commit_timeout.

  - wsrep_RSU_commit_timeout variable is dynamica variable with global scope
    and range of (5 mill-sec, 365 days).
    Unit of variable is in micro-secs so set accordingly.

  Note: RSU operation will not auto-stop node from receiving active traffic.
  This means, there could be a continous flow of active traffic while RSU
  continue to wait and that can result in RSU starvation.
  User is expected to block active RSU traffic while performing operation.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc56.test.mtr/514/

PR #638, opened 2018-05-09
BLD-1057 : PXC build is failing on Ubuntu Bionic 18.04

kennt-percona:
Issue
The build script fails to compile PXC on bionic

Solution
Add "bionic" to the list of ubuntu distros so that we can set the compile flags
correctly.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/486/

kennt-percona:
No changes needed for 5.6 (it builds fine on bionic), so this is 5.7 only.

PR #640, opened 2018-05-14
PXC-2098 : Retry of Store Proc causes assert with next_trx_id

kennt-percona:
Issue:
The assert is 5.7 only.  However, when the stored procedure
fails (due to LCF/BFA), the retry paths are different.
A BFA will cause the sproc to be retried, but not for an LCF.

Solution
Have the LCF and BFA error code paths work the same.  Make sure that
the BFA error case is also handled.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/516/

PR #645, opened 2018-05-24
- PXC#2113: Make PXC depends on PXB-2.4.11+

kbauskar:
- XB-2.4.11 is now compatible to support data-encryption-at-rest
    and vault-plugin.

  - In order to faciliate this new development, XB added a new param
    xtrabackup-plugin-dir. If this params is not specified during
    SST invocation of PXC it would use the default plugin directory.

    [Normally if XB is installed as package then plugin directory
     could be found but if XB is installed as binary then plugin
     directory is not found in case like this specifying the said
     param help XB to find the needed plugins]

   - With this we also enforce need of XB-2.4.11+ for future PXC releases.

  --------------------

  - XB-2.4.11 also introduced concept of transition-key that can be used
    to re-encrpyt the table/tablespaces/logs while taking backup (in our
    case SST).

  - this transition key help remove dependency on master-key there-by
    allowing different master-key on DONOR and JOINER and also
    probable mix-match of keyring_plugins (file and vault).

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -ue
+
 # Copyright (C) 2013 Percona Inc
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


kbauskar:
LGTM
mtr test: https://jenkins.percona.com/job/pxc57.test.mtr/496/

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1349,11 +1394,28 @@ then
...
     sst_info_file_path="${donor_tmpdir}/${SST_INFO_FILE}"
+    transition_key=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
     echo "[sst]" > "$sst_info_file_path"
     echo "galera-gtid=$WSREP_SST_OPT_GTID" >> "$sst_info_file_path"
     echo "binlog-name=$(basename "$WSREP_SST_OPT_BINLOG")" >> "$sst_info_file_path"
     echo "mysql-version=$MYSQL_VERSION" >> "$sst_info_file_path"
 
+
+    # append transition_key only if keyring is being used.
+    if [[ $keyring_plugin -eq 1 && $encrypt -ge 0 ]]; then
```
kennt-percona: Not sure why the "$encrypt -ge 0" test is needed here.  The code below tests for the encryption status.  Also the check here should be "$encrypt -gt 0"

kbauskar: Fixed. You are right. -ge 0 is not needed. Check for no presence of encrypted channel is now moved above this block.

PR #648, opened 2018-06-11
sst script fixed to respect SELinux

y-trudeau:
The way the current SST script is written, it is nearly impossible to enable SELinux because the "ss" command in the wait_for_listen function looks at all the running processes.  In SELinux audit log, you have tons of:

type=AVC msg=audit(1527792830.991:138): avc:  denied  { getattr } for  pid=3683 comm="ss" scontext=system_u:system_r:mysqld_t:s0 tcontext=system_u:system_r:syslogd_t:s0 tclass=process

type=SYSCALL msg=audit(1527792830.991:138): arch=c000003e syscall=0 success=yes exit=31 a0=5 a1=17de070 a2=fff a3=63 items=0 ppid=3680 pid=3683 auid=4294967295 uid=997 gid=995 euid=997 suid=997 fsuid=997 egid=995 sgid=995 fsgid=995 tty=(none) ses=4294967295 comm="ss" exe="/usr/sbin/ss" subj=system_u:system_r:mysqld_t:s0 key=(null)

The new script respects the "limits" set by SELinux and do no provoke any denied access, allowing a secure configuration of xtraDB cluster.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -931,8 +931,20 @@ wait_for_listen()
 
     for i in {1..300}
     do
-        ss -p state listening "( sport = :$PORT )" | grep -qE 'socat|nc' && break
         sleep 0.2
+        # List only our (mysql user) processes to avoid triggering SELinux
+        for cmd in $(ps -u $(id -u) -o pid,comm | sed 's/^\s*//g' | tr ' ' '|' | grep -E 'socat|nc')
+        do
+            pid=$(echo $cmd | cut -d'|' -f1)
+            ls -l /proc/$pid/fd > /tmp/socat_fd
+            cat /proc/$pid/net/tcp > /tmp/socat_tcp
+            # List the sockets of the pid
+            sockets=$(ls -l /proc/$pid/fd | grep socket | cut -d'[' -f2 | cut -d ']' -f1 | tr '\n' '|')
+            if [[ -n $sockets ]]; then
+                # Is one of these sockets listening on the SST port? If so, we need to break from 2 loops
+                grep -E "${sockets:0:-1}" /proc/$pid/net/tcp | grep "00000000:$(printf '%X' $PORT)" > /dev/null && break 2
+            fi
+        done
     done
 
     echo "ready ${HOST}:${PORT}/${MODULE}//$sst_ver"


y-trudeau:
Those 2 lines:

￼ +            ls -l /proc/$pid/fd > /tmp/socat_fd
￼ +            cat /proc/$pid/net/tcp > /tmp/socat_tcp

should be removed, they were used for debugging.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -931,8 +931,20 @@ wait_for_listen()
 
     for i in {1..300}
     do
-        ss -p state listening "( sport = :$PORT )" | grep -qE 'socat|nc' && break
         sleep 0.2
+        # List only our (mysql user) processes to avoid triggering SELinux
+        for cmd in $(ps -u $(id -u) -o pid,comm | sed 's/^\s*//g' | tr ' ' '|' | grep -E 'socat|nc')
```
kennt-percona, line 936: This looks for processes that belong to the mysql user, however this breaks our MTR runs, since we run multiple test cases concurrently so multiple SSTs may be happening at the same time.
This should probably look at child processes only (ps --ppid $$)

y-trudeau: If you have multiple SST running, they can't all be on the same TCP port so the following grep on line 945 will sort out.  I realize some changes are required to account for different listening IP instead of 0.0.0.0 and likely also for IPv6.

kennt-percona: Yes, but the port check comes after the pid listing.  What happens is that we get a list of processes from the ps, some of those processes are then shut down since they are for a different SST, and we then get an error when trying to do the "ls -l /proc/$pid/fd" on a process that is no longer there.

RoelVdP:
@y-trudeau let's not let get this work to waste. Can we continue? @hrvojem fyi, you may have thoughts and/or be able to contribute. Also see mail in experts on enforcing selinux.

scripts/wsrep_sst_xtrabackup-v2.sh
RoelVdP: Would `| grep -E $(whoami) |` help also?

y-trudeau:
Of course Roel, I just have limited bandwidth when I booked 40h a week with customers.

scripts/wsrep_sst_xtrabackup-v2.sh
y-trudeau: Guys, do I need to be involved here or you have enough to move forward.  I mean, this is a security issue, if it breaks the tests.... modify the tests to run a single SST at a time.

kennt-percona: This PR should be closed and refiled as a high priority jira bug instead.  PRs are meant for us to merge into the github tree directly (and we can't add commits to your branch to make changes).

y-trudeau: Ok, will do.

kennt-percona:
Closing, Jira bug created: https://jira.percona.com/browse/PXC-2164

PR #651, opened 2018-06-18
- PXC#2133: Deprecate wsrep-forced-binlog-format and wsrep_sst_method…

kbauskar:
…="mysqldump"

  - PXC annouces deprecation of wsrep-forced-binlog-format
    and SST through mysqldump.

  - No change in current behavior but user should avoid using it
    as these features will be removed in next major release.


kbauskar:
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/510/
LGTM

PR #655, opened 2018-06-19
- PXC#2135: If PXC fails to load a keyring plugin then don't server t…

kbauskar:
…he node/server

  - If some of the nodes of PXC fails to load keyring plugin
    then it could create inconsistency in cluster if the encrypted
    objects are used from the node that has keyring plugin loaded.

  - In ordered to limit the damange, PXC now refuse to start the node
    if keyring plugin params are specified but PXC fails to load them.

  - If user doesn't specify keyring param on some of the node then
    the issue of inconsistency still remains but that's difficult
    to control as it is user induced behavior.
    Also, PXC self protection will shutdown such node on detecting
    inconsistency there-by keeping the data-safe.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/511/

PR #656, opened 2018-06-19
5.7-pxc-2127 : PXC shutdown hangs when pool-of-threads is set

kennt-percona:
Issue:
When the threadpool is used, the client connection threads are not
being shutdown correctly by wsrep.  Thus if a client connection is
in use at time of shutdown, the server will hang waiting for the
connection to be killed.

Solution:
On server shutdown, the wsrep code will not attempt to close the
connection, but will rely on the mysql shutdown code to close the
connections.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/512/

kbauskar:
Approved.

PR #657, opened 2018-06-19
PXC-2111 : PXC support for vault plugin

kennt-percona:
Issue:
The vault plugin is a new feature.

Solution:
Adding tests of the vault plugin for PXC.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/514/

kbauskar:
Percona-Vault server continues to hit issues so pushing test-case will cause mtr to fail. For now, delaying push of test-case. The feature has been tested by QA and dev and found it to be working as expected.

PR #673, opened 2018-06-29
- PXC#2154: rsync sst shouldn't be used with keyring_vault plugin for…

kbauskar:
… encryption

  Vault keyring plugin is not supported through rsync.
  Added error message for the same.

  Infact, Percona doesn't recommend using rsync with keyring
  as rsync doesn't use SSL channel for transmitting data.

  Also, fixed the issue off general tablespace sync through
  rsync (covers only if general tablespace is local to the data-directory).

### DIFF
--- a/scripts/wsrep_sst_rsync.sh
+++ b/scripts/wsrep_sst_rsync.sh
@@ -20,7 +20,6 @@
 
 RSYNC_PID=
 RSYNC_CONF=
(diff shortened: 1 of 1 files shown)


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/517/

PR #678, opened 2018-07-05
PXC-2155 : wsrep_sst_xtrabackup-v2 not deleting all folders on cleanup

kennt-percona:
Cherry-pick of commit 733692e06950707788c8c8f1d5acb3a5e7cf5051

Issue
If the TMPDIR environment variable is set, that location was being
used as the root folder (rather than the folder specified by
the "-p" option).  When the sst script exits, it deletes the root
folder, therefore the folders that were created in the TMPDIR were
not getting deleted.

Solution
Use the "--tmpdir" option explicitly instead of "-p" which was
deprecated.  This makes it clearer where the tmp directores are
being created.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/521/

PR #682, opened 2018-07-11
PXC-1017 : LP #1376727: Memcached access to InnoDB is not replicated …

kennt-percona:
…by Galera

Issue
The memcached API does not replicate changes (via Galera).
The innodb memcache API goes through a different code path than normal
SQL commands, thus it bypasses all the WSREP logic to replicate transactions
(it interacts directly with InnoDB at the binlog level).

Solution
A more thorough solution would involve adding the replication logic to the
memcached event handler.  However, that's a bigger feature change.  For now,
we will disable the memcached API when WSREP is being used on a node.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/533/

PR #684, opened 2018-07-11
Fix DOC-645 with better wording of xtrabackup versions sst compatibility

fiowro:


doc/source/manual/xtrabackup_sst.rst
```diff
@@ -371,7 +371,8 @@ every version is tested against a specific |PXB| version only:
 * |PXC| 5.6 requires |PXB| 2.3
 * |PXC| 5.7 requires |PXB| 2.4
 
-Other combinations are not guaranteed to work.
+Compatibility of other combinations is not guaranteed, and using the
```
kbauskar, line 374: didn't understood this " will remove the cluster."
maybe we can restrict it to simpler clause "Compatibility of other combinations is not guaranteed,"

kbauskar (changes requested):

doc/source/manual/xtrabackup_sst.rst
fiowro: As far as I understood from the issue description, the idea of the fix is to emphasis the possibility of the cluster malfunction in case of other combinations. I also think "will remove cluster" to be bad wording. May be "...will make cluster inoperable", or something like this?

kennt-percona: The impact would probably be on the joiner side, the node may not be able to join the cluster.

borys-belinsky:
Resolved in Jira ticket PXC-2197

PR #685, opened 2018-07-11
Fix DOC-769, adding libcurl as a build dependency

fiowro:


doc/source/install/compile.rst
```diff
@@ -60,6 +60,9 @@ Before you begin, make sure that the following packages are installed:
    * - socat
      - ``socat``
      - ``socat``
+   * - curl
+     - ``libcurl-dev``
+     - ``libcurl-devel``
```
kennt-percona, line 66: The names of these packages should also get added to the "sudo apt-get install" and "sudo yum install" instructions below.

dbadapt (approved):
After speaking with Dmitriy,  I believe the libcurl-dev{el} meta package will install the correct dependency.   LGTM.   Approved.

PR #688, opened 2018-07-24
- PXC#2162: Assertion failure during crash recovery

kbauskar:
* Say user is trying to run some workload and decide to run
    part of it with wsrep_on=off.

  * Server/Node hit a crash while running workload in wsrep_on=off
    and some transaction are left out in prepare state that
    could be committed on recovery.

  * Since these transactions got stamped with MySQL XID and not
    wsrep XID, on recovery, avoid updating wsrep co-ordinates
    using these transactions.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/525/

PR #689, opened 2018-07-25
- PXC#2165: Failing to set wsrep_node_address or wsrep_sst_receive_ad…

kbauskar:
…dress

  can cause SST to fail

  - If user fail to set wsrep_node_address or wsrep_sst_receive_address
    SST could fail without any hint towards the missing configuration.

  - Of-course it doesn't fail always but in restricted cases
    as listed in documentation.

  - Post this fix, script would emit a warning to help user get an hint
    of what may be missing.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -574,6 +574,14 @@ read_cnf()
     scomp=$(parse_cnf sst compressor "")
     sdecomp=$(parse_cnf sst decompressor "")
 
+    # if wsrep_node_address is not set raise a warning
+    wsrep_node_address=$(parse_cnf mysqld wsrep_node_address "")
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -574,6 +574,14 @@ read_cnf()
     scomp=$(parse_cnf sst compressor "")
     sdecomp=$(parse_cnf sst decompressor "")
 
+    # if wsrep_node_address is not set raise a warning
+    wsrep_node_address=$(parse_cnf mysqld wsrep_node_address "")
+    wsrep_sst_recieve_address=$(parse_cnf mysqld wsrep_sst_recieve_address "")
```
kennt-percona, line 579: Variable name is misspelled
s/recieve/receive/g

PR #692, opened 2018-07-30
PXC-2164 : sst script doesn't work well with SELinux

kennt-percona:
Issue
The SST script, on the joiner, is checking to see if there is a process
waiting on the SST port.  It is using tools that check all ports on
all processes thus triggering SELinux.

Solution
Check for the network ports directly in /proc/xxx/net/tcp rather
than using a tool that scans all processes.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/541/

PR #695, opened 2018-08-08
- PXC#2199: DROP TRIGGER IF EXISTS increment unexpected GTID

kbauskar:
* MySQL logic would append drop trigger if exists to binlog
    even if there is no trigger present.

  * This early exit path didn't entered GALERA replication world.
    So the said statement got registered only with local gtid.

  * Corrected flow to register said statement early exit flow
    in GALERA replication context.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/531/

PR #697, opened 2018-08-08
- PXC#2202: Disconnected Cluster Node should still allow itself to be…

kbauskar:
… shut down

  - MySQL-5.7 introduces SQLCOM_SHUTDOWN command.

  - Once node turns non-primary, PXC allows execution of restricted command.

  - Shutdown command should be allowed but since it was introduced in 5.7
    it got missed (5.6 has different way of handling it and so it worked till 5.6).


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/532/

PR #698, opened 2018-08-20
- PXC#2209: compression dictionary is not replicated in PXC

kbauskar:
- COMPRESSION_DICTIONARY support is added in PS starting from 5.7.17

  - CREATE and DROP of this object is done through a dedicated set of
    command internally in MySQL. TOI replication was not enabled for these
    commands and so DICTIONARY object was not replicated.

  - Added relevant TOI replication hook.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc57.test.mtr/533/

PR #699, opened 2018-08-22
- PXC#2213: NULL/VOID DDL can commit out-of-order

kbauskar:
- With recent performance optimization we expect all
    transaction to pass through ordered_commit.

  - This is true for all transaction including DML and DDL.

  - Unfortunately, if DDL transaction fails in pre-check
    then there is nothing to commit but this DDL statement
    is replicated before execution or pre-check starts
    so it still try to commit.

  - This could cause a race between transaction passing
    through ordered_commit maintaining commit ordering
    vs above short-circuit null/void DDL transaction.

  - Avoid update of wsrep co-ordinate through such
    null/void DDL transaction.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/535/

PR #711, opened 2018-09-25
Doc fix warnings

borys-belinsky:


kbauskar:
@borys-belinsky 
Something not right the PR. Your PR has code change merge that is already done. Can you check

PR #712, opened 2018-09-25
Rn 5.7.23 31.31

borys-belinsky:


doc/source/release-notes/Percona-XtraDB-Cluster-5.7.23-31.31.rst
```diff
@@ -0,0 +1,47 @@
+.. rn:: 5.7.23-31.31
```
kbauskar, line 1: As replied over blog email, this may need an addition if tyler approves it.

sql/sql_parse.cc
```diff
@@ -1303,6 +1303,10 @@ static inline bool is_timer_applicable_to_statement(THD *thd)
 
 
 #ifdef WITH_WSREP
+
```
kbauskar, line 1306: why is this present here

kbauskar (changes requested):

kbauskar:
This is what I think has went wrong.
You probably created your branch from 5.7-trunk and not from release branch.
Code change that your branch is trying to merge is from 5.7-trunk that we don't plan to release in this release. Please re-create your branch from release branch only.

kbauskar:
@borys-belinsky 

Another issue.

* rel-pxc-5.7.23

grep "release =" doc/source/conf.py 
release = '5.7.22-29.26'

So the release branch doc still has old release number.

-------------

Please look into all the files that are often touched by DOC. You can check the protocol with @fiowro. Also, for quick list you can refer to the file-list embedded in the email thread (related to release).

PR #723, opened 2018-10-31
- PXC#2272: Avoid starting mysqld service (pxc) if one is running

kbauskar:
- While porting code to 5.7 and eventually during development
    some of the code that was specific to WSREP was left
    without the preprocessing directive WITH_WSREP protection.


kbauskar:
LGTM
mtr run: https://jenkins.percona.com/job/pxc57.test.mtr/548/

PR #725, opened 2018-11-08
PXC-2269 : pxc_encrypt_cluster_traffic is not dynamic, but it allows …

kennt-percona:
…set global

Issue
pxc_encrypt_cluster_traffic can be set dynamically.  This should not
be the case.

Solution
Make pxc_encrypt_cluster_traffic read-only.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/554/

PR #726, opened 2018-11-13
PXC-2274 : checking for wsrep_node_address in wsrep_sst_common uses t…

kennt-percona:
…he wrong variable

Issue
The check for wsrep_node_address looks for 'wsrep_node_address' but
parse_cnf() normalizes the names from my_print_defaults by changing
'_' to '-'.  So this means that we will never match these two names.

Solution
Change the calling code to use 'wsrep-node-address' instead of
'wsrep_node_address'.  Also remove some code that searches for
variables using underscores (since the code does nothing).

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -135,6 +135,9 @@ parse_cnf()
     local var=$2
     local reval=""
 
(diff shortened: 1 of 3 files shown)


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/557/

PR #739, opened 2019-01-22
PXC-2345 : Crash triggered with mysqldump single-transaction

kennt-percona:
Issue
Running 'mysqldump --single-transaction' on a running PXC
server causes a crash.

Solution
If WSREP is enabled, skip over the consistent snapshots.
This was regression from the previous version.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/578/

kbauskar (approved):

PR #742, opened 2019-02-06
- PXC#2388: DROP function is not replicating

kbauskar:
- DROP FUNCTION with explicit name was not replicating
    due to missing TOI call for the said flow.


kbauskar:
https://jenkins.percona.com/job/pxc56.test.mtr/564/

PR #746, opened 2019-02-11
PXC-761 :  Async slave repository tables copied via SST

kennt-percona:
Issue
In the event of an SST, the async slave repository information is
also copied to the JOINER (which can cause problems if multiple nodes
are then copying data from the async master).

Solution
If an SST occurs, remove the async slave repository information.
A "RESET SLAVE ALL" is performed on the JOINER as part of the
mysql_upgrade process.


kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/

kbauskar (approved):

PR #750, opened 2019-02-13
- PXC#2389: Failed to abort local victim thread that is busy running 

kbauskar:
dd-transaction

  As per new MySQL-8.0 flow/semantics, main action may spawn sub-actions
  to access dd-tables. To facilate this, MySQL will backup main transaction
  data and temporarily replace it with new transaction (tracked through
  attachable transaction).

  There could be chain of such attachable transactions.

  When high priority thread abort local running transaction that is busy
  with such dd action then high priority thread should reach to main
  transaction by traversing the list.

  Existing code wrongly assumed list to be of size = 1.


kbauskar:
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/51/
LGTM

PR #751, opened 2019-02-13
- PXC#2411: mysqlx plugin is not loaded if joiner is busy with SST or…

kbauskar:
… IST

  - mysqlx is loaded as plugin. As part of the plugin initialize action,
    mysqlx needs to query mysql.user table to findout some details.

  - As per original flow of mysql, mysqlx plugin will wait for
    server to start accepting connection before it fires this query.

  - In PXC, even though node open socket for accepting queries,
    it may not be able to process queries till the state of the node
    is turned to SYNCED. Normally this would happen instantly but
    if node is receiving IST then moving to SYNCED state may take time.
    (if node has received SST and there are write-sets pending in gcache
     to apply, node may take time to move to SYNCED state).

  - This condition was missing that caused mysqlx query to failed
    and so mysqlx plugin was never activiated.

  - mysqlx waits for server to start accepting connection.
    This condition is modified to now ensure that server is also ready
    to accept query and not just connection.

### DIFF
--- a/sql/srv_session_service.cc
+++ b/sql/srv_session_service.cc
@@ -29,6 +29,10 @@
 #include <stddef.h>
 #include <new>
 
+#ifdef WITH_WSREP
+#include "wsrep_mysqld.h"
+#endif /* WITH_WSREP */
+
 #include "my_dbug.h"
 /*
  service_srv_session.h should not be first to be included as it will include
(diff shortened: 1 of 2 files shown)


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/57/

PR #755, opened 2019-02-14
PXC-2406: CTAS - update info

borys-belinsky:


doc/source/features/pxc-strict-mode.rst
```diff
@@ -343,8 +343,16 @@ the following happens:
 Combining schema and data changes in a single statement
 -------------------------------------------------------
 
-|PXC| does not support ``CREATE TABLE ... AS SELECT`` (CTAS) statements,
-because they combine both schema and data changes.
+Starting from version 8.0, |PXC| supports the :abbr:`CTAS (CREATE TABLE ... AS
```
kbauskar: There is no change in this stand. We officially continue to say that under ENFORCING mode we don't support CTAS. But complete disable is not good choice so we kept these feature enabled under DISABLED or permissive mode.

kbauskar: May be we can re-word the statement to say PXC doesn;'t recommend using CTAS as it is mix of DDL and DML changes and with gtid_on MySQL too report it as error as it can affect GTID consistency. Given the said limitation, PXC has blocked use of CTAS under ENFORCING and user should avoid using CTAS.

doc/source/features/pxc-strict-mode.rst
```diff
@@ -368,8 +376,9 @@ the following happens:
 
  At runtime, any CTAS operation is denied and an error is logged.
 
-.. note:: CTAS operations for temporary tables are permitted
-   even in strict mode.
+.. note::
```
kbauskar: this is not true. Old behavior is retained. Let me clarify the comment I made in the PXC ticket.
Sorry for the confusion but why I meant was temporary table should not be used as SOURCE table in CTAS as the temporary table are not present on all the nodes.

doc/source/features/pxc-strict-mode.rst
```diff
@@ -368,8 +376,9 @@ the following happens:
...
+
+   CTAS operations for temporary tables are forbidden.
```
kbauskar, line 398: Given that behavior has changed and there is no separate dedicated to CTAS to capture it may be we can add another NOTE section and comment about new behavior here. Note: This behavior is of use only if user is trying to exercise CTAS in disabled mode.

CTAS has SOURCE and DESTINATION tables.

_[ideally user should see this note only when user uses 8.0 manual]_

- Starting PXC-8.0, CTAS replicates as TOI (vs write-set or binlog based replication as it use to be till PXC-5.7)
- SOURCE table(s) should be present on all cluster node. Each cluster node executes CTAS as an independent statement through TOI flow.

- MyISAM tables will be created and loaded even if wsrep_replicate_myisam=1. (Till PXC-5.7 if wsrep_replicate_myisam=0 MyISAM tables were not created as it involved loading too). (PXC doesn;'t recommend use of MyISAM tables and engine. Support for MyISAM is experimental and may be removed in the future release).

- Avoid using a temporary table as source table as a temporary table is not present (or replicated) on all the cluster nodes.

- If node-1 has a temporary and non-temporary table with the same name, CTAS on node-1 will use temporary and CTAS on node-2 will use non-temporary. This may create data level inconsistency.

kbauskar (changes requested):

doc/source/features/pxc-strict-mode.rst
borys-belinsky: Ready for review

kennt-percona (approved):

doc/source/features/pxc-strict-mode.rst
```diff
@@ -343,33 +343,56 @@ the following happens:
 Combining schema and data changes in a single statement
 -------------------------------------------------------
 
-|PXC| does not support ``CREATE TABLE ... AS SELECT`` (CTAS) statements,
-because they combine both schema and data changes.
+Starting from version 8.0, |PXC| supports the :abbr:`CTAS (CREATE TABLE ... AS
+SELECT)` statement in ``PERMISSIVE`` and ``DISABLED`` modes. As applies both DDL
```
kennt-percona: "As applies both DDL"
should be
"As CTAS applies both DDL"

doc/source/features/pxc-strict-mode.rst
```diff
@@ -343,33 +343,57 @@ the following happens:
...
-because they combine both schema and data changes.
+With strict mode set to ``ENFORCING``, |PXC| does not support :abbr:`CTAS
+(CREATE TABLE ... AS SELECT)` statements, because they combine both schema and
+data changes.
 
-Depending on the strict mode selected,
-the following happens:
+With strict mode set to ``PERMISSIVE`` or ``DISABLED``, |ctas| statements are
+replicated, like any DDL statement, using the :abbr:`TOI (Total Order
+Isolation)` method to ensure consistency. Note that tables in the SELECT clause
```
kbauskar: using TOI (as against PXC-5.7 approach to use DML write-sets based replication)

kbauskar (changes requested):

kbauskar (approved):

PR #756, opened 2019-02-15
- PXC#2407: Expose galera (local/apply/commit) monitor stats as part …

kbauskar:
…of show status

  - Galera has concept of monitor. Monitor maintains a queue (FIFO)
    and transaction are processed as per this order.

  - Some queue allow parallel processing. Like transaction in apply queue
    can be applied in parallel but commit queue/monitor should be applied
    striclty one-after-another.

  - Getting stats of this queue/monitor can help get important insight
    to understand how galera is processing transaction.


kbauskar:
https://jenkins.percona.com/job/pxc80.test.mtr/67/

PR #757, opened 2019-02-17
PXC-2231 : Introduce internal user for doing SST

kennt-percona:
Issue
The wsrep_sst_auth variable is needed to store the username/password
to use during an SST.  This is stored in plaintext in a cnf file and
is a security risk.

Solution
Remove the need for wsrep_sst_auth (mainly by creating an internal
user dynamically and using it only during an SST, both on the
joiner and donor).


kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/71/

sql/wsrep_service.cc
```diff
@@ -0,0 +1,392 @@
+/* Copyright (c) 2015, 2016, Oracle and/or its affiliates. All rights reserved.
```
kbauskar, line 1: this needs percona copyright

sql/wsrep_service.h
```diff
@@ -0,0 +1,1059 @@
...
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU General Public License for more details.
+
+   You should have received a copy of the GNU General Public License
+   along with this program; if not, write to the Free Software Foundation,
+   51 Franklin Street, Suite 500, Boston, MA 02110-1335 USA */
+
+
+// This file is a copy of rapid/plugin/group_replication/include/sql_service/sql_service_context_base.h
```
kbauskar: Can you help explain why we need this in the comment section.

sql/wsrep_sst.cc
```diff
@@ -868,12 +802,234 @@ static void sst_disallow_writes(THD *thd, bool yes) {
...
-  if (run_sql_command(thd, query_str)) {
+  if (run_sql_command(thd, query_str, NULL)) {
     WSREP_ERROR("Failed to disallow InnoDB writes");
   }
   return;
 }
 
+// Password characters are limited, because we are using these passwords
+// within a bash script.
+const std::string g_allowed_pwd_chars("qwertyuiopasdfghjklzxcvbnm1234567890"
```
kbauskar, line 813: Does this fit the industry security model. (Generally I see industry recommending one special character).

scripts/mysql_system_tables_fix.sql
```diff
@@ -1045,6 +1045,55 @@ INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.session', 'localhost', 'SY
 
 FLUSH PRIVILEGES;
 
+#
+# SQL commands to create the PXC SST root and role that is used by the SST process
```
kbauskar, line 1049: Not sure I understood why we need to create user through these sql when it would be anyway auto-created when sst kicks-in.

scripts/mysql_system_tables_fix.sql
```diff
@@ -1045,6 +1045,55 @@ INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.session', 'localhost', 'SY
...
+#
+# mysql.pxc.sst.root
```
kbauskar, line 1051: to get more clarity I would suggest we can rename the user to 
mysql.session_use_to_create_sst_user.root

sql/wsrep_utils.cc
```diff
@@ -331,13 +331,20 @@ static int execvpe(const char *file, char *const argv[], char *const envp[]) {
 #define STDIN_FD 0
 #define STDOUT_FD 1
 
-process::process(const char *cmd, const char *type, char **env)
-    : str_(cmd ? strdup(cmd) : strdup("")), io_(NULL), err_(0), pid_(0) {
+process::process(const char *cmd, const char *type, char **env, bool execute_immediately)
+    : str_(cmd ? strdup(cmd) : strdup("")), io_(NULL), io_w_(NULL), err_(0), pid_(0)
+  {
+    if (execute_immediately)
```
kbauskar, line 337: what is the significance of this execute immediately? Can you add a comment?

sql/wsrep_sst.cc
```diff
@@ -897,14 +1058,42 @@ static void *sst_donor_thread(void *a) {
...
+                  "sst_user=mysql.pxc.sst.user\n"
+                  "sst_password=%s\n",
+                  password.c_str());
+    if (ret < 0 || ret >= out_len)
+    {
+      WSREP_ERROR("sst_donor_thread(): snprintf() failed: %d", ret);
+      err = (ret < 0 ? ret : -EMSGSIZE);
+    }
+
+    if (!err)
```
kbauskar, line 1083: is pipe communication secure ?

kbauskar (changes requested):
can you please check the flow below and if needed update the status messages clarifying the flow
if I understand the flow correctly
1. A pxc-root user is created during database initialization/upgrade (this pxc-root user provides the security context for creating the SST user needed to run xtrabackup, so it only applies to the donor).
2. while sst start .. this pxc-root user is used to create a session to the mysql server from sst thread
3. this session is used to create an SST user with a random password
4. this user and password is then passed to the donor as part of the sst script (wsrep_sst_xtrabackup-v2.sh) through anonymous pipes (stdin)
5. the script reads these credentials and passes them to xb for backup
6. these credentials are never passed over the wire from donor to joiner
7. on the joiner side, as part of mysql upgrade, the script will recreate the SST user along with another random password (through --init-file so no need to use the pxc-root user).  The joiner SST user will have the permissions needed to run mysql_upgrade (which is different from the permissions needed to run xtrabackup).
8. this newly created user on the joiner side is used to run mysql_upgrade, as well as other operations needed.
9. finally on completion of mysql_upgrade, the SST user is removed
10. Also, the donor side user is removed after the sst is complete (failure or pass)

The donor and joiner roles reuse the sst-user (although their usage is totally separate).

sql/wsrep_sst.cc
kennt-percona: Special characters are needed to help prevent dictionary attacks, so randomly generated passwords don't need it.  For a randomly generated password, any character is just like any other character.  The amount of randomness added by a special character doesn't really add much, just complicates the usage/testing with bash.

scripts/mysql_system_tables_fix.sql
kennt-percona: Different user.  This is the user that is used to create the sst user.  The security context for the server-side code is run with this user (this session then creates the SST user).

(Can't depend on the 'root' user existing as it can be renamed and can't depend on the anonymous user as that account could be locked down).

So the mysql.pxc.sst.root creates the mysql.pxc.sst.user (and will assign the mysql.pxc.sst.role when possible).  mysql.pxc.sst.root has the minimal set of privileges needed to create the mysql.pxc.sst.user.

sql/wsrep_utils.cc
kennt-percona: The original code would execute the cmd as part of the constructor.  I moved the execution outside of the constructor (so that we could skip execution if an error occurred).  To avoid changing existing code, I left the ability to execute the code immediately as part of the constructor.  (will add a comment)

sql/wsrep_sst.cc
kennt-percona: https://unix.stackexchange.com/questions/156859/is-the-data-transiting-through-a-pipe-confidential

sql/wsrep_utils.cc
kennt-percona: comment added to wsrep_utils.h

sql/wsrep_service.cc
kennt-percona: copyright notices added

sql/wsrep_service.h
kennt-percona: Comment removed

kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/83/

PR #758, opened 2019-02-18
- PXC#2417: ALTER TABLE .. DISCARD tablespace result in crash as stmt…

kbauskar:
… is not

  tagged for replication

  - ALTER TABLE ... IMPORT/DISCARD tablespace is not replicated as it is local
    operation. Said statement being unsafe to run in cluster mode is also
    blocked under pxc_strict_mode = ENFORCING.

  - Since ALTER is not replicated it doesn't pass through TOI check.

  - ALTER command is not atomic and so wsrep_replicate (dml based binlog
    replication hook) is called called. wsrep_replicate is not equipped to
    handle non-replicating atomic ddl. This causes wsrep_replicate to return
    error when command execute (when it should just ignore this command as
    command is not running in not registered in galera context).


kbauskar:
https://jenkins.percona.com/job/pxc80.test.mtr/70/

PR #759, opened 2019-02-20
PXC-2368 : PXC 8.0 replication filters

kennt-percona:
Issue
MySQL 8.0 introduces per-channel filters (channels now have names)
as well as the global filter (in 5.7 there was only the global filter).
This causes complications for PXC due to the differences in behavior
between async replication and galera replication.

Solution
Have PXC ignore the global and per-channel filters for galera replication.

### DIFF
--- a/sql/rpl_slave.cc
+++ b/sql/rpl_slave.cc
@@ -725,6 +725,20 @@ bool start_slave_cmd(THD *thd) {
     }
 
     res = start_slave(thd);
+#ifdef WITH_WSREP
+  // Similar to GR, do not allow operations on the wsrep channel
+  } else if (lex->mi.channel && wsrep_is_wsrep_channel_name(lex->mi.channel)) {
+    const char *command = "START SLAVE FOR CHANNEL";
+    if (thd->lex->slave_thd_opt & SLAVE_IO)
+      command = "START SLAVE IO_THREAD FOR CHANNEL";
+    else if (thd->lex->slave_thd_opt & SLAVE_SQL)
+      command = "START SLAVE SQL_THREAD FOR CHANNEL";
+
+    my_error(ER_SLAVE_CHANNEL_OPERATION_NOT_ALLOWED, MYF(0), command,
+             lex->mi.channel, command);
+
+    goto err;
+#endif /* WITH_WSREP */
   } else {
     mi = channel_map.get_mi(lex->mi.channel);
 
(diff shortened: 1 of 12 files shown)


sql/rpl_slave.cc
```diff
@@ -725,6 +725,20 @@ bool start_slave_cmd(THD *thd) {
     }
 
     res = start_slave(thd);
+#ifdef WITH_WSREP
+  // Similar to GR, do not allow operations on the wsrep channel
+  } else if (lex->mi.channel && wsrep_is_wsrep_channel_name(lex->mi.channel)) {
```
kbauskar, line 730: I am not sure if these syntax are applicable to wsrep directly.
For async slave, user is allowed to start slave thread and stop it. (GR may follow the same).
In case of wsrep it is not allowed. So I don't foresee a use-case for this but if u you find one please help add it here along with test-case.

sql/rpl_slave.cc
```diff
@@ -799,6 +813,22 @@ bool stop_slave_cmd(THD *thd) {
 
   if (!lex->mi.for_channel)
     res = stop_slave(thd);
+#ifdef WITH_WSREP
```
kbauskar, line 816: same as above.

sql/rpl_slave.cc
```diff
@@ -954,7 +984,12 @@ static int find_first_relay_log_with_rotate_from_master(Relay_log_info *rli) {
...
   my_off_t master_log_pos = 0;
 
-  if (channel_map.is_group_replication_channel_name(rli->get_channel())) {
+  if (channel_map.is_group_replication_channel_name(rli->get_channel())
+  #ifdef WITH_WSREP
+      // Similar to GR, do not allow operations on the wsrep channel
+      || wsrep_is_wsrep_channel_name(rli->get_channel())
+  #endif /* WITH_WSREP */
+    ) {
     LogErr(INFORMATION_LEVEL,
```
kbauskar: Error message may create confusion since it would point to GR even though user end up using wsrep.
Can we segregate our code block along with custom error message based on how GR emits it.

kbauskar (changes requested):

sql/rpl_slave.cc
kennt-percona: I think the issue is that someone could create an async channel using the wsrep filter.

sql/rpl_slave.cc
kennt-percona: Check/error message separated out.

sql/rpl_slave.cc
kbauskar: ok but since there is no use-case can you leave a comment.

kbauskar (approved):

PR #760, opened 2019-02-20
- PXC#2418: 5.7 cache duplicate xids during recovery. 8.0 doesn't

kbauskar:
- MySQL expect unique xid for each transaction/event logged to binlog.

  - In 8.0, MySQL enforced it by changing from HASH (that allowed non-unique)
    XID to a set that enforced unique xid.

  - This change has not gone well with PXC that has duplicate xid.

  - PXC has 2 different kind of events:
    - node-local events logged with MySQL xid
    - replicating events logged with WSREP xid.

  - Both sub-system has different way to generate xids and so binlog
    may end up having 2 different event (with different GTID if enabled)
    but same xid.

  - PXC can't detect this xid conflict before hand as during recovery
    xid event only logs xid numeral and not complete.

  From long term perspective in order to comply with original MySQL design
  PXC has decided to avoid binlogging node-local statements.

  PXC can have node local statement originating by setting wsrep_on=off
  or non-replicating-atomic-ddl.

  Fix will block both of these aspect to ensure they don't cause bin logging
  of event.


kbauskar:
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/79/
LGTM

PR #763, opened 2019-02-22
- PXC#2390: Non-atomic SE parallel DDL can race with atomic SE DML while

kbauskar:
updating wsrep co-ordinates

  - MySQL-8.0, executes DDL commit for non-atomic storage engine objects
    at 2 level: intermediate commit and final commit.
    [Reason for intermediate commit is not clear from code].

  - This non-atomic storage engine DDL is also replicated by PXC using TOI.

  - This causes double persistence of the same wsrep co-ordinates

  - Patch works a solution to avoid persistence of wsrep co-ordinates
    during intermediate commit.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/

PR #764, opened 2019-02-26
PXC-2430 : Centos6 : if SST fails in mysql_upgrade, script continues …

kennt-percona:
…rather than exits

PXC-2433 : PXC-8.0 SST is failing with "use_mysql_upgrade_conf _suffix: unbound variable" error

Issue
In Centos6, a subshell is spawned when running mysql_upgrade.  This causes
the "exit" statement, to exit the subshell and not the script.  Thus the
process continues on after an error has occurred (and we expected to fully exit).

Solution
Change the function to "return" rather than "exit".  In the calling script
handle the return values and exit properly.

Also additional script fixes (PXC-2433)
Also an optimization, do not run mysql_upgrade if donor/joiner have the same mysql version
(major.minor.revision are the same).

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -334,27 +334,18 @@ function get_length_in_bytes()
 {
     local str=$1
     local byte_len=0
-    local orig_lang=""
-    local orig_locale=""
 
-    if [ -n "${LANG:-}" ]; then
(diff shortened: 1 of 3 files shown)


kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/89/

PR #765, opened 2019-02-26
- PXC#2396: handler operation is being brute force aborted

kbauskar:
- handler operation (HANDLER <table> open <handle>)
    will take explicit lock.

  - This explicit lock check was not being detected
    as part of brute-force abort flow that caused
    the said thread with handler lock to get forcefully aborted.

  - Patch now marks lock with explicit duration as non-preemptable.

Also fixed some timing issues mainly waiting for event to appear
as they are applied in async fashion.


kbauskar:
https://jenkins.percona.com/job/pxc80.test.mtr/88/

PR #769, opened 2019-02-28
PXC-2399 : If the donor node stops the SST, the joiner rsync process …

kennt-percona:
…stays around

Issue
If a failure occurs, the donor process will exit, but the joiner will be
left waiting.

Solution
If an error occurs, have the donor send the MAGIC_FILE but with
an error message. Of course, if rsync/network is the problem itself
then this doesn't help, but for other kinds of errors it will be
helpful.

Also:
Added force-ugprade option to both rsync/xtrabackup
rsync will now skip the mysql_upgrade if the version is the same
Fixed a bug where we would return success before the mysql_upgrade

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -343,6 +343,67 @@ function get_length_in_bytes()
     ) || true
 }
 
+# Returns the version string in a standardized format
+# Input "1.2.3" => echoes "010203"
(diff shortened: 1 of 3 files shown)


kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/91/

kbauskar (approved):

PR #775, opened 2019-03-02
8.0 pxc 2451

kennt-percona:


kennt-percona:
https://jenkins.percona.com/job/pxc80.build/113/

scripts/wsrep_sst_common.sh
```diff
@@ -731,7 +989,7 @@ EOF
         --defaults-file=/dev/stdin \
         --socket=$upgrade_socket \
         --unbuffered --batch --silent \
-        -e "SET sql_log_bin = OFF; ALTER USER '${WSREP_SST_OPT_USER}'@localhost ACCOUNT LOCK; SHUTDOWN;" \
+        -e "SET sql_log_bin=OFF; ALTER USER '${WSREP_SST_OPT_USER}'@localhost ACCOUNT LOCK; SHUTDOWN;" \
```
kbauskar, line 1002: please add comment why we need to lock the user before shutdown when on SHUTDOWN action we tend to remove the user.
(as we discussed during call a defensive check).

scripts/wsrep_sst_common.sh
```diff
@@ -430,6 +528,41 @@ function check_for_version()
 #   0 if successful (no errors encountered)
 #   <non-zero> otherwise
 #
+# Overview: There are two code paths:
+#   (1) Receiving from a < 8.0 donor (5.7 donor)
+#   (2) Receiving from a >= 8.0 donor
+#
+#   (1) Receiving from a 5.7 donor
```
kbauskar, line 543: Can you expand this comment to make it more explicit that 8.0 binaries will not allow CREATE USER TO work on DONOR copied 5.7 data-directory.

kbauskar (changes requested):
Can you also change the user from mysql.pxc.sst.root to mysql.pxc.internal.session.user

kbauskar:
@kennt-percona 
Can you also add force-upgrade to default mtr configuration include/default_mysqld.cnf

scripts/wsrep_sst_common.sh
kennt-percona: There's a comment below (with the detailed flow) that explains this.

scripts/wsrep_sst_common.sh
kennt-percona: comment added

kbauskar (approved):

PR #780, opened 2019-03-05
Updated SST QA script for PXC-8.0 test

rameshvs02:


kbauskar (commented):
logic to check version has been replaced in wsrep_sst_common to use better operator. Do consider if you want to change or update to it.

percona-xtradb-cluster-tests/sst/run.sh
```diff
@@ -300,16 +331,22 @@ function set_vars()
 	TAR=tar
     fi
 
-    find_program MYSQL_INSTALL_DB mysql_install_db $MYSQL_BASEDIR/bin \
-	$MYSQL_BASEDIR/scripts
+    if ! check_for_version $MYSQL_VERSION_CHECK "8.0.0" ; then
```
kbauskar: Can you add comment as to when to use find and when to use which.

kbauskar (changes requested):

percona-xtradb-cluster-tests/sst/run.sh
rameshvs02: Done

kbauskar (approved):

PR #788, opened 2019-03-28
- PXC#2490: binlog_space_limit crashes PXC to during wsrep_recover mode

kbauskar:
- As part of the "PXC#897: gitd_executed is empty post wsrep_recover"
    fix, PXC skips creation of new binlog file while running in recovery mode.

  - In light of this flow, PXC should also avoud purge action on binlog.

  - Patch adds the check for the same.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc56.test.mtr/574/

PR #791, opened 2019-03-29
- PXC#2434: Password should not be visible as part of conflict debug …

kbauskar:
…message

  - use re-written query while logging wsrep debug statement.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc56.test.mtr/575/

PR #794, opened 2019-03-29
PXC-2483 : PXC-8.0 upgrade is not initiating mysql_upgrade script wit…

kennt-percona:
…h rsync SST

Issue
When the binaries of a node are upgraded, is restarted and rejoins the cluster.
If no writes have been received, the IST/SST process is skipped
(thus not allowing us to run mysql_upgrade on the node).

Solution
If the IST/SST process is skipped, then check the wsrep_state.dat file.
If the wsrep_schema_version is different, then run wsrep_sst_upgrade.sh


kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/120/

PR #795, opened 2019-03-29
- PXC#2491: SST is failing if the donor has encrypted undo logs.

kbauskar:
- PXC-5.7 doesn't support undo log encrpytion and has no plan
    to support it in future given that the implementation has been
    replaced in PS with MySQL native implementation.

  - Infact, PXC-5.7 don't plan to support any of these
    data-at-reset encryption features

    - innodb_temp_tablespace_encrypt
    - innodb_sys_tablespace_encrypt
    - innodb_encrypt_tables
    - innodb_parallel_dblwr_encrypt
    - innodb_undo_log_encrypt
    - innodb_redo_log_encrypt

  - Support for existing features that PXC-5.7 started supporting (BETA quality)
    general tablespace, file-per-table, encrypt-tmp-files will continue under
    BETA quality tag.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc57.test.mtr/594/

PR #796, opened 2019-04-01
- PXC#2486: Avoid replication of native ddl tables internally as part of

kbauskar:
main DDL statement

  - Some of the DDL statement (like one in the use-case INSTALL COMPONENT)
    can cause creation of native table on the fly.

  - Given the parent DDL statement is replicated using TOI avoid
    replication of such derived/child tables using TOI.

    Parent replication statement will take-care of creation of such child
    tables on replicating/applying node too.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/121/

PR #799, opened 2019-04-02
PXC-2276 : wait_for_listen in wsrep_sst_xtrabackup-v2.sh makes assump…

kennt-percona:
…tion about operating system

PXC-2476 : 5.7: rsync IST fails with keyring_file
PXC-2292 : wsrep_sst_xtrabackup-v2 doesn't recognize elliptic curve key/cert

Issue
2276: The selinux fix assumes that we are using a base Linux distro, so that we
have access to /proc/<pid>/net/tcp.  However, this is not true for some
other OSes (such as SmartOS).
2476: When using rsync for an IST, the keyring file is not sent, but the
code thinks it does.  The problem is that rsync does not know if an IST or
an SST is occurring.
2292: RSA keys work but EC keys do not.

Solution
2276: If we cannot find /proc/<pid>/net/tcp, go back to using ss for
socat/nc discovery.
2476: Check the rsync transfer logs to determine if an IST or SST is occurring.
2292: Use a more general method to check the key/cert

### DIFF
--- a/scripts/wsrep_sst_rsync.sh
+++ b/scripts/wsrep_sst_rsync.sh
@@ -61,10 +61,11 @@ cleanup_joiner()
     wsrep_log_debug "Joiner cleanup. rsync PID: $PID"
     [ "0" != "$PID" ] && kill $PID && sleep 0.5 && kill -9 $PID >/dev/null 2>&1 \
     || :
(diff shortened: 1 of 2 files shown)


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/595/

PR #805, opened 2019-04-04
PXC-2464 : re-direct sst bash log messages to main mysqld for logging

kennt-percona:
Issue
MySQL 8.0 changed the way logging is done.  Since the SST scripts log
directly to stderr, they bypass the logging infrastructure introduced
in 8.0.  Thus the SST messages appear out of order.

Solution
Have the server side code that invokes the SST script, read from stderr,
and redirect the messages to the MySQL infrastructure.


kennt-percona:
https://jenkins.percona.com/job/pxc80.build/157/

sql/mysqld_thd_manager.cc
```diff
@@ -27,6 +27,7 @@
 #include "wsrep_sst.h"
 #endif /* WITH_WSREP */
 
+#include "mysql/components/services/log_builtins.h"
```
kbauskar: If we are referencing something from this and then it would be wsrep specific. 
(Can you move this to WITH_WSREP ?)
and same for all other lines added

sql/wsrep_sst.cc
```diff
@@ -273,6 +288,140 @@ struct sst_thread_arg {
...
+  wsrep_pfs_register_thread(key_THREAD_wsrep_sst_logger);
+#endif /* HAVE_PSI_INTERFACE */
+
+  int const   out_len = 1024;
+  char        out_buf[out_len];
+
+  /* Loops over the READ end of stderr.  Only stops
+     when the pipe is in an error or EOF state.
+   */
+  while (!feof(arg->err_pipe) && !ferror(arg->err_pipe))
```
kbauskar: MySQL now uses clang formatting. They ship standard clang format as part of the codebase. Do run clang formatting over codebase to be compliant with mysql. (only for wsrep-plugin can skip it for galera for now).

kbauskar (changes requested):

sql/mysqld_thd_manager.cc
kennt-percona: Moved all the log_builtins.h inside of WITH_WSREP (well those in non-wsrep code files).

sql/wsrep_sst.cc
kennt-percona: Run clang-format on wsrep_sst.c[c](url)

kennt-percona:
https://jenkins.percona.com/job/pxc80.test.mtr/130/

kbauskar (approved):

PR #809, opened 2019-04-09
- PXC#2515: deinit action failed to wait for active commit applier

kbauskar:
- User initiated deinit action by setting wsrep_provider=none
    without shutting down the node failed to wait for active commit
    applier.

  - deinit action made progress and freed memory while applier is active
    applying transaction resulting in the said assert.

  - This is regression introduced during PXC-8.0 refresh.
    Fixed logic to consider active applier even though WSREP_ON = false.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/132/

PR #810, opened 2019-04-10
- PXC#2521: ALTER TABLESPACE is not replicated

kbauskar:
- ALTER TABLESPACE command failed to replicate using TOI causing
    multiple level assert.

  - Patch fixes replication of alter tablespace and also covers
    other commands that includes CREATE/DROP/ALTER UNDO tablespace.

  (Patch also fixes following issue not related to above fix.
   If WSREP_SST emits error then immediately start flushing error log
   without waiting for log module to initialize)


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/133/

PR #811, opened 2019-04-11
- PXC#2523: Incase of SST erorr ensure log messages are flushed befor…

kbauskar:
…e aborting

  - With new MySQL logging framework, log messages are buffered and not flushed
    till log module is not initialized.

  - log-module is initialized after SST so if there is error during SST then
    it is possible that log messages are never flushed leaving now clue for
    end-user to understand SST failure reason.

  - Patch fix this scenario by forcing flush of messages if error is detected
    from sst/galera/wsrep module before SST is initialized.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/134/

PR #812, opened 2019-04-11
- PXC#2508: crash while reloading user table for user w/o hostname

kbauskar:
(Note: fix is not for the bug mentioned above but as part of the investigation
   we found another sub-issue. this fix will address it keeping original bug open)

  - If user has turned sql_log_bin = off then write_to_binlog action shouldn't
    execute. MySQL flow will generate all needed binlog events before finally
    blocking it at write-stage. This means all work spend in generating binlog
    events is of no use. (Of-course there is no harm in generating these events
    and discarding them w/o using them).

  - Fix identifies the said condition and block generation of such events
    if running in cluster mode.


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/135/

PR #813, opened 2019-04-12
- PXC#2528: mysql-upgrade log during SST should reside in data-directory

kbauskar:
- SST now start intermediate server to help process mysql-upgrade.

  - This could be a time consuming process and user may need access to server
    log during this time to understand how it is making progress.
    (Especially if donor is 5.7 then server is started 3 times).

  - Existing behavior used upgrade tmp-directory for this log.
    (this tmp-directory is removed post upgrade).
    User may also need to re-review this log later for better understanding.

  - Given all these use-cases patch now modifies the location of
    server-log, started for post-processing, in the data-directory.

  In addition to above issue following issues are fixed:

  - upgrade (mysql_upgrade output log) is also saved in data-directory.

  - upgrade selection messages based on versioning are improved and
    are now logged with info level (vs debug level).

  - If donor is 5.7 then server is started 3 times all 3 instance logs
    are appended to same log file for later review.

  (These log files will continue to reside in data-directory unless
   user decide to remove them or next sst/ist get rid of them).

  - mysqld.post.processing.log (created only if server is restarted for post-processing)
  - mysqld.upgrade.log (created only if upgrade action is executed)

### DIFF
--- a/scripts/wsrep_sst_common.sh
+++ b/scripts/wsrep_sst_common.sh
@@ -47,7 +47,7 @@ declare WSREP_LOG_DIR=""
 declare WSREP_WSREP_SCHEMA_VERSION=""
 
 # The WSREP schema version found in the local wsrep_state.dat file
-declare DATADIR_WSREP_SCHEMA_VERSION=""
+declare DATADIR_WSREP_SCHEMA_VERSION="0.0.0"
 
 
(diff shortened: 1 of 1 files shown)


kbauskar:
LGTM
mtr: https://jenkins.percona.com/job/pxc80.test.mtr/136/

PR #819, opened 2019-05-21
PXC-2537 : Nodes crash after using mysqladmin password command

kennt-percona:
Issue
The mysqladmin command uses "ALTER USER USER()" to change the password.
This will crash replica nodes since they are running on an applier
thread instead of a client thread (and thus have no SQL user).

Solution
There was a previous patch to have the server reject commands that
use USER()/CURRENT_USER().  However, this will cause the password
command in mysqladmin to fail.  This patch adds an option "--use-set-password"
that will force mysqladmin to use "SET PASSWORD" instead of "ALTER USER"
to set the password, thereby bypassing the USER/CURRENT_USER issue.


kennt-percona:
https://jenkins.percona.com/job/pxc57.test.mtr/600/

kbauskar:
LGTM

PR #834, opened 2019-07-22
PXC-2432 : userstat / I_S.{client,user}_statistics missing most of data

kennt-percona:
Issue
PXC is not updating the user stats.

Solution
The WSREP code added an if/else block, and the stat updates were being
done only in the else block.  Replicated the stat updates to the
if block.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/588/

kbauskar:
LGTM

PR #838, opened 2019-07-24
PXC-2555 : SST initialization delay

kennt-percona:
Issue
In certain cases (most notably, running the PXC docker images for
kubernetes), the SST logic to find the parent pid would fail (due
to the pid being at the end of the list of pids).

Solution
Change the grep to look for the pids correctly.


kennt-percona:
https://jenkins.percona.com/job/pxc56.test.mtr/590/

kbauskar:
lgtm

PR #842, opened 2019-08-05
Added xbstreameopts parsing on read_cnf()

vgrippa-zz:
The current version does not accept xbstream parametrization. Adding the xbstreameopts to the read_cnf() allows xbstream to be customized (using parallel for example).

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -573,6 +573,7 @@ read_cnf()
     ttime=$(parse_cnf sst time 0)
     scomp=$(parse_cnf sst compressor "")
     sdecomp=$(parse_cnf sst decompressor "")
+    xbstreameopts=$(parse_cnf sst xbstreameopts "")
 
     # if wsrep_node_address is not set raise a warning
     wsrep_node_address=$(parse_cnf mysqld wsrep-node-address "")


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -573,6 +573,7 @@ read_cnf()
     ttime=$(parse_cnf sst time 0)
     scomp=$(parse_cnf sst compressor "")
     sdecomp=$(parse_cnf sst decompressor "")
+    xbstreameopts=$(parse_cnf sst xbstreameopts "")
```
delgod, line 576: hm, looks like `xbstreameopts` become overwritten later if encrypt is enabled.
lines 1647 and 1681

kennt-percona:
Creating a separate bug for this: https://jira.percona.com/browse/PXC-2602
Do you think we should add some of these options as defaults?

vgrippa-zz:
I already created a JIRA for this:

https://jira.percona.com/browse/PXC-2597

I think the default parametrization should be single-threaded and no
encryption.

*Vinicius M. Grippa*

*Celular / Cell Phone:* + 39 349 8389 165
*Skype:* vgrippa

*Antes de imprimir pense em seu compromisso com o meio ambiente.*
*Before printing, think about your commitment to the Environment.*

kennt-percona:
Added an additional option:
  xbstream_opts (to keep it separate from the encryption options, xbstream_eopts)

See: https://github.com/percona/percona-xtradb-cluster/pull/921

PR #862, opened 2019-10-10
- PXC-2655: Setting pxc-encrypt-cluster-traffic=ON by default

kbauskar:
Issue:
  -----

  * PXC Cluster is exposed to risk of un-authorized data access (by default)
    as joining node doesn't need to handshake any credentials.

  * By turning pxc-encrypt-cluster-traffic=ON by default joining nodes
    will need approved SSL certificates. User can set it to OFF if user
    trust the network enviornment.


kbauskar:
mtr: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/63/
LGTM

PR #867, opened 2019-10-30
PXC-2667 : Updating a row with a variable-length unique key (that con…

kennt-percona:
…flicts) causes the entire cluster to go down

Issue:
This is due to a bug in the certification code. If two different rows are
updated (and are changing the unique key to the same value, in this case ''). For
any unique key, the certification code checks for differences between the old value
and new value for the row, and adds the new value for the keys to be checked
if the old value and new value are different.  However, this comparison was
not correct for variable-length keys, so the new key was not being added.

Solution:
Fix the comparison to work correctly for variable-length keys.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -10845,7 +10845,7 @@ ha_innobase::wsrep_append_keys(
 		}
 
 		for (i=0; i<table->s->keys; ++i) {
-			uint  len;
+			uint  len0, len1;
 			char  keyval0[WSREP_MAX_SUPPORTED_KEY_LENGTH+1] = {'\0'};
 			char  keyval1[WSREP_MAX_SUPPORTED_KEY_LENGTH+1] = {'\0'};
 			char* key0 		= &keyval0[1];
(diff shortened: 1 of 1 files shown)


kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/24/

storage/innobase/handler/ha_innodb.cc
```diff
@@ -10892,15 +10892,15 @@ ha_innobase::wsrep_append_keys(
 						    wsrep_thd_query(thd));
 				}
 				if (record1) {
-					len = wsrep_store_key_val_for_row(
+					len1 = wsrep_store_key_val_for_row(
 						thd, table, i, key1,
 						WSREP_MAX_SUPPORTED_KEY_LENGTH,
 						record1, &is_null, prebuilt);
-					if (!is_null && memcmp(key0, key1, len)) {
+					if (!is_null && (len0 != len1 || memcmp(key0, key1, len1))) {
```
kbauskar, line 10899: For empty string len1 = 0 amd so memcmp will continue to return 0 as the previous case was. May be I missed something when we dicussed about this fix.
len0 != len1 .. this make sense and is needed.

kennt-percona: memcmp is only called when both lengths are the same (len0 == len1), so if they're both 0, then it's ok for memcmp to return 0 (they're equal).

kennt-percona: key1 is only appended if it is different from key0, so check if the lengths are different then check the contents are different (if needed).

kbauskar (approved):

PR #885, opened 2019-12-11
PXC-2194: Upgrade procedure - reword warning about SST

borys-belinsky:
The problem reported in this (rather old) issue is that the warning at the top was not clear and did not provide specific actions for the user.


rameshvs02 (commented):
LGTM

kennt-percona (approved):

PR #903, opened 2020-01-08
- PXC-2629: crash during concurrent load.

kbauskar:
- Scenario is trying to ALTER TABLESPACE (rename) and fire dml workload
    on table that are part of the same tablespace.

  - This creates a unique problem.
    - ALTER TABLESPACE fails to projects tables that resides in TABLESPACE
      as part of its write-set signature.
    - MySQL system as per agreed semantics take needed MDL lock.
    - ALTER TABLESPACE being a TOI command if it faces problem acquiring lock
      it can bf-abort local running transaction.
    - Local transaction on hitting bf-abort evaluate for replay.
    - REPLAY logic is dependent on certification that in turn depends on
      write-set signature (object that write-set modifies).
    - Given write-set signature for ALTER TABLESPAE is not proper RELAY
      logic feel these 2 transaction don't conflict so allow local transaction
      to REPLAY.
    - Replayed transaction runs with high priority and holds needed lock
      and conflict with local running high prority ALTER TABLESPACE.

    - 2 high priority transaction is not expected scenario so flow on detecting
      it abort the server.

  - Fix:
    - ALTER TABLESPACE now projects tables that reside in tablespace as part
      of its write-set signature.
    - ALTER TABLESPACE needs to obtain this list from DD for which it needs
      a MDL lock on TABLESPACE
    - Too restrictive lock can cause deadlock with parallel running DML thread.
    - Discovery flow take IX locks on tablespace but this opens up possibility
      of new table being added to tablespace in parallel.
    - This is handled by a special mutex that help co-ordinate CREATE/ALTER
      TABLE (to a tablespace) action with ALTER TABLESPACE.
      [Mutex shouldn't affect performance here as anyway TOI will streamline
       the operation].

### DIFF
--- a/mysql-test/suite/galera/r/galera_performance_schema.result
+++ b/mysql-test/suite/galera/r/galera_performance_schema.result
@@ -2,12 +2,12 @@
 use performance_schema;
 select count(*) from setup_instruments where name like '%galera%' or name like '%wsrep%';
 count(*)
-91
+92
 #node-2
 use performance_schema;
 select count(*) from setup_instruments where name like '%galera%' or name like '%wsrep%';
 count(*)
-91
+92
 select count(*) = 5 from performance_schema.threads where name like '%galera%' or name like '%wsrep%';
 count(*) = 5
 1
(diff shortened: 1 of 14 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=903) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=903) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/percona-xtradb-cluster?pullRequest=903) it.</sub>

kbauskar:
LGTM
mtr: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/104/

PR #908, opened 2020-01-13
- PXC-2897: PXC node is crashing with replicated 'ALTER .. DISABLE KE…

kbauskar:
…YS' command

  - With introduction of atomic ddl, ddl are committed in 2 phase just like DML.

  - Normal DDL that affects data would start transaction and will eventually
    commit. Said example depict a DDL that doesn't make any modification
    at data level (metadata-only change).

  - Given this md-only change it doesn't open a transaction before it goes for
    commit (prepare + commit). As part of commit-prepare flow if the transaction
    is not started it is committed.

  - WSREP introduced a code to clear xid during transaction start but said
    code is now causing issue with metadata-only DDL as xid is preset before
    transaction is started.

  - There is no clear reason why xid should be reset during trx_start_low
    though it naturally make sense.

  - Fix moves the said code block from trx_start_low to trx_init which is called
    post transaction is committed in memory.

### DIFF
--- a/storage/innobase/trx/trx0trx.cc
+++ b/storage/innobase/trx/trx0trx.cc
@@ -225,6 +225,12 @@ static void trx_init(trx_t *trx) {
 
   trx->stats.set(false);
 
+#ifdef WITH_WSREP
+  if (trx->xid) {
+    trx->xid->reset();
+  }
+#endif /* WITH_WSREP */
+
   /* During asynchronous rollback, we should reset forced rollback flag
   only after rollback is complete to avoid race with the thread owning
   the transaction. */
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=908) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=908) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/percona-xtradb-cluster?pullRequest=908) it.</sub>

kbauskar:
LGTM
mtr run: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/107/

PR #912, opened 2020-01-15
PXC-2684 : Cluster node hangs after Stored Procedure is BF-aborted

kennt-percona:
Issue
When a stored procedure (SP) is BF-aborted, in certain cases, the
SP may continue running.  In which case a deadlock situation may
occur.  The applier thread (which caused the BF-abort), waits for
the locks held by the SP thread.  The SP thread, in turn, is waiting
for the applier thread's transaction to commit (since it is coming
after the applier threads commit).

When using an external cursor, the thread being interrupted is not
setting the error state correctly.  This leads the SP thread to assume
that the error has been handled by an exception handler, which causes
it to rollback the op (just the single op, not the full transaction)
and continue on.

Solution
Within innodb, have the code call trx_is_interrupted(), which will
set the error so that it can be handled properly.

### DIFF
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -8601,6 +8601,12 @@ report_error:
 			table->s->table_name.str);
 	}
 
+#ifdef WITH_WSREP
+	if (trx_is_interrupted(trx)) {
+		error = DB_INTERRUPTED;
+	}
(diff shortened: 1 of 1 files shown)


kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/31/

kbauskar (approved):

PR #916, opened 2020-01-20
PXC-2912 : Not able to switch to 'nc' as a transfer method

kennt-percona:
Issue
SST does not work when using nc (netcat) as the transfer method
(rather than socat).  The problem is that newer versions of nc
require the use of the '-N' flag.
  -N Shutdown the network socket after EOF on stdin

Solution
Add in the '-N' flag on the donor if nc uses the '-N' flag.


kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/34/

kbauskar (approved):

PR #921, opened 2020-01-22
PXC-2602 : add support for xbstream options with wsrep_sst_xtrabackup-v2

kennt-percona:
Issue
There is no way to set additional options for xbstream.  For performance
reasons, we would like to set "--parallel=XXX" for xbstream.

Solution
Add the 'xbstream_opts' option to the [sst] section of the config.
Note that on 5.7 and above, there is a separate xbxtream_eopts (encryption
options)

Also fixed a bug in parse_cnf() that would truncate option values that contained
an equals sign ('=')


kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/35/

kbauskar (approved):

PR #927, opened 2020-01-22
PXC-2904 pass xtrabackup version as a parameter for pxc-56

vorsel:


EvgeniyPatlan (changes requested):
for 5.6 we use pxb-2.2 which is EOL. so you can just add latest 2.2 version

vorsel:
> for 5.6 we use pxb-2.2 which is EOL. so you can just add latest 2.2 version

For now pxb version is passed via jenkins job, currently it is 2.3.10, no code adjustments are needed.

EvgeniyPatlan (approved):

PR #948, opened 2020-02-10
PXC-2950 : PXC 8 3 node setup fails to start from backup

kennt-percona:
Issue
The SST can fail if the donor datadir does not have a wsrep_state.dat
file.  This can occur if PXC 8.0 is run on a PS (or MySQL 8.0) datadir.

Solution
If MySQL has determined that an upgrade is not needed, we still
create the wsrep_state.dat file.

also fixed a bad merge (sql/sys_vars.cc)

### DIFF
--- a/sql/wsrep_utils.cc
+++ b/sql/wsrep_utils.cc
@@ -984,7 +984,24 @@ bool WSREPState::wsrep_schema_version_equals(const char *server_version) {
          wsrep_revision == server_revision;
 }
 
-bool WSREPState::load_from(const char *dir, const char *filename) {
+bool WSREPState::exists(const char *dir, const char *filename)
+{
+  const int   full_path_len = 1024;
+  char        full_path[full_path_len];
+  MY_STAT     stat_area;
+
+  if (snprintf(full_path, full_path_len,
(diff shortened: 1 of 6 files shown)


kennt-percona:
https://pxc.cd.percona.com/job/pxc80-test-mtr/143/

sql/wsrep_utils.cc
```diff
@@ -984,7 +984,24 @@ bool WSREPState::wsrep_schema_version_equals(const char *server_version) {
          wsrep_revision == server_revision;
 }
 
-bool WSREPState::load_from(const char *dir, const char *filename) {
+bool WSREPState::exists(const char *dir, const char *filename)
```
dutow, line 987: Please run clang-format-8, this isn't formatted correctly.

sql/wsrep_utils.h
```diff
@@ -387,13 +387,13 @@ class critical {
 };
 #endif
 
-class WSREPState {
+class WSREPState
+{
```
dutow, line 391: Also incorrect formatting change, the original is correct with clang-format-8.

PR #961, opened 2020-02-24
PXC-2989 : Rolling 5.7 -> 8.0 node upgrade fails during the SST

kennt-percona:
Issue
The SST script is running into an error, but the error output
is using an incorrect variable name.

Solution
Correct the script to use the correct variable.
This is a fix to the error-handling code but does not fix the
underlying problem.


kennt-percona:
This code will be removed in the PR for PXC-2997

PR #966, opened 2020-02-25
PXC-2197: PXB version dep - update

borys-belinsky:
This update removes the text which allowed for ambiguities.


kennt-percona (approved):
lgtm

EvgeniyPatlan:
But we provide pxb binaries in pxc package. so there should not be dependency between pxc and pxb

PR #976, opened 2020-03-03
PXC-2189: Virtual - Replace HAProxy

borys-belinsky:


kennt-percona (approved):
lgtm

rameshvs02 (commented):
ProxySQL2 natively supports PXC. It is handling through table [runtime_]mysql_galera_hostgroups. We should add this detail for PXC-8.0 - ProxySQL configuration. I think we do not officially support ProxySQL-1.4.x with PXC-8.0

borys-belinsky:
> ProxySQL2 natively supports PXC. It is handling through table [runtime_]mysql_galera_hostgroups. We should add this detail for PXC-8.0 - ProxySQL configuration. I think we do not officially support ProxySQL-1.4.x with PXC-8.0

@rameshvs02 In proxysql.rst, there an *important* block that informs that PXC 8.0 will not work with proxysql v1.  This section also informs about the native support

PR #977, opened 2020-03-03
PXC-2993: strict mode vs group replication - add note

borys-belinsky:


doc/source/features/pxc-strict-mode.rst
```diff
@@ -8,6 +8,10 @@ PXC Strict Mode is designed to avoid the use of
 experimental and unsupported features in |PXC|.
 It performs a number of validations at startup and during runtime.
 
+.. important:: 
+
+   The group replication plugin must be disabled for the strict mode to work correctly.
```
dutow: Sorry for the late comment, but this isn't entirely correct: strict_mode can be enabled even if the group replication plugin is loaded, if group replication isn't running.

The limiatiton is that group replication can't be running/active in pxc_strict_mode.

dutow (changes requested):
Sorry, but I have one more change request: now that I also looked at how the page looks like (https://www.percona.com/doc/percona-xtradb-cluster/LATEST/features/pxc-strict-mode.html), not just the diff, this change is at the wrong place.

* In the existing Important requirement block, we should mention that if group_replication is running, pxc_strint code can't be set to encoforcing or master
* below in the validations section, we should add a section about group replication, describing the behavior in a similar way as other validations, describing why (e.g. it is a completely different clustering product, and we only support migration from/to, not actively running them together), and how (disabled - allowed, permission - warnings, enforcing/master - can't be turned on)

kathy-williamson (approved):

PR #979, opened 2020-03-04
PXC-2997 : Remove wsrep_state.dat and supporting code from SST process

kennt-percona:
Issue
The upgrade from PS (with the same PXC version) was not creating
the mysql.pxc.internal.session user (this is because the upgrade
process is not run since it's the same internal version)

Solution
Have the PXC upgrade code always run on startup if an upgrade is
not required.  This will ensure that the internal session user
is always created.

### DIFF
--- a/sql/dd/impl/upgrade/server.cc
+++ b/sql/dd/impl/upgrade/server.cc
@@ -470,6 +470,65 @@ static void create_upgrade_file() {
 
 }  // namespace
 
+#ifdef WITH_WSREP
+
+bool pxc_fix_mysql_tables(THD *thd) {
+  const LEX_CSTRING section_start { STRING_WITH_LEN("#! PXC_SECTION::START") };
+  const LEX_CSTRING section_end { STRING_WITH_LEN("#! PXC_SECTION::END") };
+
+  const char **query_ptr;
+  bool in_pxc_section = false;
+
+  if (ignore_error_and_execute(thd, "USE mysql")) {
+    WSREP_ERROR("Could not run PXC upgrade. Could not change database to 'mysql'");
+    return true;
+  }
+
+  for (query_ptr = &mysql_fix_privilege_tables[0]; *query_ptr != NULL;
+       query_ptr++) {
+
+    if (!strncmp(*query_ptr, section_start.str, section_start.length)) {
+      in_pxc_section = true;
+    }
+    else if (!strncmp(*query_ptr, section_end.str, section_end.length)) {
+      in_pxc_section = false;
+    }
+
+    if (in_pxc_section) {
+      if (ignore_error_and_execute(thd, *query_ptr)) return true;
+    }
+  }
+  return false;
+}
+
+bool upgrade_pxc_only(THD *thd) {
+  Disable_autocommit_guard autocommit_guard(thd);
+  Bootstrap_error_handler bootstrap_error_handler;
+
+  Server_option_guard<bool> acl_guard(&opt_noacl, true);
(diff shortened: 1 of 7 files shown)


sql/dd/impl/upgrade/server.cc
```diff
@@ -470,6 +470,67 @@ static void create_upgrade_file() {
 
 }  // namespace
 
+#ifdef WITH_WSREP
+
+bool pxc_fix_mysql_tables(THD *thd) {
+  const char section_start[] = "#! PXC_SECTION::START";
+  const char section_end[] = "#! PXC_SECTION::END";
+  int section_start_len = strlen(section_start);
```
dutow: For both: could be `const LEX_CSTRING section_start = (STRING_WITH_LEN("...");`

Or at least the lens could be `const size_t` instead of `int`.

sql/dd/impl/upgrade/server.cc
```diff
@@ -470,6 +470,67 @@ static void create_upgrade_file() {
...
+
+    if (!strncmp(*query_ptr, section_start, section_start_len)) {
+      in_pxc_section = true;
+    }
+    if (!strncmp(*query_ptr, section_end, section_end_len)) {
+      in_pxc_section = false;
+    }
+
+    if (in_pxc_section) {
+      if (ignore_error_and_execute(thd, *query_ptr)) return true;
```
dutow, line 498: No additional error message in this case?

sql/dd/impl/upgrade/server.cc
```diff
@@ -470,6 +470,67 @@ static void create_upgrade_file() {
...
+bool pxc_fix_mysql_tables(THD *thd) {
+  const char section_start[] = "#! PXC_SECTION::START";
+  const char section_end[] = "#! PXC_SECTION::END";
+  int section_start_len = strlen(section_start);
+  int section_end_len = strlen(section_end);
+
+  const char **query_ptr;
+  bool in_pxc_section = false;
+
+  if (ignore_error_and_execute(thd, "USE mysql")) {
```
dutow, line 482: Shouldn't we switch back to the current db at the end?

sql/dd/impl/upgrade/server.cc
```diff
@@ -470,6 +470,67 @@ static void create_upgrade_file() {
...
+    return true;
+  }
+
+  for (query_ptr = &mysql_fix_privilege_tables[0]; *query_ptr != NULL;
+       query_ptr++) {
+
+    if (!strncmp(*query_ptr, section_start, section_start_len)) {
+      in_pxc_section = true;
+    }
+    if (!strncmp(*query_ptr, section_end, section_end_len)) {
```
kamil-holubicki: else if

sql/dd/impl/upgrade/server.cc
kennt-percona: code modified to use LEX_CSTRING

sql/dd/impl/upgrade/server.cc
kennt-percona: Our code is a copy of fix_mysql_tables().  The errors are reported within the ignore_error_and_execute() function.

sql/dd/impl/upgrade/server.cc
kennt-percona: as above, we're following the same conventions as the upgrade code (see fix_mysql_tables()).

kamil-holubicki (approved):

PR #985, opened 2020-03-05
PXC-3016: 8.0 set persist binlog_format='STATEMENT' should not succeed

kamil-holubicki:
https://jira.percona.com/browse/PXC-3016

Added validation of binlog_format value in case of SET PERSIST statement.

Note: MySQL does not validate parameters values in case of SET PERSIST_ONLY. This is by design. Unfortunately it may lead to server unable to start if the value is set to incorrect one. This is known issue and is system-wide. It is not fixed as the scope of this commit.

Another solution would be disallowing persisting of binlog_format at all (ROW is default for 8.0, PXC allows only ROW), but then the behavior would be inconsistent with upstream, which allows persisting of this value.

Having this fix, the behavior is consistent with upstream (parameter value is validated for session, global, persist and is not validated for persist_only + persisting of values is allowed)

### DIFF
--- a/mysql-test/suite/galera/r/galera_sbr.result
+++ b/mysql-test/suite/galera/r/galera_sbr.result
@@ -1,3 +1,7 @@
+SET PERSIST binlog_format = 'STATEMENT';
+ERROR 42000: Variable 'binlog_format' can't be set to the value of 'STATEMENT'
+SET PERSIST binlog_format = 'MIXED';
+ERROR 42000: Variable 'binlog_format' can't be set to the value of 'MIXED'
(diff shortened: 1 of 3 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=985) <br/>All committers have signed the CLA.

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/170/

kennt-percona (approved):
lgtm

PR #995, opened 2020-03-10
PXC-3017 : Remove these SST encryption methods. encrypt=1, encrypt=2,…

kennt-percona:
… and encrypt=3

Issue
encrypt=1,2,3 has been deprecated.  So this now removes support for these options.

Solution
Remove the code for these options from the SST scripts.  Also, the SST will
fail with a fatal error if these options are still set.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -33,15 +33,7 @@
 #
 
 # encryption specific variables.
-ealgo=""
-ekey=""
-ekeyfile=""
 encrypt=0
-ieopts=""
-xbstream_eopts=""
-xbstream_eopts_sst=""
-xbstream_eopts_other=""
-
 xbstream_opts=""
 
 nproc=1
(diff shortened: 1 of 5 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -196,110 +183,6 @@ timeit()
...
-{
-    # $encrypt -eq -1 is for internal purposes only
-    if [[ $encrypt -ge 2 || $encrypt -eq -1 ]]; then
-        return
-    fi
-
-    # TODO: it seems like encrypt=1/2/3/4 was not provided
-    if [[ $encrypt -eq 0 ]]; then
-        if $MY_PRINT_DEFAULTS -c $WSREP_SST_OPT_CONF xtrabackup | grep -q encrypt; then
-            wsrep_log_warning "Encryption configuration mis-match. SST may fail. Refer to http://www.percona.com/doc/percona-xtradb-cluster/manual/xtrabackup_sst.html for more details."
```
kamil-holubicki, line 215: Are we sure we want to get rid of this message if encryption is disabled?

kennt-percona: message moved, see below

kamil-holubicki: Sorry, but I can't see where we print the message if encrypt == 0. Maybe it is not important anymore, so just consider my comment like a question, not the recommendation.

kennt-percona: The encrypt setting doesn't matter anymore.  Using the xtrabackup encryption settings will ALWAYS cause an SST failure  (now that we've removed encrypt=1, which is the only scenario where it might have worked).

kamil-holubicki (approved):

kennt-percona:
https://pxc.cd.percona.com/job/pxc80-test-mtr/192/

PR #1009, opened 2020-03-17
PXC-3051: ACL cache transaction conflict causes assertion

kamil-holubicki:
https://jira.percona.com/browse/PXC-3051

Fixed the problem of brute force aborting event transaction at the stage of accessing ACL cache (MDL conflict). There were several problems:
1. In such case event transaction (victim thread) is not started yet, so BF thread attempted to start it twice, but providing -1 as transaction ID. -1 is special identifier for uninitialized transaction. Such start causes transaction to be in inconsistent state (-1 as id, but flagged as flag::start_transaction). Such start was done twice (in wsrep_thd_bf_abort() and wsrep_bf_abort()). The 2nd call detected inconsistent state and raised assert. Unnecessary call to wsrep_start_transaction() was removed.
2. After fixing above problem, transaction was still in inconsistent state, and was aborted. It caused another assert at the later stage when actual event execution was done. There was attempt to start transaction again (as its id was still -1), but still in inconsistent state with flags equal flag::start_transaction. As the transaction was BF aborted, it should not execute. => Added initialization of wsrep_next_trx_id, which allowed detection of transaction BF abortion after ACL cache access. Additionally improved handling transaction result at the end of event thread execution.

### DIFF
--- a/mysql-test/suite/galera/r/galera_events3.result
+++ b/mysql-test/suite/galera/r/galera_events3.result
@@ -1,0 +1,13 @@
+call mtr.add_suppression("Operation DROP USER failed");
+call mtr.add_suppression("Operation DROP USER failed");
+CREATE TABLE t1(a INT AUTO_INCREMENT, b INT, PRIMARY KEY pka(a));
+SET DEBUG_SYNC = 'now WAIT_FOR has_global_grant.got_lock'; ;
+SET @@global.debug = '+d,has_global_grant_hold_lock';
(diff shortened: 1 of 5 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=1009) <br/>All committers have signed the CLA.

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/180/

kennt-percona (approved):
lgtm

PR #1021, opened 2020-03-20
PXC-3075 : centos8 fails testcase galera_sst_xtrabackup-v2_keyring

kennt-percona:
Issue
The galera-cert.pem and galera-key.pem are causing errors when used
by socat(v1.7.3.2) on Centos8.  The same version of socat on ubuntu
works with no errors.

Solution
Have the test case use the server-cert.pem and server-key.pem intead
of the galera ones.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.cnf
+++ b/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.cnf
@@ -6,8 +6,8 @@ wsrep_debug=1
 
 [sst]
 encrypt=3
-tkey=@ENV.MYSQL_TEST_DIR/std_data/galera-key.pem
-tcert=@ENV.MYSQL_TEST_DIR/std_data/galera-cert.pem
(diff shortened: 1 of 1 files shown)


dutow:
The issue seems to be that the galera key has a 1024 bit dhparam block, which isn't secure, and rejected by openssl? And the test also provides a separate 2048 bit dhparam key, but socat isn't aware of that.

* if we want to support this scenario (having a key with a small dhparam block, but using a separate dhparam key), we should fix the issue
* if we no longer want to support this, we should change the dhparam block in the galara key to a 2048 bit block, and then we can continue using that key

Also related: based on your blog post from 2017, the ssl_dhparams parameter was introduced to fix the reverse issue (older versions of socat couldn't handle large keys). But it seems that all platforms pxc 8.0 support have at least socat 1.7.3, do we still need this parameter?

dutow:
Also this will conflict with #995

kamil-holubicki (approved):
lgtm

PR #1028, opened 2020-03-23
PXC-3094: SIGSEGV on startup with --daemonize

dutow:
Issue: daemonize calls unireg_abort, which also includes some WSREP
specific checks in PXC. Daemonize also does this quite early during
startup, when WSREP isn't initialized yet.

Fix: only do WSREP specefic checks in unireg_abort if Wsrep_server_state
is already initialized.

### DIFF
--- a/sql/wsrep_server_state.h
+++ b/sql/wsrep_server_state.h
@@ -36,6 +36,8 @@ class Wsrep_server_state : public wsrep::server_state {
   static void destroy();
   static Wsrep_server_state &instance() { return *m_instance; }
 
+  static bool initialized() noexcept { return m_instance != nullptr; }
+
   static wsrep::provider &get_provider() { return instance().provider(); }
 
   static bool has_capability(int capability) {
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -2380,8 +2380,9 @@ static void unireg_abort(int exit_code) {
   if (!daemon_launcher_quiet && exit_code) LogErr(ERROR_LEVEL, ER_ABORTING);
 
 #ifdef WITH_WSREP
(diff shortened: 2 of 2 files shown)


dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/12/#showFailuresLink

sql/wsrep_server_state.h
```diff
@@ -36,6 +36,8 @@ class Wsrep_server_state : public wsrep::server_state {
   static void destroy();
   static Wsrep_server_state &instance() { return *m_instance; }
 
+  static bool has_instance() noexcept { return m_instance != nullptr; }
```
kamil-holubicki: m_instance is created in init_once method. Would bool initialized() be the better name for this?

dutow: This method will also return false during shutdown, after the instance is already destructed. So I'm not sure about `initialized`, as technically, it was already initialized before.

kamil-holubicki: Wsrep_server_state::init_once() seems to be complementary function to Wsrep_server_state::destroy()., so init_once -> initialized -> destroy -> not initialized -> init_once -> initialized.

Anyway, this is just a matter of naming convention and in my opinion has_instance() exposes internal detais of the object implementation.

Calling of get_provider() will be problematic as well, because it calls instance() internally. So someone should call: if (wss.has_instance()) wss.get_provider() which looks strange to me.
Side note: better interface would be if instance() and get_provider() throw exception in case of internal m_instance being null, or return pointers, not references.

dutow: Changed the name.

kamil-holubicki (approved):

PR #1029, opened 2020-03-23
PXC-3072: simplify the development environment

dutow:
A first implementation of the "one command develepment environment" for
PXC.

This change is completely opt-in: without specifying additional
parameters, everything works as before (no change required for
CI/packaging/habits)

(except for the mtr & rocks/toku change, that's default)

This commit:
* adds galera (re)building to the basic build command (make/ninja)
* galera changes are automatically discovered, and the binaries are
rebuilt. No manual scons calls / file copying required.
* adds downloading, building and installing pxb 2.4 and 8.0 to the basic
build command (make/ninja)
* every binary (pxb, galela library, garbd) is installed at the correct
location for both the build tree and the install tree (mtr can be run in
both locations without manual changes)
* mtr also discovers the galera library in the build / install tree (no
manual WSREP_PROVIDER variable required after this)
* disables TokuDB/Rocksdb by default. As we never build these with PXC,
this is a better default

How to use:
* add -DWITH_GALERA_DEV=ON -DWITH_PXB_DEV=ON to the CMake
command. (independent variables, both work without the other too!)
* Execute make/ninja/whatever
* Enjoy

Note that the first build will be slower, as it downloads 2 versions of
PXB, and builds both PXB versions and galera.

Limitations:
* pxb update checks are disabled to avoid PXB rebuilds on every make
command. As PXC rarely needs changes / updates in PXC, this is a good
compromise for now, and it is easy to rebuild PXB manually when needed.
If required a "rebuild it now" option can be added later.
* Galera tests are disabled to reduce galera (re)build time. Again, an
option can be added later if needed. For now, tests can be executed
manually using scons.


dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/12/

kennt-percona (approved):
lgtm (nice to remove the need for WSREP_PROVIDER)

dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/21/#showFailuresLink

PR #1031, opened 2020-03-24
PXC-3087, PXC-3088, PXC-3079: Relax strict mode admin checks

dutow:
Issue: strict mode admin checks reported an error for non existend
tables, even when pxc_strict_mode wasn't enabled. As MySQL normally
allows non existed tables in admin commands, this resulted in many test
failures in various test suites, and also changed how ANALYZE TABLE,
OPTIMIZE TABLE and similar commands handle errors.

Fix: ignore non existent tables in the strict mode checks.


dutow:
galera_flush_local is failing on jenkins, looks like that's related why originally this bug was introduced in PXC-267, I'll fix that & update.

dutow:
Test only needed the removal of an expected error and rerecording, updated.

dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/14/

kamil-holubicki (approved):
lgtm

PR #1037, opened 2020-03-31
PXC-2999 : DML commands are not replicating to PXC-8.0 nodes through …

kennt-percona:
…MySQL shell

Issue
DML commands (coming from mysqlsh) are not replicating to the other nodes
This is because the SQL sessions/threads used by the mysqlsh are not
wsrep enabled.

Solution
Setup the threads with wsrep.

### DIFF
--- a/mysql-test/suite/galera/t/pxc_mysqlx.test
+++ b/mysql-test/suite/galera/t/pxc_mysqlx.test
@@ -1,0 +1,281 @@
+# pxc_mysqlx
+#
+# Simple tests of using the mysqlx plugin with PXC
+#
+--source include/galera_cluster.inc
+--source include/force_restart.inc
+
+--source include/have_mysqlx_plugin.inc
+--source include/xplugin_create_user.inc
+
+# Setup the tables for the test
+# (sql interface) - test basic CRUD scenarios
+# (native interface) - test basic CRUD scenarios
+
+## TEST STARTS HERE
+
+# Setup the test database and test simple CRUD operations
+# and see if they replicate
+--echo # Testing CREATE TABLE DDL and INSERT DML (SQL)
+--connection node_1
+--echo # node 1
+
+--write_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+-->sql
+ DROP DATABASE IF EXISTS dbtest;
+ CREATE DATABASE dbtest;
+ USE dbtest;
+
+ CREATE TABLE t1 (
+  id INT PRIMARY KEY,
+  f2 TEXT);
+
+ INSERT INTO t1 VALUES (1, "foo");
+ INSERT INTO t1 VALUES (2, "bar");
+-->endsql
+EOF
+
+--exec $MYSQLXTEST -u x_root --password='' --file=$MYSQL_TMP_DIR/pxc_mysqlx_script.tmp 2>&1
+--remove_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+
+--connection node_1
+--echo # node 1
+SELECT * FROM dbtest.t1;
+
+--connection node_2
+--echo # node 2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbtest' AND TABLE_NAME = 't1'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 2 FROM dbtest.t1
+--source include/wait_condition.inc
+
+SELECT * FROM dbtest.t1;
+
+
+--echo # Testing DELETE DML (SQL)
+--connection node_1
+--write_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+-->sql
+ DELETE FROM dbtest.t1 WHERE id = 1;
+-->endsql
+EOF
+
+--exec $MYSQLXTEST -u x_root --password='' --file=$MYSQL_TMP_DIR/pxc_mysqlx_script.tmp 2>&1
+--remove_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+
+--connection node_1
+--echo # node 1
+SELECT * FROM dbtest.t1;
+
+--connection node_2
+--echo # node 2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM dbtest.t1
+--source include/wait_condition.inc
+
+SELECT * FROM dbtest.t1;
+
+
+--echo # Testing UPDATE DML (SQL)
+--connection node_1
+--write_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+-->sql
+ UPDATE dbtest.t1 SET f2="baz" WHERE id = 2;
+ INSERT INTO dbtest.t1 VALUES (3, "foobar");
+-->endsql
+EOF
+
+--exec $MYSQLXTEST -u x_root --password='' --file=$MYSQL_TMP_DIR/pxc_mysqlx_script.tmp 2>&1
+--remove_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+
+--connection node_1
+--echo # node 1
+SELECT * FROM dbtest.t1;
+
+--connection node_2
+--echo # node 2
+--let $wait_condition = SELECT COUNT(*) = 2 FROM dbtest.t1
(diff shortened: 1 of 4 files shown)


mysql-test/suite/galera/include/have_mysqlx_plugin.inc
```diff
@@ -0,0 +1,9 @@
+if (`SELECT count(plugin_status)=0 FROM information_schema.plugins WHERE plugin_name LIKE 'mysqlx'`)
```
dutow: This file already exists in `mysql-test/include`

mysql-test/suite/galera/include/xplugin_create_user.inc
```diff
@@ -0,0 +1,22 @@
+#==== Usage ====
```
dutow: This also exists in `include/`

mysql-test/suite/galera/include/xplugin_drop_user.inc
```diff
@@ -0,0 +1,3 @@
+--disable_query_log
```
dutow: This also exists in include/

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+--echo # node 1
+
+--write_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+-->sql
+ DROP DATABASE if EXISTS dbtest;
+ CREATE DATABASE dbtest;
+ USE dbtest;
+
+ CREATE TABLE t1 (
+  id int PRIMARY KEY,
```
dutow: INT

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+  f2 text);
```
dutow: TEXT

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+
+# Setup the test database and test simple CRUD operations
+# and see if they replicate
+--echo ## Testing CREATE TABLE DDL and INSERT DML (SQL)
+--connection node_1
+--echo # node 1
+
+--write_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+-->sql
+ DROP DATABASE if EXISTS dbtest;
```
dutow: IF

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+ INSERT INTO t1 VALUES (2, "bar");
+-->endsql
+EOF
+
+--exec $MYSQLXTEST -u x_root --password='' --file=$MYSQL_TMP_DIR/pxc_mysqlx_script.tmp 2>&1
+--remove_file $MYSQL_TMP_DIR/pxc_mysqlx_script.tmp
+
+--connection node_1
+--echo # node 1
+select * from dbtest.t1;
```
dutow: SELECT... and so on, use upper case for SQL everywhere

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = 'dbtest' AND TABLE_NAME = 't1'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 2 FROM dbtest.t1
+--source include/wait_condition.inc
+
+select * from dbtest.t1;
+
+#
+# Now run the delete test
```
dutow: These comments above the echos are redundant

mysql-test/suite/galera/t/pxc_mysqlx.test
```diff
@@ -0,0 +1,284 @@
...
+--connection node_2
+--echo # node 2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM dbtest.t1
+--source include/wait_condition.inc
+
+select * from dbtest.t1;
+
+
+# Now run the update case
+--echo ## Testing UPDATE DML (SQL)
```
dutow: Double #

sql/srv_session.cc
```diff
@@ -792,6 +796,7 @@ bool Srv_session::open() {
   DBUG_ASSERT(state == SRV_SESSION_CREATED || state == SRV_SESSION_CLOSED);
 
   thd.push_protocol(&protocol_error);
+
```
dutow: accidental change?

sql/srv_session.cc
```diff
@@ -1107,8 +1129,16 @@ int Srv_session::execute_command(enum enum_server_command command,
...
+#ifdef WITH_WSREP
+  if (thd.wsrep_cs().state() == wsrep::client_state::s_exec) {
+    wsrep_after_command_before_result(&thd);
+  }
+  wsrep_after_command_after_result(&thd);
+#endif /* WITH_WSREP */
+
   thd.pop_protocol();
   DBUG_ASSERT(thd.get_protocol() == &protocol_error);
+
```
dutow: Unnecessary change

sql/srv_session.cc
```diff
@@ -841,6 +846,14 @@ bool Srv_session::open() {
     return true;
   }
 
+#ifdef WITH_WSREP
+  if (thd.wsrep_cs().state() != wsrep::client_state::s_none) {
+    thd.wsrep_cs().cleanup();
+  }
+  thd.wsrep_client_thread = true;
```
dutow, line 853: Shouldn't we lock LOCK_wsrep_thd before changing this? wsrep_sync_wait depends on it

dutow (changes requested):
Could we also add a case for a failing DDL/DML (to ensure that the server doesn't crash / work correctly)?

sql/srv_session.cc
```diff
@@ -1095,6 +1113,10 @@ int Srv_session::execute_command(enum enum_server_command command,
 
   mysql_audit_release(&thd);
 
+#ifdef WITH_WSREP
+  wsrep_before_command(&thd);
```
kamil-holubicki, line 1120: If we look at "normal" flow, in sql_parse.cc do_command() which is caller of dispatch_command() (so is equivalent of this function) we call wsrep_before_command() as well, but we have more sophisticated logic around this which detects BF aborted transactions done by background rollbacker.
There is also logic that blocks queries in case of snapshot was not installed.

Don't we need this logic here as well?

kennt-percona: I followed the async slave (rpl) implementation for these calls, which doesn't implement the retry logic at this level.  It's probably too complicated to add at this point.

sql/srv_session.cc
dutow: Question not answered.

sql/srv_session.cc
dutow: Still there.

sql/srv_session.cc
dutow: Still there.

mysql-test/suite/galera/t/pxc_mysqlx.test
dutow: Still there.

mysql-test/suite/galera/t/pxc_mysqlx.test
dutow: Please change the  keywords in the test file to upper case.

dutow:
"Could we also add a case for a failing DDL/DML (to ensure that the server doesn't crash / work correctly)?" - I also don't see a testcase for this?

dutow:
@kennt-percona As I said on the standup, we should fix the above comments. But as these are all about tetst (except the locking question), we can also create a separate jira ticket about fixing them and merge this so we don't block the GA build with test fixes.

mysql-test/suite/galera/t/pxc_mysqlx.test
kennt-percona: oops, forgot to push my changes

dutow: This one is still there, other test changes look good now.

dutow (approved):
Please fix that ##, otherwise the tests look good now.

PR #1040, opened 2020-04-01
PXC-3109: Shutdown trigger sig 11 assertion : Mysqld_socket_listener::close_listener

kamil-holubicki:
https://jira.percona.com/browse/PXC-3109

Problem:
If --create-admin-listener-thread=ON and --admin-address is not specified, there was still attempt to deinitialize the thread and socket in Mysqld_socket_listener::close_listener(). Thread and socket should be deinitialized only if both were initialized.

Fixed deinitialization of admin listener thread and socket

Additionally fixed admin_interface MTR test by skipping drop attempt of PXC sst user in case when server was started with --skip-grant-tables option. Such drop would fail anyway printing out error message to error log.

### DIFF
--- a/sql/conn_handler/socket_connection.cc
+++ b/sql/conn_handler/socket_connection.cc
@@ -1471,7 +1471,14 @@ void Mysqld_socket_listener::close_listener() {
     being listened is not included in the m_socket_map. Instead, this socket
     referenced by the data member m_admin_interface_listen_socket.
   */
-  if (m_use_separate_thread_for_admin) {
+
+  /*
+    See the logic in setup_listener()
+    Thread was spawned only if both --admin-address is specified
+    and --create-admin-listener-thread=ON
+   */
+  if (!m_admin_bind_address.address.empty() &&
+      m_use_separate_thread_for_admin) {
 #ifdef _WIN32
     /*
       For Windows, first close the socket referenced by the data member
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -26,6 +26,7 @@
 #include "rpl_msr.h"  // channel_map
 #include "rpl_slave.h"
 #include "sql/sql_lex.h"
+#include "sql/auth/auth_common.h"
 #include "sql_base.h"  // TEMP_PREFIX
 #include "sql_class.h"
 #include "sql_parse.h"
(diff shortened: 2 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/190/

sql/wsrep_sst.cc
```diff
@@ -1127,6 +1128,12 @@ int wsrep_remove_sst_user(bool initialize_thread) {
   int err = 0;
   MYSQL_SESSION session = NULL;
 
+  // Skip the attempt to  mysql.pxc.sst.user in case the server was started with
+  // --skip-grant-tables option. It would fail enyway with error.
+  // This will prevent writing out error to error log.
+  if (skip_grant_tables()) {
```
dutow, line 1134: This won't fix the problem completely, as there are a few other edge cases when we can't execute this statement during shutdown, I ended up removing these startup drop statements completely, as some of them can't be fixed, and we start the server with `DROP IF EXISTS` anyway.

We can keep this here, but I'll completely remove this method when I finally have a chance to finish up my bootstrap script cleanup PR.

sql/conn_handler/socket_connection.cc
```diff
@@ -1471,7 +1471,14 @@ void Mysqld_socket_listener::close_listener() {
...
   */
-  if (m_use_separate_thread_for_admin) {
+
+  /*
+    See the logic in setup_listener()
+    Thread was spawned only if both --admin-address is specified
+    and --create-admin-listener-thread=ON
+   */
+  if (!m_admin_bind_address.address.empty() &&
+      m_use_separate_thread_for_admin) {
```
dutow, line 1481: Have you verified that commands executed on the separate admin thread are replicated correctly?

kamil-holubicki: Yes, it works fine. Admin connection differs from the normal user connection only in the process of establishing connection.
1. We have dedicated connection accepting thread
2. There is no limitation for number of connections
So once the connection is established, it is processed by exactly the same mechanism as normal user connection.

sql/wsrep_sst.cc
kamil-holubicki: As discussed on slack, feel free to remove it in the future. This just resurrects the MTR test that was extended by me to test fixed functionality.

dutow (approved):

PR #1041, opened 2020-04-03
PXC-3108: Crash when UNLOCK TABLES called after FTWRL

kamil-holubicki:
https://jira.percona.com/browse/PXC-3108

Problem:
When replication was disabled for given session, and session called FTWRL and then enabled replication and executed UNLOCK TABLES, server crashed.

Cause:
When FTWRL is executed, global read lock is acquired. This causes pausing of replicatoion, as we do not want changes done on another nodes to be replicated to this node.
When UNLOCK TABLES is executed, global read lock is released. This causes resynchronization/resuming of replication.
In the given scenario, replication was not paused, but there was attempt to resume it which caused crash.

Additionally I discovered that not pausing replication server if the given session has disabled replication causes changes from another nodes to be replicated to this node even if FTWRL was executed. Replication should be paused/resumed always regardless of replication being enabled or disabled in given session. Note that wsrep_on variable specifies if changes done on this node are replicated to the cluster. Even if its value is 'OFF', changes from cluster are replicated to this node.

### DIFF
--- a/sql/lock.cc
+++ b/sql/lock.cc
@@ -1096,15 +1096,16 @@ void Global_read_lock::unlock_global_read_lock(THD *thd) {
 
     /* If node not part of the cluster avoid running the resume/pause action */
     if (server_state.state() != Wsrep_server_state::s_disconnected) {
-      if (server_state.state() == Wsrep_server_state::s_donor ||
-          (wsrep_on(thd) &&
-           server_state.state() != Wsrep_server_state::s_synced)) {
+      /*
+       Resume wsrep server even if current thread has disabled replication.
+       We allow other nodes to replicate data to this server.
+       */
+      if (server_state.state() != Wsrep_server_state::s_synced) {
         /* TODO: maybe redundant here?: */
         wsrep_locked_seqno = WSREP_SEQNO_UNDEFINED;
         server_state.resume();
         pause_provider(false);
-      } else if (wsrep_on(thd) &&
-                 server_state.state() == Wsrep_server_state::s_synced) {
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/192/

sql/lock.cc
```diff
@@ -1096,15 +1096,18 @@ void Global_read_lock::unlock_global_read_lock(THD *thd) {
...
     /* If node not part of the cluster avoid running the resume/pause action */
     if (server_state.state() != Wsrep_server_state::s_disconnected) {
+      /*
+       Resume wsrep server even if current thread has disabled replication.
+       We allow other nodes to replicate data to this server.
+       */
       if (server_state.state() == Wsrep_server_state::s_donor ||
-          (wsrep_on(thd) &&
-           server_state.state() != Wsrep_server_state::s_synced)) {
+          server_state.state() != Wsrep_server_state::s_synced) {
```
dutow: Isn't this if now basically `if(server_state.state() != Wsrep_server_state::s_synced)`? Do we really want to do this even in disconnected/initializing state?

And what if we built the server with WITH_WSREP, but aren't running a cluster? Do normal, non galera ftwrl tests pass after this change?

kamil-holubicki: This part of code is not executed when server is disconnected (see 1 "if" above)
I have checked all MTR tests which use FTWRL. No issues noticed.

dutow (approved):

PR #1042, opened 2020-04-03
PXC-3128: High priority transactions should work with WSREP

dutow:
Issue: when wsrep is running, high priority transactions weren't able to
kill normal priority transactions. This basically broke the scenario
where a PXC node is also part of a normal replication setup.

Root cause: we intentionally broke this feature in 5.7, in commit
5be95d8ff766. The comment/commit message explains that high priority
transactions aren't how galera works, an dshould be disabled.

Compared to that statement codership fixed HPTs also in 5.7, in commit
f2dd26e10785, by translating the HPT kill logic into wsrep functions.

After merging the codership fix in 5.7, we never reverted our
workaround.

For 8.0, we didn't even port f2dd26e10785, we only ported our disabling
"patch".

Fix:

* Reverting 5be95d8ff766
* Porting f2dd26e10785 to 8.0/galera4
* Removed the assert(0)
* Adding some of the high priority tests to the galera suite to ensure
that they don't break in the future
* This commit also improves the start_transactin_high_prio mtr include
so it works when used together with --debug

Remaining issue: as shown by two TODO comments in the tests, we report a
different error message in some cases compared to basic InnoDB behavior.
This doesn't result in any other behavior change however, we can fix it
later when time permits.

### DIFF
--- a/mysql-test/suite/galera/t/high_prio_trx_1.test
+++ b/mysql-test/suite/galera/t/high_prio_trx_1.test
@@ -1,0 +1,59 @@
+--source include/galera_cluster.inc
+
+--disable_query_log
+CALL mtr.add_suppression("InnoDB High Priority being used");
+--enable_query_log
+
+# Scenario:
+#  T1=({R(B), W(B)})
+#  T2=({R(B), W(B), C}, HIGH_PRIORITY).
+#
+# Outcome: T1 must abort, T2 must commit.
+
+--source include/count_sessions.inc
+
+CREATE TABLE t1 (c1 INT NOT NULL PRIMARY KEY) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (0);
+
+--connect(con1,localhost,root,,test, $NODE_MYPORT_1)
+--connect(con2,localhost,root,,test, $NODE_MYPORT_1)
+
+--echo
+--echo # On connection 1
+--connection con1
+START TRANSACTION;
+UPDATE t1 SET c1=1 WHERE c1=0;
+
+--echo
+--echo # On connection 2
+--connection con2
+--source include/start_transaction_high_prio.inc
+UPDATE t1 SET c1=2 WHERE c1=0;
+COMMIT;
+--disconnect con2
+
+--echo
+--echo # On connection 1
+--connection con1
+# TODO: this should be ER_ERROR_DURING_COMMIT
+--error ER_LOCK_DEADLOCK
+COMMIT;
+--disconnect con1
+
+--connection node_2
+--let $assert_text= 'There is a 2 in t1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM t1 WHERE t1.c1 = 2, count, 1] = 1
+--source include/assert.inc
+
+--let $assert_text= 'There is no 1 in t1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM t1 WHERE t1.c1 = 1, count, 1] = 0
+--source include/assert.inc
+
+--let $assert_text= 'There is no 0 in t1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM t1 WHERE t1.c1 = 0, count, 1] = 0
+--source include/assert.inc
+
+--connection default
+DROP TABLE t1;
+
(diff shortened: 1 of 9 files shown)


storage/innobase/trx/trx0trx.cc
```diff
@@ -3554,8 +3609,7 @@ void trx_kill_blocking(trx_t *trx) {
     trx_id_t save_id = victim_trx->id;
     victim_trx->id = victim_id;
     // TODO: G-4
```
kamil-holubicki: Do we still need this 'todo'?

storage/innobase/trx/trx0trx.cc
```diff
@@ -3400,14 +3402,67 @@ void trx_set_rw_mode(trx_t *trx) /*!< in/out: transaction that is RW */
...
+  wsrep_thd_UNLOCK(thd);
+
+  if (wsrep_thd_bf_abort(bf_thd, thd, true)) {
+    if (victim_trx->lock.wait_lock) {
+      WSREP_DEBUG("victim has wait flag: %lu", thd_get_thread_id(thd));
+      lock_t *wait_lock = victim_trx->lock.wait_lock;
+
+      if (wait_lock) {
+        WSREP_DEBUG("canceling wait lock");
+        victim_trx->lock.was_chosen_as_deadlock_victim = TRUE;
```
kamil-holubicki: true

mysql-test/suite/galera/t/high_prio_trx_1.test
```diff
@@ -0,0 +1,59 @@
...
+#  T2=({R(B), W(B), C}, HIGH_PRIORITY).
+#
+# Outcome: T1 must abort, T2 must commit.
+
+--source include/count_sessions.inc
+
+CREATE TABLE t1 (c1 INT NOT NULL PRIMARY KEY) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (0);
+
+--connect(con1,localhost,root,,test)
```
kamil-holubicki: multiple occurences.
All galera tests to be sure that we are connected to the proper node use 
--connection node_1 
and if another connection to the same node is needed they create new connection like
--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
--connection node_1a

mysql-test/suite/galera/t/high_prio_trx_2.test
```diff
@@ -0,0 +1,55 @@
...
+
+--echo
+--echo # On connection 1
+--connection con1
+# Note: DEADLOCK implies a full ROLLBACK
+--error ER_LOCK_DEADLOCK
+UPDATE t1 SET c1=1 WHERE c1=0;
+--disconnect con1
+
+--connection default
```
kamil-holubicki: multiple occurences.
Is the idea here check for replication result (use --connection node_2), or what is on the node where we did changes (in such case --connection node_1)

mysql-test/suite/galera/t/high_prio_trx_2.test
```diff
@@ -0,0 +1,55 @@
...
+--let $assert_text= 'There is no 1 in t1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM t1 WHERE t1.c1 = 1, count, 1] = 0
+--source include/assert.inc
+
+--let $assert_text= 'There is no 0 in t1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM t1 WHERE t1.c1 = 0, count, 1] = 0
+--source include/assert.inc
+
+DROP TABLE t1;
+
```
kamil-holubicki, line 57: Multiple occurences:
include wait_unitl_count_sessions.inc

kamil-holubicki (changes requested):

mysql-test/suite/galera/t/high_prio_trx_2.test
dutow: The original tests work on a single node. Here I wanted to test things on another node, to see that the query correctly replicates with galera.

dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/24/

kamil-holubicki (approved):

PR #1049, opened 2020-04-08
PXC-3117 : Transition key is hardcoded

kennt-percona:
Issue
The transition-key was not being set correctly. The bash
variable is not being used properly.

Solution
Use the variables correctly.  Also need to account for compatibility
with older versions.
Also updated the required PXB version.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1284,7 +1284,8 @@ send_data_from_donor_to_joiner()
 }
 
 # Returns the version string in a standardized format
-# Input "1.2.3" => echoes "010203"
+# Input
+#   "1.2.3" => echoes "010203"
 # Wrongly formatted values => echoes "000000"
 normalize_version()
 {
@@ -1292,9 +1293,9 @@ normalize_version()
     local minor=0
     local patch=0
 
-    # Only parses purely numeric version numbers, 1.2.3 
+    # Only parses purely numeric version numbers, 1.2.3
     # Everything after the first three values are ignored
-    if [[ $1 =~ ^([0-9]+)\.([0-9]+)\.?([0-9]*)([\.0-9])*$ ]]; then
+    if [[ $1 =~ ^([0-9]+)\.([0-9]+)\.?([0-9]*)([^ ])* ]]; then
         major=${BASH_REMATCH[1]}
         minor=${BASH_REMATCH[2]}
         patch=${BASH_REMATCH[3]}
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1814,9 +1817,17 @@ then
         DONOR_MYSQL_VERSION=$(parse_sst_info "$sst_file_info_path" sst mysql-version "")
 
         transition_key=$(parse_sst_info "$sst_file_info_path" sst transition-key "")
+
+        # Use the broken key if the donor version is < 5.7.29 and is not 5.7.28-31.31-41.2
```
dutow: The other direction (fixed donor, old receiver) will be unsupported? Because then that's something we should document.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1814,9 +1817,17 @@ then
...
+        # In other words, 5.7.28-31.31-41.2 and above will use the key that was sent
+        if ! check_for_version "$DONOR_MYSQL_VERSION" "5.7.29" &&
```
dutow: The regex in check_for_version seems to require a version in the `X.Y[.Z[.A]+]` format, while this string is in `X.Y.Z-A.B-U.V`

In other occurences, the donor version was truncated before used with it because of this.

dutow (changes requested):

scripts/wsrep_sst_xtrabackup-v2.sh
kennt-percona: yes, the donor does not know the joiner version (so we don't know what transition key to use).  I have added a comment to the jira ticket.
on the other hand, the joiner knows the version the donor is sending, so we can use the correct transition-key.

scripts/wsrep_sst_xtrabackup-v2.sh
kennt-percona: The check for version will normalize the string (in X.Y.Z-A.B-U.V to X.Y.Z), in other places the version is embedded with other text so we need to extract the string.  Here, we know the format coming from the donor.

dutow: The regex in normalize_version doesn't support `-` as a separator, only `.`

dutow: See comment at https://github.com/percona/percona-xtradb-cluster/blob/5.7/scripts/wsrep_sst_xtrabackup-v2.sh#L1843

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1814,9 +1817,17 @@ then
...
+           [[ $DONOR_MYSQL_VERSION != "5.7.28-31-31.41.2" ]]; then
+            transition_key="\$transition_key"
```
kamil-holubicki: Shouldn't we set here
encrypt_move_options="--transition-key=\\$transition_key --generate-new-master-key"
as well?

scripts/wsrep_sst_xtrabackup-v2.sh
kennt-percona: that is fine.  the version check only does major-minor-patch (5.7.28).  It ignores everything else.  The version check above is looking for anything below 5.7.29

dutow (approved):

PR #1051, opened 2020-04-09
PXC-3078: Fixing issues caused by wsrep tables and users

dutow:
Issue: create table / create user / drop user statements were executed
in situations where they shouldn't be, like in read only mode, or during
force recovery.

Fix:
* Not running these statements in these situation, as in that case,
  WSREP won't work anyway, but PXC should be able to start up as a
  single node
* Not doing anything during shutdown: donor thread startup starts
  with DROP IF EXISTS anyway, dropping during shutdown isn't required.
  And rhe way we executed drop statements at the end can't be fixed
  easily in some configurations.
* Also rerecorded many main/innodb testruns


sql/wsrep_sst.cc
```diff
@@ -1297,8 +1301,6 @@ static void *sst_donor_thread(void *a) {
                 strerror(err));
   }
```
kamil-holubicki: Isn't it the security issue that after SST the user remains?

sql/wsrep_sst.cc
```diff
@@ -1132,13 +1133,15 @@ int wsrep_remove_sst_user(bool initialize_thread) {
   // --skip-grant-tables option. It would fail enyway with error.
   // This will prevent writing out error to error log.
   if (skip_grant_tables()) {
-      return ECANCELED;
+    return ECANCELED;
```
kamil-holubicki, line 1136: do we still need wsrep_remove_sst_user() function? It has been removed from sql_parse.cc shutdown() and from sst_donor_thread(). Any other callers?

sql/wsrep_schema.cc
```diff
@@ -539,38 +540,14 @@ Wsrep_schema::Wsrep_schema() {}
...
+bool Wsrep_schema::init(THD *thd) {
   DBUG_ENTER("Wsrep_schema::init()");
-  int ret;
-  THD *thd = new THD;
-  thd->set_new_thread_id();
-  if (!thd) {
-    WSREP_ERROR("Unable to get thd");
-    DBUG_RETURN(1);
+
+  handlerton *ddse = ha_resolve_by_legacy_type(thd, DB_TYPE_INNODB);
```
kamil-holubicki: Do we need it here again? It seems that caller does the same.

kamil-holubicki (commented):
>startup starts with DROP IF EXISTS anyway

Where is it exactly happening?

sql/wsrep_sst.cc
dutow: I readded the sst user removal to the end of the donor thread, as it's safe there. This means that with normal code flow, the sst user will always be removed once sst is done, only the extra removal from server shutdown is gone.

sql/wsrep_schema.cc
dutow: Yes, if I remember correctly it called from somewhere else too - and without adding it to the called, we would create the object, but wouldn't initialize it, which is again an issue.

sql/wsrep_schema.cc
```diff
@@ -539,38 +540,14 @@ Wsrep_schema::Wsrep_schema() {}
...
+  if (ddse->is_dict_readonly && ddse->is_dict_readonly()) {
+    LogErr(WARNING_LEVEL, ER_DD_NO_WRITES_NO_REPOPULATION, "InnoDB", " ");
+    return false;
```
kennt-percona: Shouldn't the return statements be DBUG_RETURN(x)

dutow: Good catch, yes.

sql/wsrep_schema.cc
kamil-holubicki: Declaration of wsrep_schema::init is in wsrep_schema.h. It changed - thd parameter was added. The only place I see its invocation changed is in caller - wsrep_init_schema. So it can not be called from any other place.

dutow: Fixed, it's really enough at one place. The failure I was seeing was related to the missing DBUG_RETURN noticed by Kenn.

sql/wsrep_sst.cc
dutow: The SST user is removed when the donor thread exits, I'm not sure why the other call was there. And even if it remains, it shouldn't be an issue, as the password isn't written down anywhere.

kamil-holubicki (approved):

PR #1052, opened 2020-04-13
PXC-3152: SST fails with the password validation component

dutow:
Issue:
The password validation plugin requires a symbol in the password by
default. PXC doesn't add symbols as some symbols could cause issues with
the use of passwords in bash scripts, and the inclusion of numbers or
upper case letters is also random.

Fix:
We add a few fixed characters to the start of the password, which
guarantees a lower case, an upper case, a symbol and a number in the
password. This way we satisfy the requirements of the validation plugin
without changing the rest of the code.
The fixed string added this way also isn't included in password
dictionaries, which ensures that even the strong setting of the password
validation plugin will work.

Non standard settings, such as requiring more symobls, won't be
supported even with this patch.


kennt-percona (approved):
lgtm

sql/wsrep_sst.cc
```diff
@@ -946,6 +946,8 @@ const std::string get_allowed_pwd_chars() { return g_allowed_pwd_chars; }
 static void generate_password(std::string *password, int size) {
   std::stringstream ss;
   bool srnd;
+  ss << "yx9!A-";
```
kamil-holubicki: Maybe it is worth considering something like
`constexpr const char *prefix = "yx9!A-";`
`ss << prefix;`
`ss << strlen(prefix)`

That will get rid of hardcodded magic values and give  compiler the chance to compile time optimize strlen().

kamil-holubicki (approved):
Approved with minor improvement to consider.

sql/wsrep_sst.cc
dutow: Done.

PR #1055, opened 2020-04-14
PXC-3150: Update the dev build scripts to match the RE team

kennt-percona:
Issue
Dev and RE are generating different version strings.  This is creating
problems (since the differing version strings are messing up the
version comparisons in the SST script)

Solution
Update the dev scripts to generate the correct version strings.


EvgeniyPatlan (approved):

dutow (approved):
Do we also need this for 8.0?

kennt-percona:
Note that this is not needed for 8.0 since we use a different version naming scheme for 8.0 (we do not add on a PXC-specific extension, we use the PS extension).

PR #1056, opened 2020-04-14
8.0 pxc 3117

kennt-percona:


dutow (changes requested):
Null merges should be separate.

Please either:
* Create a separate PR for a null merge first
* Move the 5.7 PR to the current GCA point
* Or cherry-pick the 5.7 commit instead of merging

dutow (approved):

PR #1057, opened 2020-04-14
PXC-3153: SQL failure in SST donor thread leaves the server hanging

dutow:
Issue: if the donor thread exits with an error before the SST process
was started, it leaves the spawning thread waiting for a signal forever.
This results in the entire server hanging.

Fix: also send this signal if donor bootstrap sql statements fail.

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -25,8 +25,8 @@
 #include "mysqld.h"
 #include "rpl_msr.h"  // channel_map
 #include "rpl_slave.h"
-#include "sql/sql_lex.h"
 #include "sql/auth/auth_common.h"
+#include "sql/sql_lex.h"
 #include "sql_base.h"  // TEMP_PREFIX
 #include "sql_class.h"
 #include "sql_parse.h"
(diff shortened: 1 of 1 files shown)


sql/wsrep_sst.cc
```diff
@@ -1194,6 +1193,14 @@ static void *sst_donor_thread(void *a) {
 #ifdef HAVE_PSI_INTERFACE
     wsrep_pfs_delete_thread();
 #endif /* HAVE_PSI_INTERFACE */
+
```
kamil-holubicki, line 1196: When node becomes  a donor its state is shifted:
Shifting DONOR/DESYNCED -> JOINED

If Galera applier thread fails to start sst_donor_thread, it rejoins the cluster. However when sst_donor_thread is started it is detached and it is up to this thread to signal sst finish and trigger rejoining. This is done by calling
Wsrep_server_state::instance().sst_sent(gtid, -err);

I'm afraid that  doing just this will  left the node in desync state.

kamil-holubicki (changes requested):

sql/wsrep_sst.cc
dutow: It works, the server notices the failure and shifts the status back to synced, I verified that.

kamil-holubicki: I did this:

if (true) {
//   if (err) {


Then started cluster (node1) and then joined with node2.
node1 went 

[Galera] server n1 state change: synced -> donor

and stayed in this state.

dutow: As we discussed on slack, in this case, the error isn't set, which is different compared to the real error case.

kamil-holubicki (approved):

PR #1062, opened 2020-04-15
PXC-3157 : 8.0 update build scripts for different pxb tarball names

kennt-percona:
Issue
The pxb tarball names have slightly different formats.  We parse
the names in order to get the full directory name.

Solution
Made the name extraction more generic.


EvgeniyPatlan (changes requested):
it might break release builds. For release build we download  needed package(rpm or deb depends on OS) and prepare tar.gz with correct name percona-xtranackup-{24,80}.tar.gz
so it is better to improve development builds to follow this approach

EvgeniyPatlan (approved):

PR #1063, opened 2020-04-15
PXC-3154 - Thread pooling hangs on shutdown

altmannmarcelo:
Problem
When wsrep_close_client_connections is called, it closes connections
abruptly. This is ok on shutdown, however, same function is also called
when we set wsrep_cluster_address dynamically and when setting
wsrep_reject_queries=ALL_KILL, which fails to deliver
proper notification to thread_pool threads.

Fix
Adjust wsrep_close_client_connections to first signal the thd as killed
then attempt to close the connection.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -2038,6 +2038,25 @@ class Call_close_conn : public Do_THD_Impl {
 
 #ifdef WITH_WSREP
 /**
+  This class implements callback function used by
+  wsrep_close_client_connections() to set KILL_CONNECTION
+  flag on all client thds and awake the thread.
+*/
+class Set_wsrep_kill_client_conn : public Do_THD_Impl {
+ public:
+  Set_wsrep_kill_client_conn() {}
+
+  virtual void operator()(THD *killing_thd) {
+    if (killing_thd->get_protocol()->connection_alive() &&
+       (WSREP(killing_thd) || wsrep_thd_is_local(killing_thd)) &&
+       killing_thd != current_thd) {
+      mysql_mutex_lock(&killing_thd->LOCK_thd_data);
+      killing_thd->awake(THD::KILL_CONNECTION);
+      mysql_mutex_unlock(&killing_thd->LOCK_thd_data);
+    }
+  }
+};
+/**
   This class implements callback function used by close_connections()
   to set KILL_CONNECTION flag on all thds in thd list.
   If m_kill_dump_thread_flag is not set it kills all other threads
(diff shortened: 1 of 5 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=1063) <br/>All committers have signed the CLA.

altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/30/ - full test suite
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/32/ - testing only thread_pool_shutdown (it was force pushed)

sql/mysqld.cc
```diff
@@ -9390,19 +9391,27 @@ int wsrep_wait_committing_connections_close(int wait_time) {
...
-  sql_print_information("Giving %d client threads a chance to die gracefully",
-                        static_cast<int>(thd_manager->get_thd_count()));
 
-  Call_wsrep_close_client_conn call_wsrep_close_client_conn(server_shutdown);
-  thd_manager->do_for_all_thd(&call_wsrep_close_client_conn);
+  /*
+   * wsrep_close_client_connections can be used when there is a cluster
+   * reconfiguration or when we set wsrep_reject_queries=ALL_KILL
+   * on those cases if we are using thread pooling,
+   * we cannot close the connections abruptly otherwise the threa pool workers
```
percona-ysorokin:
```suggestion
   * we cannot close the connections abruptly otherwise the thread pool workers
```

sql/mysqld.cc
```diff
@@ -9390,19 +9391,27 @@ int wsrep_wait_committing_connections_close(int wait_time) {
...
+   * wont get signaled and will hang forever when we shutdown the server.
```
percona-ysorokin:
```suggestion
   * won't get signaled and will hang forever when we shutdown the server.
```

sql/mysqld.cc
```diff
@@ -2057,6 +2057,7 @@ class Set_wsrep_kill_conn : public Do_THD_Impl {
   int get_dump_thread_count() const { return m_dump_thread_count; }
 
   virtual void operator()(THD *killing_thd) {
+    if (killing_thd == current_thd) return;
```
percona-ysorokin: Was it OK before without this check added in existing places where `Set_wsrep_kill_conn` was instantiated? I mean in everywhere else rather than in `wsrep_close_client_connections()` that you added in this commit?

altmannmarcelo: Set_wsrep_kill_conn is currently not used anywhere in the code.
If we look to the comments on `wsrep_close_client_connections ()` https://github.com/percona/percona-xtradb-cluster/blob/8.0/sql/mysqld.cc#L9393 - it was meant to mimic the behavior of `close_connections()` https://github.com/percona/percona-xtradb-cluster/blob/8.0/sql/mysqld.cc#L2232 - which first sends a kill, and later tries to close it abruptly.

`if (killing_thd == current_thd) return;` ensures that the thread executing `SET GLOBAL wsrep_cluster_address = ''` or `SET GLOBAL wsrep_reject_queries=ALL_KILL` will not be also killed. 

However, I've re-checked what Set_wsrep_kill_conn and Call_wsrep_close_client_conn will signal/close and they are not doing the same thing. I will create a new method to signal only client `_client_` part from Call_wsrep_close_client_conn and will leave Set_wsrep_kill_conn intact.

altmannmarcelo:
@percona-ysorokin adjusted the code a bit. Even though `Set_wsrep_kill_con`n is not used anywhere, I left it untouched as it has different parameters that serve other purposes, instead, I've created a new class `Set_wsrep_kill_client_conn` that will follow the same criteria as `Call_wsrep_close_client_conn` but instead of closing the connection, will be signaling it as killed.

https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/33/

sql/mysqld.cc
kamil-holubicki: > or SET GLOBAL wsrep_reject_queries=ALL_KILL will not be also killed.

From https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-reject-queries:
ALL_KILL The node enables this feature. It rejects all queries and kills existing client connections without waiting, including the current connection.

altmannmarcelo: I would say this is one of the differences between pxc and galera. Currently we don't kill the thd calling the SET:

https://www.percona.com/doc/percona-xtradb-cluster/5.7/wsrep-system-index.html#wsrep_reject_queries. 

New kill function and Call_wsrep_close_client_conn (currently used) have the same criteria about which threads to kill/close.

kamil-holubicki:
Do we still need wsrep_close_threads(NULL); in void kill_mysql(void) after wsrep_close_client_connections(false, true);?
It looks like it is done now inside wsrep_close_client_connections() by Set_wsrep_kill_client_conn (THD::awake() and so on).

altmannmarcelo:
@kamil-holubicki `wsrep_close_threads` will call `Call_wsrep_close_wsrep_threads` which closes wsrep system threads, while `wsrep_close_client_connections` will call `Set_wsrep_kill_client_conn /Call_wsrep_close_client_conn` and kill/close user threads.

kamil-holubicki:
OK, you are right. I didn't notice `if (thd->wsrep_applier)` inside `Call_wsrep_close_wsrep_threads`

percona-ysorokin (approved):
LGTM

kamil-holubicki (approved):

PR #1065, opened 2020-04-16
PXC-2959: Disable optimistic_pa and modify related mtr tests

dutow:
According to support, this change will reduce data inconsistency issues
with workloads with foreign keys.


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "percona-xtradb-cluster-galera"]
 	path = percona-xtradb-cluster-galera
-	url = https://github.com/percona/galera
-	branch = 4.x-8.0
+	url = https://github.com/dutow/galera
```
kamil-holubicki: https://github.com/percona/galera

kamil-holubicki (changes requested):
Could you please add message comment describing the issue solved here?

.gitmodules
dutow: I will update this after the galera commit is merged. I needed this change for jenkins testing

kamil-holubicki (approved):
lgtm

PR #1067, opened 2020-04-16
PXC-3171: Fix PXC 5.7 -> 8.0 upgrade

dutow:
Issue: PXC upgrade failed because the server executed sql statements on
the mysql db before it was updated to the 8.0 schema

Fix: only run those statements on an upgraded schema


kamil-holubicki (approved):

kennt-percona (approved):
lgtm

PR #1068, opened 2020-04-17
PXC-3175: Improve group replication error reporting

dutow:
Issue: INSTALL PLUGIN didn't report why group replication can't be
installed to the user, only to the error log.

Fix: send the related error/warning also to the user.


sql/rpl_group_replication.cc
```diff
@@ -103,12 +103,27 @@ int group_replication_init() {
...
         "Using group replication with Percona XtraDB Cluster is only supported "
         "for migration. Pease make sure that group replication is turned off "
         "once all data is migrated to Percona XtraDB Cluster.");
+    if (current_thd) {
+      push_warning_printf(
+          current_thd, Sql_condition::SL_WARNING, ER_PXC_GR_INIT,
+          ER_THD(current_thd, ER_PXC_GR_INIT),
+          "Using group replication with Percona XtraDB Cluster is only "
+          "supported "
+          "for migration. Pease make sure that group replication is turned off "
```
kennt-percona: 'Pease' -> 'Please'

dutow: Thanks, fixed

kennt-percona (approved):
lgtm

mysql-test/suite/galera/t/pxc_install_gr.test
```diff
@@ -0,0 +1,50 @@
+#
+# Test Galera as a slave to a MySQL master
```
kamil-holubicki: Is this test description proper one?

mysql-test/suite/galera/t/pxc_install_gr.test
```diff
@@ -0,0 +1,50 @@
...
+call mtr.add_suppression("Plugin group_replication reported");
+call mtr.add_suppression("Group replication cannot be used with Percona XtraDB Cluster in strict mode.");
+call mtr.add_suppression("Plugin 'group_replication' init function returned error.");
+call mtr.add_suppression("Using group replication with Percona XtraDB Cluster is only supported for migration");
+--enable_query_log
+
+# Default
+SET GLOBAL pxc_strict_mode = 0;
+
+--sleep 2
```
kamil-holubicki: Why do we need these sleeps?

sql/rpl_group_replication.cc
```diff
@@ -103,12 +103,28 @@ int group_replication_init() {
...
+          ER_PXC_GR_INIT, MYF(0),
+          "Group replication cannot be used with Percona XtraDB Cluster in "
+          "strict mode.");
+    }
     return 1;
   } else {
     WSREP_WARN(
         "Using group replication with Percona XtraDB Cluster is only supported "
-        "for migration. Pease make sure that group replication is turned off "
+        "for migration. Please make sure that group replication is turned off "
```
kamil-holubicki, line 114: We have the same string message twice. It is probably better to use const char *msg ="...." and then use msg.

mysql-test/suite/galera/t/pxc_install_gr.test
dutow: We only really need the last one, and even that's a rare failure: if we change node_2's strict mode back to 0 before it executes the install plugin command, it will succeed, causing check_mysqld to fail. But this is just an instability, as next time the test will most likely pass with retry-pass.

I'll just remove them for now, as I've only seen this error once from quite many executions.

mysql-test/suite/galera/t/pxc_install_gr.test
dutow: fixed

sql/rpl_group_replication.cc
dutow: fixed

mysql-test/suite/galera/t/pxc_install_gr.test
kamil-holubicki: I agree that this is not critical, as code fix is OK, but anyway potential source of MTR instability which, well..., will cause constant pain and probably will never be fixed as "it will fail only sometimes".
Maybe this can help: https://www.percona.com/doc/percona-xtradb-cluster/LATEST/wsrep-system-index.html#wsrep_sync_wait

Or at least it seems to be ease to synchronize using DEBUG_SYNC...

kamil-holubicki (approved):

PR #1069, opened 2020-04-17
5.7 - PXC-3165 - Allow COM_FIELD_LIST to be executed when WSREP is not ready

altmannmarcelo:
Problem
Currently we allow users to execute some `SHOW` and `SET` commands.
`SHOW FIELDS FROM [table]` is one of them. However, `mysql` and `mysqlshow`
clients currently utilize `mysql_list_fields` which executes
`COM_FIELD_LIST` instead of `COM_QUERY` `SHOW FIELDS FROM [table]`.

Solution
Allow `COM_FIELD_LIST` to be executed while WSREP is not ready or
`wsrep_reject_queries` is set to anything rather than `NONE`.


altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc57-test-mtr/51/

kamil-holubicki (approved):
lgtm

PR #1072, opened 2020-04-21
PXC-2500: ALTER USER REPLACE password stmt is crashing PXC nodes

kamil-holubicki:
https://jira.percona.com/browse/PXC-2500

Background:
1. REPLACE clause is allowed only when changing password for current user.
2. TOI transactions are replicated before actual validation/action is done on local node.
3. Check for 1. uses current thd security context to determine current user. On slave node thd is Galera applier thread.

Problem:
1. There is no current user context on slave node. Applier thread has no user name set. This caused segfault in sql_user.cc validate_password_require_current() when thd username and altered username were compared.
2. ALTER USER ... REPLACE should not be replicated, if it will fail locally. This is because there is no way to distinguish 'bad' and 'good' replace on slave node - we have no current user context.

Solution:
1. Skip check if we are changing password for another user if executing thread is Galera applier thread.
2. Before starting TOI transaction on master node, perform validation of REPLACE clause. Only part of validation involving current user context is needed here. Other validations that are not dependent on current user context can be done on slave node side.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_alter_user.result
+++ b/mysql-test/suite/galera/r/pxc_alter_user.result
@@ -1,0 +1,48 @@
+call mtr.add_suppression("Operation ALTER USER failed for 'User3'@'localhost'");
+call mtr.add_suppression("Operation ALTER USER failed for 'User3'@'localhost'");
+# connection node_1, root
+CREATE USER User1@localhost IDENTIFIED BY 'old_pass';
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/206/

kennt-percona (approved):
lgtm

PR #1074, opened 2020-04-22
PXC-3179: CREATE USER ... RANDOM PASSWORD is not repicating as expected in PXC

kamil-holubicki:
https://jira.percona.com/browse/PXC-3179

Problem:
CREATE/ALTER USER ... IDENTIFIED BY RANDOM PASSWORD creates user on master and slave nodes, but with different passwords. This is because TOI transactions are replicated as statements, so different passwords are generated on different nodes.

Solution:
Forbid using of RANDOM PASSWORD if node replicates to the cluster.

### DIFF
--- a/sql/auth/sql_user.cc
+++ b/sql/auth/sql_user.cc
@@ -1258,6 +1258,19 @@ bool set_and_validate_user_attributes(
     inbuflen = (unsigned)Str->auth.length;
     std::string gen_password;
     if (Str->has_password_generator) {
+#ifdef WITH_WSREP
+      if (WSREP(thd)) {
+        const char *msg =
+            "Percona XtraDB Cluster doesn't allow use of"
+            " RANDOM PASSWORD for CREATE/ALTER USER operation"
+            " while operating in cluster mode";
+        WSREP_ERROR("%s", msg);
+        my_message(ER_UNKNOWN_ERROR, msg, MYF(0));
+        plugin_unlock(0, plugin);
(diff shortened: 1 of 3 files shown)


sql/auth/sql_user.cc
```diff
@@ -1258,6 +1258,17 @@ bool set_and_validate_user_attributes(
...
     std::string gen_password;
     if (Str->has_password_generator) {
+#ifdef WITH_WSREP
+      if (WSREP(thd)) {
+        const char *msg = "Percona XtraDB Cluster doesn't allow use of"
+            " RANDOM PASSWORD for CREATE/ALTER USER operation"
+            " while operating in cluster mode";
+        WSREP_ERROR("%s", msg);
+        my_message(ER_UNKNOWN_ERROR, msg, MYF(0));
+        plugin_unlock(0, plugin);
```
kennt-percona, line 1269: The other places where we return an error after the plugin lock, I see this is getting set
  what_to_set.m_what = NONE_ATTR;
should we also do this here?

kamil-holubicki: Shouldn't be a big deal, because in case of error it is not used, and at the top of this function it is set to NONE_ATR, but anyway, good to stick to convention in this function.
Fixed.

kamil-holubicki: Fixed.

kennt-percona (approved):
lgtm

PR #1075, opened 2020-04-22
PXC-3170 - Backport to 5.7 - Thread pooling hangs on shutdown

altmannmarcelo:
Problem
When wsrep_close_client_connections is called, it closes connections
abruptly. This is ok on shutdown, however, same function is also called
when we set wsrep_cluster_address dynamically and when setting
wsrep_reject_queries=ALL_KILL, which fails to deliver
proper notification to thread_pool threads.

Fix
Adjust wsrep_close_client_connections to first signal the thd as killed
then attempt to close the connection.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -1124,6 +1124,29 @@ private:
 
 #ifdef WITH_WSREP
 /**
+  This class implements callback function used by
+  wsrep_close_client_connections() to set KILL_CONNECTION
+  flag on all client thds and awake the thread.
+*/
+class Set_wsrep_kill_client_conn : public Do_THD_Impl
+{
+ public:
+  Set_wsrep_kill_client_conn()
+  {}
+
+  virtual void operator()(THD *killing_thd)
+  {
+    if (killing_thd->get_protocol()->connection_alive() &&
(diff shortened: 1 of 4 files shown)


altmannmarcelo:
mtr - https://pxc.cd.percona.com/job/pxc57-test-mtr/53/

peterschwaller:
I've edited the title to avoid matches with the original Jira issue.

sql/mysqld.cc
```diff
@@ -1123,6 +1123,26 @@ class Call_close_conn : public Do_THD_Impl
 };
 
 #ifdef WITH_WSREP
+/**
+  This class implements callback function used by
+  wsrep_close_client_connections() to set KILL_CONNECTION
+  flag on all client thds and awake the thread.
+*/
+class Set_wsrep_kill_client_conn : public Do_THD_Impl {
```
kamil-holubicki: Code formatting does not conform clang 5.0

kamil-holubicki (changes requested):

kamil-holubicki:
Please remember about doing null merge 5.7 -> 8.0

kamil-holubicki (approved):

PR #1078, opened 2020-04-23
PXC-2939: Upgrade - configuration files different

borys-belinsky:


doc/source/howtos/upgrade_guide.rst
```diff
@@ -21,50 +21,50 @@ and variables mentioned in these documents when upgrading to |PXC| 8.0.
...
+
+   If downtime is not possible, the rolling upgrade is supported but ensure the
+   traffic is controlled during the upgrade and writes are directed only to 5.7
+   nodes until all nodes are upgraded to 8.0.
+
+The configuration file layout has changed in |pxc| |version|
+   All configuration settings are stored in the default |MySQL| configuration
+   file:
+
+   * Path on Debian and Ubuntu: :file:`/etc/mysql/mysql.conf.d/mysqld.cnf`
```
tplavcic: On debian/ubuntu the file is `/etc/mysql/my.cnf` which then includes files in directory `/etc/mysql/mysql.conf.d` which has file `/etc/mysql/mysql.conf.d/mysqld.cnf` which in the end has all the new default options.
It's still not 3 files like before, but users should always check /etc/mysql/my.cnf and start from there since that is the main one and could include the wrong config/subdirectory.
I see your point that settings are actually in `/etc/mysql/mysql.conf.d/mysqld.cnf`, but it would be good to mention that the main file is `/etc/mysql/my.cnf` since from what I see in major upgrade that one can include wrong subdirectory.

tplavcic (changes requested):

tplavcic (approved):
LGTM

PR #1082, opened 2020-04-24
PXC-3182 - disallow write on 8.0 while 5.7 nodes are still on the clu…

altmannmarcelo:
…ster

This commit introduces the control of blocking queries that will modify
any data while cluster is operating with protocol lower than 4 and
pxc_strict_mode set to ENFORCING or MASTER.
It also sets pxc_maint_mode to MAINTENANCE so proxies won't redirect
traffic to this node while operating in multi major version mode.

Attempting to disable pxc_maint_mode while still on pxc_strict_mode
is not allowed.

Caveats:
Async replication is also blocked (this can cause the same issue on 5.7 nodes)
Disabling pxc_strict_mode, allow users to set pxc_maint_mode to DISABLED. Re-enabling pxc_strict_mode again doesn't automatically SET pxc_maint_mode to MAINTENANCE again, unless there is a reconfiguration of cluster view.

### DIFF
--- a/mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
+++ b/mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
@@ -1,0 +1,80 @@
+#
+# PXC-3182 - disallow write on 8.0 while 5.7 nodes are still on the cluster
+#
+
+--source include/galera_cluster.inc
+--source include/have_debug.inc
+
+--connection node_2
+--echo #node-2
+CALL mtr.add_suppression("Percona-XtraDB-Cluster doesn't recommend use of multiple major versions while accepting write workload with pxc_strict_mode = PERMISSIVE");
+CALL mtr.add_suppression("pxc_maint_mode cannot be changed while using of multiple major versions");
+CALL mtr.add_suppression("Percona-XtraDB-Cluster prohibits use of multiple major versions while accepting write workload with pxc_strict_mode = ENFORCING");
+SET @global_saved_strict_tmp = @@global.pxc_strict_mode;
+SET GLOBAL pxc_strict_mode = ENFORCING;
+CREATE TABLE tb1 (ID INT PRIMARY KEY);
+INSERT INTO tb1 VALUES (1),(2),(3);
+
+# provoke reconfiguration 
+SET GLOBAL debug="+d,simulate_wsrep_multiple_major_versions";
+--disable_query_log
+SET GLOBAL wsrep_cluster_address=@@wsrep_cluster_address;
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
+SET SESSION wsrep_sync_wait = DEFAULT;
+--enable_query_log
+
+# Test Error trying to change pxc_maint_mode
+--error ER_UNKNOWN_ERROR
+SET GLOBAL pxc_maint_mode = DISABLED;
+--let $assert_text= Assert that the expected entry is in the error log when pxc_strict_mode = ENFORCING
+--let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.2.err
+--let $assert_only_after = Synchronized with group, ready for connections
+--let $assert_select= .* \[ERROR\] .* pxc_maint_mode cannot be changed while using of multiple major versions
+--let $assert_count= 1
+--source include/assert_grep.inc
+
+# Test Error while writing data
+--error ER_UNKNOWN_ERROR
+INSERT INTO tb1 VALUES (4);
+--error ER_UNKNOWN_ERROR
+UPDATE tb1 SET ID = 5 WHERE ID = 3;
+--error ER_UNKNOWN_ERROR
+DELETE FROM tb1 WHERE ID = 2;
+
+--let $assert_text= Assert that the expected entry is in the error log when pxc_strict_mode = ENFORCING
+--let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.2.err
+--let $assert_only_after = Synchronized with group, ready for connections
+--let $assert_select= .* \[ERROR\] .* Percona-XtraDB-Cluster prohibits use of multiple major versions while accepting write workload with pxc_strict_mode = ENFORCING or MASTER
+--let $assert_count= 3
+--source include/assert_grep.inc
+
+# Test warning while writing data
+SET GLOBAL pxc_strict_mode = PERMISSIVE;
+INSERT INTO tb1 VALUES (4);
+UPDATE tb1 SET ID = 5 WHERE ID = 3;
+DELETE FROM tb1 WHERE ID = 2;
+--let $assert_text= Assert that the expected entry is in the error log when pxc_strict_mode = ENFORCING
+--let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.2.err
+--let $assert_only_after = Synchronized with group, ready for connections
+--let $assert_select= .* \[Warning\] .* Percona-XtraDB-Cluster doesn't recommend use of multiple major versions while accepting write workload with pxc_strict_mode = PERMISSIVE
+--let $assert_count= 3
+--source include/assert_grep.inc
+
+# Cleanup 
+--disable_query_log
+SET GLOBAL pxc_strict_mode = @global_saved_strict_tmp;
+SET GLOBAL debug="-d,simulate_wsrep_multiple_major_versions";
+SET GLOBAL wsrep_cluster_address=@@wsrep_cluster_address;
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
+SET SESSION wsrep_sync_wait = DEFAULT;
+DROP TABLE tb1;
+--enable_query_log
+
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1605,6 +1605,54 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
 
   return block;
 }
+/*
+ * This function should be called to determine if cluster still has nodes
+ * with lower versions of wsrep_protocol_version (eg.: version 3).
+ * Writing on a new node can crash nodes with lower version 8.0->5.7
+ */
+static bool block_write_while_in_rolling_upgrade(THD *thd) {
+  /* Ignore check while server is not ready or for background wsrep applier too
+   * (like slave thread) */
+  if (!wsrep_node_is_synced() || (WSREP(thd) && thd->wsrep_applier))
+    return false;
+
+  bool block = false;
+  LEX *lex = thd->lex;
+  if (sql_command_flags[lex->sql_command] & CF_CHANGES_DATA) {
+    bool multi_version_cluster = wsrep_protocol_version < 4;
+    if (multi_version_cluster ||
+        DBUG_EVALUATE_IF("simulate_wsrep_multiple_major_versions", true,
+                         false)) {
+      const char *msg;
+      switch (pxc_strict_mode) {
+        case PXC_STRICT_MODE_DISABLED:
+          /* Do nothing */
+          break;
+        case PXC_STRICT_MODE_PERMISSIVE:
+          msg =
+              "Percona-XtraDB-Cluster doesn't recommend use of multiple major"
+              " versions while accepting write workload"
+              " with pxc_strict_mode = PERMISSIVE";
+          WSREP_WARN("%s", msg);
+          push_warning_printf(thd, Sql_condition::SL_WARNING, ER_UNKNOWN_ERROR,
+                              "%s", msg);
+          break;
+        case PXC_STRICT_MODE_MASTER:
+        case PXC_STRICT_MODE_ENFORCING:
+        default:
+          block = true;
+          msg =
+              "Percona-XtraDB-Cluster prohibits use of multiple major versions"
+              " while accepting write workload with pxc_strict_mode = "
+              "ENFORCING or MASTER";
+          WSREP_ERROR("%s", msg);
+          my_message(ER_UNKNOWN_ERROR, msg, MYF(0));
+          break;
+      }
+    }
+  }
+  return block;
+}
 #endif /* WITH_WSREP */
 
 /**
@@ -3329,6 +3377,9 @@ int mysql_execute_command(THD *thd, bool first_level) {
         !wsrep_is_show_query(lex->sql_command)) {
       my_message(ER_UNKNOWN_COM_ERROR,
                  "WSREP has not yet prepared node for application use", MYF(0));
+      return -1;
+    }
+    if (block_write_while_in_rolling_upgrade(thd)) {
       return -1;
     }
   }
--- a/sql/wsrep_server_service.cc
+++ b/sql/wsrep_server_service.cc
@@ -186,6 +186,33 @@ void Wsrep_server_service::log_view(
     global_system_variables.auto_increment_increment = view.members().size();
   }
   wsrep_protocol_version = view.protocol_version();
+  bool not_shutdown = (pxc_maint_mode != PXC_MAINT_MODE_SHUTDOWN);
+  bool multi_version_cluster = wsrep_protocol_version < 4;
+  if (not_shutdown &&
+      ((multi_version_cluster &&
+        (pxc_strict_mode > PXC_STRICT_MODE_PERMISSIVE)) ||
+       DBUG_EVALUATE_IF("simulate_wsrep_multiple_major_versions", true,
+                        false))) {
+    std::ostringstream os;
+    os << "Detected Protocol version: " << wsrep_protocol_version
+       << " Changing pxc_maint_mode to "
+          "MAINTENANCE.";
+    WSREP_INFO("%s", os.str().c_str());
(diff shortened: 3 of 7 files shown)


altmannmarcelo:
mtr - https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/209/

sql/sql_parse.cc
```diff
@@ -1605,6 +1605,59 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
...
+          push_warning_printf(thd, Sql_condition::SL_WARNING, ER_UNKNOWN_ERROR,
+                              "Percona-XtraDB-Cluster doesn't recommend use of"
+                              " multiple major versions while accepting write"
+                              " workload with pxc_strict_mode = PERMISSIVE");
+          break;
+        case PXC_STRICT_MODE_MASTER:
+        case PXC_STRICT_MODE_ENFORCING:
+        default:
+          block = true;
+          WSREP_ERROR(
```
kamil-holubicki: const char *msg = "....";
WSREP_ERROR("%s", msg);
my_message(ERR_UNKNOWN_ERROR, msg, MYF(0));

sql/sql_parse.cc
```diff
@@ -1605,6 +1605,59 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
...
+    DBUG_EXECUTE_IF("simulate_wsrep_multiple_major_versions",
+                    debug_simulate_wsrep_multiple_major_versions = true;);
+    if (wsrep_protocol_version < 4 ||
+        debug_simulate_wsrep_multiple_major_versions) {
+      switch (pxc_strict_mode) {
+        case PXC_STRICT_MODE_DISABLED:
+        /* Do nothing */
+        break;
+        case PXC_STRICT_MODE_PERMISSIVE:
+          WSREP_WARN(
```
kamil-holubicki: const char *msg = "....";
WSREP_ERROR("%s", msg);
push_warning_printf(thd, Sql_condition::SL_WARNING, ER_UNKNOWN_ERROR, msg);

sql/wsrep_var.cc
```diff
@@ -1003,6 +1003,22 @@ bool pxc_maint_mode_check(sys_var *, THD *, set_var *var) {
     my_message(ER_UNKNOWN_ERROR, message, MYF(0));
     return true;
   }
+  /* pxc-maint-mode cannot be changed if operating with multiple major versions
+   * and pxc_strict_mode set to ENFORCING or MASTER.
+   * check Wsrep_server_service::log_view() */
+  if (wsrep_pxc_maint_mode != PXC_MAINT_MODE_DISABLED &&
+      pxc_strict_mode >= PXC_STRICT_MODE_ENFORCING) {
+    WSREP_ERROR(
```
kamil-holubicki: const char *msg = "....";
WSREP_ERROR("%s", msg);
my_message(ERR_UNKNOWN_ERROR, msg, MYF(0));

sql/wsrep_server_service.cc
```diff
@@ -186,6 +186,32 @@ void Wsrep_server_service::log_view(
...
+  if ((view.protocol_version() < 4 &&
+       (pxc_strict_mode > PXC_STRICT_MODE_PERMISSIVE)) ||
+      debug_simulate_wsrep_multiple_major_versions) {
+    std::ostringstream os;
+    os << "Detected Protocol version: " << wsrep_protocol_version
+       << " Changing pxc_maint_mode to "
+          "MAINTENANCE.";
+    WSREP_INFO("%s", os.str().c_str());
+    pxc_maint_mode = PXC_MAINT_MODE_MAINTENANCE;
+    wsrep_pxc_maint_mode = PXC_MAINT_MODE_MAINTENANCE;
```
kamil-holubicki: This variable acts like bool flag
PXC_MAINT_MODE_MAINTENANCE => true
PXC_MAINT_MODE_DISABLED => false

Wouldn't it be more readable if it was something like:
bool wsrep_pxc_maint_mode_forced = false;

and then

if (wsrep_pxc_maint_mode_forced) { .... } else { ... }
?

sql/wsrep_server_service.cc
```diff
@@ -186,6 +186,32 @@ void Wsrep_server_service::log_view(
...
+  } else {
+    /* if pxc_maint_mode was previously set by wsrep, we should reset */
```
kamil-holubicki, line 204: Isn't it possible that we are executing shutdown, so pxc_maint_mode == PXC_MAINT_MODE_SHUTDOWN and then we get here?

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,63 @@
...
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
+SET SESSION wsrep_sync_wait = DEFAULT;
+--enable_query_log
+
+
+# Test Error trying to change pxc_maint_mode
+--error 1105 # ER_UNKNOWN_ERROR
```
kamil-holubicki: Multiple occurences:
--error ER_UNKNOWN_ERROR

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,63 @@
...
+SET GLOBAL pxc_strict_mode = PERMISSIVE;
+INSERT INTO tb1 VALUES (4);
+UPDATE tb1 SET ID = 5 WHERE ID = 3;
+DELETE FROM tb1 WHERE ID = 2;
+
+
+# Cleanup 
+--disable_query_log
+# MTR requires pxc_strict_mode DISABLED
+SET GLOBAL pxc_strict_mode = DISABLED;
```
kamil-holubicki: SET @global_saved_tmp = @@global.pxc_strict_mode;
... do the test and change global variable ....
SET @@global.pxc_strict_mode = @global_saved_tmp;

kamil-holubicki (commented):
General note about pxc_maint_mode:
I looked briefly how this variable can change and:
1. It is set to PXC_MAINT_MODE_SHUTDOWN in mysqld.cc  signal_hand(), then shutdown is delayed.
2. It can be changed by user. Transition is validated in pxc_maint_mode_check() without LOCK_global_system_variables mutex and then updated in sys_var::update() under LOCK_global_system_variables.
3. Its value can also change in Wsrep_server_service::log_view under LOCK_global_system_variables.

Aren't there possible concurrency issues here?

sql/sql_parse.cc
```diff
@@ -1605,6 +1605,59 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
...
+    return false;
+
+  bool block = false;
+  LEX *lex = thd->lex;
+  if (sql_command_flags[lex->sql_command] & CF_CHANGES_DATA) {
+    bool debug_simulate_wsrep_multiple_major_versions = false;
+    DBUG_EXECUTE_IF("simulate_wsrep_multiple_major_versions",
+                    debug_simulate_wsrep_multiple_major_versions = true;);
+    if (wsrep_protocol_version < 4 ||
+        debug_simulate_wsrep_multiple_major_versions) {
```
venkatesh-prasad-v: We can replace
```
    bool debug_simulate_wsrep_multiple_major_versions = false;
    DBUG_EXECUTE_IF("simulate_wsrep_multiple_major_versions",
                    debug_simulate_wsrep_multiple_major_versions = true;);
    if (wsrep_protocol_version < 4 ||
        debug_simulate_wsrep_multiple_major_versions) {
```
with
```
    if (wsrep_protocol_version < 4 ||
            DBUG_EVALUATE_IF("simulate_wsrep_multiple_major_versions", true, false) {
```

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
altmannmarcelo: From my tests, `--error` is only valid for the next statement. If I remove the ones below, the test fails because of the statements return `ER_UNKNOWN_ERROR`

george-lorch: @altmannmarcelo is correct, --error only applies to the next statement and is cleared after execution

sql/wsrep_server_service.cc
altmannmarcelo: It was indeed possible, either by having a reconfiguration while still working on multi major versions or having a reconfig caused by the last 5.7 leaving the cluster. This is not been taken care of.

altmannmarcelo:
> General note about pxc_maint_mode:
> I looked briefly how this variable can change and:
> 
> 1. It is set to PXC_MAINT_MODE_SHUTDOWN in mysqld.cc  signal_hand(), then shutdown is delayed.
> 2. It can be changed by user. Transition is validated in pxc_maint_mode_check() without LOCK_global_system_variables mutex and then updated in sys_var::update() under LOCK_global_system_variables.
> 3. Its value can also change in Wsrep_server_service::log_view under LOCK_global_system_variables.
> 
> Aren't there possible concurrency issues here?

We now validate Wsrep_server_service::log_view that we are not under shutdown (currently the only blocked transition).

sql/sql_parse.cc
```diff
@@ -1605,6 +1605,54 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
...
+static bool block_write_while_in_rolling_upgrade(THD *thd) {
+  /* Ignore check while server is not ready or for background wsrep applier too
+   * (like slave thread) */
+  if (!wsrep_node_is_synced() || (WSREP(thd) && thd->wsrep_applier))
+    return false;
+
+  bool block = false;
+  LEX *lex = thd->lex;
+  if (sql_command_flags[lex->sql_command] & CF_CHANGES_DATA) {
+    bool multi_version_cluster = wsrep_protocol_version < 4;
```
venkatesh-prasad-v, line 1622: IMO, with the use of hard-coded version numbers, it becomes difficult to search for its occurrences when the version changes in the future.

How about we define an `enum` for the same? something like:
```
enum enum_wsrep_version {
  WSREP_VERSION_V2 = 2,
  WSREP_VERSION_V3 = 3,
  WSREP_VERSION_V4 = 4
};
```
and use it like
`bool multi_version_cluster = wsrep_protocol_version < WSREP_VERSION_V4;`

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,62 @@
...
+DELETE FROM tb1 WHERE ID = 2;
+
+# Test warning while writing data
+SET GLOBAL pxc_strict_mode = PERMISSIVE;
+INSERT INTO tb1 VALUES (4);
+UPDATE tb1 SET ID = 5 WHERE ID = 3;
+DELETE FROM tb1 WHERE ID = 2;
+
+
+# Cleanup
```
venkatesh-prasad-v, line 67: I suggest you to assert that the error messages are logged into error log as well. You can do it with the use of `assert_grep.inc`. Something like:
```
--let $assert_text= Assert that the expected entry is in the error log when pxc_maint_mode = DISABLED
--let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.2.err
--let $assert_only_after = Synchronized with group, ready for connections
--let $assert_select= .* \[ERROR\] .* Percona-XtraDB-Cluster prohibits use of multiple major versions while accepting write workload
--let $assert_count= 3
--source include/assert_grep.inc

--let $assert_text= Assert that the expected entry is in the error log when pxc_strict_mode = PERMISSIVE
--let $assert_file=$MYSQLTEST_VARDIR/log/mysqld.2.err
--let $assert_only_after = Synchronized with group, ready for connections
--let $assert_select= .* \[Warning\] .* Percona-XtraDB-Cluster doesn't recommend use of multiple major versions while accepting write workload
--let $assert_count= 3
--source include/assert_grep.inc
```

Even though we have added the suppression for the above messages, but they doesn't ensure that the message is actually being logged.

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,62 @@
...
+
+# provoke reconfiguration 
+SET GLOBAL debug="+d,simulate_wsrep_multiple_major_versions";
+--disable_query_log
+SET GLOBAL wsrep_cluster_address=@@wsrep_cluster_address;
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
```
venkatesh-prasad-v, line 26: These lines seem to be present in every galera test. If this is very common, isn't it good to have an `.inc` file for the same?
We can create an .inc file in `mysql-test/include` which does
1. Accept the cluster size in a variable. Something like `--let $cluster_size= 1|2|3`.
2. Let the wait happen in the .inc file until the following condition becomes true. 
```
--let $wait_condition = SELECT VARIABLE_VALUE = $cluster_size FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
--source include/wait_condition.inc
```
Once this is done, we can use it like
```
--let $cluster_size= 2
--source inclulde/wait_for_cluster_size.inc
```

We can do create a similar inc file for waiting for `wsrep_local_state_comment` as well.

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,62 @@
...
+CALL mtr.add_suppression("Percona-XtraDB-Cluster prohibits use of multiple major versions while accepting write workload with pxc_strict_mode = ENFORCING");
+SET @global_saved_strict_tmp = @@global.pxc_strict_mode;
+SET GLOBAL pxc_strict_mode = ENFORCING;
+CREATE TABLE tb1 (ID INT PRIMARY KEY);
+INSERT INTO tb1 VALUES (1),(2),(3);
+
+# provoke reconfiguration 
+SET GLOBAL debug="+d,simulate_wsrep_multiple_major_versions";
+--disable_query_log
+SET GLOBAL wsrep_cluster_address=@@wsrep_cluster_address;
```
venkatesh-prasad-v, line 21: Why do we set `wsrep_cluster_address`? Does its value change during the course for the test? Just asking my understanding.

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
```diff
@@ -0,0 +1,62 @@
...
+SET GLOBAL pxc_strict_mode = @global_saved_strict_tmp;
+SET GLOBAL debug="-d,simulate_wsrep_multiple_major_versions";
+SET GLOBAL wsrep_cluster_address=@@wsrep_cluster_address;
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
+SET SESSION wsrep_sync_wait = DEFAULT;
+DROP TABLE tb1;
```
venkatesh-prasad-v, line 78: Shouldn't we wait for this statement to be replicated to all members of the cluster? How does galera tests ensure proper cleanup in such cases?

venkatesh-prasad-v (commented):
Hi Marcelo,

Thank you for working on this issue. 

Even though I am not an official reviewer, I started reviewing this as I thought reviews will increase my overall understanding and also they serve as a better and an effective way to increasing my knowledge on PXC.

I had a few minor comments, questions and suggestions on this. Please go through them when you get time.

PS: The suggestions are not really blockers and I leave it to your discretion whether or not to accept them.

Thank you for your time!

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
altmannmarcelo: This is to provoke a reconfiguration of the cluster and make node2 read the debug flag to force protocol v3

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
altmannmarcelo: sounds like a good idea. We should raise a JIRA for it and change the current test cases to make use of it.

sql/sql_parse.cc
altmannmarcelo: I think this deserves a more broad refactor. This logic is used in other parts of the code. We should do it and make sure the other functionalities still behaving as expected. Do you wanna track it over a JIRA ticket so we don't forget?

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
altmannmarcelo: This is replicated as TOI, which means it will get applied at the same point on all nodes. Once a DDL has started as TOI it has to complete.

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
altmannmarcelo: Done

mysql-test/suite/galera/t/pxc_block_write_while_in_rolling_upgrade.test
kamil-holubicki: Yes, I know. What I meant here is using 
--error IDENTIFIER
instead of
--error HARDCODED_NUMBER
so
--error ER_UNKNOWN_ERROR
instead of 
--error 1105

george-lorch: Ahh, yes, this^^

altmannmarcelo: @kamil-holubicki Done

kamil-holubicki (approved):
lgtm

PR #1083, opened 2020-04-24
PXC-3189 SST fails if super_read_only

altmannmarcelo:
Problem
If a joiner request an SST, as part of run_post_processing_steps
its required to start a temporary standalone mysqld server. In case
the user has super_read_only set on my.cnf the checks performed on that
routine will fail.

Solution
Always start temporary server with super_read_only=OFF.


altmannmarcelo:
mtr - https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/212/

kennt-percona (approved):
lgtm

PR #1086, opened 2020-04-27
Pxc 2979 new cluster encrypt cluster traffic 8.0

borys-belinsky:


doc/source/security/encrypt-traffic.rst
```diff
@@ -61,54 +63,62 @@ Replication traffic refers to the inter-node traffic which includes
...
 
 .. _ssl-auto-conf:
 
 SSL Automatic Configuration
 ===========================
 
+The automatic configuration of the SSL encryption needs a key and certificate
+files. MySQL generates a default key and certificate files and places them in
+the data directory. It is important that your cluster use the same SSL
+certificates on all nodes.
```
rameshvs02: It will be great if we highlight this `It is important that your cluster use the same SSL
certificates on all nodes.`

tplavcic (approved):
LGTM

doc/source/security/encrypt-traffic.rst
```diff
@@ -61,54 +63,70 @@ Replication traffic refers to the inter-node traffic which includes
...
+ 
 .. _enabling_encrypt-cluster-traffic:
 
 Enabling :variable:`pxc-encrypt-cluster-traffic`
 ------------------------------------------------
 
-|PXC| includes the :variable:`pxc-encrypt-cluster-traffic` variable that
-enables automatic configuration of SSL encryption there-by encrypting
+|PXC| includes the |pxc-encrypt-cluster-traffic| variable that
+enables automatic configuration of SSL encryption thereby encrypting
```
hrvojem: "automatic" where I need to manually copy over cert files to all the nodes, stop each node, change the config file and start it again before it automatically works?

hrvojem (changes requested):
This still doesn't show steps how to setup the encryption on the cluster. When do I need to have the certs in place? How do I roll the encryption to the cluster?

doc/source/configure.rst
```diff
@@ -25,61 +17,93 @@ Configuration examples assume there are three |PXC| nodes:
...
 
-If you are running Debian or Ubuntu,
-add the following configuration variables to :file:`/etc/mysql/my.cnf`
-on the first node::
+1. Stop the |PXC| server:
+
+   .. code-block:: bash
+
+      $ sudo service mysql stop
+
```
hrvojem, line 25: For 8.0 service isn't started after installation, so this step isn't needed anymore (unless user manually started the service after installation)

doc/source/security/encrypt-traffic.rst
```diff
@@ -42,73 +49,84 @@ for each node:
...
 ===========================
 
+The automatic configuration of the SSL encryption needs a key and certificate
+files. MySQL generates a default key and certificate files and places them in
+the data directory.
+
+.. important::
+
+   It is important that your cluster use the same SSL certificates on all nodes.
+
```
hrvojem, line 85: Note that encryption can't be enabled on a running cluster. If it was disabled before the cluster was bootstrapped, cluster needs to stopped, encryption set up, and bootstrapped again.

doc/source/configure.rst
hrvojem, line 26: in 8.0 service isn't automatically started after installation, so this step isn't needed (unless user manually started the service).

hrvojem (changes requested):

PR #1088, opened 2020-04-27
PXC-3187: Tarball installation - add dependencies

borys-belinsky:


rameshvs02 (commented):
`rsync` package is only required for 5.6 and 5.7. We can remove rsync from the 8.0 doc.

EvgeniyPatlan (approved):
lgtm

PR #1092, opened 2020-04-28
PXC-3204 pxc incorrectly set wsrep_protocol_version when wsrep_auto_i…

altmannmarcelo:
…ncrement_control is disabled

Problem
PXC will instantiate wsrep_protocol_version = wsrep_max_protocol_version
and will adjust the variable when there is a reconfiguration of the
cluster at Wsrep_server_service::log_view. However
wsrep_protocol_version will only change if wsrep_auto_increment_control
is enabled.

Solution
Move wsrep_protocol_version = view.protocol_version() line to outside
the if that checks for wsrep_auto_increment_control, this way the
variable will always have the latest protocol version passed via the
view reconfiguration.


altmannmarcelo:
mtr - https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/213/

kennt-percona (approved):
lgtm

PR #1110, opened 2020-05-06
PXC-3026: Fix most of WITH_WSREP issues in the 5.6 branch

dutow:
* Introduces scripts under the check_src directory, which detect
  differences between the pxc WITH_WSREP=OFF source and PS. For the same
  upstream versions, there shouldn't be any changes except for the
  existing whitelists.
  The script also disables WITH_INNODB_DISALLOW_WRITES, as that's
  another 3rd party path not present in PS.
* Whitelisted product name changes, as PXC should be called PXC even
  without WITH_WSREP.
* Whitelisted the use of the WSREP_BINLOG_FORMAT macro, as it defaults
  to a no-op without WITH_WSREP.
* Fixed incorrect naming of the WITH_WSREP macro at a few places
* Added WITH_WSREP if(n)defs everywhere where PXC added/deleted logic
* Removed cosmetic differences compared to PS.

### DIFF
--- a/check_src/check_file.sh
+++ b/check_src/check_file.sh
@@ -1,0 +1,50 @@
+# Compares a single PS/PXC source file
+# The script will print out an error and a diff if there is any difference 
+# after removing the PXC specific code, not counting whitespace and empty lines
+# The script also allows whitelisting expected differences, such as different
+# product names (Percona Server vs Percona Xtradb Cluster)
+
+# Usage: bash check_src/check_file.sh <file_name> <PS 5.6 source directory>
+# The script should be run from the root directory of the PXC source
+# The PS source should be the currently merged PS tag
+
+# Extending/changing the whitelist:
+# The whitelist directory contains files in a single directory (without
+# subdirectories), each file whitelist a single real source file, with the
+# same name.
+# The whitelist file contains the lines starting with > or < of the expected
+# diff: only the change without the actual line/offset numbers, to prevent
+# unrelated changes from breaking the whitelisted change.
+if [ ! -s $2/$1 ] ; then
+  exit 0
+fi
+F=`mktemp`
+gawk -f check_src/rem.awk $1 > $F
+
+D=`mktemp`
+D2=`mktemp`
+diff -wB $F $2/$1 > $D
+
+if [ -s $D ] ; then
+  WHITELIST_NAME=check_src/whitelist/${1##*/}
+  if [ -f $WHITELIST_NAME ] ; then
+    D3=`mktemp`
+    cat $D | grep "^[<>].*" > $D3
+    diff -wB $WHITELIST_NAME $D3 > $D2
+    rm $D3
+    if [ ! -s $D2 ] ; then
+      echo -n "" > $D
+    fi
+  fi
+fi
+
+if [ -s $D ] ; then
+  echo "============"
+  echo "Error: $1"
+  echo "============"
+  cat $D
+fi
+
+rm $F
+rm $D
+rm $D2
--- a/check_src/run.sh
+++ b/check_src/run.sh
@@ -1,0 +1,9 @@
+# Compares every source file between PS/PXC for differences without the if WITH_WSREP
+# blocks.
+# Usage:  bash check_src/run.sh <PS_DIRECTORY>
+PS_56_LOCATION=$1
+find . -type f -name '*.h' -exec bash check_src/check_file.sh '{}' $PS_56_LOCATION \;
+find . -type f -name '*.cc' -exec bash check_src/check_file.sh '{}' $PS_56_LOCATION \;
+find . -type f -name '*.c' -exec bash check_src/check_file.sh '{}' $PS_56_LOCATION \;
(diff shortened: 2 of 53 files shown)


check_src/check_file.sh
```diff
@@ -0,0 +1,34 @@
+
```
kamil-holubicki, line 6: I think it would be useful to add short description at the top of each 'check_src' script, describing what it does and how to use it.

Additionally it would be useful to have the description (maybe in main check_src file) how to add whitelist file, what it contains, etc.

Just to be able to use this functionality without first reverse engineering how it works.

check_src/run.sh
```diff
@@ -0,0 +1,6 @@
+PS_56_LOCATION=$1
```
kamil-holubicki, line 4: We do not check *.c files?

sql/lock.cc
```diff
@@ -208,11 +208,13 @@ lock_tables_check(THD *thd, TABLE **tables, uint count, uint flags)
     if (!(flags & MYSQL_LOCK_IGNORE_GLOBAL_READ_ONLY) && !t->s->tmp_table)
     {
       if (t->reginfo.lock_type >= TL_WRITE_ALLOW_WRITE &&
-          enforce_ro && opt_readonly && !thd->slave_thread
 #ifdef WITH_WSREP
+          enforce_ro && opt_readonly && !thd->slave_thread
```
kamil-holubicki, line 212: Actually this was ok. I understand that the problem was with the trailing 'if' bracket, right?

sql/log.cc
```diff
@@ -2901,8 +2901,8 @@ int TC_LOG_MMAP::open(const char *opt_name)
     pg->ptr= pg->start=(my_xid *)(data + i*tc_log_page_size);
 #ifdef WITH_WSREP
     if (!WSREP_ON) 
-#endif /* WITH_WSREP */
     pg->end=(my_xid *)(pg->start + tc_log_page_size);
+#endif /* WITH_WSREP */
```
kamil-holubicki: It can be done with 1 ifdef
```
#ifdef WITH_WSREP
  if (!WSREP_ON) pg->end=(my_xid *)(pg->start + tc_log_page_size);
  else pg->end=pg->start + pg->size;
#else
  pg->end=pg->start + pg->size; 
#endif
```

sql/signal_handler.cc
```diff
@@ -73,11 +73,11 @@ extern "C" sig_handler handle_fatal_signal(int sig)
 
   segfaulted = 1;
 
+#ifdef WITH_WSREP
```
kamil-holubicki, line 76: #endif - probably more occurrences without /* WITH_WSREP */

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2651,6 +2676,18 @@ innobase_trx_init(
 	trx->check_unique_secondary = !thd_test_options(
 		thd, OPTION_RELAXED_UNIQUE_CHECKS);
 
+#ifndef WITH_WSREP
```
kamil-holubicki, line 2680: Did we remove fake_changes part from PXC on purpose?

storage/innobase/include/srv0srv.h
```diff
@@ -305,7 +305,7 @@ extern uint	srv_flush_log_at_timeout;
 extern char	srv_use_global_flush_log_at_trx_commit;
 extern char	srv_adaptive_flushing;
 
-#ifdef WITH_INNODB_DISALLOW_WRITES
+#if defined(WITH_WSREP) AND defined(WITH_INNODB_DISALLOW_WRITES)
```
kamil-holubicki: I think this part can be removed. There is no srv_allow_writes_event variable anywhere :)

storage/innobase/os/os0file.cc
```diff
@@ -1256,8 +1256,10 @@ os_file_create_simple_func(
 
 #else /* __WIN__ */
 	int		create_flag;
+#ifdef WITH_INNODB_DISALLOW_WRITES
```
kamil-holubicki, line 1259: Where from is WITH_INNODB_DISALLOW_WRITES? PS does not have this.

storage/innobase/handler/ha_innodb.cc
dutow: Yes, at least it is documented.

storage/innobase/os/os0file.cc
dutow: It's from the wsrep source, it is a read-only like patch.

sql/lock.cc
dutow: yes, as that's not whitespace

check_src/run.sh
dutow: Added.

sql/signal_handler.cc
dutow: Yeah, but in 5.6 many endif already don't have WITH_WSREP comments, and I don't think it adds much value for these very short ifdefs.

kamil-holubicki (approved):
lgtm

PR #1113, opened 2020-05-07
PXC-3034: wsrep_debug: update

borys-belinsky:


doc/source/wsrep-system-index.rst
```diff
@@ -292,18 +292,40 @@ Defines ``DBUG`` options to pass to the wsrep provider.
...
 * Under ``[sst]`` it enables debug logging for the SST script only
 
+This variable may be set to one of the following values:
+
+NONE
+   No debug-level messages.
+
+SERVER
+
+   debug-level messages from wsrep-lib *server_state* and *others* parts. Galera
```
kamil-holubicki: 'others' refer to the diagram I attached to the Jira ticket. In my opinion it can be not clear for the user what is it.
Maybe it is more clear to say here:
SERVER - wsrep-lib general debug-level messages and detailed debug-level messages from 'server_state' part are printed out. 
Galera debug-level logs are printed out.

TRANSACTION -  Same as SERVER + wsrep-lib *transaction* part

... and so on

kamil-holubicki (approved):
lgtm

PR #1123, opened 2020-05-12
PXC-3080: SLOW SLAVE STATUS timeout

venkatesh-prasad-v:
Problem
-------
The test cases `rpl_gtid.rpl_stop_slave_partial_trx` and
`rpl_mts_gtids_restrart_slave_io_lost_trx` fail with test case
timeout on PXC-8.0 builds.

Background
----------
When the IO thread reconnects to a master using GTID auto
positioning while in the middle of a transaction, it will leave
the partial events on the relaylog as is and will fully retrieve
the same transaction again.

When such thing happens, MTS co-ordinator thread handles partial
events by rolling back the incomplete transaction by queuing
QUERY(ROLLBACK) event to make the worker gracefully finish its
work.  Once this is done, the MTS coordinator goes ahead and
applies the ROTATE_LOG_EVENT in synchronous mode.

Analysis & Fix
--------------
With the regression introduced by the initial patch for Galera-4
refresh (commit id - a5af2f8ab50ad70e7590fafb40b40f4853c89c90),
the above behavior was changed and it caused the MTS co-ordinator
thread to perform an early return from the
`exec_relay_log_event()` function without processing the
ROTATE_LOG_EVENT and without unlocking the
Relay_log_info::data_lock thus causing SHOW SLAVE STATUS command
to hang.

With this fix, the SQL thread now processes the ROTATE_LOG_EVENT
synchronously (like before) and performs proper cleanup.

### DIFF
--- a/sql/rpl_slave.cc
+++ b/sql/rpl_slave.cc
@@ -4854,6 +4854,7 @@ static int exec_relay_log_event(THD *thd, Relay_log_info *rli,
   if (ev) {
 #ifdef WITH_WSREP
     if (wsrep_before_statement(thd)) {
+      mysql_mutex_unlock(&rli->data_lock);
       WSREP_INFO("Wsrep before statement error");
       return 1;
     }
@@ -4943,13 +4944,17 @@ static int exec_relay_log_event(THD *thd, Relay_log_info *rli,
         will not try to get the rli->data_lock again.
       */
 #ifdef WITH_WSREP
+      {
         wsrep_after_statement(thd);
 #endif /* WITH_WSREP */
-      /*
-        In the case of an error, coord_handle_partial_binlogged_transaction
-        will not try to get the rli->data_lock again.
-      */
-      return 1;
+        /*
+          In the case of an error, coord_handle_partial_binlogged_transaction
+          will not try to get the rli->data_lock again.
+        */
+        return 1;
+#ifdef WITH_WSREP
+      }
+#endif /* WITH_WSREP */
     }
 
     /* ptr_ev can change to NULL indicating MTS coorinator passed to a Worker */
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=1123) <br/>All committers have signed the CLA.

sql/rpl_slave.cc
```diff
@@ -4937,19 +4938,18 @@ static int exec_relay_log_event(THD *thd, Relay_log_info *rli,
     if (ev->get_type_code() == binary_log::FORMAT_DESCRIPTION_EVENT &&
         ev->server_id != ::server_id && ev->common_header->log_pos != 0 &&
         rli->is_parallel_exec() && rli->curr_group_seen_gtid) {
-      if (coord_handle_partial_binlogged_transaction(rli, ev))
-      /*
```
dutow: The PXC source without WITH_WSREP should be identical to the PS source, except for whitespace differences. I will add a script / CI test soon that checks/enforces this (and also fix existing differences).

Moving the comment, and adding the `{` to the if line breaks this rule. A possible fix is to:

 * keep the comment where it is
* add the `{` as a separate line into the existing WITH_WSREP block
* add a separate ifdef WITH_WSREP around the closing `}`

sql/rpl_slave.cc
```diff
@@ -6691,6 +6691,7 @@ static int slave_start_workers(Relay_log_info *rli, ulong n, bool *mts_inited) {
    Coordinator finalizes with its MTS running status to reset few objects.
 */
 static void slave_stop_workers(Relay_log_info *rli, bool *mts_inited) {
+  DBUG_TRACE;
```
dutow, line 6700: Same comment as above: as this DBUG_TRACE doesn't exists in PS, we have to add it inside WITH_WSREP if we need it.

sql/rpl_slave.cc
venkatesh-prasad-v: Hi Zsolt, Thank you for reviewing the changes. There is a duplication of comment in PS and PXC. Do you suggest to keep them as is?

dutow: Yes, if it is duplicated in PS too, we should keep it duplicated.

venkatesh-prasad-v:
Jenkins link: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/50/

dutow (approved):

PR #1124, opened 2020-05-13
PXC-2935: Assertion `server_state_.rollback_mode() == wsrep::server_s…

kamil-holubicki:
…tate::rm_async' failed

https://jira.percona.com/browse/PXC-2935

Problem:
This problem occurred only if server was started with option --thread_handling=pool-of-threads.
1. Connection 1: start explicit transaction, execute 1st statement that acquires MDL lock
2. Connection 1: Execute 2nd statement. In threadpool_common.cc thread_attach() we wait for potential transaction that is being rolled back by background_rollbacker to finish. After this we acquire thread ownership and client_state's state is shifted to s_exec.
3. Connection 2: Just after acquiring ownership of the thread and moving state to s_exec by Connection 1 execute another query that conflicts with the MDL lock taken by Connection 1. This causes BF eviction of Connection 1 transaction, but as it is in s_exec state, rollback is not requested to be done by background_rollbacker. Transaction state is shifted to s_must_abort and is expected to be aborted in before_command() call.
4. Connection 1: the flow moves on and gets to before_command(). Transaction is in s_must_abort state and we hit the control path with assert that is wrong.

Solution:
Removed wrong assert.
Added MTR test.
Relaxed assert in client_service::debug_sync() which prevented setting debug_sync from thread another than the one owning client_service object. That prevented setting debug_sync in abort_bf when executing thread is aborter thread.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_thread_pool_bf_abort.result
+++ b/mysql-test/suite/galera/r/pxc_thread_pool_bf_abort.result
@@ -1,0 +1,14 @@
+CREATE TABLE t1(a int);
+BEGIN;
+SELECT * FROM t1 WHERE a = 0;
+a
+SET DEBUG_SYNC="thread_attach_after_acquire_ownership SIGNAL reached WAIT_FOR continue";
+SELECT * FROM t1 WHERE a = 0;;
+SET DEBUG_SYNC="now WAIT_FOR reached";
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/214/

altmannmarcelo:
LGTM

altmannmarcelo (approved):

PR #1129, opened 2020-05-15
PXC-3169: wsrep_reject_queries - update

borys-belinsky:


altmannmarcelo:
@borys-belinsky people tend to google for their error messages. My idea with PXC-3169 is that if someone gets `ERROR 1047 (08S01): WSREP has not yet prepared node for application use ` they will land on this page.

altmannmarcelo (changes requested):
@borys-belinsky  people tend to google for their error messages. My idea with PXC-3169 is that if someone gets `ERROR 1047 (08S01): WSREP has not yet prepared node for application use` they will land on this page.

altmannmarcelo (approved):

PR #1132, opened 2020-05-21
PXC-3271: Fixed MTR tests

kamil-holubicki:
https://jira.percona.com/browse/PXC-3271

Most failures cleaned up by:
1. Fixed/adjusted for PXC needs the MTR execution engine Perl scripts
2. Disabled applicable tests (InnoDB Fake Changes related)
3. Rerecorded results for tests that produce result dependent of binlog format (PXC default is ROW)
4. Fixed wrong upstream MTR tests merges
5. Adjusted/Rerecorded results for tests related to behavior specific for PXC (e.g. pool_of_threads)
6. Fixed wrong code merge in sql_prepare.cc

Some tests are still failing:
--- PXC-3279:
rpl.rpl_stm_mixed_mts_rec_crash_safe_checksum
rpl.rpl_stm_mixed_mts_rec_crash_safe
rpl.rpl_stm_mixed_mts_crash_safe
rpl.rpl_stm_mixed_crash_safe
rpl.rpl_crash_safe_master
rpl.rpl_row_mts_rec_crash_safe
rpl.rpl_row_mts_crash_safe
rpl.rpl_row_crash_safe
rpl.rpl_stm_mixed_crash_safe
innodb.innodb_bug59641

--- PXC-3280:
main.mysql_client_test
main.subquery_sj_all
main.subquery_sj_all_bka
main.subquery_sj_all_bkaunique
main.subquery_sj_dupsweed
main.subquery_sj_dupsweed_bka
main.subquery_sj_dupsweed_bkaunique
main.subquery_sj_firstmatch
main.subquery_sj_firstmatch_bka
main.ps_ddl
main.ps_ddl1
main.subquery_sj_firstmatch_bkaunique
main.subquery_sj_loosescan
main.subquery_sj_loosescan_bka
main.subquery_sj_loosescan_bkaunique
main.subquery_sj_mat
main.subquery_sj_mat_bka
main.subquery_sj_mat_bka_nixbnl
main.subquery_sj_mat_bkaunique
main.percona_statement_set

--- PXC-3283:
innodb.innodb-autoinc

--- PXC-3288:
main.alter_table


mysql-test/r/subquery_sj_all_bka_nixbnl.result
```diff
@@ -2105,9 +2105,10 @@ DEALLOCATE PREPARE stmt;
 EXPLAIN SELECT * FROM t1 WHERE (t1.i) IN 
 (SELECT t3.i FROM t2 RIGHT JOIN t3 ON t2.i=t3.i);
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
-1	SIMPLE	t3	ALL	NULL	NULL	NULL	NULL	2	Start temporary
-1	SIMPLE	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
-1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where; End temporary
+1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where
+1	SIMPLE	<subquery2>	eq_ref	<auto_key>	<auto_key>	5	test.t1.i	1	NULL
+2	MATERIALIZED	t3	ALL	NULL	NULL	NULL	NULL	2	NULL
+2	MATERIALIZED	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
```
venkatesh-prasad-v, line 2111: Does it mean that there is a complete change in the query plan itself?

sql/sql_prepare.cc
```diff
@@ -4149,25 +4149,6 @@ bool Prepared_statement::execute(String *expanded_query, bool open_cursor)
...
-    second log entry we output values of parameter markers.
-
-    Rewriting/password obfuscation:
-
-    - Any passwords in the "Execute" line should be substituted with
-      their hashes, or a notice.
-
-  */
-  if (error == 0)
-    log_execute_line(thd);
```
venkatesh-prasad-v, line 4169: Why this change?

mysql-test/r/tc-heuristic-recover.result
```diff
@@ -20,5 +20,4 @@ Restarting with --tc-heuristic-recover=COMMIT
 SELECT * FROM t1;
 a
 1
-2
```
venkatesh-prasad-v: The result file in PS-5.6 has this row -- meaning that server was able to recover the row after the server crash by using --tc-heuristic-recover=COMMIT during restart. So, it looks like there is some problem in recovery code in PXC. Until we find the cause, can we mark this as experimental or disable this test?

venkatesh-prasad-v (commented):
Hi Kamil,

Thank you for working on this issue. Nice work!
All changes look good. But, I have a few questions. If you don't mind, please see them below.

sql/sql_prepare.cc
kamil-holubicki: We have exactly the same comment and call to log_execute_line(thd); as well.
This is consequence of bad merge done in commit 520b939d1b86b8cd5c906fb7f1711b96555f12f8 where PS commit 195d638f08eba9f797c19f4f33b220546c52081e was merged into PXC. 

PS change was introduced together with merging of MySQL upstream 5.6.24 in commit 195d638f08eba9f797c19f4f33b220546c52081e

This was discovered by failing mtr test and fixed.

mysql-test/r/subquery_sj_all_bka_nixbnl.result
kamil-holubicki: PS-5.6 uses STATEMENT binlog format as the default.
PXC-5.6 uses ROW binlog format as the default.

I didn't investigate the exact reason  why the output differs. What I found is that when I set PXC-5.6 binlog format back to STATEMENT, the test produces exactly the same result as PS-5.6. 
On the other hand if I change the default in PS-5.6 to ROW it produces the same output as PXC.

This is why I concluded that this is caused by binlog format change (without getting into details).
My plan is to investigate exact reasons, but as 5.6 will be EOL soon, this has low priority.

mysql-test/r/tc-heuristic-recover.result
kamil-holubicki: You are right. There is something wrong here. 
It works fine on PS-5.6 with statement/row binlog format.
It always fails in the same way on PXC-5.6.
I've created PXC-3289 and revoked this change.

venkatesh-prasad-v (approved):
Looks good to me! Please make sure you test it on Jenkins once.

PR #1136, opened 2020-05-30
Fixed PXC-3270 (PXC8, inconsistency does not crash node)

altmannmarcelo:
https://jira.percona.com/browse/PXC-3270

Fixed wsrep_ignore_apply_errors variable default to
restore 5.x behaviour. As part of the default change, a few test cases
were failing because the output relayed on the old value. Fixed those.


altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc80-test-mtr/216/

Test failures are due to https://jira.percona.com/browse/PXC-3298

mysql-test/suite/galera/t/pxc_install_gr.test
```diff
@@ -11,6 +11,8 @@ call mtr.add_suppression("Plugin group_replication reported");
 call mtr.add_suppression("Group replication cannot be used with Percona XtraDB Cluster in strict mode.");
 call mtr.add_suppression("Plugin 'group_replication' init function returned error.");
 call mtr.add_suppression("Using group replication with Percona XtraDB Cluster is only supported for migration");
+call mtr.add_suppression("Event [0-9]* Query apply failed: [0-9]*, seqno [0-9]*");
+call mtr.add_suppression("PLUGIN group_replication does not exist");
```
kamil-holubicki: Why do we need this suppression? Changing default value for wsrep_ignore_apply_errors should not trigger this warning. In current trunk works without this suppression.

sql/sys_vars.cc
```diff
@@ -7818,7 +7818,7 @@ static Sys_var_bool Sys_wsrep_dirty_reads(
 static Sys_var_uint Sys_wsrep_ignore_apply_errors(
     "wsrep_ignore_apply_errors", "Ignore replication errors",
     GLOBAL_VAR(wsrep_ignore_apply_errors), CMD_LINE(REQUIRED_ARG),
-    VALID_RANGE(WSREP_IGNORE_ERRORS_NONE, WSREP_IGNORE_ERRORS_MAX), DEFAULT(7),
+    VALID_RANGE(WSREP_IGNORE_ERRORS_NONE, WSREP_IGNORE_ERRORS_MAX), DEFAULT(0),
```
kamil-holubicki: DEFAULT(WSREP_IGNORE_ERRORS_NONE)

altmannmarcelo: Done

mysql-test/suite/galera/t/pxc_install_gr.test
altmannmarcelo: I assume you mean `call mtr.add_suppression("PLUGIN group_replication does not exist");` . That was indeed intriguing me. And it seems to be a matter of synchronization issue. Adding a simple sleep in between INSTALL and UNISTALL fixes the issue. I've raised a ticket for us to investigate it ( https://jira.percona.com/browse/PXC-3270 ) as it is not related to the default change (well spotted btw).

kamil-holubicki (approved):
lgtm

PR #1142, opened 2020-06-09
PXC-3026: Fix most of WITH_WSREP issues in the 5.7 branch

dutow:
* Introduces scripts under the check_src directory, which detect
  differences between the pxc WITH_WSREP=OFF source and PS. For the same
  upstream versions, there shouldn't be any changes except for the
  existing whitelists.
  The script also disables WITH_INNODB_DISALLOW_WRITES, as that's
  another 3rd party path not present in PS.
* Whitelisted product name changes, as PXC should be called PXC even
  without WITH_WSREP.
* Whitelisted the use of the WSREP_BINLOG_FORMAT macro, as it defaults
  to a no-op without WITH_WSREP.
* Added WITH_WSREP if(n)defs everywhere where PXC added/deleted logic
* Removed cosmetic differences compared to PS.

### DIFF
--- a/storage/innobase/include/trx0trx.ic
+++ b/storage/innobase/include/trx0trx.ic
@@ -357,12 +357,14 @@ trx_is_high_priority(const trx_t* trx)
 @return the transaction that will be rolled back, null don't care */
 UNIV_INLINE
 const trx_t*
-trx_arbitrate(const trx_t*	requestor,
-	      const trx_t*	holder
 #ifdef WITH_WSREP
-	      ,my_bool		sync
-#endif /* WITH_WSREP */
+trx_arbitrate(const trx_t*	requestor,
+	      const trx_t*	holder,
+	      my_bool		sync
 	     )
+#else
+trx_arbitrate(const trx_t*	requestor, const trx_t*	holder)
+#endif /* WITH_WSREP */
 {
 	ut_ad(!trx_is_autocommit_non_locking(holder));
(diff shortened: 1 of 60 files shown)


storage/innobase/include/trx0trx.ic
```diff
@@ -357,12 +357,14 @@ trx_is_high_priority(const trx_t* trx)
 @return the transaction that will be rolled back, null don't care */
 UNIV_INLINE
 const trx_t*
-trx_arbitrate(const trx_t*	requestor,
-	      const trx_t*	holder
 #ifdef WITH_WSREP
-	      ,my_bool		sync
-#endif /* WITH_WSREP */
+trx_arbitrate(const trx_t*	requestor,
```
kamil-holubicki, line 361: In the header file we have ifndef for this function declaration. It is consistent now, but makes a confusion.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1803,11 +1809,11 @@ RecLock::create(
 		trx_mutex_enter(trx);
 	}
 
-	lock_add(lock, add_to_hash
-#ifdef WITH_WSREP
-		 , c_lock, thr
+#ifndef WITH_WSREP
```
kamil-holubicki, line 1812: More common is #ifdef WITH_WSREP ... #else .... Maybe worth swapping here?

kamil-holubicki (approved):
lgtm. Just a few minor comments.

PR #1145, opened 2020-06-09
grep fails when system doesn't support IPv6

manwe-pl:
There's no check if tcp6 pseudo-file exists, failing on systems without IPv6 support.


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-xtradb-cluster?pullRequest=1145) <br/>All committers have signed the CLA.

kamil-holubicki (approved):

dutow (approved):

PR #1148, opened 2020-06-10
PXC-3280: Different rows order in SELECT result in comparison to PS

kamil-holubicki:
https://jira.percona.com/browse/PXC-3280

Problem:
Many MTR tests fail because order of rows returned by SELECT is different than the recorded one.

Cause:
1. Despite of InnoDB being the default storage engine, there is the logic in MTR framework that forces use of MyISAM if the test does not contain include/have_innodb.inc
2. The default binglog format for PS (and upstream MySQL) is STMT
3. For PXC the default binlog format is ROW
4. Affected tests used STMT + MyISAM when results were recorded, but in PXC they use ROW + MyISAM
5. For STMT binlog format 'DELETE FROM t' resolves to ha_delete_all_rows(). This is written to binlog as a single event. (sql_delete.cc mysql_delete())
6. When we use ROW binlog format, 'DELETE FROM t' is executed as a series of ha_delete_row() for each loop, which turns out into a series of binlog events.
7. For STMT binlog format, ha_delete_all_rows() causes MyISAM to trim the table file. Next inserts just start from file offset 0 as it was the new table.
8. For ROW binlog format, ha_delete_row() causes marking rows in the file as deleted, and adding them to the list of free rows. Next insert gets the last free item from the list and stores new row there. That is why we see rows in reverse order when random scan is done.

Solution:
Rerecord tests results.

### DIFF
--- a/mysql-test/r/subquery_sj_mat_bka_nixbnl.result
+++ b/mysql-test/r/subquery_sj_mat_bka_nixbnl.result
@@ -2155,9 +2155,10 @@ DEALLOCATE PREPARE stmt;
 EXPLAIN SELECT * FROM t1 WHERE (t1.i) IN 
 (SELECT t3.i FROM t2 RIGHT JOIN t3 ON t2.i=t3.i);
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
-1	SIMPLE	t3	ALL	NULL	NULL	NULL	NULL	2	Start temporary
-1	SIMPLE	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
-1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where; End temporary
+1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where
+1	SIMPLE	<subquery2>	eq_ref	<auto_key>	<auto_key>	5	test.t1.i	1	NULL
+2	MATERIALIZED	t3	ALL	NULL	NULL	NULL	NULL	2	NULL
+2	MATERIALIZED	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
 SELECT * FROM t1 WHERE (t1.i) IN 
 (SELECT t3.i FROM t2 RIGHT JOIN t3 ON t2.i=t3.i);
 i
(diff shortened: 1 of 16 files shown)


mysql-test/r/subquery_sj_mat_bka_nixbnl.result
```diff
@@ -2155,9 +2155,10 @@ DEALLOCATE PREPARE stmt;
 EXPLAIN SELECT * FROM t1 WHERE (t1.i) IN 
 (SELECT t3.i FROM t2 RIGHT JOIN t3 ON t2.i=t3.i);
 id	select_type	table	type	possible_keys	key	key_len	ref	rows	Extra
-1	SIMPLE	t3	ALL	NULL	NULL	NULL	NULL	2	Start temporary
-1	SIMPLE	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
-1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where; End temporary
+1	SIMPLE	t1	ALL	NULL	NULL	NULL	NULL	2	Using where
+1	SIMPLE	<subquery2>	eq_ref	<auto_key>	<auto_key>	5	test.t1.i	1	NULL
+2	MATERIALIZED	t3	ALL	NULL	NULL	NULL	NULL	2	NULL
+2	MATERIALIZED	t2	ALL	NULL	NULL	NULL	NULL	0	Using where
```
venkatesh-prasad-v, line 2161: Hi Kamil,

Thank you for working on this. Great analysis!! 
While all other changes looked fine, I did not understand one thing. 

Here the query is returning 4 rows instead of 3. I tried running the test in statement format. It returns 3 rows in statement format.
```
    1 # 
    2 # Run subquery_sj_mat.test with BKA enabled 
    3 #
+   4 SET BINLOG_FORMAT=STATEMENT;
    5 set optimizer_switch='batched_key_access=on,block_nested_loop=off,mrr_cost_based=off';
    6 
    7 --source t/subquery_sj_mat.test
    8 
    9 set optimizer_switch=default;
```
The test passed with expected result content mismatch.
```
CURRENT_TEST: main.subquery_sj_mat_bka_nixbnl
--- /home/venkatesh.prasad/work/pxc56/mysql-test/r/subquery_sj_mat_bka_nixbnl.result	2020-06-10 14:53:56.136797117 +0300
+++ /home/venkatesh.prasad/work/pxc56/bld/mysql-test/var/log/subquery_sj_mat_bka_nixbnl.reject	2020-06-10 15:22:24.059598330 +0300
@@ -1,3 +1,4 @@
+SET BINLOG_FORMAT=STATEMENT;
 set optimizer_switch='batched_key_access=on,block_nested_loop=off,mrr_cost_based=off';
 set optimizer_switch='semijoin=on,materialization=on';
 set @old_opt_switch=@@optimizer_switch;

mysqltest: Result content mismatch
```
I'm not understanding why changing the binlog format is returning different result in this particular case. Is this an expected behavior?

kamil-holubicki: As discussed on slack EXPLAIN SELECT * FROM t is affected by binlog format as well. In server layer we have 'ifs' like if(binlog_format == stmt) .... else.... in explain path.

venkatesh-prasad-v (approved):
Thanks for the explanation. Changes look good to me. Approved!

PR #1150, opened 2020-06-11
PXC-3292: Enable the whole MTR test suite for PXC 5.7

kamil-holubicki:
https://jira.percona.com/browse/PXC-3292

Enabled all MTR tests. Fixes done:
1. MTR framework Perl scripts fixed
2. Tests results rerecorded because of output change caused by PXC additions
3. Tests results rerecorded because of binlog_format default changed to ROW in PXC
4. Disabled tests that use variables that are read only in PXC
5. PFS_MAX_STAGE_CLASS changed 150 -> 180 to allow PXC related instrumentation
6. Group replication tests fixed. This involved adjusting code logic in lock0lock.cc and rpl_gtid_state.cc
7. Fixed test percona_rpl_mts_backup_lock_ps3345 (change in binlog.cc)
8. -D_GLIBCXX_ASSERTIONS disabled for CentOS 6/7 debug build (change in Galera repo)

Note that several (groups of) tests still fail:
PXC-3295 - HP transactions related tests
PXC-3323 - percona_extended_innodb_status.test
PXC-3297 - keyring_vault related tests
PXC-3296 - merge.test and merge_mmap.test
PXC-3283 - innodb-autoinc.test


kamil-holubicki:
I will squash before merge. Left as separate commits to make it easier to review.

mysql-test/r/mysqldump.result
```diff
@@ -679,7 +679,7 @@ drop database mysqldump_test_db;
 # --default-character-set=xxx. However, we should dump in UTF8
 # if it is explicitely set.
 CREATE TABLE t1 (a  CHAR(10));
-INSERT INTO t1  VALUES (_latin1 '??');
+INSERT INTO t1  VALUES (_latin1 '����');
```
dutow, line 682: This is some accidental change, maybe because of the editor encoding?

sql/rpl_gtid_state.cc
```diff
@@ -545,12 +546,14 @@ enum_return_status Gtid_state::generate_automatic_gtid(THD *thd,
     Gtid automatic_gtid= { specified_sidno, specified_gno };
 
 #ifdef WITH_WSREP
-    /* If the trx has been executed in wsrep then get wsrep_sidno
-    and not the normal server_sid no. */
-    if (WSREP(thd) && thd->wsrep_trx_meta.gtid.seqno != -1 &&
+    /* If the trx has been executed in wsrep then get wsrep_sidno and not the
+     * normal server_sid no. Retain the same sidno if group_replication is
```
dutow: typo: server_sidno

storage/innobase/lock/lock0lock.cc
```diff
@@ -1983,20 +1983,26 @@ RecLock::add_to_waitq(const lock_t* wait_for, const lock_prdt_t* prdt)
 	gathering) transactions. The problem is that we don't currently
 	block them using the TrxInInnoDB() mechanism. */
 
+	bool	high_priority = trx_is_high_priority(m_trx);
 #ifdef WITH_WSREP
-	if (wsrep_on(m_trx->mysql_thd) &&
+
+    /* Allow high priority transactions to jump over the low priority locks and
+     * rollback them if Galera is not loaded. */
```
dutow: High priority transactions also work with codership mysql, I based the 8.0 fix on that. We could use the same patch here.

mysql-test/r/mysqldump.result
kamil-holubicki: This is OK. '??' is wrong caused by editor. 
This file looks like:
image
and in fact is the same as upstream (PS) version of it.

storage/innobase/lock/lock0lock.cc
kamil-holubicki: Hi @venkatesh-prasad-v , can you please check it out?

venkatesh-prasad-v: Hi @dutow!! I and Kamil decided to handle HPTs as part of PXC-3295. So, we have reverted the changes done in lock0lock.cc and will work towards evaluating the backport of the 8.0 fix to 5.7.

dutow (approved):
Approved after squashing.

PR #1162, opened 2020-06-17
PXC-3326 Enhance Data At Rest Encryption Documentation

patrickbirch:
Modified management/data_at_rest_encryption

In 5.7, only file-per-tablespaces and temporary files

### DIFF
--- a/doc/source/management/data_at_rest_encryption.rst
+++ b/doc/source/management/data_at_rest_encryption.rst
@@ -4,8 +4,6 @@
 Data at Rest Encryption
 =======================
 
-This feature is considered **BETA** quality.
-
 .. contents::
    :local:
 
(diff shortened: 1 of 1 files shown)


kathy-williamson (changes requested):
This sentence doesn't make sense.  What is meant by what is after the )? 
Data-in-transit can be encrypted using SSL connections (details are available in the `encrypt traffic documentation <https://www.percona.com/doc/percona-xtradb-cluster/LATEST/security/encrypt-traffic.html>`_) and data-at-rest only.


This sentence needs to be corrected - If the user has a bootstrapped a node with the keyring enabled, then the upcoming nodes of the
cluster either inherit the keyring (encrypted key) from the donor node
has a - the a should be deleted 
inherit should be inherits 

5.6 does not have encrypted tablespace - please make tablespace plural 

change - 'and SST on joiner is' to 'and SST on a joiner is to'

change 'recommends using same keyring_plugin ' to 'recommends using the same keyring_plugin'
change 'an error is notified during' to 'an error is raised during'

this is confusing - If some nodes of the cluster are unable to connect to the vault-server, this
relates only to these specific nodes: e.g. if node-1 can connect, and
node-2 is not, only node-2 refuses to start. If the server has
pre-existing encrypted objects and, on reboot, the server fails to connect to the
vault-server, the objects are not accessible.
Suggested change - 
If any node of the cluster is unable to connect to the vault-server the encrypted objects on that server will not be accessible.

patrickbirch:
Made requested changes.

mohitj1988 (approved):
LGTM

kathy-williamson (approved):

PR #1172, opened 2020-06-25
PXC-3295: [5.7] [MTR] Make high priority transactions work with PXC

venkatesh-prasad-v:
This is a backport of PXC-3128: High priority transactions should work
with WSREP.

Issue: when wsrep is running, high priority transactions weren't able to
kill normal priority transactions. This basically broke the scenario
where a PXC node is also part of a normal replication setup.

Root cause: We intentionally broke this feature in 5.7, in commit
5be95d8ff766. The comment/commit message explains that galera used a
different mechanism and thus it was disabled.

Compared to that statement codership fixed HPTs also in 5.7, in commit
f2dd26e10785, by translating the HPT kill logic into wsrep functions.

After merging the codership fix in 5.7, we never reverted our
workaround.

Fix:

* Reverted 5be95d8ff766
* Backported 03a80d175c70b from PXC-8.0.
* Backported some of the high priority tests to the galera suite to ensure
  that they don't break in the future
* This commit also improves the start_transactin_high_prio mtr include
  so it works when used together with --debug

Testing Link: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc57-test-mtr/63/console
Note: Most tests are failing because of missing dependencies. I have reported PXC-3342 for tracking them.

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -1983,20 +1983,21 @@ RecLock::add_to_waitq(const lock_t* wait_for, const lock_prdt_t* prdt)
 	gathering) transactions. The problem is that we don't currently
 	block them using the TrxInInnoDB() mechanism. */
 
+	bool	high_priority = trx_is_high_priority(m_trx);
 #ifdef WITH_WSREP
 	if (wsrep_on(m_trx->mysql_thd) &&
 	    m_trx->lock.was_chosen_as_deadlock_victim) {
 		return(DB_DEADLOCK);
 	}
 
-	lock = create(m_trx, true, true, prdt,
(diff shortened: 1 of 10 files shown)


storage/innobase/lock/lock0lock.cc
```diff
@@ -2716,11 +2716,9 @@ RecLock::make_trx_hit_list(
 
 			/* Assert that it is not waiting for current record. */
 			ut_ad(trx->lock.wait_lock != next);
-#ifdef UNIV_DEBUG
```
venkatesh-prasad-v: Self Review: It was added for debugging. Revert this change.

dutow (approved):

PR #1184, opened 2020-07-09
PXC-3243 : BF aborting a stored procedure hangs

kennt-percona:
Issue
When running a SP with a transaction insde, BF-aborting the SP
can cause the system to hang.

Solution
This happens because the transaction is holding onto a lock,
but the BF-abort is aborting a statement, which may not be the
one holding the lock.
This change will force the BF-abort to propagate and abort and retry
the SP, not the statement.

### DIFF
--- a/mysql-test/suite/galera/t/galera_migrate.test
+++ b/mysql-test/suite/galera/t/galera_migrate.test
@@ -233,3 +233,8 @@ DROP USER sst;
 CALL mtr.add_suppression("InnoDB: Error: Table \"mysql\"\\.\"innodb_index_stats\" not found");
 CALL mtr.add_suppression("Hostname '[0-9a-zA-Z\.\-]+' does not resolve to '127.0.0.2'.");
 CALL mtr.add_suppression("Host name '[0-9a-zA-Z\.\-]+' could not be resolved.*");
+
+--connection node_4
+CALL mtr.add_suppression("InnoDB: Error: Table \"mysql\"\\.\"innodb_index_stats\" not found");
+CALL mtr.add_suppression("Hostname '[0-9a-zA-Z\.\-]+' does not resolve to '127.0.0.2'.");
+CALL mtr.add_suppression("Host name '[0-9a-zA-Z\.\-]+' could not be resolved.*");
--- a/mysql-test/suite/galera/t/galera_sp_bf_abort2.inc
+++ b/mysql-test/suite/galera/t/galera_sp_bf_abort2.inc
@@ -1,0 +1,76 @@
+
+# Setup the test on node 1
+--connection node_1
+--echo # Node 1
+--let $wsrep_local_bf_aborts_before = `SELECT VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_bf_aborts'`
+
+USE database1;
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY, f2 TEXT) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1, 'initial');
+INSERT INTO t1 VALUES (2, 'initial');
+INSERT INTO t1 VALUES (3, 'initial');
+INSERT INTO t1 VALUES (4, 'initial');
+
+# Invoke the SP
+--connection node_1a
+--echo # Node 1a
+USE database1;
+
+SET DEBUG_SYNC = "ha_innobase_end_of_write_row SIGNAL entered WAIT_FOR continue";
+--send_eval CALL $galera_sp_bf_abort_proc
+
+# Ensure that we have reached the sync point
+--connection node_1
+--echo # Node 1
+SET SESSION wsrep_sync_wait = 0;
+SET DEBUG_SYNC = "now WAIT_FOR entered";
+
+# Force the thread on node1 to be aborted (by making a change on node2)
+--connection node_2
+--echo # Node 2
+UPDATE database1.t1 SET f2 = "from node2" WHERE f1 < 4;
+
+
+--connection node_1
+--echo # Node 1
+# Ensure that the update from node2 has reached node1
+--let $wait_condition = SELECT f2 = 'from node2' FROM database1.t1 WHERE f1 = 1;
+--source include/wait_condition.inc
+
+# Signal the thread to continue
+# The update from node2 should have bf aborted the op on node1
+SET DEBUG_SYNC = "now SIGNAL continue";
+
+--connection node_1a
+--echo # Node 1a
+
+if ($galera_sp_bf_abort2_expect_error)
+{
+	--error ER_LOCK_DEADLOCK
+	--reap
+}
+if (!$galera_sp_bf_abort2_expect_error)
+{
+	--reap	
+}
+
+--connection node_1
+--echo # Node 1
+
+--let $wsrep_local_bf_aborts_after = `SELECT VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_bf_aborts'`
+--let $assert_text = wsrep_local_bf_aborts has been incremented once
+--let $assert_cond = $wsrep_local_bf_aborts_after - $wsrep_local_bf_aborts_before = 1 AS wsrep_local_bf_aborts_increment;
+--source include/assert.inc
+
+
+SELECT * FROM database1.t1;
+
+SET DEBUG_SYNC = "RESET";
+
+--connection node_2
+--echo # Node 2
+SELECT * FROM database1.t1;
+
+--connection node_1
+DROP TABLE t1;
+
--- a/mysql-test/suite/galera/t/galera_sp_bf_abort2.test
+++ b/mysql-test/suite/galera/t/galera_sp_bf_abort2.test
@@ -1,0 +1,151 @@
+#
+# PXC-3243: Test BF-aborting during a transaction in an SP
+# This tests the behavior while an SP is BF-aborted during
+# an operation in a transaction within the SP.
+#
+
+--source include/have_debug_sync.inc
+--source include/have_innodb.inc
+--source include/galera_cluster.inc
+
+
+# Test setup
+--echo # Node 1
+--connection node_1
+CREATE DATABASE database1;
+USE database1;
+
+DELIMITER |;
+CREATE PROCEDURE proc_with_transaction()
+BEGIN
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp");
+  COMMIT;
+END|
+
+CREATE PROCEDURE proc_with_transaction_continue_handler()
+BEGIN
+  DECLARE CONTINUE HANDLER FOR SQLEXCEPTION BEGIN END;
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp continue");
+  COMMIT;
+END|
+
+CREATE PROCEDURE proc_with_transaction_exit_handler()
+BEGIN
+  DECLARE EXIT HANDLER FOR SQLEXCEPTION BEGIN END;
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp exit");
+  COMMIT;
+END|
+DELIMITER ;|
+
+# Determine initial number of connections (set $count_sessions)
+--source include/count_sessions.inc
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+
+# ------------------------------------
+# Test 1a: no handlers, wsrep_retry_autocommit=0
+# The call to the SP will fail and return an error.
+#
+--connection node_1a
+--echo # Node 1a
+USE database1;
+SET SESSION wsrep_retry_autocommit = 0;
+--let $galera_sp_bf_abort_proc = proc_with_transaction
+--let $galera_sp_bf_abort2_expect_error = 1
+--source galera_sp_bf_abort2.inc
+
+
+# ------------------------------------
+# Test 1b: exit handler, wsrep_retry_autocommit=0
+# The call to the SP will fail and return an error.
+#
+--connection node_1a
+--echo # Node 1a
+USE database1;
+SET SESSION wsrep_retry_autocommit = 0;
+--let $galera_sp_bf_abort_proc = proc_with_transaction_exit_handler
+--let $galera_sp_bf_abort2_expect_error = 1
+--source galera_sp_bf_abort2.inc
+
+
+# ------------------------------------
+# Test 1c: continue handler, wsrep_retry_autocommit=0
+# The call to the SP will fail and return an error.
+#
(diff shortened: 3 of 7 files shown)


sql/sql_parse.cc
```diff
@@ -6341,13 +6362,28 @@ case SQLCOM_PREPARE:
...
+        PXC-3243
+        To handle the error in an SP, the thd->killed must be propagated
+        back up to the SP call so that the trans_rollback_stmt()
+        can be invoked for the SP call.
+
+        So leaving thd->killed unchanged will ensure that sp_head::execute
+        will set the err_status to true.
+       */
+      if (conflict_state != MUST_ABORT) {
+#endif /* WITH_WSREP */
```
kamil-holubicki: Indentation

sql/sql_parse.cc
```diff
@@ -6310,28 +6310,49 @@ case SQLCOM_PREPARE:
 
   if (! thd->in_sub_stmt)
   {
+    int killed_errno;
```
kamil-holubicki: @dutow created the tool that compares PS code vs PXC code and expects everything PXC related to be within WITH_WSREP block. Everything outside should equal what we have in PS. That makes it easier to spot merging problems.
Having that said, my proposal is to just transform all these changes into #ifdef WITH_WSREP .... #else .... #endif. Original code is just a few lines, so keeping logic separately for PXC and not PXC will make it more readable than chunking like this.

sql/sql_parse.cc
```diff
@@ -6341,13 +6362,28 @@ case SQLCOM_PREPARE:
...
+      /*
+        PXC-3243
+        To handle the error in an SP, the thd->killed must be propagated
+        back up to the SP call so that the trans_rollback_stmt()
+        can be invoked for the SP call.
+
+        So leaving thd->killed unchanged will ensure that sp_head::execute
+        will set the err_status to true.
+       */
+      if (conflict_state != MUST_ABORT) {
```
kamil-holubicki: Shouldn't we check has_sp_runtime_ctx here?

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,475 @@
+#
```
kamil-holubicki, line 1: This test has a lot of duplication (setup, checks). Would it be possible to extract common parts into inc(s) files and have them in one place?

mysql-test/suite/galera/t/galera_migrate.test
```diff
@@ -233,3 +233,8 @@ DROP USER sst;
 CALL mtr.add_suppression("InnoDB: Error: Table \"mysql\"\\.\"innodb_index_stats\" not found");
 CALL mtr.add_suppression("Hostname '[0-9a-zA-Z\.\-]+' does not resolve to '127.0.0.2'.");
 CALL mtr.add_suppression("Host name '[0-9a-zA-Z\.\-]+' could not be resolved.*");
+
```
kamil-holubicki, line 236: Why we need this? How is it related to the fix?

kamil-holubicki (changes requested):

storage/innobase/handler/ha_innodb.cc
```diff
@@ -8716,6 +8716,8 @@ ha_innobase::write_row(
 func_exit:
 	innobase_active_small();
 
+	DEBUG_SYNC(user_thd, "ha_innobase_end_of_write_row");
```
kamil-holubicki, line 8720: #ifdef WITH_WSREP

kamil-holubicki (dismissed):

mysql-test/suite/galera/t/galera_migrate.test
kennt-percona: Not related, but MTR tests were failing without this (this was fixed in 5.7 but not in 5.6)

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kennt-percona: changed

sql/sql_parse.cc
kennt-percona: changed code to make separate wsrep blocks

storage/innobase/handler/ha_innodb.cc
kennt-percona: fixed

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,144 @@
...
+# Test the operation of DDLs that affect multiple database objects
+#
+
+--source include/galera_cluster.inc
+--source include/have_innodb.inc
+--source include/have_debug_sync.inc
```
venkatesh-prasad-v: I suggest to include restrictive incs in the beginning. 
```suggestion
--source include/have_debug_sync.inc
--source include/have_innodb.inc
--source include/galera_cluster.inc
```

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,144 @@
...
+    INSERT INTO t1 VALUES (10, "from sp exit");
+  COMMIT;
+END|
+DELIMITER ;|
+
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+--connection node_1a
+--echo # Node 1a
+--let $wsrep_retry_autocommit_orig = `SELECT @@SESSION.wsrep_retry_autocommit;`
```
venkatesh-prasad-v: No need to store and restore the session variables.

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,144 @@
...
+
+
+# ------------------------------------
+# Test 1a: no handlers, wsrep_retry_autocommit=0
+# The call to the SP will fail and return an error.
+#
+--connection node_1a
+--echo # Node 1a
```
venkatesh-prasad-v, line 56: Not a strong opinion though! We are already on `node_1a`

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,144 @@
...
+USE database1;
+set SESSION wsrep_retry_autocommit = 0;
```
venkatesh-prasad-v: All SQL commands should be in uppercase. 
```suggestion
SET SESSION wsrep_retry_autocommit = 0;
```
Please change it in all occurences.

mysql-test/suite/galera/t/galera_sp_bf_abort2.inc
```diff
@@ -0,0 +1,76 @@
...
+
+if ($galera_sp_bf_abort2_expect_error)
+{
+	--error ER_LOCK_DEADLOCK
+	--reap
+}
+if (!$galera_sp_bf_abort2_expect_error)
+{
+	--reap	
+}
```
venkatesh-prasad-v, line 55: Even though no error shall be reported when we re-execute the SP, it still doesn't enusre that retry was performed. So, would like to validate that retry was successul. My suggestion is to change it to
```suggestion
if ($galera_sp_bf_abort2_expect_error)
{
    --error ER_LOCK_DEADLOCK
    --reap

    # Ensure that retry has not been done.
    --let $assert_text = Retry has not been done.
    --let $assert_cond = COUNT(*) = 3 FROM database1.t1 WHERE f2 = "from node2"
    --source include/assert.inc
}
if (!$galera_sp_bf_abort2_expect_error)
{
    --reap  

    # Ensure that retry has been performed.
    --let $assert_text = Retry has been performed.
    --let $assert_cond = COUNT(*) = 1 FROM database1.t1 WHERE f2 = "from node2"
    --source include/assert.inc
}
```

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,144 @@
+#
+# Test the operation of DDLs that affect multiple database objects
```
venkatesh-prasad-v: Doesn't sync with the test body and also a missing reference to the JIRA issue.

sql/sql_parse.cc
```diff
@@ -6310,45 +6310,91 @@ case SQLCOM_PREPARE:
...
+    else
+    {
+      /* If commit fails, we should be able to reset the OK status. */
+      thd->get_stmt_da()->set_overwrite_status(true);
+      trans_commit_stmt(thd);
+      thd->get_stmt_da()->set_overwrite_status(false);
+    }
+    if (killed == THD::KILL_QUERY ||
+        killed == THD::KILL_TIMEOUT ||
+        killed == THD::KILL_BAD_DATA)
```
venkatesh-prasad-v: Shouldn't it be `thd->killed`?

sql/sql_parse.cc
```diff
@@ -6310,45 +6310,91 @@ case SQLCOM_PREPARE:
...
     {
+      trans_rollback_stmt(thd);
+
       /*
-        The error was cleared, but THD was aborted by wsrep and
-        wsrep_conflict_state is still set accordingly. This
-        situation is expected if we are running a stored procedure
-        that declares a handler that catches ER_LOCK_DEADLOCK error.
-        In which case the error may have been cleared in method
-        sp_rcontext::handle_sql_condition().
```
venkatesh-prasad-v: Can we retain this comment if it is still valid after this bugfix?

sql/sql_parse.cc
```diff
@@ -6310,45 +6310,91 @@ case SQLCOM_PREPARE:
...
+
+    mysql_mutex_lock(&thd->LOCK_wsrep_thd);
+    conflict_state= wsrep_thd_conflict_state(thd);
+    has_sp_runtime_ctx= (thd->sp_runtime_ctx != NULL);
+    killed_errno= thd->killed_errno();
+    is_error= thd->is_error();
+    killed= thd->killed;
+    mysql_mutex_unlock(&thd->LOCK_wsrep_thd);
+
     /* report error issued during command execution */
```
venkatesh-prasad-v, line 6325: This line fails the `/check_src/check_file.sh` script.

sql/sql_parse.cc
```diff
@@ -6310,45 +6310,91 @@ case SQLCOM_PREPARE:
...
+      /*
+        PXC-3243
+        To handle the error in an SP, the thd->killed must be propagated
+        back up to the SP call so that the trans_rollback_stmt()
+        can be invoked for the SP call.
+
+        So leaving thd->killed unchanged will ensure that sp_head::execute
+        will set the err_status to true.
+       */
+      if (!has_sp_runtime_ctx || conflict_state != MUST_ABORT) {
```
venkatesh-prasad-v: Let us keep it uniform with the rest of the code
```suggestion
      if (!has_sp_runtime_ctx || conflict_state != MUST_ABORT)
      {
```

sql/sql_parse.cc
```diff
@@ -6310,45 +6310,91 @@ case SQLCOM_PREPARE:
...
-        sp_rcontext::handle_sql_condition().
+        PXC-3243
+        If BF-aborted, do not clear the conflict-state/killed variables.
+        Let the error propagate upward (if this was called from an SP),
+        to ensure that the locks are cleared.
       */
-      trans_rollback_stmt(thd);
-      thd->wsrep_conflict_state= NO_CONFLICT;
-      thd->killed= THD::NOT_KILLED;
+      if (conflict_state != MUST_ABORT) {
```
venkatesh-prasad-v: Let us keep it uniform with the rest of the code.
```suggestion
      if (conflict_state != MUST_ABORT)
      {
```

venkatesh-prasad-v (changes requested):
Hi Kenn,

Thank you working on this issue. Have a few suggestions, if you don't mind.

sql/sql_parse.cc
kamil-holubicki: +1

sql/sql_parse.cc
kamil-holubicki: +1

sql/sql_parse.cc
kamil-holubicki: +1

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kamil-holubicki: +1

kamil-holubicki (commented):
I think it would be good to address comments from Venkatesh. Especially this one: "Shouldn't it be thd->killed?". Other than that it is OK for me.

sql/sql_parse.cc
kamil-holubicki: +1

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kennt-percona: fixed

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kennt-percona: fixed

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kennt-percona: fixed

sql/sql_parse.cc
kennt-percona: good catch.  fixed

sql/sql_parse.cc
kennt-percona: readded the comment

sql/sql_parse.cc
kennt-percona: fixed

sql/sql_parse.cc
kennt-percona: fixed

sql/sql_parse.cc
kennt-percona: fixed

mysql-test/suite/galera/t/galera_sp_bf_abort2.inc
kennt-percona: this would be redundant, since we do a "select * from t1" to check what is on both nodes.  It would probably make things more explicit though.

venkatesh-prasad-v: IMHO, we always should not depend on the output of the select commands as it is not error proof and is also prone to false positive errors.

Using asserts makes the test more robust to test failures. Reason being,
1. For example, let's say that the retry logic gets broken in the future for some reasons. Even though the test may fail in the beginning, but it shall pass on rerecording. 
2. We need not wait for the MTR to complete the test and to perform diff in the end if there is an issue with the retry. With asserts, it is faster as the test fails immediately when the condition fails.

We can avoid both these issues by using asserts. 

Having said that, I do agree that asserts have little overhead, but it is worth to ensure correctness of the results. But this doesn't really block you, as I'm fine with the current approach as well. So, I leave it to your discretion to use the assert.

venkatesh-prasad-v (approved):
Hi Kenn,

Thank you for addressing all the comments. Patch looks good now. Aprroved from my side.

kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/52/

PR #1190, opened 2020-07-20
PXC-3352 [5.7]: Unexpected ERROR 1205 modifying a child table in a FK relationship

venkatesh-prasad-v:
PXC-3352: Unexpected ERROR 1205 modifying a child table in a FK relationship

https://jira.percona.com/browse/PXC-3352

Problem
-------
When deleting/updating from a child table in a FK relationship, if the
parent table's referenced rows are locked, the operation on a child table
failed with lock wait timeout error when the parent table is unlocked.

Analysis
--------
1. For DELETE query

This is the stacktrace of the DELETE query while waiting for the parent
rows to be unlocked.
```
__pthread_cond_wait
os_event::wait
os_event::wait_low
os_event_wait_low
lock_wait_suspend_thread
row_ins_check_foreign_constraint
wsrep_row_upd_check_foreign_constraint 
row_upd_sec_index_entry
row_upd_sec_step
row_upd
row_upd_step
row_update_for_mysql_using_upd_graph
row_update_for_mysql
```
However, the same DELETE query doesn't wait in PS-5.7. 

In PS-5.7, when a row is being modified by an UPDATE/DELETE query, it even
updates the secondary index if there any index defined on the table. While
doing that, it first marks the row as deleted in the secondary index and
inserts into the index with the new value if it is an UPDATE query. This
updation of secondary index is necessary only for UPDATE queries and there
is no need to do the same for DELETE queries.

However, the same was not being done properly in PXC-5.7 because of the
special handling in WSREP for the foreign keys in InnoDB. This made even
the DELETE query to inadvertently update the secondary index
(row_ins_check_foreign_constraint()) thereby trying to lock the parent
table and thus causing lock wait timeout error.

2. For UPDATE query

During Update/Delete, whenever we fail to acquire lock on a row, the
transaction status is to DB_LOCK_WAIT and transaction enters into wait
state waiting for the row to be unlocked; and when the row is unlocked, the
operation is retried.

However, in case of PXC, the temporary DB_LOCK_WAIT during foreign key
checks was being translated into a permanent DB_LOCK_WAIT_TIMEOUT because
of the fallthrough in the error handling code thereby causing the queyr to
fail without retrying the operation.

Fix
---
1. The server is made to behave same way as of the PS-5.7 by removing the
check for DELETE query in `wsrep_row_upd_check_foreign_constraints()`
function which caused the thread to enter the wait.

2. DB_LOCK_WAIT error is now properly handled and the server retries the
operation.

Additionally, this patch 

1. Adds new foreign key tests in galera suite invoking their respective
test from InnoDB with galera enabled.
2. Re-enables the disabled `galera_fk_multitable` and
`galera_toi_ddl_fk_insert` test cases.

Testing
---
v1 - Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/51/testReport/
v2 - Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/57/testReport/

Note to Reviewers:
---
* Many tests in the above link are failing because the galera suite was run with higher value of `--parallel` causing making ports to be unavailable during the test run. To address this issue, [PXC-3364](https://jira.percona.com/browse/PXC-3364) has been reported.

### DIFF
--- a/storage/innobase/row/row0upd.cc
+++ b/storage/innobase/row/row0upd.cc
@@ -420,10 +420,8 @@ wsrep_row_upd_check_foreign_constraints(
 		NOT break the constraint. */
 
 		if (foreign->foreign_index == index
-		    && (node->is_delete
-			|| row_upd_changes_first_fields_binary(
-				entry, index, node->update,
-				foreign->n_fields))) {
+		    && row_upd_changes_first_fields_binary(
+		        entry, index, node->update, foreign->n_fields)) {
 
 			if (foreign->referenced_table == NULL) {
 				foreign->referenced_table = 
@@ -2496,6 +2494,16 @@ row_upd_sec_index_entry(
 				case DB_NO_REFERENCED_ROW:
 					err = DB_SUCCESS;
 					break;
+				/* We can expect DB_LOCK_WAIT when the transaction that blocked
+				our lock was rolled back or committed while we were waiting
+				for the lock. In such cases, we can retry the operation.
+
+				See row_mysql_handle_errors() and
+				row_update_for_mysql_using_upd_graph()
+				for more details.*/
+
+				case DB_LOCK_WAIT:
+					break;
 				case DB_DEADLOCK:
 					if (wsrep_debug) {
 						ib::warn() << "WSREP: sec index FK check fail for deadlock"
@@ -2504,7 +2512,6 @@ row_upd_sec_index_entry(
 					}
 					break;
 				case DB_LOCK_WAIT_TIMEOUT:
-				case DB_LOCK_WAIT:
 					err = DB_LOCK_WAIT_TIMEOUT;
 					break;
 				default:
--- a/mysql-test/suite/galera/r/galera-index-online-fk.result
+++ b/mysql-test/suite/galera/r/galera-index-online-fk.result
@@ -1,0 +1,750 @@
+CREATE TABLE parent (a INT PRIMARY KEY, b INT NOT NULL) ENGINE = InnoDB;
+INSERT INTO parent VALUES(1,2),(2,3);
+CREATE INDEX tb ON parent(b);
+INSERT INTO parent VALUES(10,20),(20,30);
+CREATE TABLE child (a1 INT PRIMARY KEY, a2 INT) ENGINE = InnoDB;
+CREATE INDEX tb ON child(a2);
+INSERT INTO child VALUES(10,20);
+ALTER TABLE child ADD FOREIGN KEY(a2) REFERENCES parent(b),
+ALGORITHM = INPLACE;
+ERROR 0A000: ALGORITHM=INPLACE is not supported. Reason: Adding foreign keys needs foreign_key_checks=OFF. Try ALGORITHM=COPY.
+SET foreign_key_checks = 0;
+ALTER TABLE child ADD CONSTRAINT fk_1 FOREIGN KEY (a2)
+REFERENCES parent(b) ON DELETE SET NULL ON UPDATE CASCADE,
+ALGORITHM = INPLACE;
+SELECT * FROM information_schema.INNODB_SYS_FOREIGN;
+ID	FOR_NAME	REF_NAME	N_COLS	TYPE
+test/fk_1	test/child	test/parent	1	6
+SELECT * FROM information_schema.INNODB_SYS_FOREIGN_COLS;
+ID	FOR_COL_NAME	REF_COL_NAME	POS
+test/fk_1	a2	b	0
+ALTER TABLE child ADD CONSTRAINT fk_1 FOREIGN KEY (a2)
+REFERENCES parent(b) ON DELETE SET NULL ON UPDATE CASCADE,
+ALGORITHM = INPLACE;
+ERROR HY000: Duplicate foreign key constraint name 'test/fk_1'
+SET foreign_key_checks = 1;
+INSERT INTO child VALUES(1,2),(2,3);
+INSERT INTO child VALUES(4,4);
+ERROR 23000: Cannot add or update a child row: a foreign key constraint fails (`test`.`child`, CONSTRAINT `fk_1` FOREIGN KEY (`a2`) REFERENCES `parent` (`b`) ON DELETE SET NULL ON UPDATE CASCADE)
+SELECT * FROM parent;
+a	b
+1	2
+2	3
+10	20
+20	30
+SET foreign_key_checks = 0;
+ALTER TABLE child ADD CONSTRAINT fk_20 FOREIGN KEY (a1, a2)
+REFERENCES parent(a, b) ON DELETE CASCADE ON UPDATE CASCADE,
+ALGORITHM = INPLACE;
+ERROR HY000: Failed to add the foreign key constaint. Missing index for constraint 'fk_20' in the referenced table 'parent'
+SHOW WARNINGS;
+Level	Code	Message
+Error	1822	Failed to add the foreign key constaint. Missing index for constraint 'fk_20' in the referenced table 'parent'
+SHOW ERRORS;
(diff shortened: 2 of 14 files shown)


storage/innobase/row/row0upd.cc
```diff
@@ -420,10 +420,8 @@ wsrep_row_upd_check_foreign_constraints(
 		NOT break the constraint. */
 
 		if (foreign->foreign_index == index
-		    && (node->is_delete
-			|| row_upd_changes_first_fields_binary(
-				entry, index, node->update,
-				foreign->n_fields))) {
+		    && row_upd_changes_first_fields_binary(
```
altmannmarcelo, line 423: I'm trying to understand why node->is_delete is not required here. I can see that on `row_upd_sec_index_entry` we call `row_upd_index_is_referenced` which is the base function for `wsrep_row_upd_check_foreign_constraints`, then if `WITH_WSREP` is defined we call `wsrep_row_upd_check_foreign_constraints` which basically runs a check from `row_ins_check_foreign_constraint` on the parent table instead of on the child table (calling `row_ins_check_foreign_constraint` with 1st parameter = TRUE). I have two questions:

1. Why this second call to check FK constrain is required only on WSREP
2. By removing `node->is_delete` we won't be calling the function that is causing the issue, what happens on an update that returns true for  row_upd_changes_first_fields_binary and has the parent table records locked?

altmannmarcelo (commented):
@venkatesh-prasad-v I have a few questions around the part of code you changed.

Also, not mandatory since its still a WIP but have a read at https://confluence.percona.com/pages/viewpage.action?pageId=68393728 (especially the part that mentions inserting the link for JIRA issue for easy access).

storage/innobase/row/row0upd.cc
altmannmarcelo: While talking with @venkatesh-prasad-v about this fix we found out that test case is failing for UPDATES:

```
setup
CREATE TABLE parent_t ( id INT AUTO_INCREMENT PRIMARY KEY );
CREATE TABLE child_t ( id INT AUTO_INCREMENT PRIMARY KEY, parent_id INT,
CONSTRAINT `FK_child_parent_id` FOREIGN KEY (`parent_id`) REFERENCES `parent_t` (`id`));
INSERT INTO parent_t VALUES (), (), (), (), ();
INSERT INTO child_t (parent_id) SELECT id FROM parent_t;
Session 1
BEGIN;
SELECT * FROM parent_t WHERE id = 1 FOR UPDATE;
Session 2
UPDATE child_t SET parent_id = 2 WHERE parent_id = 1;
Session 1
COMMIT;--(ROLLBACK;, abort, doesn't matter)
Session 2
```

altmannmarcelo (changes requested):
While talking with @venkatesh-prasad-v about this fix we found out that test case is failing for UPDATES:

```
setup
CREATE TABLE parent_t ( id INT AUTO_INCREMENT PRIMARY KEY );
CREATE TABLE child_t ( id INT AUTO_INCREMENT PRIMARY KEY, parent_id INT,
CONSTRAINT `FK_child_parent_id` FOREIGN KEY (`parent_id`) REFERENCES `parent_t` (`id`));
INSERT INTO parent_t VALUES (), (), (), (), ();
INSERT INTO child_t (parent_id) SELECT id FROM parent_t;
Session 1
BEGIN;
SELECT * FROM parent_t WHERE id = 1 FOR UPDATE;
Session 2
UPDATE child_t SET parent_id = 2 WHERE parent_id = 1;
Session 1
COMMIT;--(ROLLBACK;, abort, doesn't matter)
Session 2
```

storage/innobase/row/row0upd.cc
venkatesh-prasad-v: Hi Marcelo,

Thank you for your time on reviewing this.

> 1. Why this second call to check FK constrain is required only on WSREP

The call to `row_upd_index_is_referenced()` function is for checking if there are any references to the index in the current table. i.e in simple words, it only means that if there are any child tables referring to the current table object.

Whereas the the call to `wsrep_row_upd_index_is_foreign()` function is for checking if the current table has a any foreign key defined.

Both these calls required since they will is used by the `wsrep_append_foreign_key` for generating writesets for the update operation.

> 2. By removing `node->is_delete` we won't be calling the function that is causing the issue, what happens on an update that returns true for row_upd_changes_first_fields_binary and has the parent table records locked?

In such cases, updates also wait for the lock on the parent row to be released. However there was an issue in the error handling code that made the UPDATE queries to fail even after the parent rows were released. This has been fixed in the latest version of the patch.

venkatesh-prasad-v:
> While talking with @venkatesh-prasad-v about this fix we found out that test case is failing for UPDATES:
> 
> ```
> setup
> CREATE TABLE parent_t ( id INT AUTO_INCREMENT PRIMARY KEY );
> CREATE TABLE child_t ( id INT AUTO_INCREMENT PRIMARY KEY, parent_id INT,
> CONSTRAINT `FK_child_parent_id` FOREIGN KEY (`parent_id`) REFERENCES `parent_t` (`id`));
> INSERT INTO parent_t VALUES (), (), (), (), ();
> INSERT INTO child_t (parent_id) SELECT id FROM parent_t;
> Session 1
> BEGIN;
> SELECT * FROM parent_t WHERE id = 1 FOR UPDATE;
> Session 2
> UPDATE child_t SET parent_id = 2 WHERE parent_id = 1;
> Session 1
> COMMIT;--(ROLLBACK;, abort, doesn't matter)
> Session 2
> ERROR 1205 (HY000): Lock wait timeout exceeded; try restarting transaction
> ```

This was happening because, the the temporary DB_LOCK_WAIT error was being translated to DB_LOCK_WAIT_TIMEOUT there by causing the UPDATE command to fail with the timeout error. This has been fixed in the latest revision of the patch.

```
2504                     }
2505                     break;
2506                 case DB_LOCK_WAIT_TIMEOUT:
2507                 case DB_LOCK_WAIT:
2508                     err = DB_LOCK_WAIT_TIMEOUT;
2509                     break;
2510                 default:
```

altmannmarcelo:
The new path looks good to me. UPDATE test is passing now.

altmannmarcelo (approved):

venkatesh-prasad-v:
NOTE
---
While discussing with @altmannmarcelo, we found a case where some update queries are immediate on PS and the same wait in PXC.

The issue is that if the parent table has referenced row(s) locked, then the update query on a child table will wait until the parent row is unlocked and this diverges from the PS behavior where the such updates succeed immediately.
```
CREATE TABLE parent_t (id INT AUTO_INCREMENT PRIMARY KEY )ENGINE = INNODB;
CREATE TABLE child_t (id INT AUTO_INCREMENT PRIMARY KEY, parent_id INT,
                      CONSTRAINT `FK_child_parent_id` FOREIGN KEY (`parent_id`)
                      REFERENCES `parent_t` (`id`)) ENGINE = INNODB;

INSERT INTO parent_t VALUES (), (), (), (), ();
INSERT INTO child_t (parent_id) SELECT id FROM parent_t;

--con1
BEGIN;
SELECT * FROM parent_t WHERE id = 1 FOR UPDATE;

--con2
UPDATE child_t SET parent_id = 1 WHERE id = 1; - Both Succeed
UPDATE child_t SET parent_id = 2 WHERE id = 1; - PXC Waits
UPDATE child_t SET parent_id = 3 WHERE id = 1; - PXC Waits
UPDATE child_t SET parent_id = 4 WHERE id = 1; - PXC Waits
UPDATE child_t SET parent_id = 5 WHERE id = 1; - PXC Waits
```

However, this is not in the context of the current bug and shall be handled as part of https://jira.percona.com/browse/PXC-3374.

PR #1193, opened 2020-07-22
5.7 - Implemented PXC-3092 (Log warning at startup if keyring is specified …

altmannmarcelo:
…but cluster traffic encryption is turned off)

https://jira.percona.com/browse/PXC-3092

Added a new WSREP warning to notify customer that keyring is enabled but
pxc_encrypt_cluster_traffic is disabled.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
+++ b/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
@@ -8,9 +8,17 @@
 --source include/big_test.inc
 --source include/galera_cluster.inc
 --source include/have_innodb.inc
+CALL mtr.add_suppression("You have enabled keyring plugin. SST encryption is mandatory.");
 
 # Force a restart at the end of the test
 --source include/force_restart.inc
+
+--let $assert_text = Keyring plugin requires SST encryption
+--let $assert_select = You have enabled keyring plugin. SST encryption is mandatory.
+--let $assert_count = 1
+--let $assert_file = $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $assert_only_after = CURRENT_TEST
+--source include/assert_grep.inc
 
 SELECT 1;
 
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -5033,6 +5033,23 @@ a file name for --log-bin-index option", opt_binlog_index_name);
   }
   plugins_are_initialized= TRUE;  /* Don't separate from init function */
 
+#ifdef WITH_WSREP
+  static const LEX_CSTRING keyring_vault_name= {
+      C_STRING_WITH_LEN("keyring_vault")};
+  static const LEX_CSTRING keyring_name= {C_STRING_WITH_LEN("keyring_file")};
+  if (!pxc_encrypt_cluster_traffic &&
+      (plugin_is_ready(keyring_vault_name, MYSQL_KEYRING_PLUGIN) ||
+       plugin_is_ready(keyring_name, MYSQL_KEYRING_PLUGIN)))
+  {
+    WSREP_WARN(
+        "You have enabled keyring plugin. SST encryption is mandatory. "
+        "Please enable pxc_encrypt_cluster_traffic. Check "
+        "https://www.percona.com/doc/percona-xtradb-cluster/%u.%u/security/"
+        "encrypt-traffic.html#encrypt-sst for more details.",
+        MYSQL_VERSION_MAJOR, MYSQL_VERSION_MINOR);
+  }
+#endif
+
   Session_tracker session_track_system_variables_check;
   LEX_STRING var_list;
   char *tmp_str;
--- a/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_keyring.result
+++ b/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_keyring.result
@@ -1,3 +1,5 @@
+CALL mtr.add_suppression("You have enabled keyring plugin. SST encryption is mandatory.");
+include/assert_grep.inc [Keyring plugin requires SST encryption]
 SELECT 1;
 1
 1


altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/56/

sql/mysqld.cc
```diff
@@ -5033,6 +5033,21 @@ a file name for --log-bin-index option", opt_binlog_index_name);
   }
   plugins_are_initialized= TRUE;  /* Don't separate from init function */
 
+  LEX_CSTRING keyring_vault_name= {C_STRING_WITH_LEN("keyring_vault")};
```
venkatesh-prasad-v: @dutow created the tool that compares PS code vs PXC code and expects all PXC related code to be within WITH_WSREP block. Everything outside should be same as we have in PS. That makes it easier to spot merging problems.

So, please place all these changes inside the `#ifdef WITH_WSREP .... #endif` block.

sql/mysqld.cc
```diff
@@ -5033,6 +5033,21 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+  LEX_CSTRING keyring_name= {C_STRING_WITH_LEN("keyring_file")};
+  if ((plugin_is_ready(keyring_vault_name, MYSQL_KEYRING_PLUGIN) ||
+       plugin_is_ready(keyring_name, MYSQL_KEYRING_PLUGIN)) &&
+      !pxc_encrypt_cluster_traffic)
```
venkatesh-prasad-v: I would suggest to reorder the checks performed here. `plugin_is_ready()` locks `LOCK_plugin` and therefore is a costly operation. So, here is my suggestion.
```suggestion
  if (!pxc_encrypt_cluster_traffic &&
      (plugin_is_ready(keyring_vault_name, MYSQL_KEYRING_PLUGIN) ||
       plugin_is_ready(keyring_name, MYSQL_KEYRING_PLUGIN))
```

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
```diff
@@ -8,6 +8,7 @@
 --source include/big_test.inc
 --source include/galera_cluster.inc
 --source include/have_innodb.inc
+call mtr.add_suppression("You have keyring plugin enabled. SST encryption is mandatory.");
```
venkatesh-prasad-v: `call` is an SQL statement. So it should be in uppercase.

venkatesh-prasad-v: Adding a suppression doesn't really ensure that message is actually being logged making the test to pass even without the fix. Also that we are adding a new warning message here, I see it is better to have an `assert_grep.inc` in the test to verify that message is actually being logged.

sql/mysqld.cc
```diff
@@ -5033,6 +5033,21 @@ a file name for --log-bin-index option", opt_binlog_index_name);
...
+  {
+    WSREP_WARN(
+        "You have keyring plugin enabled. SST encryption is mandatory. "
```
venkatesh-prasad-v: Not a strong opinion. But was thinking how about we say 
`SST encryption is mandatory when keyring plugin is enabled. So, please enable pxc_encrypt_cluster_traffic or make sure you have SST encryption configured under [SST] section of your my.cnf - Check https://www.percona.com/doc/percona-xtradb-cluster/5.7/security/encrypt-traffic.html#encrypt-sst for more details.
`

venkatesh-prasad-v (changes requested):
Hi Marcelo!

Thank you for working on this fix. I am fine with the solution and have a few suggestions. Please see below, if you don't mind.

sql/mysqld.cc
kennt-percona: This is only checking pxc_encrypt_cluster_traffic and does not check for SST encryption (as indicated by the comment).  So this check would fail if only SST encryption was setup (with pxc_encrypt_cluster_traffic = OFF).

altmannmarcelo: @kennt-percona - mysqld binary only reads mysqld section of my.cnf files. This fix was intended to keep that way. 
Are you suggesting we should change the logic and make it read [sst] section as well? If so this seems rather a big change as we will need to make sure it only reads it on a temporary buffer/variable and do not start the server with those options

sql/mysqld.cc
altmannmarcelo: @venkatesh-prasad-v will put @patrickbirch in the loop to validate the proper wording.
@patrickbirch can you please provide us with the syntactically correct wording for above warning message?

patrickbirch: You have enabled the keyring plugin. SST encryption is mandatory.

sql/mysqld.cc
altmannmarcelo: Also, the user can change my.cnf adding [sst] section after mysqld has started in order to make SST works. So I'm not sure how valid is to implement a check to read [sst] section of my.cnf at startup since it can be changed dynamically.

venkatesh-prasad-v (approved):
Hi Marcelo!!

Thank you for addressing the review comments. Patch looks good and I have no more comments. 

But please wait till the discussion https://github.com/percona/percona-xtradb-cluster/pull/1193#discussion_r459804809 is resolved as I believe Kenn may have something to say.

Other than that, it is approved from my side.

sql/mysqld.cc
kennt-percona: The issue is that we are telling the user how to fix it, but if they do it that way, the warning will still appear.  Sounds like a recipe for a frustrated user (we're saying to the user, there are two ways to fix the issue, but just one way will remove the warning).
It may be better to just remove the portion about enabling it via the [sst] section.

altmannmarcelo: @kennt-percona ack, fixed

kennt-percona: lgtm

kennt-percona (approved):
lgtm

sql/mysqld.cc
```diff
@@ -5033,6 +5033,22 @@ a file name for --log-bin-index option", opt_binlog_index_name);
   }
   plugins_are_initialized= TRUE;  /* Don't separate from init function */
 
+#ifdef WITH_WSREP
+  LEX_CSTRING keyring_vault_name= {C_STRING_WITH_LEN("keyring_vault")};
```
kamil-holubicki: Not a big deal here but what about
`static const LEX_CSTRING keyring_vault_name= {C_STRING_WITH_LEN("keyring_vault")};`
?

patrickbirch (approved):
LGTM

PR #1201, opened 2020-07-24
PXC-3243 : BF aborting a stored procedure hangs

kennt-percona:
Issue
When running a SP with a transaction insde, BF-aborting the SP
can cause the system to hang.

Solution
This happens because the transaction is holding onto a lock,
but the BF-abort is aborting a statement, which may not be the
one holding the lock.
This change will force the BF-abort to propagate and abort and retry
the SP, not the statement.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sp_bf_abort2.test
+++ b/mysql-test/suite/galera/t/galera_sp_bf_abort2.test
@@ -1,0 +1,151 @@
+#
+# PXC-3243: Test BF-aborting during a transaction in an SP
+# This tests the behavior while an SP is BF-aborted during
+# an operation in a transaction within the SP.
+#
+
+--source include/have_debug_sync.inc
+--source include/have_innodb.inc
+--source include/galera_cluster.inc
+
+
+# Test setup
+--echo # Node 1
+--connection node_1
+CREATE DATABASE database1;
+USE database1;
+
+DELIMITER |;
+CREATE PROCEDURE proc_with_transaction()
+BEGIN
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp");
+  COMMIT;
+END|
+
+CREATE PROCEDURE proc_with_transaction_continue_handler()
+BEGIN
+  DECLARE CONTINUE HANDLER FOR SQLEXCEPTION BEGIN END;
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp continue");
+  COMMIT;
+END|
+
+CREATE PROCEDURE proc_with_transaction_exit_handler()
+BEGIN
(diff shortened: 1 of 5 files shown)


sql/sql_parse.cc
```diff
@@ -6206,31 +6206,84 @@ case SQLCOM_PREPARE:
                        first_level ? "MYSQL_AUDIT_QUERY_STATUS_END" :
                                      "MYSQL_AUDIT_QUERY_NESTED_STATUS_END");
 #endif /* !EMBEDDED_LIBRARY */
+#ifdef WITH_WSREP
+    int killed_errno;
```
dutow: These variables are only set once. Please initialize them at the declaration and make them const.

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,141 @@
...
+  DECLARE EXIT HANDLER FOR SQLEXCEPTION BEGIN END;
+  START TRANSACTION;
+    UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+    INSERT INTO t1 VALUES (10, "from sp exit");
+  COMMIT;
+END|
+DELIMITER ;|
+
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
```
dutow, line 49: Add count_sessions / wait_until_count_sessions

dutow (changes requested):

sql/sql_parse.cc
venkatesh-prasad-v: +1

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
venkatesh-prasad-v: +1

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
```diff
@@ -0,0 +1,141 @@
+#
+# PXC-3243: Test BF-aborting during a transaction in an SP
+# This tests the behavior while an SP is BF-aborted during
+# an operation in an transaction within the SP.
```
venkatesh-prasad-v:
```suggestion
# an operation in a transaction within the SP.
```

venkatesh-prasad-v (commented):
Hi Kenn! I guess it is better to address Zsolt's comments. Also, if you are addressing these comments, then please update the same in the 5.6 version as well.

sql/sql_parse.cc
kennt-percona: fixed

mysql-test/suite/galera/t/galera_sp_bf_abort2.test
kennt-percona: fixed

venkatesh-prasad-v:
Hi Kenn! Can you please add the Jenkins link in the description?

kennt-percona:
https://pxc.cd.percona.com/job/pxc57-test-mtr/72/

dutow (approved):

PR #1209, opened 2020-07-31
PXC-3352 [8.0]: Unexpected ERROR 1205 modifying a child table in a FK relationship

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3352

Problem
-------
When deleting/updating from a child table in a FK relationship, if the
parent table's referenced rows are locked, the operation on a child table
failed with lock wait timeout error when the parent table is unlocked.

Analysis
--------
1. For DELETE query

This is the stacktrace of the DELETE query while waiting for the parent
rows to be unlocked.
```
__pthread_cond_wait
os_event::wait
os_event::wait_low
os_event_wait_low
lock_wait_suspend_thread
row_ins_check_foreign_constraint
wsrep_row_upd_check_foreign_constraint 
row_upd_sec_index_entry
row_upd_sec_step
row_upd
row_upd_step
row_update_for_mysql_using_upd_graph
row_update_for_mysql
```
However, the same DELETE query doesn't wait in PS-5.7. 

In PS-5.7, when a row is being modified by an UPDATE/DELETE query, it even
updates the secondary index if there any index defined on the table. While
doing that, it first marks the row as deleted in the secondary index and
inserts into the index with the new value if it is an UPDATE query. This
updation of secondary index is necessary only for UPDATE queries and there
is no need to do the same for DELETE queries.

However, the same was not being done properly in PXC-5.7 because of the
special handling in WSREP for the foreign keys in InnoDB. This made even
the DELETE query to inadvertently update the secondary index
(row_ins_check_foreign_constraint()) thereby trying to lock the parent
table and thus causing lock wait timeout error.

2. For UPDATE query

During Update/Delete, whenever we fail to acquire lock on a row, the
transaction status is to DB_LOCK_WAIT and transaction enters into wait
state waiting for the row to be unlocked; and when the row is unlocked, the
operation is retried.

However, in case of PXC, the temporary DB_LOCK_WAIT during foreign key
checks was being translated into a permanent DB_LOCK_WAIT_TIMEOUT because
of the fallthrough in the error handling code thereby causing the queyr to
fail without retrying the operation.

Fix
---
1. The server is made to behave same way as of the PS-5.7 by removing the
check for DELETE query in `wsrep_row_upd_check_foreign_constraints()`
function which caused the thread to enter the wait.

2. DB_LOCK_WAIT error is now properly handled and the server retries the
operation.

Additionally, this patch 

1. Adds new foreign key tests in galera suite invoking their respective
test from InnoDB with galera enabled.
2. Re-enables the disabled `galera_fk_multitable` and
`galera_toi_ddl_fk_insert` test cases.

Testing
---

v1 - Jenkins: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/74/testReport/

Note to Reviewers:
---
* Many tests in the above link are failing because the galera suite was run with higher value of `--parallel` causing making ports to be unavailable during the test run. To address this issue, [PXC-3364](https://jira.percona.com/browse/PXC-3364) has been reported.

### DIFF
--- a/mysql-test/suite/galera/r/galera-index-online-fk.result
+++ b/mysql-test/suite/galera/r/galera-index-online-fk.result
@@ -1,0 +1,750 @@
+CREATE TABLE parent (a INT PRIMARY KEY, b INT NOT NULL) ENGINE = InnoDB;
+INSERT INTO parent VALUES(1,2),(2,3);
+CREATE INDEX tb ON parent(b);
+INSERT INTO parent VALUES(10,20),(20,30);
+CREATE TABLE child (a1 INT PRIMARY KEY, a2 INT) ENGINE = InnoDB;
+CREATE INDEX tb ON child(a2);
+INSERT INTO child VALUES(10,20);
+ALTER TABLE child ADD FOREIGN KEY(a2) REFERENCES parent(b),
+ALGORITHM = INPLACE;
+ERROR 0A000: ALGORITHM=INPLACE is not supported. Reason: Adding foreign keys needs foreign_key_checks=OFF. Try ALGORITHM=COPY.
+SET foreign_key_checks = 0;
+ALTER TABLE child ADD CONSTRAINT fk_1 FOREIGN KEY (a2)
+REFERENCES parent(b) ON DELETE SET NULL ON UPDATE CASCADE,
+ALGORITHM = INPLACE;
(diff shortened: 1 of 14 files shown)


altmannmarcelo (approved):
LGTM

PR #1215, opened 2020-08-27
PXC-3312: Prevent cleanup of statement diagnostic area in case of tra…

kamil-holubicki:
…nsaction replay.

Fixed issues:
1. When transaction was marked to be replied its diagnostic area was cleaned up. Then, after replaying this cleaned diagnostic are was restored causing diagnostics to be empty.
2. MW-329 test stabilized. It was observed under heavy CPU load that test failed with timeout. It was caused be test flow, which expects at least 5 updates to end up with replay. However under high load it happened that test table grown to large for update transaction to be BF aborted after replication but before certification which ended up with deadlock error without replay. Modified test flow to prevent test table size to grow too much.

### DIFF
--- a/mysql-test/suite/galera/t/MW-329.test
+++ b/mysql-test/suite/galera/t/MW-329.test
@@ -1,14 +1,25 @@
 #
 # #MW-329 Fix incorrect affected rows count after replay
 #
+# The following test starts insert thread on node_1 and parallel update thread on node_2
+# Then it is expected to update to be BF aborted on node_2 after it is replicated, but before it is certified.
+# This will cause update transaction to be replayed.
+# However under high CPU load, when running this test like
+# ./mtr MW-329{,,,,,,,,,,} --parallel=11 --mem --testcase-timeout=10
+# it happened sometimes that t1 on some instances grown so much that update transaction was BF aborted always before
+# being replicated (no replay), so we ended up with infinite update loop and test finished with timeout.
+# To prevent this update thread is paused, t1 data is pruned after reaching threshold and update thread is resumed.
 
 --source include/galera_cluster.inc
 --source include/force_restart.inc
 
 CREATE TABLE t1 (f1 INTEGER, f2 CHAR(20) DEFAULT 'abc') ENGINE=InnoDB;
+CREATE TABLE sig_t (can_update INTEGER, can_delete INTEGER) ENGINE=InnoDB;
 
 # We start with a populated table
 INSERT INTO t1 (f1) VALUES (1),(65535);
+
+INSERT INTO sig_t(can_update, can_delete) values (1, 0);
 
 # Clear the wsrep_local_replays counter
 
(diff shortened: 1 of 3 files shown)


mysql-test/suite/galera/t/MW-329.test
```diff
@@ -1,15 +1,26 @@
 #
 # #MW-329 Fix incorrect affected rows count after replay
 #
+# The following test starts insert thread on node_1 and parallel update thread on node_2
+# Then it is expected to update to be BF abborted on node_2 after it is replicated, but before it is certified.
```
venkatesh-prasad-v: 1. Did you mean that 
```
# Then it is expected that update to be BF aborted on node_2 after it is replicated, but before it is certified.
```

2. `s/abborted/aborted/`

mysql-test/suite/galera/t/MW-329.test
```diff
@@ -47,17 +91,26 @@ select sleep(0.5);
...
-	--error 0,ER_LOCK_DEADLOCK
-	--eval UPDATE t1 SET f2 = '$signature'
-	--enable_query_log
-	--let $row_count = `SELECT ROW_COUNT()`
-	if (`SELECT @@error_count = 0`) {
-		if (`SELECT $row_count = 0`) {
-			--die ROW_COUNT() = 0
-		}
-	}
+        # If we updated block was sig_ted, stop and allow deleteion from t1.
```
venkatesh-prasad-v: `s/deleteion/deletion/`

mysql-test/suite/galera/t/MW-329.test
```diff
@@ -1,15 +1,26 @@
 #
 # #MW-329 Fix incorrect affected rows count after replay
 #
+# The following test starts insert thread on node_1 and parallel update thread on node_2
+# Then it is expected to update to be BF abborted on node_2 after it is replicated, but before it is certified.
+# This will cause update transaction to be replayed.
+# However under high CPU load, when running this test like 
+# ./mtr MW-329{,,,,,,,,,,} --parallel=11 --mem --testcase-timeout=10
```
venkatesh-prasad-v: Applying the patch shows these warnings. Please remove the trailing whitespaces in the end of these two lines.

```
1215.diff:67: trailing whitespace.
# However under high CPU load, when running this test like 
1215.diff:68: trailing whitespace.
# ./mtr MW-329{,,,,,,,,,,} --parallel=11 --mem --testcase-timeout=10 
1215.diff:147: trailing whitespace.
        # If we updated block was sig_ted, stop and allow deleteion from t1. 
warning: 3 lines add whitespace errors.
```

mysql-test/suite/galera/t/MW-329.test
venkatesh-prasad-v: Remove the trailing whitespace in the end.

sql/binlog.cc
```diff
@@ -8452,7 +8452,7 @@ TC_LOG::enum_result MYSQL_BIN_LOG::commit(THD *thd, bool all) {
       gtid_state->update_on_rollback(thd);
       thd_get_cache_mngr(thd)->reset();
       // Reset the thread OK status before changing the outcome.
-      if (thd->get_stmt_da()->is_ok())
+      if (!wsrep_must_replay(thd) && thd->get_stmt_da()->is_ok())
```
venkatesh-prasad-v, line 8455: 1. At this point, if thd already has any error set, then calling `my_error()` will make the DA will make the assert `!set` in `DA::set_error_status()`. Is it expected that there shall be no error when we reach here?

2. This will cause the thread to not reset the DA if the transaction must be replayed. So, if we are replaying the transaction, then from where does the DA get cleared before the replay?

venkatesh-prasad-v (commented):
Hi Kamil! 

Thank you for working on this issue. It is indeed a tricky one to fix. Nice job!

If you don't mind, could you please limit the commit message to 80 characters wide, just like other commit messages, so that is more readable?

Apart from that, I also have a few questions and a few suggestions. Please see below.

Thank you for your time.

mysql-test/suite/galera/t/MW-329.test
kamil-holubicki: done

mysql-test/suite/galera/t/MW-329.test
kamil-holubicki: done

mysql-test/suite/galera/t/MW-329.test
kamil-holubicki: done

mysql-test/suite/galera/t/MW-329.test
kamil-holubicki: done

sql/binlog.cc
kamil-holubicki: 1. The callstack here is
```
MYSQL_BIN_LOG::commit(THD*, bool)
ha_commit_trans(THD*, bool, bool)
trans_commit_stmt(THD*, bool)
mysql_execute_command(THD*, bool)
mysql_parse(THD*, Parser_state*, bool)
```

DA allows overwriting the result (I've put comment in details in the code)
 
2. Please take a looka at Wsrep_replayer_service constructor and destructor and m_da_shadow in wsrep_high_priority_service.cc

3. Commit message first line is Jira issue summary

mysql-test/suite/galera/t/MW-329.test
```diff
@@ -22,10 +33,43 @@ SELECT VARIABLE_VALUE = 0 FROM performance_schema.global_status WHERE VARIABLE_N
...
+                        # sig_t update thread to resume
+                        UPDATE sig_t SET can_update = 1;
+
+                        # Wait for update thread to resume. When resumed, it will signal that deletes are not allowed anymore.
+                        SELECT can_delete FROM sig_t INTO delete_allowed;
+                        WHILE delete_allowed DO
+                                SET @foo = (SELECT SLEEP(0.1));
+                                SELECT can_delete FROM sig_t INTO delete_allowed;
+                        END WHILE;
+		END IF;
```
venkatesh-prasad-v, line 72: If it isn't too much to ask, then please replace tabspaces with whitespaces for better readability.

sql/binlog.cc
venkatesh-prasad-v: Regarding the comments on the commit message, I actually meant to wrap the message body width to 80 characters. something like below.

```
PXC-3312: Prevent cleanup of statement diagnostic area in case of transaction replay.

Fixed issues:
1. When transaction was marked to be replied its diagnostic area was cleaned up. 
Then, after replaying this cleaned diagnostic are was restored causing
diagnostics to be empty. 

2. MW-329 test stabilized. It was observed under heavy CPU load that test failed
with timeout. It was caused be test flow, which expects at least 5 updates to
end up with replay. However under high load it happened that test table grown to
large for update transaction to be BF aborted after replication but before
certification which ended up with deadlock error without replay. Modified test
flow to prevent test table size to grow too much.
```

venkatesh-prasad-v (approved):
Hi Kamil,

Thank you for addressing the comments. 
The patch looks good now and have a few cosmetic suggestions below.

Apart from that, it is approved from my side.

Thank you.

sql/binlog.cc
kamil-holubicki: done

PR #1216, opened 2020-09-01
PXC-3249 Upgrade Documentation Changes (8.0)

patrickbirch:
modified:   source/howtos/upgrade_guide.rst


kathy-williamson (changes requested):
I don't beleive this is correct and would like to have examples of how it needs to be done

PR #1221, opened 2020-09-09
Merge 5.7.31, Part 2

dutow:


kamil-holubicki (commented):
Why we have doc files here as conflict resolution part? I think it should be left unchanged when doing merge with conflicts.

kamil-holubicki (approved):

PR #1224, opened 2020-09-15
Pxc rel 8.0.20

kennt-percona:


build-ps/build-binary.sh
```diff
@@ -479,63 +468,6 @@ fi
...
-        -DMYSQL_DATADIR="/usr/local/$PRODUCT_FULL/data" \
-        -DROUTER_INSTALL_LIBDIR="/usr/local/$PRODUCT_FULL/lib/mysqlrouter/private" \
-        -DROUTER_INSTALL_PLUGINDIR="/usr/local/$PRODUCT_FULL/lib/mysqlrouter/plugin" \
-        -DCOMPILATION_COMMENT="$COMMENT" \
-        -DWITH_PAM=ON \
-        -DWITH_ROCKSDB=ON \
-        -DWITH_INNODB_MEMCACHED=ON \
-        -DWITH_ZLIB=system \
-        -DWITH_NUMA=ON \
-        -DWITH_LDAP=system \
```
dutow: This diff changed the `WITH_LDAP` argument, which is missing from PXC. We should add it, as we already received requests for using LDAP auth with pxc, so at least we should test that it works in development builds.

client/mysql.cc
```diff
@@ -3165,14 +3165,8 @@ static int com_help(String *buffer MY_ATTRIBUTE((unused)),
   put_info(
       "\nFor information about Percona products and services, visit:\n"
       "   http://www.percona.com/\n"
-<<<<<<< HEAD
       "Percona XtraDB Cluster manual: "
       "http://www.percona.com/doc/percona-xtradb-cluster/8.0/\n"
-||||||| merged common ancestors
-      "Percona Server manual: http://www.percona.com/doc/percona-server/%d.%d\n"
```
dutow: We should change the version string to the same `%d.%d` pattern.

mysys/my_md5.cc
```diff
@@ -78,32 +78,18 @@ then a md5-hash (16 bytes) string is generated using the complete record.
 Following functions act as helper function in generation of this md5-hash. */
 
 void *wsrep_md5_init() {
-#if defined(HAVE_YASSL)
```
dutow, line 81: This is not a conflict marker resolution. Please mention non trivial changes such as this in the commit message.

share/messages_to_clients.txt
```diff
@@ -9382,16 +9382,12 @@ ER_EXCLUDE_ENCRYPTION_THREADS_RUNNING
 ER_EXPLICIT_DECRYPTION_OF_ONLINE_ENCRYPTED_TABLESPACE
   eng "Tablespace %s is encrypted by encryption threads and cannot be explicitly decrypted. Please use default-table-encryption=ONLINE_FROM_KEYRING_TO_UNENECRYPTED to decypt the tablespace"
 
-<<<<<<< HEAD
 ER_PXC_GR_INIT
   eng "%s"
 
-||||||| merged common ancestors
-=======
 ER_SEQUENCE_TABLE_SIZE_LIMIT
```
dutow: I think we should change the order here. PXC specific error messages should come after PS error messages, to ensure that for PS error messages, the error codes are the same in PS and PXC.

sql/wsrep_binlog.cc
```diff
@@ -156,13 +156,17 @@ static int wsrep_write_cache_inc(THD *const thd,
     should be appended to write-set. */
     Gtid_log_event gtid_event(thd, true, 0, 0, false, 0, 0,
                               UNKNOWN_SERVER_VERSION, UNKNOWN_SERVER_VERSION);
-    uchar gtid_buf[Gtid_log_event::MAX_EVENT_LENGTH];
-    uint32 gtid_len = gtid_event.write_to_memory(gtid_buf);
-
-    if (thd->wsrep_cs().append_data(wsrep::const_buffer(gtid_buf, gtid_len)))
+    int error = 0;
```
dutow: Looks like another non trivial change.

sql/wsrep_mysqld.cc
```diff
@@ -2283,15 +2283,17 @@ void wsrep_to_isolation_end(THD *thd) {
       (long long)wsrep_thd_trx_seqno(req), wsrep_thd_client_mode_str(req),   \
       wsrep_thd_client_state_str(req), wsrep_thd_transaction_state_str(req), \
       req->get_command(), req->lex->sql_command,                             \
-      (req->rewritten_query.length() ? req->rewritten_query.c_ptr_safe()     \
-                                     : req->query().str),                    \
+      (req->rewritten_query().length()                                       \
+                ? const_cast<String&>(req->rewritten_query()).c_ptr_safe()   \
+                : req->query().str),                    \
```
dutow: Formatting. Have you ran clang-format on this file? That `\` should go to the same column as the others.

sql/wsrep_mysqld.h
```diff
@@ -310,9 +310,9 @@ extern wsrep_seqno_t wsrep_locked_seqno;
     if (victim_thd) WSREP_LOG_CONFLICT_THD(victim_thd, "Victim thread");      \
   }
 
-#define WSREP_QUERY(thd)                                   \
-  ((!opt_general_log_raw) && thd->rewritten_query.length() \
-       ? thd->rewritten_query.c_ptr_safe()                 \
+#define WSREP_QUERY(thd)                                           \
+  ((!opt_general_log_raw) && thd->rewritten_query().length()         \
```
dutow: Formatting.

storage/innobase/lock/lock0lock.cc
```diff
@@ -1186,46 +1177,12 @@ static void lock_rec_insert_to_waiting(hash_table_t *lock_hash, lock_t *lock,
   ut_ad(lock->is_waiting());
   ut_ad(rec_id.matches(lock));
   ut_ad(lock_mutex_own());
-<<<<<<< HEAD
-
-#ifdef WITH_WSREP
-  /* Disable CATS for galera for now */
```
dutow: This should be also mentioned in the commit message.

dutow (changes requested):

dutow:
`PS_REFRESH_VERSION` should also be updated.

sql/mysqld.cc
```diff
@@ -9495,7 +9495,7 @@ extern "C" void *start_wsrep_THD(void *arg) {
   thd_manager->add_thd(thd);
   thd_added = true;
 
-#ifdef SKIP_INNODB_HP
+#ifndef SKIP_INNODB_HP
```
dutow: The commit message only mentions test fixes. Please also mention the code changes in the commit message.

VERSION
```diff
@@ -1,13 +1,5 @@
...
-MYSQL_VERSION_EXTRA=-10
-MYSQL_CLUSTER_VERSION_EXTRA=-rc
-||||||| merged common ancestors
-MYSQL_VERSION_PATCH=19
-MYSQL_VERSION_EXTRA=-10
-=======
 MYSQL_VERSION_PATCH=20
 MYSQL_VERSION_EXTRA=-11
->>>>>>> Percona-Server-8.0.20-11
+MYSQL_CLUSTER_VERSION_EXTRA=
```
kamil-holubicki, line 5: Conflict resolves to MYSQL_CLUSTER_VERSION_EXTRA=-rc

build-ps/build-binary.sh
kamil-holubicki: @dutow Do we have Jira ticket for this? I think we should handle LDAP thing separately from merge (at least in case when switching it ON causes problems now)

client/mysql.cc
kamil-holubicki: @dutow Looks like upstream PS has changed it form %d.%d to the hard-coded 8.0

sql/log_event.cc
```diff
@@ -4962,8 +4962,8 @@ int Query_log_event::do_apply_event(Relay_log_info const *rli,
 #ifdef WITH_WSREP
         rli->report(ERROR_LEVEL, ER_ERROR_ON_MASTER,
                     ER_THD(thd, ER_ERROR_ON_MASTER), expected_error,
-                    (!opt_general_log_raw) && thd->rewritten_query.length()
-                        ? thd->rewritten_query.c_ptr_safe()
+                    (!opt_general_log_raw) && thd->rewritten_query().length()
+                        ? const_cast<String&>(thd->rewritten_query()).c_ptr_safe()
```
kamil-holubicki: Please see pxc 5.7 wsrep_mysqld.cc:: wsrep_thd_rewritten_query()
We should use this function returning non-const copy of rewritten query instead of const_casting here (and in several other places)

sql/wsrep_mysqld.cc
kamil-holubicki: wsrep_thd_rewritten_query()...

client/mysql.cc
dutow: You are right, this is good then.

VERSION
dutow: I don't think this should be an rc version. It was probably a merge mistake even for the previous release.

build-ps/build-binary.sh
dutow: Just by building it, it won't become part of the release package automatically, so enabling it shouldn't cause any issues. And then we can create a jira ticket for testing it, but as this is just an authentication plugin, it shouldn't cause any issues.

VERSION
venkatesh-prasad-v: This is required only when we are building mysql-cluster and is not required for building PS or MySQL. Even the upstream doesn't have this line. See https://github.com/mysql/mysql-server/blob/mysql-8.0.20/VERSION
If it is not required, shall we remove it?

sql/log_event.cc
kennt-percona: hmm.  that change didn't make it to 8.0 (no wsrep_thd_rewritten_query in 8.0)

build-ps/build-binary.sh
kennt-percona: Adding it as "-DWITH_LDAP=system", as far as I can tell from cmake, it's either system or a path to a custom ldap build.

mysys/my_md5.cc
kennt-percona: fixed

share/messages_to_clients.txt
kennt-percona: fixed

sql/wsrep_binlog.cc
kennt-percona: fixed

storage/innobase/lock/lock0lock.cc
kennt-percona: fixed

sql/mysqld.cc
kennt-percona: fixed

VERSION
kennt-percona: removed

mysql-test/suite/galera/t/galera_events3.test
```diff
@@ -15,6 +15,7 @@ call mtr.add_suppression("Operation DROP USER failed");
 call mtr.add_suppression("Event [0-9]* Query apply failed: [0-9]*, seqno [0-9]*");
 CREATE TABLE t1(a INT AUTO_INCREMENT, b INT, PRIMARY KEY pka(a));
 --send SET DEBUG_SYNC = 'now WAIT_FOR has_global_grant.got_lock'; 
+--sleep 1
```
dutow, line 18: Is this sleep needed, will it be stable on jenkins?

kennt-percona: The sleep is there to make the test more reliable.  It failed (for me) occasionally under load.

dutow (approved):

kennt-percona:
https://pxc.cd.percona.com/job/pxc80-test-mtr/226/

PR #1242, opened 2020-09-22
PXC-3419 pxc-strict-mode requires tx_isolation=SERIALIZABLE that was …

patrickbirch:
…deprecated for transaction_isolation (5.7)

modified:   source/wsrep-system-index.rst

### DIFF
--- a/doc/source/wsrep-system-index.rst
+++ b/doc/source/wsrep-system-index.rst
@@ -115,7 +115,8 @@ or the node is bootstrapping, then :variable:`pxc_strict_mode` defaults to
    * ``wsrep_replicate_myisam=OFF``
    * ``binlog_format=ROW``
    * ``log_output=FILE`` or ``log_output=NONE`` or ``log_output=FILE,NONE``
-   * ``tx_isolation=SERIALIZABLE``
+  
+   The ``SERIALIZABLE`` method of isolation is not allowed in ``ENFORCING`` mode.
 
 For more information, see :ref:`pxc-strict-mode`.
 


kathy-williamson (approved):

doc/source/wsrep-system-index.rst
```diff
@@ -115,7 +115,7 @@ or the node is bootstrapping, then :variable:`pxc_strict_mode` defaults to
    * ``wsrep_replicate_myisam=OFF``
    * ``binlog_format=ROW``
    * ``log_output=FILE`` or ``log_output=NONE`` or ``log_output=FILE,NONE``
-   * ``tx_isolation=SERIALIZABLE``
+   * ``transaction_isolation=SERIALIZABLE``
```
kamil-holubicki: I would rather say that this should be like
`transaction_isolation != SERIALIZABLE`

It is the same for 5.7

In other words: PXC does not recommend using SERIALIZABLE transaction isolation, so it is not allowed in ENFORCING mode

```
mysql> show variables like 'pxc_strict_mode';
+-----------------+------------+
| Variable_name   | Value      |
+-----------------+------------+
| pxc_strict_mode | PERMISSIVE |
+-----------------+------------+
1 row in set (0.01 sec)

mysql> show global variables like 'transaction_isolation';
+-----------------------+--------------+
| Variable_name         | Value        |
+-----------------------+--------------+
| transaction_isolation | SERIALIZABLE |
+-----------------------+--------------+
1 row in set (0.00 sec)

mysql> set global pxc_strict_mode=ENFORCING;
ERROR 1105 (HY000): Can't change pxc_strict_mode to ENFORCING as isolation level is SERIALIZABLE

```

kamil-holubicki (changes requested):

doc/source/wsrep-system-index.rst
patrickbirch: Updated.

kamil-holubicki (approved):
lgtm

PR #1243, opened 2020-09-22
PXC-3424: Fix error handling when donor is not able to serve SST

kamil-holubicki:
1. Improved error handling in wsrep_sst_donate_cb
2. Fixed sst script termination on joiner side. sst_process variable was
   cleared just after sst script startup. That prevented signaling sst
   script from wsrep_sst_cancel()
3. Removed 'timeout' use sfrom wsrep_sst_xtrabackup-v2.sh. Having
   'timeout' prevented graceful and instant script termination.
   As 'timeout' creates its own process group it was not killed when
   SIGTERM signal was sent to wsrep_sst_xtrabackup-v2 script, causing
   socat to keep opened ports till the end of timeout and mysqld waiting
   in 'defunct' state for its child processes to terminate. If during
   this time mysqld was started again, SST script failed as sockets were
   still used by old instance of socat.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_failure.test
+++ b/mysql-test/suite/galera/t/galera_sst_failure.test
@@ -1,0 +1,45 @@
+# Test if donor node goes back to SYNC state after it desyncs for donating
+# another node, but failing to start SST.
+
+--source include/have_debug.inc
+--source include/galera_cluster.inc
+
+--connection node_1
+CALL mtr.add_suppression(".*WSREP: SST failed.*");
+CALL mtr.add_suppression(".*WSREP: .* State transfer to .* failed: -125 .Operation canceled.*");
+
+--let $assert_text=node_1 should be in SYNC state after serving SST to node_2
+--let $wsrep_local_state_comment = query_get_value(SHOW STATUS LIKE 'wsrep_local_state_comment', Value, 1)
+--let $assert_cond= "$wsrep_local_state_comment" = "Synced"
+--source include/assert.inc
+
+# Setup node_1 to fail SST on next SST request
+SET GLOBAL debug = "+d,wsrep_sst_donate_cb_fails";
+
+# Restart node 2 forcing SST
+--connection node_2
+--source include/shutdown_mysqld.inc
+
+# Remove the grastate.dat file to force an SST
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+# As SST fails on donor side, the joiner will abort
+--let $error_log= $MYSQLTEST_VARDIR/tmp/test_error_log.err
+--let $mysqld2=$MYSQLD --defaults-group-suffix=.2 --defaults-file=$PATH_CONFIG_FILE --wsrep-provider-options='base_port=$NODE_GALERAPORT_2' --console > $error_log 2>&1
+--error 134
+--exec $mysqld2
+--remove_file $error_log
+
+# node_1 should go back to SYNC state
+--connection node_1
+--let $assert_text=node_1 should go back to SYNC state
+--let $wsrep_local_state_comment = query_get_value(SHOW STATUS LIKE 'wsrep_local_state_comment', Value, 1)
+--let $assert_cond= "$wsrep_local_state_comment" = "Synced"
+--source include/assert.inc
+
+# cleanup
+SET GLOBAL debug = "-d,wsrep_sst_donate_cb_fails";
+
+--connection node_2
+--echo cleanup
+--source include/start_mysqld.inc
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -107,6 +107,9 @@ MAGIC_FILE="${DATA}/${INFO_FILE}"
 # from the donor to the joiner
 SST_INFO_FILE="sst_info"
 
+# Used to pass status of pipelined processes executed in background
+PIPESTATUS_FILE="pipestatus"
+
 # Setting the path for ss and ip
 export PATH="/usr/sbin:/sbin:$PATH"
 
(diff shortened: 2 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.6-param/51/

sql/wsrep_sst.cc
```diff
@@ -1293,6 +1301,12 @@ wsrep_cb_status_t wsrep_sst_donate_cb (void* app_ctx, void* recv_ctx,
   char uuid_str[37];
   wsrep_uuid_print (&current_gtid->uuid, uuid_str, sizeof(uuid_str));
 
+  DBUG_EXECUTE_IF("wsrep_sst_donate_cb_fails", 
+  {
+    fprintf(stderr, "KH: returning WSREP_CB_FAILURE\n");
```
dutow: leftover debug print?

mysql-test/suite/galera/t/galera_sst_failure.test
```diff
@@ -0,0 +1,38 @@
+# Test if donor node goes back to SYNC state after it desyncs for donating
+# another node, but failing to start SST.
+
+--source include/galera_cluster.inc
+--source include/have_debug.inc
```
venkatesh-prasad-v, line 4: `include/have_debug.inc` is a restrictive inc. So please source it in the beginning of the test case.

mysql-test/suite/galera/t/galera_sst_failure.test
```diff
@@ -0,0 +1,38 @@
...
+
+# As SST fails on donor side, the joiner will abort
+--let $error_log= $MYSQLTEST_VARDIR/tmp/test_error_log.err
+--let $mysqld2=$MYSQLD --defaults-group-suffix=.2 --defaults-file=$PATH_CONFIG_FILE --wsrep-provider-options='base_port=$NODE_GALERAPORT_2' --console > $error_log 2>&1
+--error 134
+--exec $mysqld2
+
+# node_1 should go back to SYNC state
+--connection node_1
+SHOW STATUS LIKE 'wsrep_local_state_comment';
```
venkatesh-prasad-v: Please use assert.inc to validate the state.

mysql-test/suite/galera/t/galera_sst_failure.test
```diff
@@ -0,0 +1,38 @@
...
+
+# Restart node 2 forcing SST
+--connection node_2
+--source include/shutdown_mysqld.inc
+
+# Remove the grastate.dat file to force an SST
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+# As SST fails on donor side, the joiner will abort
+--let $error_log= $MYSQLTEST_VARDIR/tmp/test_error_log.err
```
venkatesh-prasad-v, line 27: Delete the file in the end of the test. MTR doesn't report this in 5.6, but it fails in check-testcase on 8.0.

venkatesh-prasad-v (commented):
Nice work, Kamil! 

Please see below for a few minor suggestions.

sql/wsrep_sst.cc
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_sst_failure.test
kamil-holubicki: done

mysql-test/suite/galera/t/galera_sst_failure.test
kamil-holubicki: done

mysql-test/suite/galera/t/galera_sst_failure.test
kamil-holubicki: No need for this, as MTR framework will compare produced output with result file and validate it.

venkatesh-prasad-v: Right! MTR does compare the output with the result file. But this phase happens only in the end of the test run and doesn't happen immediately. But, IMHO we always should not depend on the output of the SELECT/SHOW commands as it is not fully error proof.

Result file comparion causes the MTR to wait till the test to complete and then performs diff in the end. If the test is bigger with many small scenarios, MTR can take more time to report the test failure even when there is a failure in the first scenario. Whereas with asserts, it is faster as the test fails immediately when the condition fails. 

Also, using asserts makes the test more robust to test failures. For example, let's say this SST logic gets broken in the future for some reasons( eg: a big upstream merge). Then, it is very likely that the test failure may go unnoticed if someone re-records the failing testcases and merges them. Given the advantages we have with asserts, I think it is worth to use them.

kamil-holubicki: MTR test adjusted to use proposed validation pattern.

venkatesh-prasad-v (approved):
Thank you for addressing all the comments. 
The patch looks good and is approved from my side.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -133,6 +136,90 @@ timeit(){
...
+        fi 
+
+        sleep 1
+        tmt=$((tmt-1))
+    done
+    
+    if [[ $result -eq 0 ]];then
+      piperesult=(`cat ${dir}/$PIPESTATUS_FILE`)
+    fi
+
```
kennt-percona, line 218: Should we remove the PIPESTATUS_FILE at this point?

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -133,6 +136,90 @@ timeit(){
...
+    piperesult=( $result "${piperesult[@]}" )
```
kennt-percona, line 219: Doesn't piperesult need to be defined before this ('local piperesult')?

scripts/wsrep_sst_xtrabackup-v2.sh
kamil-holubicki: It is created in temp directory created by the script ( initialize_tmpdir()) and will be removed on script exit. 

Edit: OK, I see another point of possible confusion:
man sh:
`The shell uses dynamic scoping, so that if you make the variable x local to function f, which then calls function g, references to the variable x made inside g will refer to the variable x declared inside f, not to the global variable named x.`

'dir' here is the local dir variable of parent function.

I have changed it to 'tempdirbase' explicitly to avoid confusion where we create temp file.

scripts/wsrep_sst_xtrabackup-v2.sh
kamil-holubicki: Good point. The way it is now, piperesult becomes global variable, which is not necessary. 
I declared it local. 
I adjusted timeout_delegate() local variables in the same way.

kennt-percona (approved):
lgtm

PR #1247, opened 2020-09-24
PXC-3424: Fix error handling when donor is not able to serve SST (5.7)

kamil-holubicki:


kamil-holubicki:
This PR shows merge conflict with galera repo pointer, however everything merges fine. This is probably because of galera repo pointer being from my fork. 
Having proper pointer in my branch is needed for Jenkins job to run properly. It has to be updated anyway after approval and merge of galera part.

dutow:
This is a GCA merge of the 5.6 PR? Github showd 2 commits, but no merges.

kamil-holubicki:
@dutow Yes, it is. I changed the message of merge commit to the message of actual commit. I always did like this. Should I leave merge message like "merge branch into branch..." It will show something like "Merge branch kamil-holubicki/PXC-ABC-56 into kamil-holubicki/PXC-ABC-57

dutow (approved):

PR #1254, opened 2020-09-29
PS 3439 Update 5.7 to 8.0 Upgrade instructions (8.0)

patrickbirch:
modified:   source/howtos/upgrade_guide.rst

### DIFF
--- a/doc/source/howtos/upgrade_guide.rst
+++ b/doc/source/howtos/upgrade_guide.rst
@@ -4,11 +4,17 @@
 Upgrading Percona XtraDB Cluster
 ================================
 
-.. This guide describes the procedure for upgrading |PXC| without downtime
-   (*rolling upgrade*) to the |PXC| 8.0.
+This guide describes the procedure for upgrading |PXC| without downtime
+(*rolling upgrade*) to the |PXC| 8.0. A "rolling upgrade" means
+there is no need to take down the complete cluster during the upgrade.
 
-.. A "rolling upgrade" means there is no need to take down the complete
-   cluster during the upgrade.
+.. important::
+
+    There is no downgrade path.
+    
+    If you must revert a upgrade between 8.0 versions, take an 8.0 backup,
+    using mysqldump from the upgraded cluster and restore the backup to the earlier 8.0
+    version.
 
 The following documents contain details about relevant changes in the 8.0 series
 of MySQL and |percona-server|. Make sure you deal with any incompatible features
@@ -33,7 +39,7 @@ Important changes in |pxc| |version|
 .. contents:: 
    :local:
 
-.. _upgrade-guide-changed-traffic-encryption:
+ _upgrade-guide-changed-traffic-encryption:
 
 Traffic encryption is enabled by default
 --------------------------------------------------------------------------------
@@ -41,10 +47,9 @@ Traffic encryption is enabled by default
 The :variable:`pxc_encrypt_cluster_traffic` variable, which enables traffic
 encryption, is set to ``ON`` by default in |pxc| |version|. 
 
-Unless you configure a node accordingly (each node in your cluster must use the
-same SSL certificates) or try to join a cluster running |pxc| 5.7 which
-unencrypted cluster traffic, the node will not be able to join resulting in an
-error.
+All nodes have the same SSL certificates. The 5.7 traffic is, by default, not
+encrypted. If you attempt to connect an "out of a box" 8.0 node, this results
+in an error:
 
 .. code-block:: text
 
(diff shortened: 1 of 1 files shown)


doc/source/howtos/upgrade_guide.rst
```diff
@@ -53,53 +50,49 @@ error.
...
 
-After you check the log for any experimental or unsupported features and fix any
-encountered incompatibilities, you can set the variable back to ``ENFORCING`` at
+After a time, check the log for any experimental or unsupported features and fix any
+encountered incompatibilities. You can then set the variable to ``ENFORCING`` at
 run time:
 
 .. code-block:: mysql
 
    mysql> SET pxc_strict_mode=ENFORCING;
```
rameshvs02: Replace `SET pxc_strict_mode=ENFORCING;` with `SET GLOBAL pxc_strict_mode=ENFORCING;`

doc/source/howtos/upgrade_guide.rst
```diff
@@ -126,54 +118,84 @@ authentication plugin in these cases.
 
 Be sure you are running on the latest 5.7 version before you upgrade to 8.0.
 
-|mysql-upgrade| is part of :term:`SST`
---------------------------------------------------------------------------------
+.. upgrading.auto::
 
-|mysql-upgrade| is now run automatically as part of :term:`SST`. You do not have
-to run it manually when upgrading your system from an older version.
+Auto-upgrade
```
rameshvs02: Remove this section.
Removed wsrep_state.dat and supporting code from SST process. https://jira.percona.com/browse/PXC-2997

doc/source/howtos/upgrade_guide.rst
```diff
@@ -126,54 +118,84 @@ authentication plugin in these cases.
...
 
-.. #. Back up :file:`grastate.dat`, so that you can restore it
-   if it is corrupted or zeroed out due to network issue.
+#. Back up :file:`grastate.dat`. If the file is corrupted or zeroed out due to
+   network issues in the upgrade process you may need to restore it.
 
-.. #. Now, start the cluster node with 8.0 packages installed, |pxc| will upgrade
-   the data directory as needed - either as part of the startup process or a
-   state transfer (IST/SST).
+#. Restart the ``mysqld`` service outside the cluster in standalone mode by
```
rameshvs02, line 167: Remove this step.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -126,54 +118,84 @@ authentication plugin in these cases.
...
+   .. note::
 
+      To prevent any users from accessing this node while performing
+      work on it, you may add `--skip-networking
+      <https://dev.mysql.com/doc/refman/8.0/en/server-options.html#option_mysqld_skip-networking>`_
+      to the startup options and use a local socket to connect, or
+      alternatively you may divert any incoming traffic from
+      your application to other operational nodes.
+      
+#. When the upgrade is complete, stop the ``mysqld`` process. You can either run
```
rameshvs02, line 183: Remove this steps

doc/source/howtos/upgrade_guide.rst
```diff
@@ -53,53 +50,49 @@ error.
 
 .. seealso:: sections :ref:`encrypt-traffic`, :ref:`configure`
```
rameshvs02, line 60: I think we should also add `wsrep_sst_auth` removal changes here. PXC-8.0 will create an SST user internally for data transfer, so the user should remove `wsrep_sst_auth` from my.cnf before starting the node with 8.0 packages.
Detailed info is available in this blog post: https://www.percona.com/blog/2019/10/03/percona-xtradb-cluster-8-0-new-feature-wsrep_sst_auth-removal/

rameshvs02 (changes requested):

doc/source/howtos/upgrade_guide.rst
```diff
@@ -53,53 +50,49 @@ error.
...
+nodes, then remove and re-create the cluster from scratch. Alternatively,
+you can perform a `rolling upgrade from PXC 5.6 to 5.7 <https://www.percona.com/doc/percona-xtradb-cluster/5.7/howtos/upgrade_guide.html>`_.
+and then follow the current procedure to upgrade from 5.7 to 8.0.
 
 Not recommended to mix |pxc| 5.7 nodes with |pxc| 8.0 nodes
 --------------------------------------------------------------------------------
 
-Shut down the cluster and upgrade each node to |pxc| 8.0. It is
-important that you make backups before attempting an upgrade.
+Shut down the cluster and upgrade all nodes to |pxc| 8.0. It is important to
```
kathy-williamson: I think the statement "if downtime is acceptable.." should be added back.

doc/source/howtos/upgrade_guide.rst
```diff
@@ -53,53 +50,49 @@ error.
...
 
-.. If downtime is acceptable
-   Shut down the cluster and upgrade all nodes to |pxc|
-   8.0. It is important that you make backups before attempting an upgrade.
-
-.. If downtime is not possible
-   The rolling upgrade is supported but ensure the
-   traffic is controlled during the upgrade and writes are directed only to 5.7
-   nodes until all nodes are upgraded to 8.0.
+The rolling upgrade is supported,but you must ensure the traffic is controlled
```
kathy-williamson: supported,but - please correct the spacing

doc/source/howtos/upgrade_guide.rst
```diff
@@ -126,54 +118,84 @@ authentication plugin in these cases.
...
-   5.7 to the new major version 8.0.
+.. code-block:: text
 
-   To upgrade the cluster, follow these steps for each node:
+   # *wsrep_state.dat* WSREP state file (created on initialization)
+   version=1.0
+   wsrep_schema_version=8.0.16
 
-   #. Make sure that all nodes are synchronized.
+The |wsrep_state_dat| file represents that the data directory is compatible
```
kathy-williamson: Why is that version specifically mentioned -we are at a higher level?

doc/source/howtos/upgrade_guide.rst
```diff
@@ -187,14 +209,23 @@ to run it manually when upgrading your system from an older version.
...
-      
-.. #. Repeat this procedure for the next node in the cluster until you upgrade all nodes.
+       
+#. Repeat the steps for the other nodes.
+
+   .. note::
+   
+      The last node upgrade may take time to join since the node invites IST for
+      configuration changes. Starting with ``Galera`` 4, configuration changes
+      are cached to gcache and dontated as part of IST/SST to help the
```
kathy-williamson: dontated?  should it be donated?

doc/source/howtos/upgrade_guide.rst
```diff
@@ -218,20 +249,17 @@ Scenario: No active parallel workload or with read-only workload
 --------------------------------------------------------------------------------
 
 If there is no active parallel workload or the cluster has read-only workload
-while upgrading the nodes, complete the following procedure for each node in the
```
kathy-williamson, line 221: Does this mean there are no writes being done in the cluster?  If so this needs to be revisited - if no writes - no split brain.   Please explain when this would occur.

kathy-williamson (changes requested):
As soon as the last 5.7 node shuts down, the configuration of the remaining
two nodes is updated to use protocol version 4 - Is this really true or is it when the last node joins as a 8.0 node.  Node could come back up as 5.7 

penultimate - please change this to something else - it is not a word many will know 

It is not recommended to restart the last node without upgrading it. - This needs to be bolder and reworded to be clear that this is a bad option

The upgrade process - should be process

Joining a 5.7 node to a |pxc| |version| cluster is not supported. - Needs to be more prominant - called out - bold - something 

In this case, you need to configure the source-replica replication. Make one of the
   PXC nodes as an async replica. Upgrade the |pxc| async replica node and check the
   replication status.  - Needs to be expanded and scenario detailed as the others 

What is considered a Minor Upgrade?  - 
Minor upgrade

I would like to review again once a. changes are made

george-lorch:
> As soon as the last 5.7 node shuts down, the configuration of the remaining
> two nodes is updated to use protocol version 4 - Is this really true or is it when the last node joins as a 8.0 node. Node could come back up as 5.7

As I understand it, once the **last** 5.7 node disconnects from the cluster (for any reason, crash, kill, shutdown, whatever), the remaining 8.0 nodes upgrade their communication protocol to v4.  At that point, a 5.7 node can no longer join/re-join the cluster.

doc/source/howtos/upgrade_guide.rst
patrickbirch: Added.

doc/source/howtos/upgrade_guide.rst
patrickbirch: Removed section on auto-upgrade.

doc/source/howtos/upgrade_guide.rst
patrickbirch: Added to the steps.

doc/source/howtos/upgrade_guide.rst
patrickbirch: added.

doc/source/howtos/upgrade_guide.rst
patrickbirch: fixed.

doc/source/howtos/upgrade_guide.rst
patrickbirch: removed

doc/source/howtos/upgrade_guide.rst
patrickbirch: fixed.

PR #1260, opened 2020-10-01
PXC-3442: Fixed wsrep checkpoint assertion when DML executed in TOI (5.6)

kamil-holubicki:
Problem:
is executed as TOI. This is done by pt-table-checksum tool.

Solution:
and logic to prevent double storing	of the same value. This solution is

### DIFF
--- a/storage/innobase/trx/trx0sys.cc
+++ b/storage/innobase/trx/trx0sys.cc
@@ -361,7 +361,17 @@ trx_sys_update_wsrep_checkpoint(
             read_wsrep_xid_uuid(xid, xid_uuid);
             if (!memcmp(xid_uuid, trx_sys_cur_xid_uuid, 8))
             {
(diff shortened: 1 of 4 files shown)


storage/innobase/trx/trx0sys.cc
```diff
@@ -361,7 +361,17 @@ trx_sys_update_wsrep_checkpoint(
             read_wsrep_xid_uuid(xid, xid_uuid);
             if (!memcmp(xid_uuid, trx_sys_cur_xid_uuid, 8))
             {
-                ut_ad(xid_seqno > trx_sys_cur_xid_seqno);
+				/* We have the case when REPLACE INTO ... SELECT or
```
venkatesh-prasad-v: Just to make it look uniform with the rest of the function, can we use whitespaces here?

venkatesh-prasad-v (approved):
Apart from the above cosmetic change, I do not have anything to comment on. So, it is approved from my side.

PR #1261, opened 2020-10-05
PXC-3442: Fixed wsrep checkpoint assertion when DML executed in TOI (5.7)

kamil-holubicki:
Problem:
is executed as TOI. This is done by	pt-table-checksum tool.

Solution:
and logic to prevent double storing	of the same value. This solution is


venkatesh-prasad-v (approved):
LGTM

PR #1262, opened 2020-10-05
PXC-3442: Fixed wsrep checkpoint assertion when DML executed in TOI (8.0)

kamil-holubicki:
Problem:
is executed as TOI. This is done by pt-table-checksum tool.

Solution:
and logic to prevent double storing of the same value. This solution is

### DIFF
--- a/mysql-test/suite/galera/t/pxc_log_slave_updates.test
+++ b/mysql-test/suite/galera/t/pxc_log_slave_updates.test
@@ -1,0 +1,22 @@
+#
+# Test that executing INSERT INTO ... SELECT in TOI mode, like pt-table-checksum
+# tool does, does not cause cluster node to crash if started with 
+# log_slave_updates=ON option.
+#
+--source include/galera_cluster.inc
+
+--connection node_1
+CREATE TABLE t1 (a INT PRIMARY KEY);
+CREATE TABLE t2 (a INT PRIMARY KEY);
+INSERT INTO t1 VALUES (0);
+
+--connection node_2
+INSERT INTO t2 SELECT /*!99997*/ * FROM t1;
+
+--connection node_1
+--let $assert_text = node_1 should be alive and t2 should contain one row
+--let $assert_cond = [SELECT COUNT(*) FROM t2] = 1
+--source include/assert.inc
+
+# cleanup
+DROP TABLE t1, t2;
--- a/storage/innobase/trx/trx0sys.cc
+++ b/storage/innobase/trx/trx0sys.cc
@@ -361,7 +361,17 @@ trx_sys_update_wsrep_checkpoint(
             read_wsrep_xid_uuid(xid, xid_uuid);
             if (!memcmp(xid_uuid, trx_sys_cur_xid_uuid, 8))
             {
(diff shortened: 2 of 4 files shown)


storage/innobase/trx/trx0sys.cc
```diff
@@ -431,12 +431,23 @@ void trx_sys_update_wsrep_checkpoint(
         /* Wait and unregister from wsrep group commit */
         wsrep_wait_for_turn_in_group_commit(current_thd);
 
-        /* DDL transaction are executed as TOI.
+        /* Case 1:
+        DDL transaction are executed as TOI.
```
venkatesh-prasad-v: s/transaction/transactions/

mysql-test/suite/galera/t/pxc_log_slave_updates.test
```diff
@@ -0,0 +1,22 @@
...
+#
+--source include/galera_cluster.inc
+
+--connection node_1
+CREATE TABLE t1 (a INT PRIMARY KEY);
+CREATE TABLE t2 (a INT PRIMARY KEY);
+INSERT INTO t1 VALUES (0);
+
+--connection node_2
+INSERT INTO t2 SELECT /*!99997*/ * FROM t1;
```
venkatesh-prasad-v, line 14: Not an issue. Just leaving a note here that we don't forget. 

Executing the above INSERT..SELECT statement twice causes the server to crash with an assertion failure in `wsrep::transaction::before_rollback()`

Test case:
```

--connection node_1
CREATE TABLE t1 (a INT PRIMARY KEY);
CREATE TABLE t2 (a INT PRIMARY KEY);
INSERT INTO t1 VALUES (0);

--connection node_2
INSERT INTO t2 SELECT /*!99997*/ * FROM t1;
INSERT INTO t2 SELECT /*!99997*/ * FROM t1;
```

```
2020-10-05T16:58:50.501787Z 15 [Note] [MY-000000] [WSREP] Query (INSERT INTO t2 SELECT /*!99997*/ * FROM t1) with write-set (7) and exec_mode: toi replicated in TO Isolation mode
2020-10-05T16:58:50.501863Z 15 [Note] [MY-000000] [WSREP] wsrep: TO isolation initiated for write set (7)
mysqld: /home/venki/work/pxc/3418/80/wsrep-lib/src/transaction.cpp:621: int wsrep::transaction::before_rollback(): Assertion `0' failed.
2020-10-05T16:58:50.505057Z 15 [Note] [MY-000000] [WSREP] Initiating SST cancellation

wsrep::transaction::before_rollback()
wsrep::client_state::before_rollback()
ha_rollback_low(THD*, bool)
MYSQL_BIN_LOG::rollback(THD*, bool)
ha_rollback_trans(THD*, bool)
trans_rollback_stmt(THD*)
mysql_execute_command(THD*, bool)
mysql_parse(THD*, Parser_state*, bool)
dispatch_command(THD*, COM_DATA const*, enum_server_command)
do_command(THD*)
```

venkatesh-prasad-v (approved):
Hi Kamil, 

Apart from the above minor suggestion, I do not have anything to comment on.

So, it is approved from my side.

PR #1273, opened 2020-10-15
8.0 PXC-3191 : cluster cannot update wsrep_* tables if in super_read_only

kennt-percona:
Issue
If super-read-only=ON, the system cannot start up correctly. There are
two problems (1) we cannot modify the wsrep_xxx system tables and (2) the
SST donor cannot create/modify the mysql.pxc.sst.user to perform the
backup

Solution
(1) skip the readonly checks if we are modifying a wsrep_xxx table
(2) skip the readonly checks if we are running as the PXC internal
    session user.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
+++ b/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
@@ -1,26 +1,235 @@
 #
-# This test checks the pxc-encrypt-cluster-traffic option (auto SSL config).
-# Initial SST happens via xtrabackup, so there is not much to do in the body of the test
+# This test checks that the cluster starts up (without error)
+# if super-read-only=ON.  This will check server startup
+# as well as the SST process.
 #
 
 --source include/big_test.inc
+--source include/have_util_sed.inc
 --source include/galera_cluster.inc
-
-SELECT 1;
 
 --let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size';
 --source include/wait_condition.inc
 
+SELECT 1;
+
+# --------------------------------
+# Test 1: Joiner node is super-read-only
+# --------------------------------
+
+--let $SUPER_READ_ONLY_CONFIG_FILE = $MYSQLTEST_VARDIR/tmp/galera_sst_xtrabackup-v2_super_read_only.test.cnf
+--copy_file $MYSQLTEST_VARDIR/my.cnf $SUPER_READ_ONLY_CONFIG_FILE
+--append_file $SUPER_READ_ONLY_CONFIG_FILE
+	[mysqld]
+	super_read_only=ON
+EOF
+
+# --------------------------------
+# Test 1a: Joiner (node2) is super-read-only (no SST)
+# --------------------------------
+
+# restart the node2 with super-read-only=ON (no SST)
 --connection node_2
+--echo [connection node_2]
+
+--echo "# restarting node2 (no SST)"
+--let $restart_hide_args=1
+--let $restart_parameters = "restart: --defaults-file=$SUPER_READ_ONLY_CONFIG_FILE"
+--source include/restart_mysqld.inc
+
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size';
+--source include/wait_condition.inc
+
+SELECT 1;
+
+# We should not find the super-read-only warnings in the logs
+--let $assert_text = Could not find warnings about super-read-only
+--let $assert_select = super-read-only
+--let $assert_count = 0
+--let $assert_file = $MYSQLTEST_VARDIR/log/mysqld.2.err
+--let $assert_only_after = CURRENT_TEST
+--source include/assert_grep.inc
+
+
+# --------------------------------
+# Test 1b: Joiner (node2) is super-read-only (with SST)
+# --------------------------------
+--echo "# shutting down node2"
 SET SESSION wsrep_sync_wait = 0;
 --source include/shutdown_mysqld.inc
+
+# remove the grastate.dat file to force an SST
 --remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
---exec sed -i '/^\[mysqld.2\]/a super_read_only=ON' $MYSQLTEST_VARDIR/my.cnf
+
+# restart the server with super-read-only=ON (SST)
+--echo "# restarting node2 (with SST)"
+--let $restart_hide_args=1
+--let $start_mysqld_params = "--defaults-file=$SUPER_READ_ONLY_CONFIG_FILE"
 --let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.2.expect
 --source include/start_mysqld.inc
 
-#cleanup
+SET SESSION wsrep_sync_wait = 0;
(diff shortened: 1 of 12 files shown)


mysql-test/suite/galera/t/galera_super_read_only.test
```diff
@@ -0,0 +1,24 @@
...
+# super-read-only=ON
+
+--source include/galera_cluster.inc
+--source include/force_restart.inc
+
+--connection node_2
+--echo [connection node_2]
+--let $restart_parameters = "restart: --super-read-only"
+--source include/restart_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: This is redundant as `include/restart_mysqld.inc` does poll the server until it becomes online.

mysql-test/suite/galera/t/galera_super_read_only.test
```diff
@@ -0,0 +1,24 @@
+# Test handling of super-read-only on server startup
+# This test will shutdown node2 and then restart it with
```
venkatesh-prasad-v: I guess is better we add a scenario where the donor has `super-read-only` set and the joiner tries to join the cluster.

sql/wsrep_mysqld.cc
```diff
@@ -344,6 +344,12 @@ PSI_file_info all_galera_files[] = {
 typedef std::vector<void *> wsrep_psi_key_vec_t;
 static wsrep_psi_key_vec_t wsrep_psi_key_vec;
 
+LEX_CSTRING PXC_INTERNAL_SESSION_USER =
```
venkatesh-prasad-v: The output of clang-format-8 on all the updated files seems to be differing from the current patch. Can you check this?
```
-LEX_CSTRING PXC_INTERNAL_SESSION_USER =
-    {STRING_WITH_LEN("mysql.pxc.internal.session")};
-LEX_CSTRING PXC_INTERNAL_SESSION_HOST =
-    {STRING_WITH_LEN("localhost")};
-
+LEX_CSTRING PXC_INTERNAL_SESSION_USER = {
+    STRING_WITH_LEN("mysql.pxc.internal.session")};
+LEX_CSTRING PXC_INTERNAL_SESSION_HOST = {STRING_WITH_LEN("localhost")};

```

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
```diff
@@ -1,26 +1,97 @@
...
+# --------------------------------
+# Test 1a: Restart node2 (no sst)
+# --------------------------------
+
+# restart the server with super-read-only=ON (no SST)
+--echo "# restarting node2 (no SST)"
+--let $restart_hide_args=1
+--let $restart_parameters = "restart: --defaults-file=$NODE2_CONFIG_FILE"
+--source include/restart_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: Wouldn't just this be enough? Is it really required to create a temporary config file?
```
+ --echo "# restarting node2 (no SST)"
+ --let $restart_parameters = "restart: --super-read-only=ON"
+ --source include/restart_mysqld.inc
```

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
```diff
@@ -1,26 +1,97 @@
...
 --remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
---exec sed -i '/^\[mysqld.2\]/a super_read_only=ON' $MYSQLTEST_VARDIR/my.cnf
+
+# restart the server with super-read-only=ON (SST)
+--echo "# restarting node2 (with SST)"
+--let $restart_hide_args=1
+--let $start_mysqld_params = "--defaults-file=$NODE2_CONFIG_FILE"
 --let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.2.expect
 --source include/start_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: Same as above.
```
+ # restart the server with super-read-only=ON (SST)
+ --echo "# restarting node2 (with SST)"
+ --let $restart_parameters = "restart: --super-read-only=ON"
  --let $_expect_file_name= $MYSQLTEST_VARDIR/tmp/mysqld.2.expect
  --source include/start_mysqld.inc
```

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
```diff
@@ -1,26 +1,97 @@
...
 #cleanup
 --connection node_2
-SET GLOBAL super_read_only = OFF;
-SET GLOBAL read_only = OFF;
---exec sed -i '/^super_read_only=ON/d' $MYSQLTEST_VARDIR/my.cnf
+--echo [connection node_2]
+
+--let $restart_parameters = "restart:"
+--source include/restart_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: This is redundant. We can avoid including `wait_until_connected_again.inc` after using either the `start_mysqld.inc` and `restart_mysqld.inc`, as the same check is already done inisde the inc files.

sql/table.h
```diff
@@ -3959,6 +3965,20 @@ inline bool is_user_table(TABLE *table) {
...
 }
 
+#ifdef WITH_WSREP
+inline bool is_wsrep_system_table(const char *db_name, size_t db_len,
+                                  const char *table_name, size_t table_len) {
+  return (WSREP_DB_NAME.length == db_len &&
+          !my_strcasecmp(system_charset_info, WSREP_DB_NAME.str,
+                         db_name) &&
+          WSREP_TABLE_PREFIX.length < table_len &&
+          !native_strncasecmp(table_name, WSREP_TABLE_PREFIX.str,
```
kamil-holubicki: why we use strcmp and strNcmp? 
I would prefer using strNcmp, but I see is_perfschema_db uses strcmp (so probably it is ensured by the caller that argument is the proper, null-terminated string).

At least I would stick to one of them if there is no reason for using different comparison functions.

sql/auth/sql_authorization.cc
```diff
@@ -1940,12 +1940,20 @@ bool check_readonly(THD *thd, bool err_if_readonly) {
...
   if (thd->slave_thread || thd->is_cmd_skip_readonly()) return false;
 
+  Security_context *sctx = thd->security_context();
 #ifdef WITH_WSREP
   /* Ignore readonly for background wsrep applier too (like slave thread) */
   if (WSREP(thd) && thd->wsrep_applier) return false;
+
+  /* Ignore readonly for PXC internal user */
+  if (sctx->user().str != nullptr &&
+      !strcmp(sctx->user().str, PXC_INTERNAL_SESSION_USER.str) &&
```
kamil-holubicki: Suggestion: Is it safe to use strcmp here? Maybe strNcmp would be better?

kamil-holubicki (approved):
lgtm

kennt-percona:
https://pxc.cd.percona.com/job/pxc80-test-mtr/230/

venkatesh-prasad-v (approved):
Nice work on the tests, Kenn! Approved from my side.

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
kennt-percona: see the comment below (it matters more for the SST process).

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_super_read_only.test
kennt-percona: The SST script does not have access to all of the command-line parameters, so it reads from the config file.  Thus we need the config file to test how the SST handles the super-read-only option.

PR #1274, opened 2020-10-15
PXC-3445 : MTR test failures in 8.0.20

kennt-percona:
Issue
Tests fail (sometimes) in PXC 8.0.20

Solution
Fix the tests:
galera.MW-328B : code can fail with ER_ERROR_DURING_COMMIT due to locking changes
galera_sr.galera-features#56 : code can fail on both nodes (but more often on node2)


dutow (approved):

venkatesh-prasad-v (approved):
Looks good

kamil-holubicki (approved):

PR #1277, opened 2020-10-16
PXC-3392: Donor uses invalid SST methods (5.6)

kamil-holubicki:


dutow (approved):

sql/wsrep_sst.cc
```diff
@@ -1276,6 +1414,78 @@ static int sst_donate_other (const char*   method,
...
+  }
+
+  std::string method= msg.substr(0, method_len);
+
+  // Is this method allowed?
+  std::vector<std::string>::iterator res= std::find(
+      allowed_sst_methods.begin(), allowed_sst_methods.end(), method);
+  if (res == allowed_sst_methods.end())
+  {
+    return false;
```
kennt-percona, line 1441: Maybe an error message here describing how to fix this would be good (saying that they need to add the method to the list of allowed methods). Otherwise, they wouldn't know how to fix the error.  Everything else looks good.

sql/wsrep_sst.cc
```diff
@@ -82,17 +83,18 @@ static Regex sst_allowed_methods_regex;
    characters and the ways they could be used to inject the command to
    the OS. However this approach seems to be too error prone.
    Instead of this we will just allow alpha-num + a few special characters
-   (colon, slash, dot, underscore, square brackets).
+   (colon, slash, dot, underscore, square brackets, hyphen).
 
-   It is the same as ^[\w:\/\.\[\]]+$    */
+   It is the same as ^[\w:/.[\]-]+$    */
```
venkatesh-prasad-v, line 88: Shouldn't this be
```suggestion
   It is the same as ^[\w:/.\[\]-]+$    */
```

sql/wsrep_sst.cc
```diff
@@ -82,17 +83,18 @@ static Regex sst_allowed_methods_regex;
...
 
    /* For some reason regex engine is not able to handle the following pattern
-   allowing square brackets as well:
-   "^[[:alnum:]:\\/\\._\\[\\]]+$"
-   We will substitute square brackets with lt/gt before validation
-   and we will revert to square brackets after validation. */
+   allowing square brackets:
+   "^[[:alnum:]:/._[\\]-]+$"
```
venkatesh-prasad-v, line 92: Shouldn't this be
```suggestion
   "^[[:alnum:]:/._\[\]-]+$"
```

mysql-test/suite/galera/t/galera_sst_allowed_methods.test
```diff
@@ -88,13 +88,18 @@
...
+
 --echo #
 --echo # Sending custom sst request
 --echo #
 
 --let $expect_rejection = no
---let $sst_allowed_methods = xtrabackup-v2,custom-sst_method,mysqldump
---let $sst_request_method = custom-sst_method
+--let $sst_allowed_methods = xtrabackup-v2,custom-sst_method.py,mysqldump
+--let $sst_request_method = custom-sst_method.py
```
venkatesh-prasad-v, line 102: Nice improvement :+1:

sql/wsrep_sst.cc
kamil-holubicki: No, we don't need to escape the opening square bracket.

venkatesh-prasad-v: You're right! https://www.regular-expressions.info/charclass.html#special says it is not required.

venkatesh-prasad-v (approved):
Great job! Thank you for addressing all the review comments. The patch looks good now. Approved from my side.

sql/wsrep_sst.cc
kamil-holubicki: In case of rejection, the caller function writes the error message "Invalid sst_request:" and dumps the whole request string. I don't think we should add more hints about what is wrong. We need to document the new parameter anyway. And honestly, I don't think there are no external users having their own sst scripts :)

PR #1278, opened 2020-10-16
PXC-3392: Donor uses invalid SST methods (5.7)

kamil-holubicki:


dutow (approved):

mysql-test/suite/galera/r/galera_defaults.result
```diff
@@ -42,6 +42,7 @@ WSREP_RSU_COMMIT_TIMEOUT	5000
 WSREP_SLAVE_FK_CHECKS	ON
 WSREP_SLAVE_THREADS	1
 WSREP_SLAVE_UK_CHECKS	OFF
+WSREP_SST_ALLOWED_METHODS	mysqldump,rsync,skip,xtrabackup,xtrabackup-v2
```
venkatesh-prasad-v, line 45: Not an issue, but as an improvement, can we extend this test case to print the count of wsrep variables we support, similar to the global status; so that it becomes handy for future references?

Something like having the below line would be sufficient.
```
SELECT COUNT(*) FROM performance_schema.global_variables WHERE VARIABLE_NAME LIKE 'wsrep_%';
```

venkatesh-prasad-v (approved):
Hi Kamil,

Apart from a minor improvement mentioned below (not a blocker though), the patch looks good. Approved from my side.

PR #1291, opened 2020-10-29
8.0 PXC-3146 : Galera/SST is not looking default data directory location …

kennt-percona:
…for SSL certs

Issue
SST does not use relative paths in the SSL key/cert/ca options.
So it requires the full absolute path.

Solution
Have the SST script convert the relative paths into absolute paths
(assuming that the current working directory is the data dir).

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_pxc_encrypt.test
+++ b/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_pxc_encrypt.test
@@ -2,20 +2,75 @@
 # This test checks the pxc-encrypt-cluster-traffic option (auto SSL config).
 # Initial SST happens via xtrabackup, so there is not much to do in the body of the test
 #
+# Also tests certs that are specified using relative paths (certs are in the datadir)
+#
 
 --source include/big_test.inc
 --source include/galera_cluster.inc
 
---disable_query_log
---connection node_1
-call mtr.add_suppression("Failed to initialize TLS for channel");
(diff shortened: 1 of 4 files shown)


mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_pxc_encrypt.test
```diff
@@ -2,20 +2,75 @@
 # This test checks the pxc-encrypt-cluster-traffic option (auto SSL config).
 # Initial SST happens via xtrabackup, so there is not much to do in the body of the test
 #
+# Also tests certs that are specified using relative paths (certs are in the datadir)
+#
 
 --source include/big_test.inc
 --source include/galera_cluster.inc
```
venkatesh-prasad-v, line 9: The same testcase in 5.7 has `--source include/force_restart.inc` included, and is missing here. Is it intentional?

venkatesh-prasad-v (approved):
Thank you for working on this issue. I only have a minor comment. Other than that I don't have any comments. So, it is good to go from my side.

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_pxc_encrypt.test
kennt-percona: It may have been removed during the merge.  node2 is restarted at the end of the test, so the force_restart.inc is not needed (should have been removed from the 5.7 branch).

PR #1295, opened 2020-11-05
PXC-3466 Review and Update - Experimental Features (8.0)

patrickbirch:
modified:   configure.rst
	modified:   features/pxc-strict-mode.rst
	modified:   howtos/centos_howto.rst
	modified:   howtos/ubuntu_howto.rst
	modified:   howtos/upgrade_guide.rst
	modified:   index.rst
	modified:   security/simple-ldap.rst
	modified:   wsrep-system-index.rst

### DIFF
--- a/doc/source/features/pxc-strict-mode.rst
+++ b/doc/source/features/pxc-strict-mode.rst
@@ -5,7 +5,7 @@ PXC Strict Mode
 ===============
 
 PXC Strict Mode is designed to avoid the use of
-experimental and unsupported features in |PXC|.
+tech preview features and unsupported features in |PXC|.
 It performs a number of validations at startup and during runtime.
 
 Depending on the actual mode you select,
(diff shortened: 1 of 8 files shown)


kamil-holubicki (approved):

doc/source/features/pxc-strict-mode.rst
```diff
@@ -303,7 +305,7 @@ Depending on the selected mode, the following happens:
 Explicit table locking
 ----------------------
 
-|PXC| has only experimental support for explicit table locking operations,
+|PXC| supports explicit table locking operations,
```
george-lorch: This is suspect. AFAIK PXC does not correctly and fully support explicit table locks in that the locks are not propagated and taken on all nodes, only the node which received the lock request.  This may lead to unexpected certification failures later on that leads to a different observable behavior of a cluster vs a non-cluster instance of PS.

patrickbirch: Changed to tech preview support

PR #1303, opened 2020-11-12
[8.0] PXC-3418: Cluster node locked up with alter table and concurrent RW load

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3418
https://jira.percona.com/browse/PXC-2382

Problem
-------
Cluster node can hang when there is a concurrent read-write load on the table
and there is a parallel inplace ALTER TABLE is happening on the table.

Analysis
--------
The deadlock can be seen even on a single node cluster with

- An Inplace ALTER TABLE query in one session.
- A DML on same table in from another session.

1. In general, as per PXC protocol, DDL in TOI/RSU is a blocking operation.
   This means when DDL is active no other transaction is allowed to make
   progress.

2. When DDL starts, it will block upcoming transactions from committing till
   it is completed. If these transactions are valid (non-conflicting) then
   generally these transactions will end up waiting for commit and this would
   cause flow-control to kick-in.

3. If a conflicting DML is executed on the same machine, in most cases, MDL locks
   can protect it from concurrent execution. But, in case of an inplace ALTER,
   the query doesn't need EXCLUSIVE MDL lock for complete tenure. So, DMLs can
   take SHARED_WRITE metadata lock on the same table and can even update the
   rows. However, if the DML is conflicting, it shall fail during the
   certification stage thereby causing a local cert failure. 

4. These transactions then enter into a wait state inside the local_monitor in
   Galera untill its turn. In this case, it can proceed only after ALTER TABLE
   exits the local monitor.

futex_wait_cancelable,__pthread_cond_wait_common,__pthread_cond_wait,
gu::Lock::wait,galera::Monitor::enter,galera::ReplicatorSMM::cert,
galera::ReplicatorSMM::cert_and_catch,galera::ReplicatorSMM::certify,
galera_certify,wsrep::wsrep_provider_v26::certify(wsrep_provider_v26.cpp:778),
wsrep::transaction::certify_commit(transaction.cpp:1380),
wsrep::transaction::before_prepare(transaction.cpp:311),
wsrep::client_state::before_prepare(client_state.hpp:421),
wsrep_before_prepare(wsrep_trans_observer.h:241),ha_prepare_low(handler.cc:2491),
MYSQL_BIN_LOG::prepare(binlog.cc:8401),ha_commit_trans(handler.cc:1817),
trans_commit_stmt(transaction.cc:546),mysql_execute_command(sql_parse.cc:5683),
mysql_parse(sql_parse.cc:6391),wsrep_mysql_parse(sql_parse.cc:7662),
dispatch_command(sql_parse.cc:2128),do_command(sql_parse.cc:1436),

5. In the meanwhile, if the inplace ALTER TABLE tries to upgrade the lock to
   EXCLUSIVE during its execution, the acquisition fails (due to the above
   transactions holding the SHARED MDLs) and thus ALTER query waits until it
   gets the required LOCK on the table.

Stack trace (pt-pmp):

futex_abstimed_wait_cancelable,__pthread_cond_wait_common,
__pthread_cond_timedwait,native_cond_timedwait(thr_cond.h:100),
my_cond_timedwait(thr_cond.h:149),inline_mysql_cond_timedwait(mysql_cond.h:217),
MDL_wait::timed_wait(mdl.cc:1880),MDL_context::acquire_lock(mdl.cc:3631),
MDL_context::upgrade_shared_lock(mdl.cc:3896),wait_while_table_is_used(sql_base.cc:2480),
mysql_inplace_alter_table,mysql_alter_table(sql_table.cc:17083),
Sql_cmd_alter_table::execute(sql_alter.cc:608),mysql_execute_command(sql_parse.cc:5476),
mysql_parse(sql_parse.cc:6391),wsrep_mysql_parse(sql_parse.cc:7662),
dispatch_command(sql_parse.cc:2128),do_command(sql_parse.cc:1436),

In simple, this is a 2 way dead lock where-in 

- The client threads wait for their turn (can happen only when ALTER TABLE
  finishes its execution).

- The alter table query is trying to upgrade to an exclusive MDL lock on the
  target table and is waiting for all the users of the table to release the
  MDL.

Solution
--------
Make all the inplace alter table to take the MDL for the whole duration.

Observed behavioral changes post fix:

All the DMLs now will now wait for the metadata lock on the table if there is
an ongoing inplace ALTER TABLE on the same table. With this change we wouldn't
be deviating much away from the existing behavior as we are already blocking
the DMLs during the execution of a DDL. As a result, the conflicting DMLs will
now wait instead of the erroring out with deadlock error.

Note to reviewers
---------
1. In 8.0, Deadlock happens only when the ALTER TABLE is executing in RSU and doesn't happen with TOI. In case of TOI, the DML fails the certification process, rolls back the statement and will wait inside the commit_order_enter when it tries to relase the commit_order.

```
futex_wait_cancelable,__pthread_cond_wait_common,__pthread_cond_wait,gu::Lock::wait,galera::Monitor::enter,galera::ReplicatorSMM::commit_order_enter_local,galera_commit_order_enter,wsrep::wsrep_provider_v26::commit_order_enter(wsrep_provider_v26.cpp:812),wsrep::transaction::release_commit_order(transaction.cpp:680),wsrep::transaction::after_statement(transaction.cpp:783),wsrep::client_state::after_statement(client_state.cpp:245),wsrep_after_statement(wsrep_trans_observer.h:452),wsrep_mysql_parse(sql_parse.cc:7686),dispatch_command(sql_parse.cc:2128),do_command(sql_parse.cc:1436),handle_connection(connection_handler_per_thread.cc:312),pfs_spawn_thread(pfs.cc:2855),start_thread,clone
```

### DIFF
--- a/mysql-test/suite/galera/r/galera_alter_table_big.result
+++ b/mysql-test/suite/galera/r/galera_alter_table_big.result
@@ -1,0 +1,4452 @@
+SET sql_mode = 'NO_ENGINE_SUBSTITUTION';
+SET SESSION information_schema_stats_expiry=0;
+create table t1 (
+col1 int not null auto_increment primary key,
+col2 varchar(30) not null,
+col3 varchar (20) not null,
+col4 varchar(4) not null,
+col5 enum('PENDING', 'ACTIVE', 'DISABLED') not null,
+col6 int not null, to_be_deleted int);
+insert into t1 values (2,4,3,5,"PENDING",1,7);
+alter table t1
+add column col4_5 varchar(20) not null after col4,
+add column col7 varchar(30) not null after col5,
+add column col8 datetime not null, drop column to_be_deleted,
+change column col2 fourth varchar(30) not null after col3,
+modify column col6 int not null first;
+select * from t1;
+col6	col1	col3	fourth	col4	col4_5	col5	col7	col8
+1	2	3	4	5		PENDING		0000-00-00 00:00:00
+drop table t1;
+create table t1 (bandID MEDIUMINT UNSIGNED NOT NULL PRIMARY KEY, payoutID SMALLINT UNSIGNED NOT NULL) engine=myisam;
+insert into t1 (bandID,payoutID) VALUES (1,6),(2,6),(3,4),(4,9),(5,10),(6,1),(7,12),(8,12);
+alter table t1 add column new_col int, order by payoutid,bandid;
+select * from t1;
+bandID	payoutID	new_col
+6	1	NULL
+3	4	NULL
+1	6	NULL
+2	6	NULL
+4	9	NULL
+5	10	NULL
+7	12	NULL
+8	12	NULL
+alter table t1 order by bandid,payoutid;
+select * from t1;
+bandID	payoutID	new_col
+1	6	NULL
+2	6	NULL
+3	4	NULL
+4	9	NULL
+5	10	NULL
+6	1	NULL
+7	12	NULL
+8	12	NULL
+drop table t1;
+create table t1 (bandID MEDIUMINT UNSIGNED NOT NULL PRIMARY KEY, payoutID SMALLINT UNSIGNED NOT NULL) engine=InnoDB;
+insert into t1 (bandID,payoutID) VALUES (1,6),(2,6),(3,4),(4,9),(5,10),(6,1),(7,12),(8,12);
+alter table t1 add column new_col int, order by payoutid,bandid;
+Warnings:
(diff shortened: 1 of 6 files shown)


venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/113/testReport/

dutow (approved):

kennt-percona (approved):

PR #1304, opened 2020-11-16
PXC-3381 : GTID functions broken in PXC 8.0.19

kennt-percona:
Issue
When calling WSREP_LAST_WRITTEN_GTID() and WSREP_LAST_SEEN_GTID()
from the MySQL client, they would return a hexadecimal string, instead
of a UUID:seqno.

Solution
The solution is to change the functions to use the latin1 charset instead
of the bin charset.

### DIFF
--- a/mysql-test/suite/galera/t/galera_gtid_functions.cnf
+++ b/mysql-test/suite/galera/t/galera_gtid_functions.cnf
@@ -1,0 +1,8 @@
+!include ../galera_2nodes.cnf
+
+[mysqld]
+gtid-mode=ON
+log-bin=mysqld-bin
+log-slave-updates
+enforce-gtid-consistency
+binlog-format=ROW
--- a/mysql-test/suite/galera/t/galera_gtid_functions.test
+++ b/mysql-test/suite/galera/t/galera_gtid_functions.test
@@ -1,0 +1,149 @@
+#
+# Test the WSREP GTID-related functions
+#  WSREP_LAST_SEEN_GTID()
+#  WSREP_LAST_WRITTEN_GTID()
+#  WSREP_SYNC_WAIT_UPTO_GTID()
+#
+
+--source include/have_debug.inc
+--source suite/galera/include/galera_have_debug_sync.inc
+--source include/have_log_bin.inc
+--source include/have_util_sed.inc
+--source include/galera_cluster.inc
+
+--connection node_1
+--echo [connection node_1]
+
+--let $outfile = $MYSQLTEST_VARDIR/tmp/galera_gtid_functions.out
+
+CREATE TABLE t1 (id INT PRIMARY KEY);
+INSERT INTO t1 VALUES (1);
+
+#
+# Test 1 : WSREP_LAST_SEEN_GTID
+#
+
+# Retrieve the last seen gtid
+--let $last_seen_gtid=`SELECT WSREP_LAST_SEEN_GTID()`
+
+# Verify that we get the same result using the MySQL client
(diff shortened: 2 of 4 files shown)


kennt-percona:
https://pxc.cd.percona.com/job/pxc80-test-mtr/232/

mysql-test/suite/galera/t/galera_gtid_functions.cnf
```diff
@@ -0,0 +1,8 @@
+!include ../galera_2nodes_as_master.cnf
```
venkatesh-prasad-v: This makes the test to start with 3 servers, which is not required for this test case. Please change it to 
```suggestion
!include ../galera_2nodes.cnf
```

mysql-test/suite/galera/t/galera_gtid_functions.test
```diff
@@ -0,0 +1,148 @@
...
+--disable_query_log
+--echo # Call to WSREP_SYNC_WAIT_UPTO_GTID() which should fail (with bad arguments)
+--error ER_WRONG_ARGUMENTS
+--eval SELECT WSREP_SYNC_WAIT_UPTO_GTID('00000000:-1', 1) AS WSREP_SYNC_WAIT_UPTO_GTID_EXPECTED;
+--enable_query_log
+
+# WSREP_SYNC_WAIT_UPTO_GTID() should fail (because the applier is blocked)
+--disable_query_log
+--echo # Call to WSREP_SYNC_WAIT_UPTO_GTID() which should fail (timeout due to blocked applier)
+--error ER_LOCK_WAIT_TIMEOUT
```
venkatesh-prasad-v, line 120: Not an issue. May be we should consider `ER_QUERY_TIMEOUT` to report a timeout.

venkatesh-prasad-v (commented):
Hi Kenn! Thank you for working on this. Good work on the test cases.

Regarding the commit message, I think it is better if also we include the reason for this regression so that it helps in better understanding of the problem. Something like the message in the https://github.com/percona/percona-server/commit/fa47b9488c6536b9ad7b0214ff6609cf9a1ff8e4 would be fine.

kamil-holubicki (commented):
I agree with @venkatesh-prasad-v comments. Nothing else from my side.

mysql-test/suite/galera/t/galera_gtid_functions.test
kennt-percona: The Galera docs say that ER_LOCK_WAIT_TIMEOUT is returned on timeout: https://github.com/codership/documentation/issues/145

mysql-test/suite/galera/t/galera_gtid_functions.cnf
kennt-percona: fixed

kennt-percona:
@venkatesh-prasad-v   Updated the commit message (you'll have to go see the actual commit).

mysql-test/suite/galera/t/galera_gtid_functions.test
```diff
@@ -0,0 +1,148 @@
...
+# Test the WSREP GTID-related functions
+#  WSREP_LAST_SEEN_GTID()
+#  WSREP_LAST_WRITTEN_GTID()
+#  WSREP_SYNC_WAIT_UPTO_GTID()
+#
+
+--source include/have_debug.inc
+--source include/have_log_bin.inc
+--source include/have_util_sed.inc
+--source include/galera_cluster.inc
```
venkatesh-prasad-v: Please source `suite/galera/include/galera_have_debug_sync.inc`

kennt-percona: fixed

venkatesh-prasad-v (approved):
LGTM

PR #1307, opened 2020-11-23
Null-merge 8.0 PXC-3459 

altmannmarcelo:
https://jira.percona.com/browse/PXC-3459

Problem:
In row_ins_foreign_check_on_constraint() we pass clustered index record
to wsrep_append_foreign_key() after the call of mtr_commit. In case of
other thread modifying it, wsrep_append_foreign_key might crash the
server.

Fix:
Pass cascade->pcur->old_rec instead. This has a copy of clustered index
record.


venkatesh-prasad-v (approved):
Thanks for working on this. Approved from my side.

PR #1318, opened 2020-12-09
PXC 3115: AppArmor and SELinux configuration

kamil-holubicki:


build-ps/debian/percona-xtradb-cluster-server.postinst
```diff
@@ -85,9 +85,34 @@ case "$1" in
...
+		if [ "${CNF_ACTION}" = "Use NEW my.cnf" ]; then
+			update-alternatives --force --install /etc/mysql/my.cnf my.cnf "/etc/mysql/mysql.cnf" 200
+		fi
+
+		PROFILE_ACTION="Use NEW AppArmor profile"
+		# If the existing AppArmor module/local profile is the proper file, we back it up
+		if [ -f "/etc/apparmor.d/usr.sbin.mysqld" -o -f "/etc/apparmor.d/local/usr.sbin.mysqld" -o -f "/etc/apparmor.d/usr.bin.wsrep_sst_xtrabackup-v2" ]; then
+			cp /etc/apparmor.d/usr.sbin.mysqld /etc/apparmor.d/usr.sbin.mysqld.bak 2> /dev/null || true
+			cp /etc/apparmor.d/local/usr.sbin.mysqld /etc/apparmor.d/local/usr.sbin.mysqld.bak 2> /dev/null || true
+			cp /etc/apparmor.d/usr.bin.wsrep_sst_xtrabackup-v2 /etc/apparmor.d/usr.bin.wsrep_sst_xtrabackup-v2.bak 2> /dev/null || true
```
EvgeniyPatlan: maybe it is better to copy to backup based on $RET result.
Also as a suggestion - add date for backup file.
As this action would be performed on each install and update and we could replace file which is needed for end user

kamil-holubicki: I followed the same pattern as we have for /etc/mysql/my.cnf a few lines above. But I agree that the proposed way is better. So I think we should change it for both.

kamil-holubicki: I decided to apply the suggested change only for AA part. The backup of my.cnf is similar, but not the same as the file is potentially modified in the meantime and that would need more rework.

EvgeniyPatlan (approved):
LGTM

PR #1321, opened 2020-12-16
PXC-3502: Crash while listing indexes

kamil-holubicki:
https://jira.percona.com/browse/PXC-3502

Problem:
For tables having more than one index, listing indexes causes server
crash.

Cause:
Wsrep commit hooks are called because of wrong condition in function
thd_binlog_format(). They are called even if binlog is disabled.
The problem is caused by wrong initial merge from 5.7 branch.

Solution:
Fixed condition in thd_binlog_format() function.


venkatesh-prasad-v (approved):
Nice catch. Approved!

PR #1322, opened 2020-12-16
PXC-3509: Do not try to create SSL files when --help is specified

dutow:
Issue: the server tries to create ssl files even when the --help option
is specified, but in this case, the working directory isn't changed to
the datadir.

This results in:
1. the server creating SSL certificates in an incorrect location (at the
  current working directory when executing mysqld --help)
2. if the current user has no write permission to the current working
  directory, startup will fail with permission denied errors.

Fix: only initialize wsrep ssl if the help option isn't specified


kamil-holubicki (approved):
lgtm

PR #1323, opened 2020-12-18
PXC-2913: Fix MDL locks assertion when wsrep provider is unloaded

kamil-holubicki:
https://jira.percona.com/browse/PXC-2913

Problem:
This problem is visible only for debug builds and the case when wsrep
provider is not loaded (SET GLOBAL wsrep_provider = 'none')

1. Backup lock is taken
2. Transaction is started. This causes wsrep transaction object to be
created
3. Without involving any storage engine execute statement that causes
error. This causes wsrep transaction object transition to 'aborted'
state
4. Because wsrep provider is not loaded, wsrep_after_statement() is not
called and wsrep transaction id is not cleared
5. Execute next statement. wsrep_before_command() is executed and it
detects aborted transaction which causes MDL locks check (problematic
assert). However backup locks are allowed at this point.

Cause:
Wrong assert condition.

Solution:
Assert condition disallowing any MDL locks changed to the one
disallowing stmt/transactional locks. Explicit locks are allowed.

Note:
We still have the effect of the failure of the statement next to the one
causing wsrep transaction rollback. It fails with ER_LOCK_DEADLOCK. This
is caused by explained above detection of aborted transaction in
wsrep_before_command(). As this is the corner case and is the same as we
already have in release build, I decided not to change it. It would need
bigger logic rework, which seems not to be necessary for now.

### DIFF
--- a/mysql-test/suite/galera/t/galera_lock_table.test
+++ b/mysql-test/suite/galera/t/galera_lock_table.test
@@ -66,3 +66,83 @@ show create table wsrep_cluster;
 #
 use mysql;
 select count(*) from wsrep_cluster;
+
+#
+# Test scenario:
+# 1) unload wsrep_provider
+# 2) lock tables/instance for backup
+# 3) start transaction
+# 4) execute statement that causes WSREP transaction rollback, but does not
+#    involve any storage engine
+# 5) check that backup lock is still present
+#
+
+--let $no_backup_locks_txt = No backup locks expected
+--let $backup_locks_txt = One backup lock expected
+
+--connection node_2
+--let $wsrep_provider_orig = `SELECT @@wsrep_provider`
+--let $wsrep_cluster_address_orig = `SELECT @@wsrep_cluster_address`
+
+SET GLOBAL wsrep_provider='none';
+
+
+--let $assert_text = $no_backup_locks_txt
+--let $assert_cond = [SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE="BACKUP TABLES"] = 0
+--source include/assert.inc
+LOCK TABLES FOR BACKUP;
+--let $assert_text = $backup_locks_txt
+--let $assert_cond = [SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE="BACKUP TABLES"] = 1
+--source include/assert.inc
+BEGIN;
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SET GLOBAL log_bin='wrong value';
+--error ER_LOCK_DEADLOCK
+SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE='BACKUP TABLES';
+
+--let $assert_text = $backup_locks_txt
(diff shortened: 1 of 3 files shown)


mysql-test/suite/galera/t/galera_lock_table.test
```diff
@@ -66,3 +66,78 @@ show create table wsrep_cluster;
...
+--let $assert_cond = [SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE='BACKUP LOCK'] = 1
+BEGIN;
+--error ER_INCORRECT_GLOBAL_LOCAL_VAR
+SET GLOBAL log_bin='wrong value';
+--error ER_LOCK_DEADLOCK
+SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE='BACKUP LOCK';
+
+--let $assert_text = $backup_locks_txt
+--let $assert_cond = [SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE='BACKUP LOCK'] = 1
+UNLOCK TABLES;
```
venkatesh-prasad-v: Shouldn't this be UNLOCK INSTANCE?

mysql-test/suite/galera/t/galera_lock_table.test
```diff
@@ -66,3 +66,78 @@ show create table wsrep_cluster;
...
+# MySQL transaction is still ongoing, only WSREP transaction was aborted.
+# Finish it before reloading wsrep provider.
+ROLLBACK;
+
+# Cleanup
+--disable_query_log
+--eval SET GLOBAL wsrep_provider = '$wsrep_provider_orig';
+--enable_query_log
+
+--disable_query_log
```
venkatesh-prasad-v: No need to enable and disable it again. If possible, can we remove this?

mysql-test/suite/galera/t/galera_lock_table.test
```diff
@@ -66,3 +66,78 @@ show create table wsrep_cluster;
...
+
+--connection node_2
+--let $wsrep_provider_orig = `SELECT @@wsrep_provider`
+--let $wsrep_cluster_address_orig = `SELECT @@wsrep_cluster_address`
+
+SET GLOBAL wsrep_provider='none';
+
+
+--let $assert_text = $no_backup_locks_txt
+--let $assert_cond = [SELECT COUNT(*) FROM performance_schema.metadata_locks WHERE OBJECT_TYPE='BACKUP TABLES'] = 0
```
venkatesh-prasad-v: In all relevant places, please source `assert.inc`.

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/galera_lock_table.test
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_lock_table.test
kamil-holubicki: What a silly mistake!. Fixed

venkatesh-prasad-v (approved):
LGTM

PR #1337, opened 2021-01-04

kennt-percona:
…TX_set_tmp_dh" from socat

Issue
used with OpenSSL 1.1.1.

Solution
old certs.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -442,6 +442,18 @@ get_transfer()
             if check_for_version "$SOCAT_VERSION" "1.7.3"; then
                 donor_extra=',commonname=""'
             fi
+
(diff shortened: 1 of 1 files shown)


kamil-holubicki (commented):
Looks good, but do we have MTR test that tests it?

kennt-percona:
https://pxc.cd.percona.com/job/pxc56-test-mtr/60/
All nodes failing but due to:
galera_sst_allowed_methods (all nodes)
galera_sst_mysqldump_with_key (on buster, which I thinks is due to some SSL build issue with mysqldump)

kamil-holubicki (commented):
LGTM, but I have 1 question: shouldn't we use GCA for this? I see the same done for 5.7
https://github.com/percona/percona-xtradb-cluster/pull/1338

kennt-percona:
@kamil-holubicki  Redid the 5.6 branch using the GCA process

kamil-holubicki (approved):

PR #1338, opened 2021-01-04

kennt-percona:
…TX_set_tmp_dh" from socat

Issue
used with OpenSSL 1.1.1.

Solution
old certs.


kamil-holubicki (commented):
Looks good.
Do we have Jenkins run for this?

kennt-percona:
https://pxc.cd.percona.com/job/pxc-5.7-param/96/#showFailuresLink
Failures are all in galera_3nodes.galera_innobackupex_backup, which does not use SSL (so the fix doesn't affect that). However, they are failing for the newer distros only (similar to what this PR is trying to fix). However, I haven't been able to repro this locally, so that is still under investigation.

venkatesh-prasad-v:
Hi @kennt-percona the test `galera_3nodes.galera_innobackupex_backup` was fixed by https://github.com/percona/percona-xtradb-cluster/commit/820686c15#diff-74a60453c07b0a25990bd45258d798295498623a99954bc78daf3969ce17c06e and it doesn't fail anymore. Can you check once on the latest 5.7 trunk? 

Edit: It fails due to https://jira.percona.com/browse/PXB-2381

kennt-percona:
@venkatesh-prasad-v 
Very nice. Everything passes (on the galera-only tests)!

https://pxc.cd.percona.com/job/pxc-5.7-param/100/

kamil-holubicki:
We have a similar change in 5.6. Shouldn't we use GCA?

kennt-percona:
@kamil-holubicki 5.7 cherry-picked from the 5.6 branch

kennt-percona:
https://pxc.cd.percona.com/job/pxc-5.7-param/119/

The tests that were failing on newer distros are no longer failing.
galera_sst_xtrabackup-v2_encrypt_with_key
galera_sst_xtrabackyp-v2_keyring

kamil-holubicki (approved):

PR #1355, opened 2021-01-12

kennt-percona:
…TX_set_tmp_dh" from socat

Issue
used with OpenSSL 1.1.1.

Solution
old certs.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -505,6 +505,18 @@ get_transfer()
             if compare_versions "$SOCAT_VERSION" ">=" "1.7.3"; then
                 donor_extra=',commonname=""'
             fi
+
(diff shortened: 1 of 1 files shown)


kamil-holubicki (commented):
Looks good,
Do we have MTR for this?
Do we have Jenkins run for this?

kennt-percona:
https://pxc.cd.percona.com/job/pxc-5.7-param/96/#showFailuresLink
The tests are still failing for the newer distros (focal, centos8, buster), but they don't use SSL, so the tests are failing for a different reason.

kamil-holubicki (approved):

kennt-percona:
https://pxc.cd.percona.com/job/pxc-8.0-param/182/#showFailuresLink

The tests that were failing on newer distros are no longer failing.
galera_sst_xtrabackup-v2_encrypt_with_key
galera_sst_xtrabackyp-v2_keyring

PR #1365, opened 2021-01-25
PXC-3561: Bug#32045681: Heap-use-after-free in triggers Item_splocal::this_item()

dutow:
This problem requires several subsequent sessions to use the same
trigger procedure, containing a local variable. Since Items are
prepared only on first installment in the server, and the m_thd member
of Item_splocal is initialized only on preparation, on second use
m_thd is different from the current THD and a failure is provoked.

There are at least two ways to fix this problem. One is to assign
m_thd when binding the procedure to the new session. However, the
solution may be considered vulnerable and we may risk ending with
the same problem. The strategy chosen here is to remove the m_thd
member and use current_thd instead. The existing code warns about it,
however accessing current_thd as a thread-local variable is
sufficiently fast now, and the same principle is used several other
places in the server. Still, passing a THD as context argument would
be even better, but is far too intrusive for a bugfix.

Reviewed by: Dmitry Lenev <Dmitry.Lenev@oracle.com>

### DIFF
--- a/sql/item.cc
+++ b/sql/item.cc
@@ -1609,17 +1609,9 @@ bool Item::is_blob_field() const {
 *****************************************************************************/
 
 Item_sp_variable::Item_sp_variable(const Name_string sp_var_name)
-    : m_thd(nullptr),
-      m_name(sp_var_name)
-#ifndef DBUG_OFF
-      ,
-      m_sp(nullptr)
-#endif
-{
-}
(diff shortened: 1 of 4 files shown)


dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/162/

kamil-holubicki (approved):

PR #1371, opened 2021-02-04
PXC-3565: PXC is more than 60% slower than PS

dutow:

select. This macro had a strlen call in it, which was executed two times
for each row.

instead of directly using strlen it introduces the wsrep_provider_set
changes.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -6627,6 +6627,11 @@ int mysqld_main(int argc, char **argv)
 #endif
   }
 
+#ifdef WITH_WSREP
+  // Initialize wsrep_provider_set before anything else wsrep related
+  wsrep_provider_set = wsrep_provider != NULL && strcmp(wsrep_provider, WSREP_NONE) != 0;
+#endif
+
   /* 
    The subsequent calls may take a long time : e.g. innodb log read.
    Thus set the long running service control manager timeout
@@ -6639,6 +6644,7 @@ int mysqld_main(int argc, char **argv)
     unireg_abort(1);
 
 #ifdef WITH_WSREP /* WSREP AFTER SE */
+
   if (wsrep_recovery)
   {
     select_thread_in_use= 0;
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -512,6 +512,7 @@ int wsrep_init()
   {
     // enable normal operation in case no provider is specified
     wsrep_ready_set(TRUE);
+    wsrep_provider_set = false;
     global_system_variables.wsrep_on = 0;
     wsrep_init_args args;
     args.logger_cb = wsrep_log_cb;
--- a/sql/wsrep_mysqld.h
+++ b/sql/wsrep_mysqld.h
@@ -213,10 +213,11 @@ extern "C" bool wsrep_is_wsrep_on(void);
 /* Other global variables */
 extern wsrep_seqno_t wsrep_locked_seqno;
 
-#define WSREP_ON                         \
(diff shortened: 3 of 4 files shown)


sql/mysqld.cc
```diff
@@ -6639,6 +6639,9 @@ int mysqld_main(int argc, char **argv)
     unireg_abort(1);
 
 #ifdef WITH_WSREP /* WSREP AFTER SE */
+  // Initialize wsrep_provider_set before anything else wsrep related
+  wsrep_provider_set = wsrep_provider != NULL && strcmp(wsrep_provider, WSREP_NONE) != 0;
```
kamil-holubicki: I think here wsrep_provider is always NULL (not initialized yet, so we probably can commit this at all (as wsrep_provier_set is already set to false)

kamil-holubicki (approved):
LGTM

sql/mysqld.cc
dutow: No, at this point sysvars are already initialized, this is where we set its initial value to true when it is specified at the command line / as a parameter.

sql/wsrep_mysqld.cc
```diff
@@ -488,6 +488,7 @@ int wsrep_init()
       WSREP_ERROR("wsrep_load() failed to load the provider('%s'): %s (%d). Reverting to no provider.",
                   wsrep_provider, strerror(rcode), rcode);
 
+      wsrep_provider_set = false;
```
venkatesh-prasad-v: I am not sure, I think we should also set `wsrep_provider_set = false` in case if `wsrep_load()` is successful and  `wsrep_provider` is set "none"/empty?

i.e, Inside this block https://github.com/percona/percona-xtradb-cluster/blob/ccd2185ba1247b3460f6074a40ec7582ccbf4f61/sql/wsrep_mysqld.cc#L515-L518

venkatesh-prasad-v (commented):
In the commit message, it is written that
>This macro had a strlen call in it, which was executed two times
for each row.

I guess you meant `strcmp`.

sql/wsrep_mysqld.cc
dutow: That's a good question. For WSREP_NONE, that's already done in mysqld.cc (that's where wsrep_init is called from). For the empty string... there are a few places where wsrep_mysqld.cc checks that, but this condition wasn't included in the WSREP_ON macro previously, so I'm not sure.

venkatesh-prasad-v: When the server is started with `wsrep_provider=""`, `wsrep_load()` fails and we enter this block https://github.com/percona/percona-xtradb-cluster/blob/ccd2185ba1247b3460f6074a40ec7582ccbf4f61/sql/wsrep_mysqld.cc#L491-L499

Here, we recursively call `wsrep_init()` with `wsrep_provider=none` and this second call to the `wsrep_init()` will be successful.

Considering this flow(with the proposed solution), we are setting `wsrep_provider_set = false` only when the server is started with `wsrep_provider=""` and it is not set if the server is started with `wsrep_provider=none`.

I think it is probably better to move `wsrep_provider_set = false` to this block, so that we set it correctly for both none and empty cases. https://github.com/percona/percona-xtradb-cluster/blob/ccd2185ba1247b3460f6074a40ec7582ccbf4f61/sql/wsrep_mysqld.cc#L518-L525

dutow: Done.

venkatesh-prasad-v (approved):
Thanks for working on this. Approved!

dutow:
https://pxc.cd.percona.com/view/PXC%205.6/job/pxc-5.6-param/116/

PR #1372, opened 2021-02-08
PXC-3565: PXC is more than 60% slower than PS (8.0)

dutow:

select. This macro had a strcmp call in it, which was executed two times
for each row.

instead of directly using strcmp it introduces the wsrep_provider_set
changes.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -6635,10 +6635,16 @@ int mysqld_main(int argc, char **argv)
   Service.SetSlowStarting(slow_start_timeout);
 #endif
 
+#ifdef WITH_WSREP
+  // Initialize wsrep_provider_set before anything else wsrep related
+  wsrep_provider_set = wsrep_provider != NULL && strcmp(wsrep_provider, WSREP_NONE) != 0;
+#endif
+
   if (init_server_components())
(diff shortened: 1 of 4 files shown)


kamil-holubicki (commented):
Please add a proper/descriptive PR title and description.

No other comments

dutow:
@kamil-holubicki description updated.

sql/mysqld.cc
```diff
@@ -8054,6 +8054,11 @@ int mysqld_main(int argc, char **argv)
     unireg_abort(MYSQLD_ABORT_EXIT); /* purecov: inspected */
   }
 
+#ifdef WITH_WSREP
+  // Initialize wsrep_provider_set before anything else wsrep related
+  wsrep_provider_set = wsrep_provider != NULL && strcmp(wsrep_provider, WSREP_NONE) != 0;
```
venkatesh-prasad-v: Can we use `nullptr` instead of `NULL` in all checks?

venkatesh-prasad-v (approved):

dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/171/#showFailuresLink

PR #1373, opened 2021-02-08
PXC-3565: PXC is more than 60% slower than PS (5.7)

dutow:

select. This macro had a strcmp call in it, which was executed two times
for each row.

instead of directly using strcmp it introduces the wsrep_provider_set
changes.

### DIFF
--- a/sql/mysqld.cc
+++ b/sql/mysqld.cc
@@ -6635,10 +6635,16 @@ int mysqld_main(int argc, char **argv)
   Service.SetSlowStarting(slow_start_timeout);
 #endif
 
+#ifdef WITH_WSREP
+  // Initialize wsrep_provider_set before anything else wsrep related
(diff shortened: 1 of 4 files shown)


kamil-holubicki (commented):
Please add a proper/descriptive PR title and description.
PR has merge conflicts.

No other comments

dutow:
@kamil-holubicki I'll fix that conflict with the github web editor, it's just because of GCA. Description updated.

dutow:
@kamil-holubicki I moved around the change in mysqld.cc, now we also don't have a conflict.

venkatesh-prasad-v (approved):
lgtm

dutow:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/112/#showFailuresLink

PR #1378, opened 2021-02-16
PXC-3585: Galera logs are not showing new lines

dutow:
Issues: MySQL 8.0.22 introduced logging changes, which replace new lines
in log messages to spaces. This broke the galera log format.

Fix: we ported the patch that disables this functionality for WSREP
related log entries from the codership sources.


kamil-holubicki (approved):
As it comes from Codership upstream: approved.

However,
Looking at this part of the code I understand that there is the silent assumption that LOG_ITEM_SRV_SUBSYS item is located earlier than LOG_ITEM_LOG_MESSAGE in ll->item array? Nothing forces this, but I understand that caller prepares the array in a proper way. Originally subsys_len is used outside the loop.

PR #1380, opened 2021-02-19
PXC-3536

kennt-percona:
Issue
DDL queries can be killed.  If the query is killed before it is replicated,
this can lead to inconsistencies, because the primary node's DDL will have
been cancelled, but the replicas will have received and performed the
operation.

Solution
Do not allow threads/queries to be killed if the thread is in TOI.


kennt-percona:
https://pxc.cd.percona.com/job/pxc-5.7-param/118/#showFailuresLink

sql/wsrep_mysqld.cc
```diff
@@ -2412,8 +2446,15 @@ void wsrep_to_isolation_end(THD *thd)
     wsrep_cleanup_transaction(thd);
   }
 
-  if (thd->wsrep_skip_wsrep_hton)
+  if (thd->wsrep_skip_wsrep_hton) {
+    DEBUG_SYNC(thd, "wsrep_to_isolation_end_before");
+    mysql_mutex_lock(&thd->LOCK_thd_data);
```
kamil-holubicki: shouldn't we lock before 'if'? Now we are reading 'thd->wsrep_skip_wsrep_hton' without protection

mysql-test/suite/galera/t/galera_kill_ddl_query.test
```diff
@@ -0,0 +1,611 @@
+#
+# This test checks wbat happens to TOI DDL when the thread is killed.
```
kamil-holubicki: s/wbat/what

sql/sql_parse.cc
```diff
@@ -7876,9 +7876,17 @@ static uint kill_one_thread(THD *thd, my_thread_id id, bool only_kill_query)
...
-        !wsrep_thd_is_BF((void *)tmp, true))
+    if (wsrep_thd_is_in_to_isolation(tmp)) {
+      /* There is nothing special about the use of ER_QUERY_INTERRUPTED,
+         I just needed a different error code from ER_KILL_DENIED_ERROR.
+
+         See sql_kill() to see the other half of the error processing.
+      */
+      error= ER_QUERY_INTERRUPTED;
+    }
+    else if (((thd->security_context()->check_access(SUPER_ACL)) ||
```
venkatesh-prasad-v: Looks like `&& !is_utility_connection)` is omitted. It is causing build failures
```
[ 87%] Building CXX object sql/CMakeFiles/sql.dir/mysqld.cc.o
/home/venki/work/pxc/57/sql/sql_parse.cc: In function ‘uint kill_one_thread(THD*, my_thread_id, bool)’:
/home/venki/work/pxc/57/sql/sql_parse.cc:7873:16: error: unused variable ‘is_utility_connection’ [-Werror=unused-variable]
     const bool is_utility_connection =
                ^~~~~~~~~~~~~~~~~~~~~
[ 87%] Building CXX object sql/CMakeFiles/sql.dir/mysqld_thd_manager.cc.o
```

Also, looks like this branch is not based on GCA. Please check it once.

mysql-test/suite/galera/t/galera_kill_ddl_query.test
venkatesh-prasad-v: s/wbat/what/

mysql-test/suite/galera/t/galera_kill_ddl_query.test
```diff
@@ -0,0 +1,611 @@
...
+# after the TOI has ended).
+#
+
+--source include/have_debug_sync.inc
+--source include/have_innodb.inc
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+--connection node_1
+call mtr.add_suppression("WSREP: TO isolation failed");
```
venkatesh-prasad-v: `CALL` is an SQL statement. Please change it to uppercase.

mysql-test/suite/galera/t/galera_kill_ddl_query.test
```diff
@@ -0,0 +1,611 @@
...
+--send CREATE TABLE t1 (ID INT PRIMARY KEY);
+
+# Kill the CREATE TABLE query
+--connection node_1b
+--echo [connection node_1b]
+
+# The KILL will fail because we are already in TOI
+--disable_query_log
+--echo # Killing query on node1a
+--error ER_KILL_DENIED_ERROR
```
venkatesh-prasad-v, line 46: This test case can fail sporadically when there is high load.
```
[  9%] galera.galera_kill_ddl_query             w12 [ fail ]
        Test ended at 2021-02-26 14:40:28

CURRENT_TEST: galera.galera_kill_ddl_query
mysqltest: At line 43: query 'KILL QUERY $node1a_connection_id' succeeded - should have failed with errno 1095...

The result from queries just before the failure was:
call mtr.add_suppression("WSREP: TO isolation failed");
################ Test 1 ################
[connection node_1]
FLUSH TABLES WITH READ LOCK;
[connection node_1a]
CREATE TABLE t1 (ID INT PRIMARY KEY);;
[connection node_1b]
# Killing query on node1a
safe_process[147562]: Child process: 147563, exit: 1
```
Can be reproduced with 
```
./mtr galera_kill_ddl_query{,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,} --mem --parallel=auto
```

sql/wsrep_mysqld.cc
```diff
@@ -2019,7 +2019,26 @@ static int wsrep_TOI_begin(THD *thd, const char *db_, const char *table_,
...
   int buf_err;
 
+  DEBUG_SYNC(thd, "wsrep_TOI_begin_before");
+  mysql_mutex_lock(&thd->LOCK_thd_data);
+
+  /*
+    Setting wsrep_skip_wsrep_hton=true has a side-effect that this
+    query/thread cannot be killed.
+  */
   thd->wsrep_skip_wsrep_hton= true;
```
venkatesh-prasad-v, line 2029: If we are taking `thd->LOCK_wsrep_thd` for reading, then shouldn't we acquire `thd->LOCK_wsrep_thd` for writing it as well?

sql/wsrep_mysqld.cc
```diff
@@ -2396,6 +2415,21 @@ int wsrep_to_isolation_begin(THD *thd, const char *db_, const char *table_,
   return ret;
 }
 
+
+bool wsrep_thd_is_in_to_isolation(THD *thd)
+{
+  bool status = FALSE;
+  if (thd)
```
venkatesh-prasad-v, line 2420: I think we can relax this `if (thd)` condition, as this function gets called only when we have a valid thd in the `Global_THD_manager`.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Should't this be `thd->LOCK_wsrep_thd`?

sql/wsrep_mysqld.cc
```diff
@@ -2019,7 +2019,26 @@ static int wsrep_TOI_begin(THD *thd, const char *db_, const char *table_,
   size_t buf_len(0);
   int buf_err;
 
+  DEBUG_SYNC(thd, "wsrep_TOI_begin_before");
```
venkatesh-prasad-v: I would suggest renaming the sync points to `wsrep_TOI_begin_before_wsrep_skip_hton` and `wsrep_TOI_begin_after_wsrep_skip_hton`.

Using just `wsrep_TOI_begin_before` and `wsrep_TOI_begin_after` makes the reader think that these points are in the beginning and in the end, but they are not.

sql/wsrep_mysqld.cc
```diff
@@ -2412,8 +2446,15 @@ void wsrep_to_isolation_end(THD *thd)
     wsrep_cleanup_transaction(thd);
   }
 
-  if (thd->wsrep_skip_wsrep_hton)
+  if (thd->wsrep_skip_wsrep_hton) {
+    DEBUG_SYNC(thd, "wsrep_to_isolation_end_before");
```
venkatesh-prasad-v: Likewise, I would suggest renaming the sync points to `wsrep_TOI_end_before_wsrep_skip_hton` and `wsrep_TOI_end_after_wsrep_skip_hton`.

mysql-test/suite/galera/t/galera_kill_ddl_query.test
kennt-percona: fixed

mysql-test/suite/galera/t/galera_kill_ddl_query.test
kennt-percona: fixed

mysql-test/suite/galera/t/galera_kill_ddl_query.test
kennt-percona: fixed.

sql/wsrep_mysqld.cc
kennt-percona: This is to match up with kill_one_thread() which uses LOCK_thd_data.

sql/wsrep_mysqld.cc
kennt-percona: Most of the calls that use a thd (like wsrep_thd_is_BF), check the thd ptr.

sql/wsrep_mysqld.cc
kennt-percona: Actually, should probably change the read to use LOCK_thd_data

sql/wsrep_mysqld.cc
kennt-percona: agree, name changed

sql/wsrep_mysqld.cc
kennt-percona: agree, name changed

kennt-percona:
https://pxc.cd.percona.com/job/pxc-5.7-param/131/

sql/sql_parse.cc
```diff
@@ -7808,7 +7808,15 @@ static uint kill_one_thread(THD *thd, my_thread_id id, bool only_kill_query)
     */
 
 #ifdef WITH_WSREP
-    if (((thd->security_context()->check_access(SUPER_ACL)) ||
+    if (wsrep_thd_is_in_to_isolation(tmp, false)) {
```
venkatesh-prasad-v: If it is not too much to ask, can we make it uniform with the rest of the file?

sql/wsrep_mysqld.cc
```diff
@@ -2345,6 +2364,20 @@ int wsrep_to_isolation_begin(THD *thd, const char *db_, const char *table_,
...
+bool wsrep_thd_is_in_to_isolation(THD *thd, bool flock)
+{
+  bool status = false;
+  if (thd)
+  {
+    if (flock) mysql_mutex_lock(&thd->LOCK_thd_data);
+
+    status = thd->wsrep_skip_wsrep_hton;
+
+    if (flock) mysql_mutex_unlock(&thd->LOCK_thd_data);
```
venkatesh-prasad-v, line 2426: Just a thought, (not a blocker).

If we are protecting `thd->wsrep_skip_wsrep_hton` with `thd->LOCK_thd_data`, then we probably should acquire it here as well. 

https://github.com/percona/percona-xtradb-cluster/blob/5.7/sql/wsrep_hton.cc#L92-L93

Or may be it is not required since writing and reading happens in the same thread.

sql/wsrep_mysqld.cc
```diff
@@ -2361,8 +2394,15 @@ void wsrep_to_isolation_end(THD *thd)
     wsrep_cleanup_transaction(thd);
   }
 
-  if (thd->wsrep_skip_wsrep_hton)
-    thd->wsrep_skip_wsrep_hton= false;
+  if (thd->wsrep_skip_wsrep_hton) {
```
venkatesh-prasad-v: If it is not too much to ask, can we make it uniform with the rest of the file?

sql/wsrep_mysqld.cc
kamil-holubicki: This is not addressed.

kamil-holubicki (changes requested):

venkatesh-prasad-v (commented):
Hi Kenn,

Thank you for working on this issue. Nice work on the test case!

If you don't mind, please add the JIRA number, title and the link in the commit message so that we can track the commit with the JIRA issue.

The patch looks good now and I do not have any major comments except a few minor comments. Please see below.

sql/wsrep_mysqld.cc
```diff
@@ -2361,8 +2394,15 @@ void wsrep_to_isolation_end(THD *thd)
     wsrep_cleanup_transaction(thd);
   }
 
-  if (thd->wsrep_skip_wsrep_hton)
-    thd->wsrep_skip_wsrep_hton= false;
+  DEBUG_SYNC(thd, "wsrep_to_isolation_end_before_wsrep_skip_wsrep_hton");
+  mysql_mutex_lock(&thd->LOCK_thd_data);
+
+  if (thd->wsrep_skip_wsrep_hton) {
```
kamil-holubicki: I just realized that this simply evaluates to 
`thd->wsrep_skip_wsrep_hton = false;
`
No 'if' needed :)

kamil-holubicki (approved):
lgtm

sql/wsrep_mysqld.cc
kennt-percona: fixed

kennt-percona:
Also, Some of the tests were removed. This is due to the wsrep_to_isolation_end() can be called in non-TOI flows (this is 5.7 only, 8.0 fixes this).  Will come back later to fix these test cases.

PR #1383, opened 2021-02-23
PXC-3594 Develop Detailed Documentation for the Inconsistent Voting F…

patrickbirch:
…eature merged from Galera (8.0)

new file:   source/features/data-inconsistency
	modified:   source/index

### DIFF
--- a/doc/source/features/cev.rst
+++ b/doc/source/features/cev.rst
@@ -1,0 +1,81 @@
+.. _cev:
+
+====================================================
+Cluster Error Voting 
+====================================================
+
+Implemented in |PXC| 8.0.21, Cluster Error Voting (CEV) is a protocol that allows a cluster to handle replication problems with data inconsistency. The protocol is active in a cluster with any number of nodes. Data inconsistency is a failure of data integrity, such as the cluster replicating unplanned data changes. Therefore, detecting data inconsistency is always after the fact. 
+
+For example, if a node fails to apply a transaction in a five-node cluster, this node initiates a vote within the cluster. The cluster removes any nodes in the minority and continues to function. A DBA manually fixes the issue, and the minority nodes rejoin the cluster.
+
+In the source-replica topology, replication is one-way. Table updates on the source are replicated to the replica table, and replica updates are not replicated back to the source. If the clients use reads from the replica nodes for writes to the source, this action can cause inconsistency. 
+
+In the multi-master topology, a node can be both source and replica; the client interacts with any node. Thus, if one part of the cluster has inconsistent data, this data spreads throughout.
+
+Reasons for Inconsistency
+--------------------------
+
+There are many ways a cluster node can have inconsistent data. They are the following: 
+
+* Missing data in State Snapshot Transfer
+
+* Incorrect cluster variable settings
+
+* Issue in the client code
+
+:ref:`state_snapshot_transfer` (SST) is a full data copy from one node (donor) to the joining node (joiner). The SST script may miss a transaction on the donor, and this transaction is not replicated to the joiner and create an inconsistency. This inconsistency may cause either a node to fail to start or be unnoticed for a long time. 
+
+Incorrect cluster variable settings or incorrect replication filtering can also cause inconsistency. The following actions are examples:
+
+* Changing the ``wsrep_on`` variable to ``OFF``. As a result, workloads executed in that session are not replicated to other nodes.
+
+* Setting the session variable ``sql_log_bin`` to ``OFF`` disables binary logging for one node. This ``OFF`` setting also prevents binary log transactions from being assigned global transaction identifiers (GTID). If replication uses GTIDs, and the variable is reset to ``ON``, any transactions without GTIDs are lost.
+
+* Replication filters let you configure the replicas to skip selected events. If these filters are misused, they may cause inconsistency. For example, the ``binlog-do-db`` option configures a source to filter transactions that match the criteria from being written to the binary log. As a result the replicas do not see those transactions.
+
+An issue in the client code or scripts can cause data inconsistency. 
+
+Before implementing :ref:`cev`, the node with an issue during a transaction reports the error in the log and exits the cluster, the other nodes in the cluster report that node as ``out of the view``. 
+
+
+Cluster Error Voting
+-----------------------------------
+
+Using the CEV
+
+With CEV, a node that encounters a data inconsistency, the node returns an error message instead of immediately exiting the cluster. Then, the sender initiates voting. Each node informs the cluster about the operation's success or failure. 
+
+Voting can cause the following actions:
+
+* A simple majority with the same result continues. The node or nodes with different results perform an emergency shutdown. 
+
+* If there is no majority, then success wins. 
+
+* If none of the nodes have a success event, then another node is selected to continue.
+
+.. note::
+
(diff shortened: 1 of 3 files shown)


doc/source/features/data-inconsistency.rst
```diff
@@ -0,0 +1,37 @@
...
+
+* Bad SST
+
+* Cluster misuse
+
+* Issue in the code
+
+Snapshot State Transfer (SST) is part of the method of adding a new node. If one transaction is missing on the Donor, one transaction is missing on the Joiner. The inconsistent data can cause a node to not start or the error can be unnoticed for months or years. SST is built from a script. If someone modifies this script, this modification can cause inconsistent data because of an issue in the code.
+
+Cluster misuse may also cause inconsistency. If ``wsrep_on``is ``OFF`` for one node, that node does not replicate the changes. Turning off ``sql_log_bin`` disables logging for the session in one node. With no binlog events, there is nothing to replicate. Replication filtering, such as ``binlog-do-db`` or ``replicate-do-db``, if misused can cause inconsistency.
```
nastena1606:
```suggestion
Cluster misuse may also cause inconsistency. If ``wsrep_on`` is ``OFF`` for one node, that node does not replicate the changes. Turning off ``sql_log_bin`` disables logging for the session in one node. With no binlog events, there is nothing to replicate. Replication filtering, such as ``binlog-do-db`` or ``replicate-do-db``, if misused can cause inconsistency.
```

doc/source/features/data-inconsistency.rst
```diff
@@ -0,0 +1,37 @@
...
+
+====================================================
+Data Inconsistency
+====================================================
+
+Clusters can have issues with data consistency.
+
+In one type of source and replica topology, the clients read and write to the source, and the replicas are read-only. If the clients read from the replica nodes and this information is used for writes to the source, data inconsistency may occur. This issue may also occur with the multi-master topology.
+
+In the Multi-master topology, the client can connect to any node. The effect of inconsistency is immediate. If a node has bad data, all reads propagate to bad writes. When part of the cluster has bad data, this data spreads throughout the cluster, and you cannot trust any data. Any replication error could be a sign of data inconsistency. Detecting data inconsistency is always after the fact.
```
nastena1606: You have multi-master above and Multi-master here. I think these must be unified to one style

patrickbirch: fixed

doc/source/features/data-inconsistency.rst
patrickbirch: fixed

doc/source/features/data-inconsistency.rst
```diff
@@ -0,0 +1,47 @@
...
+Reasons for Inconsistency
+--------------------------
+
+There are several ways a node in a cluster can have inconsistent data. They are the following: 
+
+* Missing data in State Snapshot Transfer
+
+* Incorrect cluster variable settings
+
+* Issue in the code
```
dutow: Do you mean user or server code here?

dutow (commented):
So far this is good, but I think we should write more about how this works specifically in PXC, as the current doc is mostly generic.

We could describe things such as:
* when inconsistency voting will be triggered (failing to apply slave writeset; DDL error even on single node clusters)
* how voting works (you described the default majority. But there are edge cases (no majority), and this can also be changed by two variables, wsrep_ignore_apply_errors and gcs.vote_policy)


And optionally about (not strictly about the feature, but related - I'm not sure if we have anything about these in our docs?)
* what a user could do to avoid inconsistencies
* what a user should do when a node shuts down because of some inconsistency

doc/source/features/data-inconsistency.rst
patrickbirch: Fixed.

doc/source/features/cev.rst
```diff
@@ -0,0 +1,81 @@
...
+
+There are many ways a cluster node can have inconsistent data. They are the following: 
+
+* Missing data in State Snapshot Transfer
+
+* Incorrect cluster variable settings
+
+* Issue in the client code
+
+:ref:`state_snapshot_transfer` (SST) is a full data copy from one node (donor) to the joining node (joiner). If someone modifies the SST script, a code issue can cause a missing transaction on the donor, and this transaction is also absent on the joiner. The absent transaction may either cause a node to fail to start, or be unnoticed for a long period of time.
```
dutow: "code issue can cause a missing transaction on the donor, and this transaction is also absent on the joiner" - you mean that the SST script may miss an existing transaction on the donor, and because of this, it won't be replicated to the joiner? The transaction will still exist on the donor, it will be missing only from the joiner, and that's the inconsistency.

doc/source/features/cev.rst
```diff
@@ -0,0 +1,81 @@
...
+
+* Incorrect cluster variable settings
+
+* Issue in the client code
+
+:ref:`state_snapshot_transfer` (SST) is a full data copy from one node (donor) to the joining node (joiner). If someone modifies the SST script, a code issue can cause a missing transaction on the donor, and this transaction is also absent on the joiner. The absent transaction may either cause a node to fail to start, or be unnoticed for a long period of time. 
+
+Incorrect cluster variable settings or incorrect replication filtering can also cause inconsistency. The following actions are examples:
+
+* Making a schema change or performing an upgrade, the DBA sets ``wsrep_on`` to ``OFF`` on node 2 while this node is live. If the cluster continues to receive workloads, these workloads are not replicated to node 2. When the DBA sets ``wsrep_on`` to ``ON``, the cluster is healthy. The error is not found until someone queries a table on node 2 and discovers missing data or extra rows.
```
dutow: With PXC, wsrep_on is a session variable, not global (as in codership/mariadb). You can't disable it globally, and disabling it for a single session won't cause issues for other sessions.

You can disable it for a session, and change data on that session, but that's a more restricted case than what you describe here.

doc/source/features/cev.rst
```diff
@@ -0,0 +1,81 @@
...
+* Setting the session variable ``sql_log_bin`` to ``OFF`` on one node disables binary logging for one node. This ``OFF`` setting also prevents binary log transactions from being assigned global transaction identifiers (GTID). If replication uses GTIDs, and the variable is reset to ``ON``, any transactions without GTIDs are lost.
+
+* Replication filters lets you configure the replicas to skip selected events. If these filters are misused, they may cause inconsistency. For example, the ``binlog-do-db`` option configures a source to filter transactions that match the criteria from being written to the binary log. The replicas do not see those transactions.
+
+An issue in the client code or scripts can cause data inconsistency. 
+
+Before the implementation of CEV, the node with an issue during a transaction reports the error in the log and exits the cluster. The other nodes in the cluster report that node as out of the view. 
+
+
+Cluster Error Protocol
```
dutow: Cluster Error Voting?

doc/source/features/cev.rst
patrickbirch: Rewritten.

doc/source/features/cev.rst
patrickbirch: Fixed.

doc/source/features/cev.rst
patrickbirch: Updated based on Slack discussion.

PR #1385, opened 2021-02-26
PXC-3592: Make wsrep_provider and wsrep_notify_cmd read only

dutow:


dutow:
https://pxc.cd.percona.com/view/PXC%205.6/job/pxc-5.6-param/121/

mysql-test/suite/galera/r/galera_ist_rsync.result
```diff
@@ -39,7 +38,7 @@ INSERT INTO t1 VALUES ('node1_to_be_rollbacked_after');
 INSERT INTO t1 VALUES ('node1_to_be_rollbacked_after');
 INSERT INTO t1 VALUES ('node1_to_be_rollbacked_after');
 INSERT INTO t1 VALUES ('node1_to_be_rollbacked_after');
-Loading wsrep provider ...
+SET GLOBAL wsrep_cluster_address = 'gcomm://127.0.0.1:13002';;
```
kamil-holubicki: 1. double semicolon (multiple occurences)
2. Won't having an explicit port number here cause problems when running tests in parallel?

Both will be solved by fixing galera_load_provider.inc

mysql-test/suite/galera/include/galera_load_provider.inc
```diff
@@ -1,8 +1,3 @@
---echo Loading wsrep provider ...
-
---disable_query_log
```
kamil-holubicki, line 3: this line was removed unnecessarily.

kamil-holubicki (changes requested):

mysql-test/suite/galera/include/galera_load_provider.inc
dutow: Good catch and this causes port 13002 to appear, I'll fix this.

dutow: Fixed.

kamil-holubicki (approved):
lgtm

PR #1392, opened 2021-03-10
PXC-3554: Fixed handling multiple servers specified in bind-address

kamil-holubicki:
https://jira.percona.com/browse/PXC-3554

1. WSREP_ERROR macro fixed. Added missing check for Wsrep_server_state
instance before using it.

2. MySQL >= 8.0.13 allows multiple IPs to be specified as 'bind-address'
parameter value. This parameter is used by wsrep part to guess node
address, however implementation assumed a single IP value. Implementation
was fixed to choose the first IP in case

### DIFF
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -835,10 +835,19 @@ static std::string wsrep_server_incoming_address() {
        !strcmp(wsrep_node_incoming_address, WSREP_NODE_INCOMING_AUTO))) {
     bool is_ipv6 = false;
     unsigned int my_bind_ip = INADDR_ANY;  // default if not set
+    char *single_addr = nullptr;
 
     if (my_bind_addr_str && strlen(my_bind_addr_str) &&
         strcmp(my_bind_addr_str, "*") != 0) {
-      my_bind_ip = wsrep_check_ip(my_bind_addr_str, &is_ipv6);
+      size_t single_addr_buf_size = strlen(my_bind_addr_str) + 1;
+      single_addr =
+          (char *)my_malloc(key_memory_wsrep, single_addr_buf_size, MYF(0));
+      if (!single_addr) goto done;
+
+      wsrep_get_single_address(my_bind_addr_str, single_addr,
+                               single_addr_buf_size);
+
+      my_bind_ip = wsrep_check_ip(single_addr, &is_ipv6);
     }
 
     if (INADDR_ANY != my_bind_ip) {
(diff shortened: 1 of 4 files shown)


sql/wsrep_mysqld.cc
```diff
@@ -835,10 +835,18 @@ static std::string wsrep_server_incoming_address() {
        !strcmp(wsrep_node_incoming_address, WSREP_NODE_INCOMING_AUTO))) {
     bool is_ipv6 = false;
     unsigned int my_bind_ip = INADDR_ANY;  // default if not set
+    char *single_addr = nullptr;
 
     if (my_bind_addr_str && strlen(my_bind_addr_str) &&
         strcmp(my_bind_addr_str, "*") != 0) {
-      my_bind_ip = wsrep_check_ip(my_bind_addr_str, &is_ipv6);
+      size_t single_addr_buf_size = strlen(my_bind_addr_str) + 1;
+      single_addr = (char *)malloc(single_addr_buf_size);
```
venkatesh-prasad-v: Here and for all occurences in this patch:

We need to instrument every dynamic memory allocation so that it can be observed through perfschema.
Please use `my_malloc` instead of `malloc` and `my_free` instead of `free`
```suggestion
      single_addr = (char *) my_malloc(key_memory_wsrep, single_addr_buf_size, MYF(0));
```

venkatesh-prasad-v (commented):
Hi Kamil,

Thank you for working on this. 

The patch looks great and I do not have any comments other than the one mentioned below.

sql/wsrep_mysqld.cc
kamil-holubicki: I have changed it to my_malloc, but as we discussed on Slack I don't see a big advantage here.
my_malloc add instrumentation can be useful in non-trival scenarios (we allocate memory here, free it somewhere else, etc). Here we do malloc/do_quick_simple_task/free with no possibility to malloc but not free. I'm not sure if there is any advantage in instrumenting such allocations. It is comparable to having  std::string object on the stack, which memory is not instrumented as well. 

Additionally in my opinion it should not be the rule of thumb that we should always use my_malloc over malloc.
Internally my_malloc/my_free  use atomics, so if we use them in places executed in parallel by several threads we will pay the performance penalty. In such cases, it is better to use malloc/free. my_malloc/my_free should be used for the memory we want to track (e.g. not trivial allocation/free patterns), not everywhere.

venkatesh-prasad-v (approved):
LGTM

PR #1396, opened 2021-03-16
[5.7] PXC-3226: Results from CHECK, ANALYZE from PXC server is causing the client libraries to crash

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3226

Problem
-------
PXC node can send malformed packets to client and can cause client to fail with
asserion

  `!check_buffer || (vio_pending(net->vio) <= 1)' in net_clear().

Background
----------
For any query that sends result set to the client, the server does the following
things as per classic protocol.

1. Result metadata specifying the number of fields in the result set
   i.e, `protocol->start_result_metadata()`.
2. The field metadata specifying the types of fields in the result set
   i.e, `protocol->send_field_metadata()` followed by
   `protocol->end_result_metadata()`.
3. Actual row data (any data sent between `protocol->start_row()` and
   `protocol->end_row()`
4. An OK/EOF packet indicating the end of result set.
   `protocol->send_eof()` / `protocol->send_ok()` usually called from
   `THD::send_statement_status()` in the end of `dispatch_command()`

Analysis
--------
When the server is running in autocommit mode and when it is executing a query
that sends some result set to client programs and it was BF aborted by TOI or
high priority transactions, the `wsrep_retry_autocommit` mechanism comes into
effect and the server retries the autocommit query withouy returning the error
to the client.

However, when the server is retrying the query, it is possible that the client
program may have already received partial result from server and may have been
already waiting for the OK/EOF packet to report it to the user (i.e, Steps 1-3
are over and waiting for the step 4 to happen).

In such a scenario, when a retry is performed, the server shall start executing
from Step-1 to Step-4 and if the query execution is successful, it sends OK/EOF
packet in the end to indicate that the query is complete. But on the client
side, this causes the client program to receive unexepected result metadata in
place of an OK/EOF packet and thus causes the client to error out with Malformed
packet error.

Note:
This issue was mostly seen with CHECK TABLE query when it's execution was
interrupted by a TOI. However, this issue is not seen with commands that run
in TOI. So, we can infer that this can happen on all queries that return result,
cannot run in TOI and can be killed by a PXC.

Fix
---
Disable wsrep_retry_autocommit mechanism for CHECK TABLE and instead report
ER_DEADLOCK_ERROR.

Note: As of now, we only disable it for CHECK TABLE and if the problem exists for
other commands, we can add them later.

Testing Done
---
Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/135/testReport/
Failing Tests: 
1. galera.galera_alter_table_big -- Unrelated to this fix

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -7835,6 +7835,39 @@ static uint kill_one_thread(THD *thd, my_thread_id id, bool only_kill_query)
 }
 
 #ifdef WITH_WSREP
+static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command)
+{
+  /*
+    We are here could mean that the query resulted in a cluster-wide
+    conflict and had to be aborted. While it happened, it is possible that
+    the client may have already received partial data from server and may
+    have been waiting for the OK/EOF packet (usually sent by
+    THD::send_statement_status() in dispatch_command()) to report it to the
+    user.
+
+    When retry is performed in such a case, the server shall start sending
+    result and field metadata once again and this would cause the client
+    program to receive unexpected metadata information in place of an
+    OK/EOF packet and thus causes the client to error out with Malformed
+    packet error.
+
+    So, we avoid retries for such queries that return result set to client,
+    but cannot be run in TOI and can be killed by a TOI.
+
+    As of now, we only do this check for CHECK TABLE and SELECT, and if the
+    same symptom is found for other commands, then please add it to the
+    below list.
+  */
+  switch (sql_command)
+  {
+    case SQLCOM_CHECK:
+    case SQLCOM_SELECT:
+      return false;
+    default:
+      return true;
+  }
+}
+
 static void wsrep_mysql_parse(THD *thd, const char *rawbuf, uint length,
                               Parser_state *parser_state, bool update_userstat)
 {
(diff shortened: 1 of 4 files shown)


sql/sql_parse.cc
```diff
@@ -7835,6 +7835,39 @@ static uint kill_one_thread(THD *thd, my_thread_id id, bool only_kill_query)
...
+    packet error.
+
+    So, we avoid retries for such queries that return result set to client,
+    but cannot be run in TOI and can be killed by a TOI.
+
+    As of now, we only do this check for CHECK TABLE and SELECT, and if the
+    same symptom is found for other commands, then please add it to the
+    below list.
+  */
+  switch (sql_command)
```
kamil-holubicki, line 7861: The ticket mentions also problems with ANALYZE TABLE. Maybe worthy adding here as well?

venkatesh-prasad-v: Even though ticket mentions about ANALYZE TABLE, I was never able to see the crash with ANALYZE TABLE. This is because ANALYZE/OPTIMIZE TABLE are replicated as TOI and can never get bf aborted by applier threads (as the whole cluster will be isolated), but problem was seen only with CHECK TABLE since is not replicated to the cluster. So, it is not required here.

kamil-holubicki (approved):

PR #1397, opened 2021-03-17
Fixed PXC-3596 (Node stuck in aborting SST)

altmannmarcelo:
https://jira.percona.com/browse/PXC-3596

Problem:
SST xtrabackup process happens with an unlimited timeout on joiner side.
In case the donor crash or has a network issue it can trigger two issues
on joiner:
1) If the connection has already been establish to socat/nc, we will
relay on TCP layer to close the connection, this can vary depending on
the nature of the network issue and OS/Kernel configuration.
2) Socat will start to listen on the SST port and will block until a new
connection is established. If donor leaves the cluster after joiner has
started to listen but before the connection has been established, joiner
socat will hang forever.

Solution:
Implement a timeout function that will use the size of .sst directory
as metric to validate if we are suffering a stall on SST. In case of
data size has not moved since the last 5 seconds, we enter stall mode.
In stall mode we keep checking the size every second, until we reach
sst-timeout seconds (default to 120 seconds). When timeout is reached,
SST is aborted on joiner side.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_failure.test
+++ b/mysql-test/suite/galera/t/galera_sst_failure.test
@@ -40,6 +40,55 @@ SET GLOBAL debug = "+d,wsrep_sst_donate_cb_fails";
 # cleanup
 SET GLOBAL debug = "-d,wsrep_sst_donate_cb_fails";
 
+# Setup node1 to skip sending backup on next SST
+SET GLOBAL debug = "+d,wsrep_sst_donor_skip";
+
+# Restart node 2 forcing SST
+--connection node_2
+
+# Remove the grastate.dat file to force an SST
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+# Adjust SST timeout
+--exec cp $MYSQLTEST_VARDIR/my.cnf $MYSQLTEST_VARDIR/my_sst.cnf
+append_file $MYSQLTEST_VARDIR/my_sst.cnf;
+[sst]
+sst-idle-timeout=20
+EOF
+
+# Start node_2
+--let $error_log= $MYSQLTEST_VARDIR/tmp/test_error_log.err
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/136/

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -2107,6 +2159,8 @@ then
 
         XB_GTID_INFO_FILE_PATH="${DATA}/${XB_GTID_INFO_FILE}"
         wsrep_log_info "............Waiting for SST streaming to complete!"
+        ssttimeout=$(parse_cnf sst sst-timeout 120)
```
kamil-holubicki: A few lines above we have 
recv_data_from_donor_to_joiner "${KEYRING_FILE_DIR}" "${stagemsg}-keyring" 0 2
which can be stuck in the same way. Probably we should monitor there as well.

mysql-test/suite/galera/t/galera_sst_failure.test
```diff
@@ -40,6 +40,55 @@ SET GLOBAL debug = "+d,wsrep_sst_donate_cb_fails";
...
+--connection node_2
+
+# Remove the grastate.dat file to force an SST
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+# Adjust SST timeout
+--exec cp $MYSQLTEST_VARDIR/my.cnf $MYSQLTEST_VARDIR/my_sst.cnf
+append_file $MYSQLTEST_VARDIR/my_sst.cnf;
+[sst]
+sst-timeout=20
```
kamil-holubicki: Proposal: name it sst-idle-timeout or sst-inactivity-timeout. Now it suggests that it is the timeout for the whole SST to finish.

kamil-holubicki (approved):
lgtm

PR #1398, opened 2021-03-22
PXC-3559: Fix unstable galera_sr.GCF-1043B test

kamil-holubicki:
https://jira.percona.com/browse/PXC-3559

Problem:
Assertion is triggered when empty local transaction is BF aborted by
replicated transaction (originally reproduced by galera_sr.GCF-1043B
test ran in parallel)

Cause:
We removed check for TRX_FORCE_ROLLBACK_DISABLE from 'if' condition in
lock_make_trx_hit_list(), so now HP transaction can BF other transaction
even if the flag is set.
It is OK in general case if local transaction is not empty (so it is
replicated and certified). If it conflicts with HP transaction it will
not be certified (so will be discarded), and if it was certified, there
is no possibility for other transaction to BF it because that other
transaction should fail certification.
If local transaction is the empty transaction, it is not replicated
and certified as it does not generate any writestes. So it follows the
normal transaction path. But empty transaction can get some locks.
If at the same time there is replicated transaction attempting to get
the same locks, it can BF local transaction at any stage, even if local
transaction is commiting. (note that it can happen only for empty local
transaction)

Solution:
Allow trx_commit_for_mysql() to return DB_FORCED_ABORT.

### DIFF
--- a/mysql-test/suite/galera/t/kill_empty_transaction.test
+++ b/mysql-test/suite/galera/t/kill_empty_transaction.test
@@ -1,0 +1,45 @@
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+--connect node_2a, 127.0.0.1, root, , test, $NODE_MYPORT_2
+
+--connection node_1
+--echo #node_1
+CREATE TABLE test_table (f1 INTEGER PRIMARY KEY AUTO_INCREMENT, f2 CHAR(255)) ENGINE=InnoDB;
+
+--connection node_2
+--echo #node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.INNODB_TABLES WHERE NAME LIKE 'test/test_table';
+--source include/wait_condition.inc
+
+SET @@GLOBAL.debug = "+d,pause_commit_after_update_single_table";
+
+--send UPDATE test_table SET f2 = 'y' WHERE f1 = 1 OR f1 = 2;
+
+--connection node_2a
+--echo #node_2a
+SET DEBUG_SYNC = "now WAIT_FOR update_commit_waiting";
+# do not pause next time when retry autocommit takes place
+SET @@GLOBAL.debug = "-d,pause_commit_after_update_single_table";
+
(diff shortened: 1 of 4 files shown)


mysql-test/suite/galera/t/kill_empty_transaction.test
```diff
@@ -0,0 +1,40 @@
...
+SET DEBUG_SYNC = "now WAIT_FOR update_commit_waiting";
+
+--connection node_1
+--echo #node_1
+INSERT INTO test_table VALUES (1,'y'),(2,'x');
+
+# We do not need to signal innobase_commit_low_begin point to continue, as the
+# thread waiting on it was BF killed, and threads waiting on debug sync points
+# were already continued automatically. 
+# Also nothing to reap from node_2 connection
```
venkatesh-prasad-v: Not a review comment though, I was testing this and I was surprised to see that the test fails with 
```
CURRENT_TEST: galera.kill_empty_transaction
mysqltest: At line 34: Cannot run query on connection between send and reap
```
when I execute any query on node_2
```
--connection node_2
--echo #node_2
SHOW TABLES;
```

But the same test hangs when I try to reap from connection `node_2`.

Did you also see this behavior? Shouldn't it fail with `ER_LOCK_DEADLOCK` when we try to reap the connection?

kamil-holubicki: Hi @venkatesh-prasad-v . I have improved the test. The reason of hanging reap was pause_commit_after_update_single_table still being active which was unnecessarily blocking test execution when retry autocommit took place.

As for ER_LOCK_DEADLOCK, no, retry will succeed as there is nothing preventing update to take place
Query OK, 1 row affected (37.81 sec)
Rows matched: 2  Changed: 1  Warnings: 0

venkatesh-prasad-v (approved):
Great work on this, Kamil! Approved!

Just a note that the `rpl_late_abort` testcase was also hitting the same assert mentioned in the description. But after this change, it is now failing with timeout and doesn't crash anymore. This will be handled in PXC-3555.

PR #1402, opened 2021-03-24
PXC-3551: enforce_storage_engine doesn't work on startup with PXC

dutow:
Issue: enforce_storage_engine ignores some tables, but wsrep tables
aren't on the list. Because of this, enforcing any other engine than
innodb will prevent the creation of wsrep tables on startup.

Fix: add an exception for wsrep tables.

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -18893,6 +18893,7 @@ static bool check_engine(THD *thd, const char *db_name, const char *table_name,
       1. for "OPTIMIZE TABLE" statements.
       2. for "ALTER TABLE" statements without explicit "... ENGINE=xxx" part
       3. Transactional data dictionary (DD) tables
+      4. WSREP tables
     */
     bool no_substitution = (!is_engine_substitution_allowed(thd));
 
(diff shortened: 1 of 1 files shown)


sql/sql_table.cc
```diff
@@ -18893,14 +18893,17 @@ static bool check_engine(THD *thd, const char *db_name, const char *table_name,
...
     */
     bool no_substitution = (!is_engine_substitution_allowed(thd));
 
     bool enforcement_forbidden =
         ((thd->lex->sql_command == SQLCOM_ALTER_TABLE) &&
          (create_info->used_fields & HA_CREATE_USED_ENGINE) == 0) ||
         (thd->lex->sql_command == SQLCOM_OPTIMIZE) ||
-        dd::get_dictionary()->is_dd_table_name(db_name, table_name);
+        dd::get_dictionary()->is_dd_table_name(db_name, table_name) ||
+        (strncmp(db_name, "mysql", 6) == 0 &&
```
venkatesh-prasad-v: Can we make use of an already existing `is_wsrep_system_table()` function here?

Ref: https://github.com/dutow/percona-xtradb-cluster/blob/pxc3551/sql/table.h#L4096-L4103

dutow: Updated.

venkatesh-prasad-v (approved):
Thanks for addressing the review comments, Approved!

kamil-holubicki (approved):
lgtm

PR #1403, opened 2021-03-24
PXC-3547: Fix SSL test failures

dutow:
Issue: ssl related MTR tests were failing because of missing ssl status
variables. These variables were likely missed during a PS->PXC merge.

Fix: re added these variables based on the PS source.


venkatesh-prasad-v (approved):
Nice find, Zsolt! Approved!!

kamil-holubicki (approved):
lgtm

PR #1406, opened 2021-04-02
Fixed PXC-3596 (Node stuck in aborting SST) - 8.0

altmannmarcelo:
https://jira.percona.com/browse/PXC-3596

Problem:
SST xtrabackup process happens with an unlimited timeout on joiner side.
In case the donor crash or has a network issue it can trigger two issues
on joiner:
1) If the connection has already been establish to socat/nc, we will
relay on TCP layer to close the connection, this can vary depending on
the nature of the network issue and OS/Kernel configuration.
2) Socat will start to listen on the SST port and will block until a new
connection is established. If donor leaves the cluster after joiner has
started to listen but before the connection has been established, joiner
socat will hang forever.

Solution:
Implement a timeout function that will use the size of .sst directory
as metric to validate if we are suffering a stall on SST. In case of
data size has not moved since the last 5 seconds, we enter stall mode.
In stall mode we keep checking the size every second, until we reach
sst-timeout seconds (default to 120 seconds). When timeout is reached,
SST is aborted on joiner side.

Also fixed an issue at cleanup_joiner code where if the sst temp dir
were created at datadir, it wouldn't be cleaned up after sst has
failed.

(cherry picked from commit 616a3b5364672f7f1ddca74b706e4d21fc3373e9)

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -270,6 +270,51 @@ interruptable_timeout() {
     echo ${piperesult[@]}
 }
 
+# Monitor progress of SST and kill the process if stalled.
+monitor_sst_progress() {
+  local tmpsstdir=$1
+  shift
+  local pid=$1
+  shift
+  local timeout=$1
+  shift
+  local current_timeout=0
+  local previous_size=0
+  local sleep=5
+  if [[ ${timeout} -eq 0 ]]; then
+    return;
+  fi
+
+  while true; do
+    kill -0 $pid 2> /dev/null
+    if [[ $? -ne 0 ]]; then
+      break;
+    fi
+    if [[ ! -d ${tmpsstdir} ]]; then
+      break;
+    fi
+
+    current_size=$(du --max-depth=1 ${tmpsstdir} | tail -n 1 | awk '{print $1}')
+    if [[ ${current_size} -eq  ${previous_size} ]]; then
+      current_timeout=$((current_timeout + 1))
+      sleep=1
+    else
+      # Reset timeout and set sleep back to 5 seconds.
+      current_timeout=0
+      sleep=5
+      previous_size=${current_size}
+    fi
+
+    if [[ ${current_timeout} -eq  ${timeout} ]]; then
+      wsrep_log_error "Killing SST ($pid) with SIGKILL after stalling for ${current_timeout} seconds"
+      pkill -SIGKILL -P ${pid} 2> /dev/null
+      break;
+    fi
+
+    sleep ${sleep};
+  done
+}
+
 #
 # If the ssl_dhparams variable is already set, uses that as a source
 # of dh parameters for OpenSSL. Otherwise, looks for dhparams.pem in the
@@ -842,6 +887,9 @@ cleanup_joiner()
     if [[ -n $progress && -p $progress ]]; then
         wsrep_log_debug "Cleaning up fifo file $progress"
         rm $progress
+    fi
+    if [[ -n "${JOINER_SST_DIR}" && -z "$WSREP_LOG_DEBUG" ]]; then
+      [[ -d "${JOINER_SST_DIR}" ]] && rm -rf "${JOINER_SST_DIR}" || true
     fi
     if [[ -n "${tmpdirbase}" ]]; then
         [[ -d "${tmpdirbase}" ]] && rm -rf "${tmpdirbase}" || true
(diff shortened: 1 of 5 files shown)


altmannmarcelo:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/201/

dutow (approved):

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -270,6 +270,51 @@ interruptable_timeout() {
...
 
+# Monitor progress of SST and kill the process if stalled.
+monitor_sst_progress() {
+  local tmpsstdir=$1
+  shift
+  local pid=$1
+  shift
+  local timeout=$1
+  shift
+  currenttimeout=0
```
kennt-percona: These three variables should be local if possible.

kennt-percona: Also, it would be good to use underscore in the names, 
  currenttimeout -> current_timeout
  previoussize -> previous_size
to make it easier to read.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -2079,7 +2135,10 @@ then
                     fi
 
                     XB_DONOR_KEYRING_FILE_PATH="${KEYRING_FILE_DIR}/${XB_DONOR_KEYRING_FILE}"
-                    recv_data_from_donor_to_joiner "${KEYRING_FILE_DIR}" "${stagemsg}-keyring" 0 2
+                    recv_data_from_donor_to_joiner "${KEYRING_FILE_DIR}" "${stagemsg}-keyring" 0 2 &
+                    jpid=$!
+                    monitor_sst_progress ${KEYRING_FILE_DIR} $jpid $ssttimeout &
```
kennt-percona: ${KEYRING_FILE_DIR} -> "${KEYRING_FILE_DIR}"  (add quotes around the path)

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -2160,6 +2219,7 @@ then
 
         XB_GTID_INFO_FILE_PATH="${DATA}/${XB_GTID_INFO_FILE}"
         wsrep_log_info "............Waiting for SST streaming to complete!"
+        monitor_sst_progress ${JOINER_SST_DIR} $jpid $ssttimeout &
```
kennt-percona: ${JOINER_SST_DIR} -> "${JOINER_SST_DIR}"   (add quotes around the path)

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -843,6 +888,9 @@ cleanup_joiner()
         wsrep_log_debug "Cleaning up fifo file $progress"
         rm $progress
     fi
+    if [[ -n "${JOINER_SST_DIR}" ]]; then
```
kennt-percona: Need to think about this.  If an error occurred, we kept this directory around for debugging.

kennt-percona: For example,
            wsrep_log_error "******************* FATAL ERROR ********************** "
            wsrep_log_error "Move failed, keeping ${DATA} for further diagnosis" \
                            "Check ${DATA}/innobackup.move.log for details"
            wsrep_log_error "Line $LINENO"
            cat_file_to_stderr "${DATA}/innobackup.move.log" "ERR" "innobackup.move.log"
            wsrep_log_error "****************************************************** "

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -2302,8 +2364,11 @@ then
 
         if [[ $errcode -ne 0 ]]; then
             wsrep_log_error "******************* FATAL ERROR ********************** "
-            wsrep_log_error "Move failed, keeping ${DATA} for further diagnosis" \
+            wsrep_log_error "Move failed."
+            if [[ -n "$WSREP_LOG_DEBUG" ]]; then
+              wsrep_log_error "Keeping ${DATA} for further diagnosis" \
```
kennt-percona, line 2371: Can you make these two error reports use the same format?  I think the wording in the move error is `better.`
I think it would also be good to mention that the ${DATA} can be removed.
So something like:
Keeping ${DATA} for further diagnosis
Check ${DATA}/XXX.log for details
${DATA} may be removed if not needed for further diagnosis.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -270,6 +270,51 @@ interruptable_timeout() {
...
+monitor_sst_progress() {
+  local tmpsstdir=$1
+  shift
+  local pid=$1
+  shift
+  local timeout=$1
+  shift
+  local current_timeout=0
+  local previous_size=0
+  sleep=5
```
kennt-percona: Can you also make 'sleep' local

kennt-percona (approved):
If you can make these changes, it looks good.

PR #1409, opened 2021-04-06
PXC-3464: Data is not propagated with SET SESSION sql_log_bin = 0

kamil-holubicki:
https://jira.percona.com/browse/PXC-3464

Problem:
When sql_log_bin session variable is set to OFF, queries are not
replicated across PXC cluster nodes

Cause:
PXC flow decides if particular statement needs to be replicated basing
on inter alia sql_log_bin value.

Solution:
All statements are replicated. However we are still blocking replication
if it was blocked internally by the server (auxiliary
OPTION_BIN_LOG_INTERNAL_OFF flag introduced)

### DIFF
--- a/mysql-test/suite/galera/t/session_logbin_off.test
+++ b/mysql-test/suite/galera/t/session_logbin_off.test
@@ -1,0 +1,72 @@
+--source include/galera_cluster.inc
+
+--connection node_1
+--let $assert_text = node_1 binlog should be enabled
+--let $sql_log_bin = query_get_value(SHOW VARIABLES LIKE 'sql_log_bin', Value, 1)
+--let $assert_cond = "$sql_log_bin" = "ON"
+--source include/assert.inc
+
+SHOW VARIABLES LIKE 'sql_log_bin';
+SHOW VARIABLES LIKE 'log_bin';
+
+CREATE TABLE t1 (a INT PRIMARY KEY);
+INSERT INTO t1 VALUES (0);
+
+SET SESSION sql_log_bin = OFF;
+
+INSERT INTO t1 VALUES (1);
+
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 2 FROM t1
+--source include/wait_condition.inc
+
+--connection node_1
+DELETE FROM t1;
+
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 0 FROM t1
+--source include/wait_condition.inc
+
+#
+# Now the same, but with --skip-log-bin. Let's use node_2 for this
+#
+--connection node_2
+--let $restart_parameters = "restart:--skip-log-bin"
+--source include/restart_mysqld.inc
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+--let $assert_text = node_2 log_bin should be OFF
+--let $log_bin = query_get_value(SHOW VARIABLES LIKE 'log_bin', Value, 1)
+--let $assert_cond = "$log_bin" = "OFF"
+--source include/assert.inc
+
+--let $assert_text = node_2 sql_log_bin should be ON
+--let $sql_log_bin = query_get_value(SHOW VARIABLES LIKE 'sql_log_bin', Value, 1)
+--let $assert_cond = "$sql_log_bin" = "ON"
+--source include/assert.inc
+
+SHOW VARIABLES LIKE 'sql_log_bin';
+SHOW VARIABLES LIKE 'log_bin';
+
+INSERT INTO t1 VALUES (0);
+
+SET SESSION sql_log_bin = OFF;
+
+INSERT INTO t1 VALUES (1);
+
+--connection node_1
+--let $wait_condition = SELECT COUNT(*) = 2 FROM t1
+--source include/wait_condition.inc
+
+--connection node_2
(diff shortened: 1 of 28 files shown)


sql/wsrep_schema.cc
```diff
@@ -108,12 +108,14 @@ class binlog_off {
       : m_thd(thd),
         m_option_bits(thd->variables.option_bits),
         m_sql_log_bin(thd->variables.sql_log_bin) {
+    thd->set_sql_log_bin_off();
```
dutow: this could use a `log_bin_off_guard` data member instead?

sql/thd_raii.h
```diff
@@ -179,6 +179,7 @@ class Disable_binlog_guard {
         m_wsrep_on(thd->variables.wsrep_on),
         m_binlog_disabled(thd->variables.option_bits & OPTION_BIN_LOG) {
     thd->variables.option_bits &= ~OPTION_BIN_LOG;
+    thd->set_sql_log_bin_off();
```
dutow: This could also use a `log_bin_off_guard` data member

sql/sql_class.cc
```diff
@@ -237,6 +237,7 @@ THD::Attachable_trx::Attachable_trx(THD *thd, Attachable_trx *prev_trx)
   // Nothing should be binlogged from attachable transactions and disabling
   // the binary log allows skipping some code related to figuring out what
   // log format should be used.
+  m_thd->set_sql_log_bin_off();
```
dutow: This could also use a `log_bin_off_guard` data member instead.

sql/rpl_info_table.cc
```diff
@@ -118,6 +118,7 @@ int Rpl_info_table::do_init_info(enum_find_method method, uint instance) {
   THD *thd = access->create_thd();
 
   saved_mode = thd->variables.sql_mode;
+  thd->set_sql_log_bin_off();
```
dutow: Why not a `log_bin_off_guard`?

sql/dd/impl/transaction_impl.cc
```diff
@@ -162,6 +162,7 @@ Update_dictionary_tables_ctx::Update_dictionary_tables_ctx(THD *thd)
     m_thd->clear_current_stmt_binlog_format_row();
 
   // Disable bin logging
+  m_thd->set_sql_log_bin_off();
```
dutow: This could use a `log_bin_off_guard` data member.

sql/rpl_info_table.cc
```diff
@@ -184,6 +186,7 @@ int Rpl_info_table::do_flush_info(const bool force) {
 
   sync_counter = 0;
   saved_mode = thd->variables.sql_mode;
+  thd->set_sql_log_bin_off();
```
dutow: Why not a `log_bin_off_guard`?

sql/sql_class.h
```diff
@@ -4955,6 +4955,40 @@ class THD : public MDL_context_owner,
...
+  void set_sql_log_bin_off() {
+      sql_log_bin_off_saved++;
+  }
+  void restore_sql_log_bin_off() {
+      DBUG_ASSERT(sql_log_bin_off_saved > 0);
+      sql_log_bin_off_saved--;
+  }
+
+  bool sql_log_bin_off() const {
+      DBUG_ASSERT(sql_log_bin_off_saved >= 0);
```
dutow: I would also add an assert to the destructor ensuring that it is 0.

sql/sql_class.h
```diff
@@ -4955,6 +4955,40 @@ class THD : public MDL_context_owner,
...
+      return (sql_log_bin_off_saved > 0);
+    }
+
+
+};
+
+class log_bin_off_guard {
+  public:
+    log_bin_off_guard(THD* thd)
```
dutow: explicit

sql/rpl_info_table.cc
kamil-holubicki: because of thd lifetime. If we used RAII here, guard is destroyed after the thd.

sql/rpl_info_table.cc
kamil-holubicki: because of thd lifetime. If we used RAII here, guard is destroyed after the thd.

sql/dd/impl/transaction_impl.cc
kamil-holubicki: Will be reworked

mysql-test/suite/galera/t/session_logbin_off.test
```diff
@@ -0,0 +1,76 @@
...
+--connection node_2
+--let $assert_text = node_2 t1 should have two rows
+--let $assert_cond = [SELECT COUNT(*) FROM t1] = 2
+--source include/assert.inc
+
+--connection node_1
+DELETE FROM t1;
+
+--connection node_2
+--let $assert_text = node_2 t1 should have zero rows
```
kennt-percona: You may want to make this into a wait condition.  This assert may fail under load if the DELETE is not replicated in time.  Similar situation below also.

kamil-holubicki: Good point. I will rework in the same way for 5.7.

kamil-holubicki:
I have reworked it completely and now it follows the same way it is done on 5.7.

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/212
No failures caused by this fix.

venkatesh-prasad-v (approved):
Thank you for working on this. Approved!

PR #1411, opened 2021-04-09
PXC-3634 Erroneous documentation on Bootstrapping XtradbCluster

patrickbirch:
Changes to be committed:
	modified:   source/configure.rst


altmannmarcelo (approved):
Change looks good to me, I would double-check this with @EvgeniyPatlan since he knows the correct path for the configuration files.

doc/source/configure.rst
```diff
@@ -47,7 +47,7 @@ on the first node::
  innodb_autoinc_lock_mode=2
 
 If you are running Red Hat or CentOS,
-add the following configuration variables to :file:`/etc/percona-xtradb-cluster.conf.d/wsrep.cnf`
+add the following configuration variables to :file:`/etc/mysql/percona-xtradb-cluster.conf.d/wsrep.cnf`
```
EvgeniyPatlan: on rpm based systems there is no mysql directory in /etc

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

PR #1213, opened 2020-08-05
8.0 - Implemented PXC-3092 (Log warning at startup if keyring is specified …

altmannmarcelo:
…but cluster traffic encryption is turned off)

https://jira.percona.com/browse/PXC-3092

Added a new WSREP warning to notify customer that keyring is enabled but
pxc_encrypt_cluster_traffic is disabled.

PR #1413, opened 2021-04-23
PXC-3643: [MTR] Tests fail with keyring warnings in error log

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3643

Background:
After PXC-3092, we log a warning to notify customer when keyring
is enabled but pxc_encrypt_cluster_traffic is disabled.

Fix: Add the keyring warning to the global suppression list.


altmannmarcelo (changes requested):
Patch looks good. I would only request to remove the suppression on tests from https://github.com/percona/percona-xtradb-cluster/pull/1213/files since they are not necessary on individual tests anymore.

altmannmarcelo (approved):
LGTM

PR #1414, opened 2021-04-28
PXC-3549: Fix high priority transactions

dutow:
Issue: pxc (and upstream mysql-wsrep) changed the bf abort logic in
8.0.22, and as part of this, also high priority abort. These changes
also didn't depend on the wsrep_on variable, and were effective always
when PXC was built with WITH_WSREP (always).

These changes were incompatible with innodb (GR) high priority
transaction, and caused related tests to fail.

Fix: rolled back some of the changes, which are also not required for
galera bf abort to function.


venkatesh-prasad-v (approved):
Thank you very much for your work on this, Great job!

Approved from my side.

PR #1421, opened 2021-05-10
PXC-3645: Deadlock during ongoing transaction and RSU

kamil-holubicki:
https://jira.percona.com/browse/PXC-3645

Problem:
1. session_1 has pending write transaction
2. session_2 performs RSU which conflicts with transaction from
session_1
3. session_1 commits
4. both sessions are blocked

Cause:
1. session_1: holds MDL lock
2. session_2: RSU is started, which causes Galera desync and then pause.
Pause causes entering into LocalOrder lock with seqno N
3. session_2: stops on MDL lock which is held by session_1
4. session_1: commits. It tries to replicate, however session_2 keeps
LocalOrder lock

In 5.7 when we called wsrep_to_isolation_begin(), regardless of TOI
or RSU we set thd->wsrep_exec_mode= TOTAL_ORDER. Then when session 2
detects deadlock, abort action is done if
thd->wsrep_exec_mode= TOTAL_ORDER. So we perform abort for both TOI
and RSU.

In 8.0 we don't have wsrep_exec_mode (we have thd->wsrep_cs().mode()
instead).
wsrep::client_state::m_toi and wsrep::client_state::m_rsu were
introduced. We set them accordingly in wsrep_to_isolation_begin().
The logic that used to check for thd->wsrep_exec_mode in 5.7 was
refactored to check wsrep_thd_is_toi(), which returns true only in case
of wsrep::client_state::m_toi.
It looks like it was some mechanical refactoring, and
wsrep::client_state::m_rsu was simply omitted.

Solution:
Bring back the behavior from 5.7

### DIFF
--- a/mysql-test/suite/galera/t/galera_rsu_simple.test
+++ b/mysql-test/suite/galera/t/galera_rsu_simple.test
@@ -46,7 +46,7 @@ UPDATE t1 SET k=k+1 WHERE id<100;
 
 --connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
 --connection node_1a
-SET wsrep_OSU_method = RSU; 
+SET wsrep_OSU_method = RSU;
 ALTER TABLE t1 ADD KEY(k);
 
 --connection node_1
@@ -54,6 +54,29 @@ ALTER TABLE t1 ADD KEY(k);
 COMMIT;
 
 DROP TABLE t1;
+
+
+###
+### RSU BF-ing local transaction within non-wsrep session (5.7 behavior)
+###
+--connection node_1
+CREATE TABLE t1(id INT PRIMARY KEY AUTO_INCREMENT, k INT);
+INSERT INTO t1(k) VALUES (1),(2),(3),(101),(102),(103);
+SET SESSION wsrep_on = OFF;
+BEGIN;
+UPDATE t1 SET k=k+1 WHERE id<100;
+
+--connection node_1a
+SET wsrep_OSU_method = RSU;
+ALTER TABLE t1 ADD KEY(k);
+
+--connection node_1
+--error ER_LOCK_DEADLOCK
+COMMIT;
+
+SET SESSION wsrep_on = ON;
+DROP TABLE t1;
+
 
 --disconnect node_1a
 --source include/wait_until_count_sessions.inc
--- a/mysql-test/suite/galera/t/galera_rsu_simple.test
+++ b/mysql-test/suite/galera/t/galera_rsu_simple.test
@@ -4,6 +4,7 @@
 
 --source include/galera_cluster.inc
 --source include/have_debug_sync.inc
+--source include/count_sessions.inc
 
 CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) Engine=InnoDB;
 
@@ -33,6 +34,29 @@ SELECT COUNT(*) = 2 FROM t1;
 
 DROP TABLE t1;
 
+
+###
+### RSU BF-ing local transaction (5.7 behavior)
+###
+--connection node_1
+CREATE TABLE t1(id INT PRIMARY KEY AUTO_INCREMENT, k INT);
+INSERT INTO t1(k) VALUES (1),(2),(3),(101),(102),(103);
+BEGIN;
+UPDATE t1 SET k=k+1 WHERE id<100;
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+--connection node_1a
+SET wsrep_OSU_method = RSU; 
+ALTER TABLE t1 ADD KEY(k);
+
+--connection node_1
+--error ER_LOCK_DEADLOCK
+COMMIT;
+
+DROP TABLE t1;
+
+--disconnect node_1a
+--source include/wait_until_count_sessions.inc
 
 ###
 ### RSU with parallely active commit connection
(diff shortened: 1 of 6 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/245/

Test failures not related to this PR (unstable tests + galera.galera_alter_table_big old, permanent failure)

mysql-test/suite/galera/t/galera_rsu_simple.test
```diff
@@ -34,6 +35,29 @@ SELECT COUNT(*) = 2 FROM t1;
...
+###
+--connection node_1
+CREATE TABLE t1(id INT PRIMARY KEY AUTO_INCREMENT, k INT);
+INSERT INTO t1(k) VALUES (1),(2),(3),(101),(102),(103);
+BEGIN;
+UPDATE t1 SET k=k+1 WHERE id<100;
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+--connection node_1a
+SET wsrep_OSU_method = RSU;
```
venkatesh-prasad-v: Remove the trailing whitespace in the end. Applying the patch generates a warning
```
$ git apply 1421.diff 
1421.diff:45: trailing whitespace.
SET wsrep_OSU_method = RSU; 
warning: 1 line adds whitespace errors.
```

sql/wsrep_thd.cc
```diff
@@ -328,10 +328,20 @@ void wsrep_fire_rollbacker(THD *thd) {
...
+  // RSU => wsrep_on=0
+  // The transaction will be aborted if:
+  // (
+  //   1. BF-ing thread has wsrep_on=1
+  //   OR
+  //   2. wsrep is globally enabled and we are in RSU
+  // )
+  // AND
+  // 3. victim thread is not aborting or already aborted
+  if (  (WSREP(bf_thd) ||
```
venkatesh-prasad-v: Please run `clang-format-10` on all `.cc` files in this commit.

mysql-test/suite/galera/t/galera_rsu_simple.test
```diff
@@ -34,6 +35,29 @@ SELECT COUNT(*) = 2 FROM t1;
...
 
 
+###
+### RSU BF-ing local transaction (5.7 behavior)
+###
+--connection node_1
+CREATE TABLE t1(id INT PRIMARY KEY AUTO_INCREMENT, k INT);
+INSERT INTO t1(k) VALUES (1),(2),(3),(101),(102),(103);
+BEGIN;
+UPDATE t1 SET k=k+1 WHERE id<100;
```
venkatesh-prasad-v, line 45: Can we add a test scenario where the victim_thd has `wsrep_on=OFF`, so that we guarantee that RSU aborts non-wsrep transactions as well?
```
--connection node_1
SET SESSION wsrep_on = OFF;
BEGIN;
UPDATE t1 SET k=k+1 WHERE id<100;

--connection node_1a
SET wsrep_OSU_method = RSU; 
ALTER TABLE t1 ADD KEY(k);

--connection node_1
--error ER_LOCK_DEADLOCK
COMMIT;
```

sql/wsrep_thd.cc
```diff
@@ -328,10 +328,20 @@ void wsrep_fire_rollbacker(THD *thd) {
...
+        (WSREP_ON && wsrep_thd_is_in_rsu(bf_thd)))
+      &&  victim_thd && !wsrep_thd_is_aborting(victim_thd) ) {
```
venkatesh-prasad-v: Just a thought.

Considering the fact that BF Abort can happen only when bf thread is either
1. An applier thread 
  OR
2. running in TOI
  OR
3. running in RSU

Do you foresee any problem if we write it like this
```
if (
    (   
    ┆ wsrep_thd_is_applying(bf_thd) ||          => BF thread has wsrep_on
    ┆ (WSREP_ON && wsrep_thd_is_rsu(bf_thd)) || => condition for RSU
    ┆ wsrep_thd_is_toi(bf_thd)                  => condition for TOI
    )   
    ┆ && victim_thd 
    ┆ && !wsrep_thd_is_aborting(victim_thd)
   )
```

sql/handler.cc
```diff
@@ -8680,8 +8680,9 @@ void handler::unlock_shared_ha_data() {
 
 int ha_wsrep_abort_transaction(THD *bf_thd, THD *victim_thd, bool signal) {
   DBUG_ENTER("ha_wsrep_abort_transaction");
-  if (!WSREP(bf_thd) && !(bf_thd->variables.wsrep_OSU_method == WSREP_OSU_RSU &&
-                          wsrep_thd_is_toi(bf_thd))) {
+
+  // do not abort if wsrep_on==0 and this BF thread is not RSU
+  if (!WSREP(bf_thd) && !wsrep_thd_is_in_rsu(bf_thd)) {
```
venkatesh-prasad-v: From the point of readability, can we change it to
```
  // Abort only if THD has wsrep_on=1 or THD is in RSU
  if (!(WSREP(bf_thd) || wsrep_thd_is_in_rsu(bf_thd))) {
``````

sql/wsrep_thd.cc
kamil-holubicki: Probably yes, but I would rather stick to the original code condition to keep it as similar to upstream as possible.

venkatesh-prasad-v (approved):
Thank you for working on this. Patch looks good and is approved from my side.

PR #1427, opened 2021-05-16
fix config file location for Ubuntu/Debian

ziegenberg:
Looking at `/usr/bin/garb-systemd` the config file on Ubuntu/Debian system is expected at `/etc/default/garb`.


ziegenberg:
This is possibly a leftover from PR #1369.

EvgeniyPatlan (approved):
Thank you for your contribution!

PR #1428, opened 2021-05-17
 PXC-3664

ynuyasha:
https://jira.percona.com/browse/PXC-3664 - clustercheck in some cases fails to identify correct node state
I think this patch can be ported to other mysql versions too.

### DIFF
--- a/scripts/clustercheck.sh
+++ b/scripts/clustercheck.sh
@@ -41,12 +41,16 @@ fi
 #
 # Perform the query to check the wsrep_local_state
 #
-WSREP_STATUS=($($MYSQL_CMDLINE -e "SHOW GLOBAL STATUS LIKE 'wsrep_%';"  \
-    2>${ERR_FILE} | grep -A 1 -E 'wsrep_local_state$|wsrep_cluster_status$' \
-    | sed -n -e '2p'  -e '5p' | tr '\n' ' '))
- 
-if [[ ${WSREP_STATUS[1]} == 'Primary' && ( ${WSREP_STATUS[0]} -eq 4 || \
-    ( ${WSREP_STATUS[0]} -eq 2 && $AVAILABLE_WHEN_DONOR -eq 1 ) ) ]]
+PXC_NODE_STATUS=($($MYSQL_CMDLINE -e "SHOW STATUS LIKE 'wsrep_local_state';SHOW VARIABLES LIKE 'pxc_maint_mode';SHOW GLOBAL STATUS LIKE 'wsrep_cluster_status';" \
+     2>${ERR_FILE} | grep -A 1 -E 'wsrep_local_state$|pxc_maint_mode$|wsrep_cluster_status$' | sed -n -e '2p' -e '5p' -e '8p' | tr '\n' ' '))
+
+# ${PXC_NODE_STATUS[0]} - wsrep_local_state
+# ${PXC_NODE_STATUS[1]} - pxc_maint_mod
+# ${PXC_NODE_STATUS[2]} - wsrep_cluster_status
+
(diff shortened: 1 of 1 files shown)


scripts/clustercheck.sh
```diff
@@ -41,12 +41,16 @@ fi
...
-if [[ ${WSREP_STATUS[1]} == 'Primary' && ( ${WSREP_STATUS[0]} -eq 4 || \
-    ( ${WSREP_STATUS[0]} -eq 2 && $AVAILABLE_WHEN_DONOR -eq 1 ) ) ]]
+PXC_NODE_STATUS=($($MYSQL_CMDLINE -e "SHOW STATUS LIKE 'wsrep_local_state';SHOW VARIABLES LIKE 'pxc_maint_mode';SHOW GLOBAL STATUS LIKE 'wsrep_cluster_status';" \
+     2>${ERR_FILE} | grep -A 1 -E 'wsrep_local_state$|pxc_maint_mode$|wsrep_cluster_status$' | sed -n -e '2p' -e '5p' -e '8p' | tr '\n' ' '))
+
+# ${PXC_NODE_STATUS[0]} - wsrep_local_state
+# ${PXC_NODE_STATUS[1]} - pxc_maint_mod
+# ${PXC_NODE_STATUS[2]} - wsrep_cluster_status
+
+if [[ ${PXC_NODE_STATUS[2]} == 'Primary' &&  ( ${PXC_NODE_STATUS[0]} -eq 4 || \
```
kamil-holubicki, line 51: If I understand the problem description in PXC-3664, it is that node is joining the cluster, it is still receiving SST/IST, but HAProxy decides to route the traffic to it which causes `WSREP has not yet prepared the node for application use`.

The condition now evaluates to:
```
IF
    node is in Primary component
    AND
    node is synced OR ( node is donor AND donor usage allowed)
    AND
    node's maintenance mode is disabled
```
Could you please explain how checking `pxc_maint_mode` helps?

ynuyasha: Hi @kamil-holubicki the patch is only to align with the script that is inside docker images, because seems that are not aligned with the docker version.
however this PR doesn't fix my issue because the problem is that haproxy when a new node is added (and is still in IST phase) for a little time pass queries until the first clustercheck says "hey i'm down". Time from haproxy reload and first healthy check.
Talking with haproxy team they suggest to create a haproxy state file where we will put the new node with a down/maintenance status, so when haproxy reload with new configuration they read that new node is already down and also first queries doesn't pass and fail.
https://www.haproxy.com/documentation/hapee/latest/api/runtime-api/show-servers-state/
I hope that now is clear the scenario.

kamil-holubicki: Hi @ynuyasha , 
I understand now. It is HAProxy that assumes the node is OK by default, even before the 1st check. So it looks like your problem is related to HAProxy behavior and should be fixed by HaProxy state file, not PXC (there is no way to fix it on PXC side).
Am I correct?

ynuyasha: yes correct @kamil-holubicki

kamil-holubicki: @ynuyasha Thanks for the confirmation and your contribution. I will accept this PR to have both scripts aligned, however, will mark PXC-3664 as invalid. Could you please squash commits and adjust the commit message to describe what it really is?

ynuyasha: @kamil-holubicki done

kamil-holubicki (approved):

kamil-holubicki:
Merged using GCA process
https://github.com/percona/percona-xtradb-cluster/pull/1670
https://github.com/percona/percona-xtradb-cluster/pull/1671

Thanks for your contribution.

PR #1431, opened 2021-05-27
PXC-3619: PXC 8.0.23 merge

kamil-holubicki:
https://jira.percona.com/browse/PXC-3619


mysql-test/extra/rpl_tests/rpl_gtids_restart_slave_io_lost_trx.test
```diff
@@ -10,14 +10,14 @@
 # IO thread right after queuing an event of a given type, starting the IO
 # thread again to fully retrieve the partially received transaction.
 #
-# Additionally, this test also tests partial transactions with MASTER_DELAY
-# configured and verifies that the applier thread doesn't wait for MASTER_DELAY
+# Additionally, this test also tests partial transactions with SOURCE_DELAY
```
kamil-holubicki, line 13: related to all changes in this commit:
Why we changed it? Does it cause any problems? It is not related to PXC and this is how it goes from upstream. It should be fixed in upstream first, then inherited to PXC. In other case it is the source of potential conflicts during the next merges.

venkatesh-prasad-v: @kamil-holubicki The test was failing on Jenkins because it had not been fixed in PS-8.0.23. However, it was fixed in 8.0.24 by  https://github.com/percona/percona-server/commit/2bc2c2c17496bbc00060379c17a9d0dd5f5df72e and the same was ported to PXC-8.0.23 upfront. So, there will not be any conflicts on future.

sql/sql_parse.cc
```diff
@@ -7416,17 +7398,17 @@ static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command) {
   }
 }
 
-static bool wsrep_mysql_parse(THD *thd, const char *rawbuf, uint length,
+static bool wsrep_dispatch_sql_command(THD *thd, const char *rawbuf, uint length,
                               Parser_state *parser_state,
                               bool update_userstat) {
-  DBUG_ENTER("wsrep_mysql_parse");
+  DBUG_ENTER("wsrep_dispatch_sql_command");
```
dutow: Minor comment: if we are changing these DBUG_ENTERs, we could use DBUG_TRACE instead.

sql/wsrep_sst.cc
```diff
@@ -1082,6 +1082,8 @@ static uint server_session_execute(MYSQL_SESSION session, std::string query,
                                    bool ignore_error = false) {
   COM_DATA cmd;
   wsp::Sql_resultset rset;
+
+  memset(&cmd, 0, sizeof(cmd));
```
dutow: I think correctly initializing the 2 new data members in COM_QUERY_DATA would be more explicit here, instead of memsetting everything.

dutow (commented):
I only added 2 optional comments, otherwise looks good.

sql/wsrep_sst.cc
kamil-holubicki: It is the same we see in upstream's plugin/sql_service_interface.cc `Sql_service_interface::execute_query`, so I think it is OK.

kamil-holubicki:
Squashed commits related to fixes of various mtr tests that do not belong to bigger groups (which are fixed as separate commits)

satya-bodapati (approved):
Reviewed MTR testcase failures. LGTM!

dutow (approved):

EvgeniyPatlan (approved):

oleksandr-kachan (approved):

PR #1433, opened 2021-05-28
PXC-3619 : PXC 8.0.23 merge

satya-bodapati:
Fix clone.remote_dml_recovery failure

Test fails because the suppressions added from clone tests didn't work.
In PXC branches, mtr.test_suppressions is changed to MyISAM engine

Clone tests, when they restore from cloned datadir, do not clone MyISAM
tables(schema replicated but not the data).So the contents of
mtr.test_suppressions table become empty

So suppression do not work and clone.remote_dml_recovery test fails.

Fix:
----
The current approach taken is to add all the clone suppressions to global_suppressions
table (which is InnoDB table and replicated by clone)


kamil-holubicki (approved):
lgtm

PR #1442, opened 2021-06-11
PXC-3675 PS 8.0.25 merge

dutow:


cmake/dev-helper.cmake
```diff
@@ -58,7 +58,7 @@ if(WITH_PXB_DEV)
 
   ExternalProject_Add(pxb80
     GIT_REPOSITORY https://github.com/percona/percona-xtrabackup.git
-    GIT_TAG percona-xtrabackup-8.0.23-16
+    GIT_TAG release-8.0.25-17
```
venkatesh-prasad-v: `release-8.0.25-17` is not a release tag. https://github.com/percona/percona-xtrabackup/tags

Please change it to `percona-xtrabackup-8.0.25-17`


Also, please set the `XB_8x_REQUIRED_VERSION` to '8.0.25' in `scripts/wsrep_sst_xtrabackup-v2.sh`

sql/binlog.cc
```diff
@@ -1197,17 +1197,12 @@ static binlog_cache_mngr *thd_get_cache_mngr(const THD *thd) {
...
   /* PXC can operate in emulation bin logging mode so suppressing the check. */
 #else
-  DBUG_ASSERT(opt_bin_log);
-#endif /* WITH_WSREP */
-||||||| 35582423e36
-  DBUG_ASSERT(opt_bin_log);
-=======
   assert(opt_bin_log);
->>>>>>> Percona-Server-8.0.25-15
+  assert(opt_bin_log);
```
venkatesh-prasad-v: assert is repeated.

sql/binlog.cc
```diff
@@ -8395,26 +8372,16 @@ int MYSQL_BIN_LOG::prepare(THD *thd, bool all) {
     active transactional storage engines, such as is the case if this is a
     replication applier and log_slave_updates=0.
   */
-<<<<<<< HEAD
 #ifdef WITH_WSREP
-  DBUG_ASSERT(thd->wsrep_applier ||
+  assert(thd->wsrep_applier ||
               (thd->slave_thread ? opt_log_slave_updates
```
venkatesh-prasad-v: This doesn't look it is clang formatted.

sql/rpl_gtid_state.cc
```diff
@@ -878,23 +878,17 @@ void Gtid_state::update_gtids_impl_own_gtid(THD *thd, bool is_commit) {
     In Group Replication the GTID may additionally be owned by another
     thread, and we won't remove that ownership (it will be rolled back later)
   */
-<<<<<<< HEAD
 #ifdef WSREP
```
venkatesh-prasad-v: Shouldn't this be `#ifdef WITH_WSREP`?

sql/rpl_slave.cc
```diff
@@ -7563,57 +7473,22 @@ wsrep_restart_point :
...
-      if (saved_skip && rli->slave_skip_counter == 0) {
-        LogErr(INFORMATION_LEVEL, ER_RPL_SLAVE_SKIP_COUNTER_EXECUTED,
-               (ulong)saved_skip, saved_log_name, (ulong)saved_log_pos,
-               saved_master_log_name, (ulong)saved_master_log_pos,
-               rli->get_group_relay_log_name(),
-               (ulong)rli->get_group_relay_log_pos(),
-               rli->get_group_master_log_name(),
-               (ulong)rli->get_group_master_log_pos());
-        saved_skip = 0;
-      }
```
venkatesh-prasad-v: Missing block??

sql/rpl_slave.cc
```diff
@@ -7897,29 +7611,13 @@ wsrep_restart_point :
...
-=======
-      For instance, user thread might be issuing show_slave_status
-      and attempting to read mi->rli->info_thd->get_proc_info().
-      Therefore thd must only be deleted after info_thd is set
-      to NULL.
-    */
     mysql_thread_set_psi_THD(nullptr);
-    delete thd;
->>>>>>> Percona-Server-8.0.25-15
+  delete thd;
```
venkatesh-prasad-v: alignment error

venkatesh-prasad-v (changes requested):
Here are the first round of comments. Please check them.

dutow:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/297/#showFailuresLink

sql/rpl_slave.cc
dutow: The alignment issue is in the line above, which isn't changed in this commit. Anyway, I fixed it.

sql/rpl_slave.cc
dutow: It's not missing, I kept the pxc version of the diff.

sql/rpl_gtid_state.cc
dutow: Maybe. But its there for a long time, and I didn't want to change it as part of a merge pr.

sql/binlog.cc
dutow: fixed.

sql/binlog.cc
dutow: fixed

cmake/dev-helper.cmake
dutow: Right, the release wasn't there when I originally changed this, fixed.

Also, looks like we missed the script change in the .23 release too.

Fixed that too.

venkatesh-prasad-v: Do you mean we missed updating the scripts other than what was changed in https://github.com/percona/percona-xtradb-cluster/commit/87046b67336211e8704d5e5caed204b24828cc0d?

sql/rpl_slave.cc
venkatesh-prasad-v: You're right! This block is around 50 above the conflict. It appeared to me that it is missing. Thanks for clarifying.

cmake/dev-helper.cmake
dutow: Uh, no, my fault, I accidentally updated the 2.4 version instead of the 8.0.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1642,7 +1642,7 @@ fi
 # 2.4.20  Transition-key fixes
 #
 
-XB_2x_REQUIRED_VERSION="2.4.20"
+XB_2x_REQUIRED_VERSION="2.4.25"
```
venkatesh-prasad-v: Latest PXB version is `2.4.23`. So, please set this to `2.4.23`.
Ref: https://github.com/percona/percona-xtrabackup/releases

sql/mysqld.cc
```diff
@@ -7488,42 +7479,7 @@ static int init_server_components() {
...
-    binlog_expire_logs_seconds = 0;
-  assert(expire_logs_days == 0 || binlog_expire_logs_seconds == 0);
-
-  if (!opt_bin_log) {
-    if (binlog_expire_logs_seconds_supplied)
-      LogErr(WARNING_LEVEL, ER_NEED_LOG_BIN, "--binlog-expire-logs-seconds");
-    if (expire_logs_days_supplied)
-      LogErr(WARNING_LEVEL, ER_NEED_LOG_BIN, "--expire_logs_days");
-    if (binlog_space_limit)
-      LogErr(WARNING_LEVEL, ER_NEED_LOG_BIN, "--binlog-space-limit");
```
venkatesh-prasad-v: I suggest to retain upstream's code here and handle `wsrep_recover` after gtid infrastructure initialization.
Please see https://github.com/percona/percona-server/commit/2dc446d46e7ca2309c09ae5bd292a14c52ce3a7e

The same has been addressed in 5.7 by https://github.com/percona/percona-xtradb-cluster/commit/4f9f7b8e405b3a05429373f41d7e68b03c0a215c#diff-2dce75ff069f612088d16147d9f58171c0614688cf5f7329789a692892c4c752

venkatesh-prasad-v (changes requested):
Few more comments.

scripts/wsrep_sst_xtrabackup-v2.sh
george-lorch: Actually, this is 8.0, why do we reference PXB 2.4 series at all, we should be exclusively requiring 8.0.25 as 2.4 does not work with the 8.0 series.

dutow: We need 2.4 for the 5.7 -> 8.0 upgrade.

sql/mysqld.cc
```diff
@@ -8383,6 +8383,13 @@ int mysqld_main(int argc, char **argv)
     if (mysql_bin_log.write_event_to_binlog_and_sync(&prev_gtids_ev))
       unireg_abort(MYSQLD_ABORT_EXIT);
 
+#ifdef WITH_WSREP
+  /* In wsrep_recovery mode, PXC avoid creation of new binlog file for
+  the reason mentioned above. In light of the said flow avoid purge
```
venkatesh-prasad-v: The reason is not mentioned above for explaining this behavior.

I think we should mention the resaon here. (Copied from an already existing comment)
```
    /* Don't spawn a new binlog file during wsrep-recovery. Why ?
    - Recovery flow is only going to read existing wsrep saved co-ordinate
      from sys_header. No other action is performed that needs binlogging.
  
    - Existing server flow looks at last binlog to set gtid_executed if server
      was shutdown abruptly. Newly create binlog will not have any such
      information and so restart post wsrep_recover will result in gtid_executed
      to be empty. */
```

dutow: I'm not sure if it's a good idea to duplicate large comments blocks, but I added it here.

venkatesh-prasad-v:
Looks good to me. Can you please run a full MTR and check we don't have any new test failures?

sql/mysqld.cc
venkatesh-prasad-v: I agree. But in this case, having a comment that says 
```
In wsrep_recovery mode, PXC avoid creation of new binlog file for
  the reason mentioned above.
```
without the actual reason above didn't make sense to me.

sql/wsrep_mysqld.cc
```diff
@@ -1783,7 +1783,7 @@ int wsrep_to_buf_helper(THD *thd, const char *query, uint query_len,
 
 static int create_view_query(THD *thd, uchar **buf, size_t *buf_len) {
   LEX *lex = thd->lex;
-  SELECT_LEX *select_lex = lex->select_lex;
+  auto *select_lex = lex->query_block;
```
kamil-holubicki: Maybe it would be better for further merges from Codership 8.0.25 to use their code here (and in all places around select_lex change)? It does the same, and we would be more aligned with their code

kamil-holubicki (commented):
Looks good, with 1 comment/question.
Can you please paste the link to MTR run results?

scripts/wsrep_sst_xtrabackup-v2.sh
george-lorch: Ahh, right, makes sense

sql/wsrep_mysqld.cc
dutow: The main difference I see is that I've used auto, while they mentioned the type.

wsrep_mysqld.cc is around 3000 lines for them, 2700 for us, the current diff between our versions is 2892 lines. I don't think there's much point in trying to keep ours versions the same at this point.

dutow:
Jenkins with latest changes: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/297/#showFailuresLink

kamil-holubicki (approved):
lgtm

PR #1445, opened 2021-06-11

kamil-holubicki:

Problem:
When ALTER TABLE (TOI) is executed in user session, sometimes it happens
that it conflicts (MDL) with high priority transaction, which causes
BF-BF abort and server termination.

Cause:
1. Let's assume we have 2 tables p1 (parent) and c1 (child). Child
having foreign key referencing p1.
2. TOI done on p1 gets replicated, but doesn't enter ApplyMonitor,
so its local execution does not start yet
3. At the same time node 2 executes and replicates DML on c1
4. Replicated DML is executed by replication thread on node 2 (high
priority).
It gets MDL locks on c1
5. ALTER TABLE (TOI) continues. It gets MDL locks on c1 as well

Solution:
Similar issue was reported to MariaDB and fixed under MDEV-21577.
The same fix was also applied to MySQL patch from Codership. The fix
assumes that DDL is locking only parent tables. ALTER TABLE however,
locks child tables as well.

1. MDEV-21577 patch has been applied (code changes + MTR test)
2. Child tables added as certification keys when ALTER TABLE is executed
(code changes + MTR test)

### DIFF
--- a/mysql-test/suite/galera/r/galera_bf_bf.result
+++ b/mysql-test/suite/galera/r/galera_bf_bf.result
@@ -1,0 +1,72 @@
+SET SESSION wsrep_sync_wait = 0;
+CREATE TABLE p1 (pk INTEGER PRIMARY KEY, f2 INTEGER);
+INSERT INTO p1 VALUES (1,1);
+CREATE TABLE c1 (pk INTEGER PRIMARY KEY, f2 CHAR(30), fk INTEGER, FOREIGN KEY (fk) REFERENCES p1(pk));
+INSERT INTO c1 VALUES (1, 'INITIAL VALUE', 1);
+INSERT INTO c1 VALUES (2, 'INITIAL VALUE', 1);
+SET GLOBAL wsrep_provider_options = 'dbug=d,apply_monitor_master_enter_sync';
+ALTER TABLE p1 ADD f1 INT;
+SET SESSION wsrep_sync_wait=0;
+SET SESSION wsrep_on = 0;
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+SET AUTOCOMMIT = ON;
+START TRANSACTION;
+UPDATE c1 SET f2 = 'changed from node 1' WHERE pk = 1;
+SELECT * FROM c1 WHERE pk = 2 FOR UPDATE;
+pk	f2	fk
+2	INITIAL VALUE	1
+SET SESSION wsrep_sync_wait = 0;
+SET GLOBAL wsrep_provider_options = 'dbug=d,apply_monitor_slave_enter_sync';
+UPDATE c1 SET f2 = 'changed from node 2' WHERE pk = 2;
+SET GLOBAL wsrep_provider_options = 'dbug=d,commit_monitor_master_enter_sync';
+COMMIT;
+SET SESSION wsrep_on = 0;
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+SET GLOBAL wsrep_provider_options = 'dbug=d,abort_trx_end';
+SET GLOBAL wsrep_provider_options = 'signal=apply_monitor_slave_enter_sync';
+SET SESSION wsrep_on = 0;
+Timeout in wait_condition.inc for SELECT 1 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_debug_sync_waiters' AND VARIABLE_VALUE = 'abort_trx_end apply_monitor_master_enter_sync commit_monitor_master_enter_sync'
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+SET GLOBAL wsrep_provider_options = 'signal=abort_trx_end';
+SET GLOBAL wsrep_provider_options = 'signal=commit_monitor_master_enter_sync';
+SET GLOBAL wsrep_provider_options = 'signal=apply_monitor_master_enter_sync';
+SET GLOBAL wsrep_provider_options = 'dbug=';
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+include/assert.inc [node_2 changes should be replicated]
+node_1:
+SELECT * FROM c1;
+pk	f2	fk
+1	INITIAL VALUE	1
+2	changed from node 2	1
+SELECT * FROM p1;
+pk	f2	f1
+1	1	NULL
+SHOW CREATE TABLE p1;
+Table	Create Table
+p1	CREATE TABLE `p1` (
+  `pk` int NOT NULL,
+  `f2` int DEFAULT NULL,
+  `f1` int DEFAULT NULL,
+  PRIMARY KEY (`pk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+node_2:
+SELECT * FROM c1;
+pk	f2	fk
+1	INITIAL VALUE	1
+2	changed from node 2	1
+SELECT * FROM p1;
+pk	f2	f1
+1	1	NULL
+SHOW CREATE TABLE p1;
+Table	Create Table
+p1	CREATE TABLE `p1` (
+  `pk` int NOT NULL,
+  `f2` int DEFAULT NULL,
+  `f1` int DEFAULT NULL,
+  PRIMARY KEY (`pk`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
+DROP TABLE c1;
+DROP TABLE p1;
--- a/mysql-test/suite/galera/r/galera_bf_bf.result
+++ b/mysql-test/suite/galera/r/galera_bf_bf.result
@@ -10,7 +10,6 @@ SET SESSION wsrep_sync_wait=0;
 SET SESSION wsrep_on = 0;
 SET SESSION wsrep_on = 1;
 SET GLOBAL wsrep_provider_options = 'dbug=';
-SET AUTOCOMMIT = ON;
 START TRANSACTION;
 UPDATE c1 SET f2 = 'changed from node 1' WHERE pk = 1;
 SELECT * FROM c1 WHERE pk = 2 FOR UPDATE;
--- a/mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
+++ b/mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
@@ -77,9 +77,9 @@ EXPECT_2
 2
 ######################################################################
 #
-# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
+# Scenario #3: 2 DMLs working on two FK parent tables try to replicate,
 #              but fails in certification for earlier DDL on child table
-#              which is child to both FK parents 
+#              which is child to both FK parents
 #
 ######################################################################
 BEGIN;
@@ -238,9 +238,9 @@ EXPECT_2
 2
 ######################################################################
 #
-# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
+# Scenario #3: 2 DMLs working on two FK parent tables try to replicate,
 #              but fails in certification for earlier DDL on child table
-#              which is child to both FK parents 
+#              which is child to both FK parents
 #
 ######################################################################
 BEGIN;
@@ -392,9 +392,9 @@ EXPECT_2
 2
 ######################################################################
 #
-# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
+# Scenario #3: 2 DMLs working on two FK parent tables try to replicate,
 #              but fails in certification for earlier DDL on child table
-#              which is child to both FK parents 
+#              which is child to both FK parents
 #
 ######################################################################
 BEGIN;
@@ -502,9 +502,9 @@ EXPECT_2
 2
 ######################################################################
 #
-# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
+# Scenario #3: 2 DMLs working on two FK parent tables try to replicate,
 #              but fails in certification for earlier DDL on child table
-#              which is child to both FK parents 
+#              which is child to both FK parents
 #
 ######################################################################
 BEGIN;
--- a/mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
+++ b/mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
@@ -1,0 +1,543 @@
+SET SESSION wsrep_sync_wait=0;
+SET SESSION wsrep_sync_wait=0;
+######################################################################
+# Test for OPTIMIZE 
+######################################################################
+######################################################################
+#
+# Scenario #1: DML working on FK parent table BF aborted by DDL
+#              over child table
+#
+######################################################################
+SET SESSION wsrep_sync_wait=0;
+CREATE TABLE p1 (pk INTEGER PRIMARY KEY, f2 CHAR(30));
+INSERT INTO p1 VALUES (1, 'INITIAL VALUE');
+CREATE TABLE p2 (pk INTEGER PRIMARY KEY, f2 CHAR(30));
+INSERT INTO p2 VALUES (1, 'INITIAL VALUE');
+INSERT INTO p2 VALUES (2, 'INITIAL VALUE');
+CREATE TABLE c1 (pk INTEGER PRIMARY KEY, fk INTEGER, FOREIGN KEY (fk) REFERENCES p1(pk));
+INSERT INTO c1 VALUES (1,1);
+CREATE TABLE c2 (pk INTEGER PRIMARY KEY, fk1 INTEGER, fk2 INTEGER, FOREIGN KEY (fk1) REFERENCES p1(pk), FOREIGN KEY (fk2) REFERENCES p2(pk));
+INSERT INTO c2 VALUES (1,1,1), (2,1,2);
+SET AUTOCOMMIT=ON;
+START TRANSACTION;
+UPDATE p1 SET f2 = 'TO DEADLOCK' WHERE pk = 1;
+SET SESSION wsrep_sync_wait=0;
+OPTIMIZE TABLE c1 ;
+Table	Op	Msg_type	Msg_text
+test.c1	optimize	note	Table does not support optimize, doing recreate + analyze instead
+test.c1	optimize	status	OK
+COMMIT;
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+######################################################################
+#
+# Scenario #2: DML working on FK parent table tries to replicate, but
+#              fails in certification for earlier DDL on child table
+#
+######################################################################
+BEGIN;
+SET GLOBAL wsrep_provider_options = 'dbug=d,apply_monitor_slave_enter_sync';
+OPTIMIZE TABLE c1 ;
+Table	Op	Msg_type	Msg_text
+test.c1	optimize	note	Table does not support optimize, doing recreate + analyze instead
+test.c1	optimize	status	OK
+SET SESSION wsrep_on = 0;
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+UPDATE p1 SET f2 = 'TO DEADLOCK' WHERE pk = 1;
+COMMIT;
+SET GLOBAL wsrep_provider_options = 'signal=apply_monitor_slave_enter_sync';
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+SELECT 'I deadlocked';
+I deadlocked
+I deadlocked
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+######################################################################
+#
+# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
+#              but fails in certification for earlier DDL on child table
+#              which is child to both FK parents 
+#
+######################################################################
+BEGIN;
+BEGIN;
+SET GLOBAL wsrep_provider_options = 'dbug=d,apply_monitor_slave_enter_sync';
+OPTIMIZE TABLE c2 ;
+Table	Op	Msg_type	Msg_text
+test.c2	optimize	note	Table does not support optimize, doing recreate + analyze instead
+test.c2	optimize	status	OK
+SET SESSION wsrep_on = 0;
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+UPDATE p1 SET f2 = 'TO DEADLOCK' WHERE pk = 1;
+COMMIT;
+UPDATE p2 SET f2 = 'TO DEADLOCK' WHERE pk = 2;
+COMMIT;
+SET GLOBAL wsrep_provider_options = 'signal=apply_monitor_slave_enter_sync';
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+SELECT 'I deadlocked';
+I deadlocked
+I deadlocked
+ERROR 40001: Deadlock found when trying to get lock; try restarting transaction
+SELECT 'I deadlocked';
+I deadlocked
+I deadlocked
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+DROP TABLE c1, c2;
+DROP TABLE p1, p2;
+######################################################################
+# Test for OPTIMIZE 
+######################################################################
+SET SESSION wsrep_sync_wait=0;
+CREATE TABLE p1 (pk INTEGER PRIMARY KEY, f2 CHAR(30));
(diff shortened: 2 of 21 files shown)


mysql-test/suite/galera/t/galera_bf_bf.test
```diff
@@ -0,0 +1,119 @@
...
+# transactions that perform DML on child table.
+#
+# node_2's UPDATE should be ordered after ALTER TABLE, as from point of view of
+# node_2 it is executed after ALTER TABLE
+# node_1's UPDATE should fail certification as its certification keys conflict
+# with ALTER TABLE certification keys.
+#
+
+--source include/galera_cluster.inc
+--source include/have_debug_sync.inc
```
venkatesh-prasad-v: Please change the order of incs. Let the restrictive incs be in the beginning.
Also, please source `--source suite/galera/include/galera_have_debug_sync.inc`
```suggestion
--source include/have_debug_sync.inc
--source suite/galera/include/galera_have_debug_sync.inc
--source include/galera_cluster.inc
```

mysql-test/suite/galera/t/galera_bf_bf.test
```diff
@@ -0,0 +1,119 @@
...
+--connection node_1
+SET SESSION wsrep_sync_wait=0;
+--let $galera_sync_point = apply_monitor_master_enter_sync
+--source include/galera_wait_sync_point.inc
+--source include/galera_clear_sync_point.inc
+
+######################################################################
+
+--connection node_1
+SET AUTOCOMMIT = ON;
```
venkatesh-prasad-v: May be this is not required.

mysql-test/suite/galera/r/galera_flush_local.result
```diff
@@ -3,6 +3,7 @@ Table	Op	Msg_type	Msg_text
 test.t0	optimize	Warning	Truncated incorrect sort_buffer_size value: '0'
 test.t0	optimize	Warning	Truncated incorrect myisam_repair_threads value: '0'
 test.t0	optimize	Error	Table 'test.t0' doesn't exist
+test.t0	optimize	Error	Table 'test.t0' doesn't exist
```
venkatesh-prasad-v, line 6: Here, same error is being reported twice. Is this expected?

mysql-test/suite/galera/r/pxc_non_existent_tables.result
```diff
@@ -11,6 +11,7 @@ test.nonexestint2	analyze	status	Operation failed
 OPTIMIZE TABLE nonexstent1;
 Table	Op	Msg_type	Msg_text
 test.nonexstent1	optimize	Error	Table 'test.nonexstent1' doesn't exist
+test.nonexstent1	optimize	Error	Table 'test.nonexstent1' doesn't exist
```
venkatesh-prasad-v, line 14: Likewise

mysql-test/suite/galera/t/galera_bf_bf.cnf
```diff
@@ -0,0 +1,2 @@
+!include suite/galera/galera_2nodes.cnf
+!include suite/galera/galera_optimistic_pa.cnf
```
venkatesh-prasad-v: Not a strong opinion though. We have disabled `optimistic_pa` by default as part of https://github.com/percona/galera/commit/adcea10ded4cd955ac6b3d564170063c78e40a83. Testing the behavior with some with specific setting may not be a good idea IMO as it may not test the default behavior.

If we really want coverage for this option, then can we rename this test into an inc file and source it from two different tests? Something like 
1. galera_bf_bf.test <- Without optimistic_pa in cnf file
2. galera_bf_bf_optimistic_pa.test <- With optimistic_pa in cnf file

mysql-test/suite/galera/t/galera_ddl_fk_conflict.test
```diff
@@ -0,0 +1,36 @@
+#
+# MDL BF-BF lock conflict
+#
+
+--source include/galera_cluster.inc
+--source include/have_debug_sync.inc
+--source suite/galera/include/galera_have_debug_sync.inc
```
venkatesh-prasad-v: Please change the order of incs. Let the restrictive incs be in the beginning.
```suggestion
--source include/have_debug_sync.inc
--source suite/galera/include/galera_have_debug_sync.inc
--source include/galera_cluster.inc
```

mysql-test/suite/galera/t/galera_ddl_fk_conflict.inc
```diff
@@ -0,0 +1,192 @@
+#
+# Test for MDL BF-BF lock conflict
+# There are some DDL statements, which take extensive MDL lock for
+# a table referenced by foreign key constraint from the actual affetec table.
```
venkatesh-prasad-v: Did you mean
```suggestion
# a table referenced by foreign key constraint from the actual affected table.
```

mysql-test/suite/galera/t/galera_ddl_fk_conflict.inc
```diff
@@ -0,0 +1,192 @@
...
+# This extensive MDL lock may cause MDL BF-BF confclict situations, if the
```
venkatesh-prasad-v:
```suggestion
# This extensive MDL lock may cause MDL BF-BF conflict situations, if the
```

mysql-test/suite/galera/t/galera_ddl_fk_conflict.inc
```diff
@@ -0,0 +1,192 @@
...
+#        $table_admin_command TABLE c;
+#
+# param: $table_admin_command_end
+#        Optional additional SQL syntax to end the SQL statement, if any
+#        $table_admin_command TABLE c $table_admin_command_end;
+#
+# scenario 1, can be used to test if a DDL statement causes such MDL locking vulnerability.
+# call this test script with some table DDL command in $table_admin_command
+# if scenario 1 passes (especially COMMIT does fail for ER_LOCK_DEADLOCK),
+# then this particular DDL is vulnerable. scenraio 2 should fail for this DDL
```
venkatesh-prasad-v:
```suggestion
# then this particular DDL is vulnerable. scenario 2 should fail for this DDL
```

sql/sql_admin.cc
```diff
@@ -534,6 +534,51 @@ static Check_result check_for_upgrade(THD *thd, dd::String_type &sname,
   return {false, result_code};
 }
 
+#ifdef WITH_WSREP
+/*
+   OPTIMIZE, REPAIR and ALTER may take MDL locks not only for the affected
+   table, but also for the table referenced by foreign key constraint.
+   ALTER additionally may take MTL lock of the tables that are referencing
```
venkatesh-prasad-v:
```suggestion
   ALTER additionally may take MDL lock of the tables that are referencing
```

venkatesh-prasad-v (changes requested):
Hi Kamil! Thank you very much for working on this.

Below are my first set of superficial review comments/suggestions, as I'm still doing a deep review and trying to understand the BF-BF conflicts.

sql/mdl.cc
```diff
@@ -5125,7 +5125,7 @@ MDL_ticket *MDL_ticket_store::materialized_front(int di) {
 #ifdef WITH_WSREP
 void MDL_ticket::wsrep_report(bool debug) {
   if (debug) {
-    WSREP_DEBUG(
+    WSREP_INFO(
```
dutow, line 5128: Is this log level change intentional?

sql/sql_admin.cc
```diff
@@ -534,6 +534,51 @@ static Check_result check_for_upgrade(THD *thd, dd::String_type &sname,
...
+   altered table.
+   This wsrep_toi_replication() function handles TOI replication for OPTIMIZE
+   and REPAIR so that certification keys for potential FK parent tables are
+   also appended in the write set.
+   ALTER TABLE case is handled in alter table execution path.
+*/
+static bool wsrep_toi_replication(THD *thd, TABLE_LIST *tables) {
```
dutow, line 548: Could we give a more specific name to this function?

sql/wsrep_mysqld.cc
```diff
@@ -1653,9 +1656,100 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
   return ret;
 }
 
+/*
+ * Collect wsrep keys corresponding to tables that are referencing given table
+ */
+static bool collect_fk_children(const dd::Table *table_def,
+                                wsrep::key_array *keys) {
+  for (const dd::Foreign_key_parent *fk : table_def->foreign_key_parents()) {
+    WSREP_INFO("(c) appended fkey %s", fk->child_table_name().c_str());
```
dutow: This doesn't seem like something we would want to print on info level?

mysql-test/suite/galera/t/galera_ddl_fk_conflict.test
kamil-holubicki: This file and all galera_ddl_fk_conflict* files come from Codership.
Adjusting as suggested anyway.

sql/sql_admin.cc
kamil-holubicki: fixed

sql/mdl.cc
kamil-holubicki: Yes, this is intentional. Please look at wsrep_report() function call. It is called like wsrep_report(wsrep_debug) most of the time. wsrep_debug is checked again inside WSREP_DEBUG macro (and uses INFO log severity anyways). So we have this debug only if wsrep_debug is != 0 (!=false, NONE).
However, in the case of BF-BF, we call it wsrep_report(true) to have this log unconditionally. (that's the intention), however, if we use WSREP_DEBUG, the log is blocked inside the macro. So having WSREP_INFO here solves the problem

sql/sql_admin.cc
kamil-holubicki: I would like to keep it as it is because this part comes from the bundle from Codership, so just to be more in sync...

sql/wsrep_mysqld.cc
kamil-holubicki: fixed (log removed)

mysql-test/suite/galera/t/galera_bf_bf.test
kamil-holubicki: fixed (removed)

mysql-test/suite/galera/t/galera_bf_bf.test
kamil-holubicki: fixed as suggested

mysql-test/suite/galera/r/galera_flush_local.result
kamil-holubicki: Yes, it is expected. I mean this is because right now we are adding parent tables as writeset keys (wsrep_append_fk_parent_table() ). To do this, we are opening the given table and traversing through its FK's. As the table is not present (this is the test) the error is pushed to the stack. Next, the original flow continues and... well... the table does not exist.
I was thinking about suppressing the 1st error, but I think it is better to have it there for diagnostic purposes.

Please also note that this behavior comes from upstream (Codership)

mysql-test/suite/galera/r/pxc_non_existent_tables.result
kamil-holubicki: Likewise

mysql-test/suite/galera/t/galera_bf_bf.cnf
kamil-holubicki: It was just an investigation's leftover. Removed.

mysql-test/suite/galera/r/galera_bf_bf.result
```diff
@@ -0,0 +1,71 @@
...
+UPDATE c1 SET f2 = 'changed from node 2' WHERE pk = 2;
+SET GLOBAL wsrep_provider_options = 'dbug=d,commit_monitor_master_enter_sync';
+COMMIT;
+SET SESSION wsrep_on = 0;
+SET SESSION wsrep_on = 1;
+SET GLOBAL wsrep_provider_options = 'dbug=';
+SET GLOBAL wsrep_provider_options = 'dbug=d,abort_trx_end';
+SET GLOBAL wsrep_provider_options = 'signal=apply_monitor_slave_enter_sync';
+SET SESSION wsrep_on = 0;
+Timeout in wait_condition.inc for SELECT 1 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_debug_sync_waiters' AND VARIABLE_VALUE = 'abort_trx_end apply_monitor_master_enter_sync commit_monitor_master_enter_sync'
```
venkatesh-prasad-v: At this point, we have 3 running transactions on node_1.
1. ALTER TABLE waiting in `apply_monitor_master_enter_sync` with `{global_seqno_ = 8, depends_seqno_ = 7, is_local_ = true, is_toi_ = true,..}`
2. Locally running Trx waiting in `commit_monitor_master_enter_sync` with `{global_seqno_ = 10, mode_ = galera::ReplicatorSMM::CommitOrder::NO_OOOC, is_local_ = true,..}`
3. Trx replicated from node_2 with `{global_seqno_ = 9, depends_seqno_ = 8, is_local_ = false, is_toi_ = false,..}` waiting in 
```
#2  __pthread_cond_wait (cond=0x7f3a1bfaf5c0, mutex=0x56052b5f3418) at pthread_cond_wait.c:638
#3  0x00007f3a25c4b7ca in gu::Lock::wait (this=0x7f3a1a5e1c80, cond=...) at galerautils/src/gu_lock.hpp:78
#4  0x00007f3a25e2faf8 in galera::Monitor<galera::ReplicatorSMM::ApplyOrder>::enter (this=0x56052b5f3418, obj=...) at galera/src/monitor.hpp:175
#5  0x00007f3a25e154c8 in galera::ReplicatorSMM::apply_trx (this=0x56052b5f22f0, recv_ctx=0x7f3a1a5e2b70, ts=...) at galera/src/replicator_smm.cpp:557
```
It looks that the trx from node_2 is waiting for the TOI to complete (its depends_seqno_=8). So, it doesn't proceed and `abort_trx_end` is not hit. Probably we should remove `abort_trx_end` from the `$galera_sync_point`.

mysql-test/suite/galera/t/galera_bf_bf.test
```diff
@@ -0,0 +1,119 @@
...
+SELECT * FROM c1 WHERE pk = 2 FOR UPDATE;
+
+# Block the applier on node #1 and issue a conflicting update on node #2
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+SET SESSION wsrep_sync_wait = 0;
+--let $galera_sync_point = apply_monitor_slave_enter_sync
+--source include/galera_set_sync_point.inc
+
+--connection node_2
+UPDATE c1 SET f2 = 'changed from node 2' WHERE pk = 2;
```
venkatesh-prasad-v: Not a strong opinion. The bug can be reproduced even without doing any transaction on node_2.
If you also think so, then can we avoid running transactions on node_2, as it only causes unnecessary deviations while debugging the issue.

Below is the test to reproduce it with only 1 node.
```
######################################################################

--connection node_1
START TRANSACTION;

UPDATE c1 SET f2 = 'changed from node 1' WHERE pk = 1;
SELECT * FROM c1 WHERE pk = 2 FOR UPDATE;

# Block the applier on node #1 and issue a conflicting update on node #2
--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
--connection node_1a

# Block the commit, send the COMMIT and wait until it gets blocked
--let $galera_sync_point = commit_monitor_master_enter_sync
--source include/galera_set_sync_point.inc

--connection node_1
--send COMMIT

--connection node_1a
--let $galera_sync_point = apply_monitor_master_enter_sync commit_monitor_master_enter_sync
--source include/galera_wait_sync_point.inc
--source include/galera_clear_sync_point.inc

# Let the transactions proceed
--let $galera_sync_point = commit_monitor_master_enter_sync
--source include/galera_signal_sync_point.inc

--let $galera_sync_point = apply_monitor_master_enter_sync
--source include/galera_signal_sync_point.inc
--source include/galera_clear_sync_point.inc

# Commit succeeds
--connection node_1
--error ER_LOCK_DEADLOCK
--reap

--connection node_1_toi
--reap
```

mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
```diff
@@ -0,0 +1,543 @@
...
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+EXPECT_1
+1
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+EXPECT_2
+2
+DROP TABLE c1, c2;
+DROP TABLE p1, p2;
+######################################################################
+# Test for OPTIMIZE
```
venkatesh-prasad-v, line 123: Applying the patch on 8.0 produces the following whitespace warnings.
```
➜  80 git:(8.0) ✗ wget https://github.com/percona/percona-xtradb-cluster/pull/1445.diff
..
➜  80 git:(8.0) ✗ git apply 1445.diff 
1445.diff:87: trailing whitespace.
# Test for OPTIMIZE 
1445.diff:163: trailing whitespace.
# Scenario #3: 2 DMLs working on two FK parent tables try to replicate, 
1445.diff:165: trailing whitespace.
#              which is child to both FK parents 
1445.diff:206: trailing whitespace.
# Test for OPTIMIZE 
1445.diff:250: trailing whitespace.
# Test for REPAIR 
warning: squelched 13 whitespace errors
warning: 18 lines add whitespace errors.
```

sql/wsrep_mysqld.h
```diff
@@ -236,6 +236,11 @@ extern enum wsrep::provider::status wsrep_sync_wait_upto_gtid(
 extern void wsrep_last_committed_id(wsrep_gtid_t *gtid);
 extern int wsrep_check_opts(int argc, char *const *argv);
 extern void wsrep_prepend_PATH(const char *path);
+void wsrep_append_fk_parent_table(THD *thd, TABLE_LIST *table,
+                                  wsrep::key_array *keys);
+void wsrep_append_child_tables(THD *thd, TABLE_LIST *tables,
+                                     wsrep::key_array *keys);
```
venkatesh-prasad-v: Please run `clang-format-10` on all `.h` and `.cc` files.

sql/sql_admin.cc
```diff
@@ -534,6 +534,51 @@ static Check_result check_for_upgrade(THD *thd, dd::String_type &sname,
...
+  /* only handle OPTIMIZE and REPAIR here */
+  switch (lex->sql_command) {
+    case SQLCOM_OPTIMIZE:
+    case SQLCOM_REPAIR:
+      break;
+    default:
+      return false;
+  }
+
+  close_thread_tables(thd);
```
venkatesh-prasad-v: I think we can avoid this redundant call to `close_thread_table()`, as this function `wsrep_toi_replication()` is called from only one place (from `mysql_admin_table()`) and we would have already called `close_thread_table()` around line number 636.

sql/sql_parse.h
```diff
@@ -147,10 +153,18 @@ extern const LEX_CSTRING command_name[];
...
+  if (WSREP(thd) && thd->wsrep_cs().state() != wsrep::client_state::s_none && \
+      !thd->lex->no_write_to_binlog &&                                        \
+      wsrep_to_isolation_begin(thd, db_, table_, table_list_))                \
+    goto wsrep_error_label;
+
+#define WSREP_TO_ISOLATION_BEGIN_FK_TABLES(db_, table_, table_list_,          \
+                                           fk_tables)                         \
+  if (WSREP(thd) && thd->wsrep_cs().state() != wsrep::client_state::s_none && \
+      !thd->lex->no_write_to_binlog &&                                        \
+      wsrep_to_isolation_begin(thd, db_, table_, table_list_, nullptr, NULL,  \
```
venkatesh-prasad-v: There is an an inconsistent usage of NULL and nullptr in this patch.

Can we use `nullptr` instead of `NULL`?

sql/sql_alter.cc
```diff
@@ -346,6 +347,43 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
   if (check_grant(thd, priv_needed, first_table, false, UINT_MAX, false))
     return true; /* purecov: inspected */
 
+#ifdef WITH_WSREP
+  /* Check if foreign keys are accessible.
+  1. Transaction is replicated first, then is done locally.
```
venkatesh-prasad-v: Just for the clarity, Does it mean that it is first replicated and then its execution is done locally? If yes, please update the comment.
```suggestion
  1. Transaction is replicated first, then is executed locally.
```

sql/sql_alter.cc
```diff
@@ -346,6 +347,43 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
...
+  That's enough, because right now we need only to know if transaction
+  should be rolled back because of lack of access and not replicated,
+  or we can replicate it and let replicated node do the rest of the job. */
+  if (alter_info.flags & Alter_info::ADD_FOREIGN_KEY) {
+    if (check_fk_parent_table_access(thd, &create_info, &alter_info, false)) {
+      return true;
+    }
+  }
+
+  TABLE *find_temporary_table(THD * thd, const TABLE_LIST *tl);
```
venkatesh-prasad-v, line 381: Not a strong opinion. From the point of readability, can we change it to 
```suggestion
 extern TABLE *find_temporary_table(THD * thd, const TABLE_LIST *tl);
```
just like https://github.com/kamil-holubicki/percona-xtradb-cluster/blob/PXC-3449/sql/sql_alter.cc#L601

sql/sql_alter.cc
```diff
@@ -371,23 +409,6 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
...
-
-  Here we check only for FK access, not if the engine supports FKs.
-  That's enough, because right now we need only to know if transaction
-  should be rolled back because of lack of access and not replicated,
-  or we can replicate it and let replicated node do the rest of the job. */
-  if (alter_info.flags & Alter_info::ADD_FOREIGN_KEY) {
-    if (check_fk_parent_table_access(thd, &create_info, &alter_info, false)) {
-      return true;
-    }
-  }
```
venkatesh-prasad-v: I do understand that this block has been moved up. But I fail to understand the reason for this change. 

Wouldn't it be better we replicate the query after checking whether the user has access to do ALTER TABLE? I mean after this block
```
    if (check_grant(thd, INSERT_ACL | CREATE_ACL, &tmp_table, false, UINT_MAX,
                    false))
      return true; /* purecov: inspected */
  }
```
Were there any problems seen when new code was added in the same place?

sql/sql_parse.cc
```diff
@@ -4811,6 +4811,22 @@ int mysql_execute_command(THD *thd, bool first_level) {
       */
       if (!handle_reload_request(thd, lex->type, first_table,
                                  &write_to_binlog)) {
+#ifdef WITH_WSREP
+        if ((lex->type & REFRESH_TABLES) &&
+            !(lex->type & (REFRESH_FOR_EXPORT | REFRESH_READ_LOCK))) {
+          /*
+            This is done after reload_acl_and_cache is because
```
venkatesh-prasad-v: Please note there is no function called `reload_acl_and_cache()` in 8.0. It was renamed to `handle_reload_request()` by https://github.com/mysql/mysql-server/commit/bf5fbb63.

Please replace all occurences of `reload_acl_and_cache` with `handle_reload_request`.

sql/sql_parse.h
```diff
@@ -129,16 +129,22 @@ extern uint sql_command_flags[];
 extern const LEX_CSTRING command_name[];
 
 #ifdef WITH_WSREP
+#include <service_wsrep.h>
```
venkatesh-prasad-v: Not a strong opinion.
```suggestion
#include "service_wsrep.h"
```

sql/sql_truncate.cc
```diff
@@ -494,20 +495,25 @@ void Sql_cmd_truncate_table::truncate_base(THD *thd, TABLE_LIST *table_ref) {
...
   // Actions needed to cleanup before leaving scope.
   auto cleanup_guard = create_scope_guard([&]() {
     end_transaction(thd, binlog_stmt, binlog_is_trans);
     cleanup_base(thd, hton);
   });
   if (mdl_locker.ensure_locked(table_ref->db)) return;
 
+#ifdef WITH_WSREP
+  wsrep::key_array keys;
+  wsrep_append_fk_parent_table(thd, table_ref, &keys);
```
venkatesh-prasad-v, line 507: Here, we are appending only the parent table keys, but not the child table keys for TRUNCATE TABLE. Is this expected?

mysql-test/suite/galera/t/galera_ddl_fk_conflict.inc
```diff
@@ -0,0 +1,192 @@
...
+--connection node_1
+--error ER_LOCK_DEADLOCK
+COMMIT;
+
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
+
+--connection node_2
+SELECT COUNT(*) AS EXPECT_1 FROM p1 WHERE f2 = 'INITIAL VALUE';
+SELECT COUNT(*) AS EXPECT_2 FROM p2 WHERE f2 = 'INITIAL VALUE';
```
venkatesh-prasad-v, line 83: Can we make use of the newly introduced `--assert()` MTR builtin function for validation
```suggestion
--assert (`SELECT COUNT(*) = 1 FROM p1 WHERE f2 = 'INITIAL VALUE'`)
--assert (`SELECT COUNT(*) = 2 FROM p2 WHERE f2 = 'INITIAL VALUE'`)

--connection node_2
--assert (`SELECT COUNT(*) = 1 FROM p1 WHERE f2 = 'INITIAL VALUE'`)
--assert (`SELECT COUNT(*) = 2 FROM p2 WHERE f2 = 'INITIAL VALUE'`)
```

venkatesh-prasad-v (changes requested):
Hi Kamil! Here is my second round of review comments. Good work on the test cases!

mysql-test/suite/galera/t/galera_bf_bf.test
kamil-holubicki: As we discussed on Slack, I resurrected the original version of the test which triggers BF-BF between TOI and applier thread and additionally I added the above test which causes BF-BF between TOI and replyer HP service

mysql-test/suite/galera/r/galera_bf_bf.result
kamil-holubicki: Tests reworked.

mysql-test/suite/galera/r/galera_ddl_fk_conflict.result
kamil-holubicki: Fixed where possible however, we cannot avoid this because how the following line works
`--echo # Test for $table_admin_command $table_admin_command_end`

sql/sql_admin.cc
kamil-holubicki: fixed.

sql/sql_parse.h
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_ddl_fk_conflict.inc
kamil-holubicki: This file comes from Galera upstream. Let's leave it as it is.

sql/sql_parse.cc
kamil-holubicki: This comes from Codership again. Fixed anyway...

sql/sql_alter.cc
kamil-holubicki: You are right. All privileges should be checked prior to replication because it will be executed in the context of root on the replica. Fixed.

sql/sql_truncate.cc
kamil-holubicki: Yes, TRUNCATE is not allowed for parent tables (if some other table is referencing it via fk), so it does not lock child tables.

sql/wsrep_mysqld.h
kamil-holubicki: done

sql/sql_alter.cc
kamil-holubicki: fixed

sql/sql_alter.cc
kamil-holubicki: Out of the scope of the current work, but I agree. Changed.

venkatesh-prasad-v (approved):
Thank you very much for addressing all the comments. The patch looks better now.
Approved from my side.

PR #1448, opened 2021-06-16
PXC-3574: Port wsrep_mode variable from Galera 8.0.22-26.5

dutow:
This variable currently has two states: empty, and ignoring slave
replication rules.


mysql-test/suite/galera/t/mysql-wsrep-372.test
```diff
@@ -0,0 +1,311 @@
+#
+# Test the case where a Galera replica to async replication uses replicate-do-db/replicate-do-table.
+# Cluster originating replication should not be affected by this.
+#
+
+--source include/have_log_bin.inc
+--source include/big_test.inc
+
+--connect node_2a, 127.0.0.1, root, , test, $NODE_MYPORT_2
```
kamil-holubicki, line 9: Do we use this connection somewhere?

source count_sessions.inc/wait_until_count_sessions.inc

mysql-test/suite/galera/t/mysql-wsrep-372.test
```diff
@@ -0,0 +1,311 @@
...
+--connection node_2
+--disable_query_log
+call mtr.add_suppression("Pending to replicate MySQL GTID event");
+--enable_query_log
+--connection node_3
+--disable_query_log
+call mtr.add_suppression("Pending to replicate MySQL GTID event");
+--enable_query_log
+
+#--connect node_3, 127.0.0.1, root, , test, $NODE_MYPORT_3
```
kamil-holubicki: Dead code

kamil-holubicki (changes requested):
1. We added a new variable. Most probably other MTR tests related to variables are affected.
2. Please add a proper commit description
3. Shouldn't we add a test for the new variable (setting it to empty string, invalid value)?
4. Consider adding the test case where wsrep_mode is empty and events are filtered

sql/log_event.cc
```diff
@@ -10850,6 +10850,7 @@ static enum_tbl_map_status check_table_map(Relay_log_info const *rli,
 #ifdef WITH_WSREP
   if ((rli->info_thd->slave_thread /* filtering is for slave only */ ||
        (WSREP(rli->info_thd) && rli->info_thd->wsrep_applier)) &&
+      wsrep_check_mode(WSREP_MODE_IGNORE_NATIVE_REPLICATION_FILTER_RULES) &&
```
venkatesh-prasad-v: Shouldn't we stop checking further and return from the function when `WSREP_MODE_IGNORE_NATIVE_REPLICATION_FILTER_RULES` is set?

Please note that, in 5.7 we have the same condition as 
```
#ifdef WITH_WSREP
  /* Check if we should ignore replication filter rules for cluster events. */
  const bool thd_is_wsrep_applier=
    ┆ WSREP(rli->info_thd) && rli->info_thd->wsrep_applier;
  if (thd_is_wsrep_applier &&
      wsrep_check_mode(WSREP_MODE_IGNORE_NATIVE_REPLICATION_FILTER_RULES))
    DBUG_RETURN(res);

  if ((rli->info_thd->slave_thread /* filtering is for slave only */ ||
    ┆  thd_is_wsrep_applier) &&
#else
  if (rli->info_thd->slave_thread /* filtering is for slave only */ &&
#endif /* WITH_WSREP */
```

mysql-test/suite/galera/t/mysql-wsrep-372.test
dutow: Fixed

mysql-test/suite/galera/t/mysql-wsrep-372.test
dutow: Yes. node1 isn't part of the cluster, so we can't initialize galera on it. We have to connect to node2 first, and initialize the cluster there.

sql/log_event.cc
dutow: I made a mistake when I copied this line from one of my checkouts to another, fixed now. Now it looks the same as in upstream codership.

dutow:
@kamil-holubicki 

1 - yes, I didn't add those changes to the commit originally, fixed.
2 - I modified it a bit, but there's not much else to explain, this is a simple variable.
3 - I added a sys_var test
4 - that is the default behavior, we already have other tests working like that.

mysql-test/suite/galera/t/mysql-wsrep-372.test
```diff
@@ -0,0 +1,315 @@
+#
```
venkatesh-prasad-v, line 1: The same test in 5.7 has the name `mysql-test/suite/galera/t/mysql-wsrep#372.test`. Can you please rename the testcase to the same name as 5.7?

sql/log_event.cc
venkatesh-prasad-v: Looks like this logic doesn't do well with native replication filters.
We have these rpl tests failing with this change.
```
[ 34%] rpl.rpl_filter_with_default_channel 'row' w3  [ fail ]
[ 39%] rpl.rpl_filters_check_counter 'row'      w3  [ fail ]
[ 60%] rpl.rpl_binlog_transaction_dependency_tracking_with_filters 'row' w3  [ fail ]
[ 68%] rpl.rpl_do_db_per_channel_filter 'row'   w3  [ fail ]
[ 72%] rpl.rpl_do_table_per_channel_filter 'row' w4  [ fail ]
[ 76%] rpl.rpl_ignore_db_per_channel_filter 'row' w1  [ fail ]
[ 83%] rpl.rpl_ignore_table_per_channel_filter 'row' w3  [ fail ]
[ 91%] rpl.rpl_wild_do_table_per_channel_filter 'row' w1  [ fail ]
[ 93%] rpl.rpl_wild_ignore_table_per_channel_filter 'row' w3  [ fail ]
```

sql/wsrep_mysqld.h
```diff
@@ -134,6 +135,9 @@ enum enum_wsrep_ignore_apply_error {
 
 extern const char *wsrep_fragment_units[];
 extern const char *wsrep_SR_store_types[];
+enum enum_wsrep_mode {
+    WSREP_MODE_IGNORE_NATIVE_REPLICATION_FILTER_RULES = 0
+};
```
venkatesh-prasad-v: Running clang-format produces a different alignment. Please run clang-format on all source files touched by this commit.
```
+enum enum_wsrep_mode { WSREP_MODE_IGNORE_NATIVE_REPLICATION_FILTER_RULES = 0 };
```

sql/log_event.cc
venkatesh-prasad-v: This is not exactly related to the mentioned code. 

The following tests are also failing. Please fix them.
1. main.mysqld--help-notwin
2. all_persisted_variables

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/mysql-wsrep-372.test
dutow: I used the name codership has in their 8.0 sources, I'm not sure if we should rename it to something different.

venkatesh-prasad-v: Okay. Then it is not a problem. Let us leave as is.

mysql-test/include/galera_cluster_master_slave3.inc
```diff
@@ -0,0 +1,10 @@
+# galera_cluster.inc
+# ==================
+#
+# Description
+# -----------
+# Configure galera cluster with 2 nodes.
```
venkatesh-prasad-v: I think you meant `3 nodes`.

mysql-test/suite/galera/t/mysql-wsrep-372.test
kamil-holubicki: I'm not sure about this part. It works but looks strange for me (mainly what inc files do vs. what they claim to do). If I understand correctly, what happens here is:
1. MTR framework starts 4 nodes. node 1 is standard MySQL node, nodes 2-4 are PXC cluster (!include ../galera_3nodes_as_slave.cnf)
2. we are connecting to node2. This is to be able to execute 2 first inc files from galera_init.inc (have_wsrep_provider.inc and have_wsrep_enabled.inc). Then galera_init.inc creates 3 connections: node_3/2/1. But node_1 has nothing to Galera! Then we connect to node_4 directly in the test (below).

so galera_cluster_master_slave3.inc does not do what it claims to do. At least it took some time to understand it. If it is not possible to rework it clearly because of what we already inherited from upstream, maybe add some comment explaining what is happening here?

Additionally, in galera_cluster.inc I see that after connecting to the particular PXC node we are checking for its readyness (galera_wait_ready.inc). Shouldn't we do the same here?

Another story is that these incs do not set up any cluster as they claim in comments, they just set up client connections to already set up cluster, but this comes from upstream, so let's leave this part as it is.

sql/wsrep_mysqld.cc
```diff
@@ -114,6 +114,7 @@ bool wsrep_restart_slave_activated = 0;  // node has dropped, and slave
                                          // restart will be needed
 bool wsrep_slave_UK_checks = 0;          // slave thread does UK checks
 bool wsrep_slave_FK_checks = 0;          // slave thread does FK checks
+ulonglong wsrep_mode                   = 0;
```
venkatesh-prasad-v: Please run clang-format on this.

dutow: Done.

kamil-holubicki (approved):
lgtm

venkatesh-prasad-v (approved):

PR #1449, opened 2021-06-16
PXC-3608 : Server crashes when due to memory access violation while a…

satya-bodapati:
…ttempting to read a FK

Problem:
We attempted to fix this problem as PXC-3459. As part of that, In row_ins_foreign_check_on_constraint()
we used a copy of old copy of record for the FK search (wsrep_append_foreign_key())
Since the page can still be modified without the latch, the problem still persisted.

Fix:
We now release the mtr latch on the page after we are done with FK operation wsrep_append_foreign_key().
This prevents the page from concurrent modifications until the mtr commit is done.

### DIFF
--- a/storage/innobase/row/row0ins.cc
+++ b/storage/innobase/row/row0ins.cc
@@ -1433,8 +1433,6 @@ row_ins_foreign_check_on_constraint(
 		btr_pcur_store_position(cascade->pcur, mtr);
 	}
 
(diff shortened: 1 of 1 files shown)


altmannmarcelo (approved):
LGTM. A customer has also been using this patch for more than a month and confirms it fixes the issue.

PR #1453, opened 2021-06-22
PXC-3688 Document NBO OSU mode (8.0)

patrickbirch:
Changes to be committed:
	new file:   source/features/online-schema-changes.rst
	modified:   source/index.rst

### DIFF
--- a/doc/source/features/nbo.rst
+++ b/doc/source/features/nbo.rst
@@ -1,0 +1,43 @@
+.. _nbo:
+ 
+================================================================================
+Non-Blocking Operations (NBO) method for Online Scheme Upgrades (OSU)
+================================================================================
+
+.. note::
+
+  The Non-Blocking Operation feature is considered **Tech Preview**. 
+
+An :ref:`online-schema-upgrade` can be a daily issue in an environment with accelerated development and deployment. The task becomes more difficult as the data grows. An ``ALTER TABLE`` statement is a multi-step operation and must run until it is complete. Aborting the statement may be more expensive than letting it complete. 
+
+The Non-Blocking Operations (NBO) method is similar to the ``TOI`` method (see :ref:`online-schema-upgrade` for more information on the available types of online schema upgrades). Every replica processes the DDL statement at the same point in the cluster transaction stream, and other transactions cannot commit during the operation.
+
+In the NBO method, the supported DDL statement acquires a metadata lock on the table or schema at a late stage of the operation. This method provides a more efficient locking strategy and avoids the ``TOI`` issue of long-running DDL statements blocking cluster updates.
+
+Attempting a State Snapshot Transfer (SST) fails during the NBO operation.
+
+To dynamically set the ``NBO`` mode in the client, run the following statement:
+
+.. sourcecode:: mysql
+
+    SET SESSION wsrep_OSU_method='NBO';
+
+.. _supported-ddl:
+
+Supported DDL Statements
+-------------------------
+
+The `NBO` method supports the following DDL statements:
+
+* ``ALTER TABLE``
+
+* ``CREATE INDEX``
+
+* ``DROP INDEX``
+
+Limitations
+-----------------
+
+The ``NBO`` method does not support running two DDL statements with conflicting locks on the same table. For example, you cannot run two ``ALTER TABLE`` statements for an ``employees`` table.
+
+See the :ref:`PXC-8.0.25-15.1` Release notes for the latest information.
--- a/doc/source/features/online-schema-upgrade.rst
+++ b/doc/source/features/online-schema-upgrade.rst
@@ -1,0 +1,51 @@
+.. _online-schema-upgrade:
+
+===================================================
+Online Schema Upgrade
+===================================================
+
+Database schemas must change as applications change. For a cluster, the schema upgrade must occur while the system is online. A synchronous cluster requires all active nodes have the same data. Schema updates are performed using Data Definition Language (DDL) statements, such as ``ALTER TABLE <table_name> DROP COLUMN <column_name>``. 
+
+The DDL statements are non-transactional, so these statements use up-front locking to avoid the chance of deadlocks and cannot be rolled back. We recommend that you test your schema changes, especially if you must run an ``ALTER`` statement on large tables. Verify the backups before updating the schemas in the production environment. A failure in a schema change can cause your cluster to drop nodes and lose data. 
+
+*Percona XtraDB Cluster* supports the following methods for making online schema changes:
+
+.. list-table::
+    :widths: 20 30 60
+    :header-rows: 1
+
+    * - Method Name
+      - Reason for use
+      - Description
+    * - `TOI` or Total Order Isolation
+    
+      - Consistency is important. Other transactions are blocked while the cluster processes the DDL statements.
+        
+      - This is the default method for the `wsrep-OSU-method <https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-osu-method>`__ variable. The isolation of the DDL statement guarantees consistency.
+        
+        The DDL replication uses a Statement format. Each node processes the replicated DDL statement at same position in the replication stream. All other writes must wait until the DDL statement is executed.
+        
+        While a DDL statement is running, any long-running transactions in progress and using the same resource receive a deadlock error at commit and are rolled back. 
+        
+        The `pt-online-schema-change <https://www.percona.com/doc/percona-toolkit/LATEST/pt-online-schema-change.html>`__ in the *Percona* Toolkit can alter the table without using locks. There are limitations: only InnoDB tables can be altered, and the ``wsrep_OSU_method`` must be ``TOI``.
+
+      
+    * - `RSU` or Rolling Schema Upgrade
+    
+      - This method guarantees high availability during the schema upgrades.
+
+      - The node desynchronizes with the cluster and disables flow control during the execution of the DDL statement. The rest of the cluster is not affected. After the statement execution, the node applies delayed events and synchronizes with the cluster.
+
+        Although the cluster is active, during the process some nodes have the newer schema and some nodes have the older schema. The RSU method is a manual operation.
+
+        For this method, the ``gcache`` must be large enough to store the data for the duration of the DDL change.
+
+    * - `NBO` or Non-Blocking Operation
+
+      - This method is used when consistency is important and uses a more efficient locking strategy.
+      
+      - Implemented in *Percona XtraDB Cluster* 8.0.25-15.1. The Non-Blocking Operation feature is in **Tech Preview** mode.
+
+        This method is similar to ``TOI``. DDL operations acquire an exclusive metadata lock on the table or schema at a late stage of the operation when updating the table or schema definition. Attempting a State Snapshot Transfer (SST) fails during the NBO operation.
+        
+        This mode uses a more efficient locking strategy and avoids the ``TOI`` issue of long-running DDL statements blocking other updates in the cluster.
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -92,6 +92,8 @@ Features
    features/highavailability
    features/multisource-replication
    features/pxc-strict-mode
+   features/online-schema-upgrade
+   features/nbo
 
 PXC Security
 ============


doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+    * - Method 
+      - Description
+    * - `TOI`
+      - Total Order Isolation
+        
+        The default DDL replication method for the :ref:`wsrep osu method` variable. The isolation of the DDL statement guarantees consistency.
+        
+        The node detects a DDL statement in the Parser stage and sends the statement to all nodes. All nodes wait for any earlier transactions to commit, and then execute the statement.
+        
+        All other writes must wait until the DDL statement is executed. While a DDL statement is running, any long-running transactions in progress and using the same resource receive a deadlock error at commit and are rolled back. There is no way to verify if individual nodes have succeeded in processing the DDL statement.
```
dutow: "There is no way to verify if individual nodes have succeeded in processing the DDL statement. " - what do you mean by this? If TOI succeeds, all nodes successfully completed the DDL statement, and it can also be verified by executing something like SHOW CREATE TABLE.

If one of the nodes failed, the DDL will fail everywhere, and the error itself won't show which nodes failed applying the DDL - but it will show up in the error log.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+        
+        The default DDL replication method for the :ref:`wsrep osu method` variable. The isolation of the DDL statement guarantees consistency.
+        
+        The node detects a DDL statement in the Parser stage and sends the statement to all nodes. All nodes wait for any earlier transactions to commit, and then execute the statement.
+        
+        All other writes must wait until the DDL statement is executed. While a DDL statement is running, any long-running transactions in progress and using the same resource receive a deadlock error at commit and are rolled back. There is no way to verify if individual nodes have succeeded in processing the DDL statement. 
+    * - `RSU`
+      - Rolling  Schema Upgrade
+
+        A DDL statement are processed on each nod locally. This action guarantees high availability. The node desynchronizes with the cluster and disables flow control while executing the DDL statement. The rest of the cluster is not affected. After the statement execution, the node applies delayed events and synchronizes with the cluster.
```
dutow: DDL statements are processed on each nod locally maybe?

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+        
+        The node detects a DDL statement in the Parser stage and sends the statement to all nodes. All nodes wait for any earlier transactions to commit, and then execute the statement.
+        
+        All other writes must wait until the DDL statement is executed. While a DDL statement is running, any long-running transactions in progress and using the same resource receive a deadlock error at commit and are rolled back. There is no way to verify if individual nodes have succeeded in processing the DDL statement. 
+    * - `RSU`
+      - Rolling  Schema Upgrade
+
+        A DDL statement are processed on each nod locally. This action guarantees high availability. The node desynchronizes with the cluster and disables flow control while executing the DDL statement. The rest of the cluster is not affected. After the statement execution, the node applies delayed events and synchronizes with the cluster. 
+
+        The cluster continues to operate, although during the process, some nodes may be updated and some nodes may have the older schema structure.
```
dutow: "may be": it's not a possibility, it is certain, since the user executes the DDL separately on each node. As long as there are more than 1 node in the cluster, there will be schema differences, and because of that, possibly consistency issues.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+
+        For this method, the ``gcache`` must be large enough to store the data for the duration of the DDL change. 
+    * - `NBO`
+      - Non-Blocking Operation
+
+        This method is similar to ``TOI``. All nodes acquire metadata locks on each table or schema for any open transactions and do not release those locks until the transaction ends. The DDL statements execute on all nodes after the transactions have completed. This method avoids the TOI issue when a long-running DDL blocks other updates in the cluster.
```
dutow: I don't understand "for any open transactions"? They acquire the MDLs required for the DDL, which at that point is just starting.

"The DDL statements execute on all nodes after the transactions have completed" - I also don't understand this. You mean that the transaction on each node commits after each node successfully executed the DDL statement?

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+
+        For this method, the ``gcache`` must be large enough to store the data for duration of the DDL statement.
```
dutow: Is this a copy paste error from RSU?

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+      - Non-Blocking Operation
+
+        This method is similar to ``TOI``. All nodes acquire metadata locks on each table or schema for any open transactions and do not release those locks until the transaction ends. The DDL statements execute on all nodes after the transactions have completed. This method avoids the TOI issue when a long-running DDL blocks other updates in the cluster.
+
+        For this method, the ``gcache`` must be large enough to store the data for duration of the DDL statement. 
+
+Using the NBO method
+----------------------
+
+You set the method for online schema changes by either using the ``wsrep_OSU_method`` variable in ``my.cnf`` or  using the ``SET`` statement in a MySQL client.
```
dutow: You shouldn't set it in my.cnf, as NBO doesn't support all DDL operations. Many DDLs would simply fail.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+
+    wsrep_osu_method=NBO
+
+Restart the server. 
+
+To dynamically set the ``NBO`` method in the client, run the following statement:
+
+.. sourcecode:: mysql
+
+    SET GLOBAL wsrep_OSU_method='NBO';
```
dutow: wsrep_OSU_method is a session variable, users most likely want to set NBO for a session, not globally.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+
+* Ensure a long transaction is not executing against the table being altered. The cluster may block the NBO from acquiring the necessary metadata locks. 
+
+* Nodes cannot be donors for SST during the non-blocking operation. 
+
+* If a node leaves the cluster during the NBO DDL operation the node's data files will be inconsistent. The node can only rejoin the cluster by SST.
+
+* NBO operations can only work with statements that involve one table at a time.
+
+* You cannot run more than one method at a time.
```
dutow: this isn't true, you can start an NBO and a TOI transaction at the same time for example, and the cluster will handle it.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+Restart the server. 
+
+To dynamically set the ``NBO`` method in the client, run the following statement:
+
+.. sourcecode:: mysql
+
+    SET GLOBAL wsrep_OSU_method='NBO';
+
+
+Limitations
```
dutow: Most of the limitations mentioned here are true for all methods.

SInce the page is called "online schema changes", not "NBO online schema change", why do we say that these are NBO limitations?

I also don't see the main NBO limitation mentioned: that it only supports some DDL statements, and that it doesn't support running two ALTER on the same table (or more specifically: two DDLs with conflicting locks) at the same time, one of them will error out.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+-----------------
+
+The following items are the limitations of using ``NBO``:
+
+* Write attempts are blocked until the ``ALTER`` statement completes. If the DDL statement is ``ALTER TABLE ... LOCK=SHARED``, read operations are allowed. If the statement is ``ALTER TABLE ... LOCK=EXCLUSIVE``, read operations are not allowed.
```
dutow: This just describes how MySQL works, what's the PXC/NBO limitation here?

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+
+* Ensure a long transaction is not executing against the table being altered. The cluster may block the NBO from acquiring the necessary metadata locks.
```
dutow: This shouldn't happen, a high priority DDL should just abort any conflicting transactions, same as with TOI.

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
...
+    * - `RSU`
+      - Rolling  Schema Upgrade
+
+        A DDL statement are processed on each nod locally. This action guarantees high availability. The node desynchronizes with the cluster and disables flow control while executing the DDL statement. The rest of the cluster is not affected. After the statement execution, the node applies delayed events and synchronizes with the cluster. 
+
+        The cluster continues to operate, although during the process, some nodes may be updated and some nodes may have the older schema structure.
+
+        For this method, the ``gcache`` must be large enough to store the data for the duration of the DDL change. 
+    * - `NBO`
+      - Non-Blocking Operation
```
dutow: @georgelorchpercona shouldn' t NBO be an experimental feature in the next release?

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,82 @@
+.. _online-schema-changes:
+
+===================================================
+Online Schema Changes
+===================================================
+
+A failure in a schema change can cause your cluster to drop nodes and lose data. |Percona XtraDB Cluster| supports the following methods for making online schema changes:
```
dutow: Do we want to start a page describing schema changes by mentioning a failure, without first explaining what an online schema change is, or how and why a failure can occur?

dutow (changes requested):

doc/source/features/online-schema-changes.rst
patrickbirch: Removed.

doc/source/features/online-schema-changes.rst
patrickbirch: Fixed.

doc/source/features/online-schema-changes.rst
patrickbirch: Fixed.

doc/source/features/online-schema-changes.rst
patrickbirch: Removed.

doc/source/features/online-schema-changes.rst
patrickbirch: Changed to SESSION

doc/source/features/online-schema-changes.rst
patrickbirch: Changed to say NBO OSC

doc/source/features/online-schema-changes.rst
patrickbirch: Removed.

doc/source/features/online-schema-changes.rst
patrickbirch: Removed.

doc/source/features/online-schema-changes.rst
patrickbirch: Added.

doc/source/features/online-schema-changes.rst
george-lorch: @dutow I thought we discussed that this is not practical to call "Tech Preview" as it is pretty core changes that can not be avoided regardless of enabled or not.

doc/source/features/online-schema-changes.rst
patrickbirch: Rewritten.

doc/source/features/online-schema-changes.rst
patrickbirch: Fixed after discussion.

doc/source/features/online-schema-changes.rst
patrickbirch: Fixed.

george-lorch (approved):

doc/source/features/online-schema-changes.rst
george-lorch: OK, yes, this feature is going to be tagged as "Tech Preview" in 8.0.25, if no serious issues present themselves, it can go to GA in 8.0.27

patrickbirch: Added.

dutow (approved):

doc/source/features/online-schema-changes.rst
```diff
@@ -0,0 +1,87 @@
...
+         
+        The node desynchronizes with the cluster and disables flow control during the execution of the DDL statement. The rest of the cluster is not affected. After the statement execution, the node applies delayed events and synchronizes with the cluster. 
+
+        Although the cluster continues to active during the process, some nodes are updated and some nodes have the older schema structure. The RSU method is a manual operation.
+
+        For this method, the ``gcache`` must be large enough to store the data for the duration of the DDL change. 
+    * - `NBO`
+      - Non-Blocking Operation
+
+        Implemented in version 25-15, |Percona XtraBackup| supports ``Non-Blocking Operation`` for :ref:`supported-ddl`.
```
venkatesh-prasad-v:
```suggestion
        Implemented in version 25-15, |Percona XtraDB Cluster| supports ``Non-Blocking Operation`` for :ref:`supported-ddl`. 
```

patrickbirch: Fixed.

PR #1454, opened 2021-06-24
Fixed PXC-3611 ("Encryption can't find master key" after SST when key…

altmannmarcelo:
…ring_file is used if keyring.backup file exists)

https://jira.percona.com/browse/PXC-3611

Problem
Under some circumstances, MySQL can leave a keyring .backup file. In
this case, once the server starts again, the .backup file will take
place of the current keyring file. This becomes a problem at SST as once
the server start again, either the keyring received from donor or the
one generated by xtrabackup will be overwritten with the staled .backup.

Fix
Since SST uses either a transaction key, which makes xtrabackup to
operate without having access to keyring file, or it receives a copy of
donor keyring file, we can remove joiner keyring file and .backup file
as part of the cleanup that precedes the SST.

### DIFF
--- a/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
+++ b/mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
@@ -50,6 +50,29 @@ INSERT INTO e1(text) VALUES('bbbbb');
 select * from e1;
 
 #
+# PXC-3611 - Test SST with presence of .backup keyring file
+#
+
+# Kill node 2 to force an SST
+--connection node_2
+--source include/kill_galera.inc
+
+# Generate a new master key that will be missing on node_2 and simulate a .backup without the new key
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://pxc.cd.percona.com/job/pxc-5.7-param/177/

Centos8 / Ubuntu focal test fail because GCA point is behind (9a22fb944ed0fd68fe10950c99ee226055464ea4)

mysql-test/suite/galera/t/galera_sst_xtrabackup-v2_keyring.test
```diff
@@ -49,6 +49,30 @@ INSERT INTO e1(text) VALUES('bbbbb');
 
 select * from e1;
 
+#
+# PXC-3611 - Test SST with presence of .backup keyring file
+#
+
+# Kill node 2 to force an SST
+# echo Killing node 2 ...
```
kamil-holubicki: I think we can delete this line, if we are not going to print the message

kamil-holubicki (approved):

PR #1455, opened 2021-06-25
Pxc 8.0 3608

satya-bodapati:


storage/innobase/row/row0ins.cc
```diff
@@ -1269,22 +1269,23 @@ static NO_INLINE MY_ATTRIBUTE((warn_unused_result)) dberr_t
...
 #ifdef WITH_WSREP
-  err = wsrep_append_foreign_key(thr_get_trx(thr), foreign,
-                                 cascade->pcur->m_old_rec, clust_index, false,
-                                 WSREP_SERVICE_KEY_EXCLUSIVE);
+  err =
+      wsrep_append_foreign_key(thr_get_trx(thr), foreign, clust_rec,
+                               clust_index, false, WSREP_SERVICE_KEY_EXCLUSIVE);
   if (err != DB_SUCCESS) {
     ib::warn() << "WSREP: foreign key append failed: " << err;
+    mtr_commit(mtr);
```
altmannmarcelo, line 1282: @satya-bodapati what about if wsrep_append_foreign_key succeeds. When do we commit the mtr?

altmannmarcelo (changes requested):

storage/innobase/row/row0ins.cc
satya-bodapati: @altmannmarcelo there is mtr_commit() on both branches.

altmannmarcelo: ah, right, it will fall the else branch.

altmannmarcelo (approved):
LGTM

PR #1459, opened 2021-06-29
8.0 Fixed PXC-3611 ("Encryption can't find master key" after SST when key…

altmannmarcelo:
…ring_file is used if keyring.backup file exists)

https://jira.percona.com/browse/PXC-3611

Problem
Under some circumstances, MySQL can leave a keyring .backup file. In
this case, once the server starts again, the .backup file will take
place of the current keyring file. This becomes a problem at SST as once
the server start again, either the keyring received from donor or the
one generated by xtrabackup will be overwritten with the staled .backup.

Fix
Since SST uses either a transaction key, which makes xtrabackup to
operate without having access to keyring file, or it receives a copy of
donor keyring file, we can remove joiner keyring file and .backup file
as part of the cleanup that precedes the SST.

### DIFF
--- a/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_keyring.result
+++ b/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_keyring.result
@@ -15,6 +15,14 @@ select * from e1;
 id	text
 1	aaaaa
 2	bbbbb
+Killing server ...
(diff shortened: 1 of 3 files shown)


altmannmarcelo:
https://pxc.cd.percona.com/job/pxc-8.0-param/272/

kamil-holubicki (commented):
LGTM

kamil-holubicki (approved):

PR #1460, opened 2021-07-02
PXC-2158 [DOC] gcomm.thread_prio documentation missing (5.7)

patrickbirch:
Changes to be committed:
	modified:   source/wsrep-provider-index.rst

### DIFF
--- a/doc/source/wsrep-provider-index.rst
+++ b/doc/source/wsrep-provider-index.rst
@@ -451,6 +451,47 @@ the :file:`galera.cache` file which is used as source for |IST|. The bigger the
 value of this variable, the better are chances that the re-joining node will
 get IST instead of |SST|.
 
+.. variable:: gcomm.thread_prio
+
+Raises the gcomm thread priority to a higher level. Use this variable when the gcomm thread does not receive enough CPU time due to other competing threads. For example, if the gcomm threads are not frequently run, a node may drop from the cluster because of the timeout.
+
+The format for this variable is: <policy>:<priority>. The policy value supports the following options: ``other``, ``fifo``, and ``rr``. The priority value is an
+integer. 
+
+.. note::
+
+   Setting a priority value of 99 is not recommended. This value blocks system threads. 
+
+An example of the variable:
+
+.. sourcecode:: text
+
+   wsrep_provider_options="gcomm.thread_prio=fifo:3"
+
+The description of the ``policy`` parameter follows:
+
+.. list-table::
+   :widths: 15 60
+   :header-rows: 1
+
+   * - Option
+     - Description
+   * - other
+     - This policy is the default Linux time-sharing scheduling. Threads run until one of the following events occur:
+
+         * Thread exit
+
+         * I/O request blocks the thread 
+
+         * Higher priority thread preempts the thread
+   * - fifo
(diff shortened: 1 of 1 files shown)


doc/source/wsrep-provider-index.rst
```diff
@@ -451,6 +451,31 @@ the :file:`galera.cache` file which is used as source for |IST|. The bigger the
 value of this variable, the better are chances that the re-joining node will
 get IST instead of |SST|.
 
+.. variable:: gcomm.thread_prio
+
+Using this option, you can raise the priority of the gcomm thread to a higher
+level than it normally uses. 
+
```
venkatesh-prasad-v, line 457: Looks good to me. But, upstream documentation has the below line which tells about when to use this variable. I feel we should also let the users know about it. What do you say?
```
You may find this useful in situations where Galera Cluster threads do not receive sufficient CPU time, due to competition with other MySQL threads. In these cases, when the thread scheduler for the operating system does not run the Galera threads frequently enough, timeouts may occur, causing the node to drop from the cluster.
```

patrickbirch: Updated.

doc/source/wsrep-provider-index.rst
```diff
@@ -451,6 +451,47 @@ the :file:`galera.cache` file which is used as source for |IST|. The bigger the
 value of this variable, the better are chances that the re-joining node will
 get IST instead of |SST|.
 
+.. variable:: gcomm.thread_prio
+
+Raises the gcomm thread priority to a higher level. Use this variable when the cluster does not receive enough CPU time and competes with other threads. For example, if the threads are not frequently run, a node may drop from the cluster because of the timeout.
```
venkatesh-prasad-v: What I feel is, this line doesn't clearly tell what competes with other threads. So, I would prefer.
```suggestion
Raises the gcomm thread priority to a higher level. Use this variable when the gcomm thread does not receive enough CPU time due to other competing threads. For example, if the gcomm threads are not frequently run, a node may drop from the cluster because of the timeout.
```

patrickbirch: fixed.

doc/source/wsrep-provider-index.rst
```diff
@@ -451,6 +451,47 @@ the :file:`galera.cache` file which is used as source for |IST|. The bigger the
...
+   * - other
+     - This policy is the default Linux time-sharing scheduling. Threads run until one of the following events occur:
+
+         * Thread exit
+
+         * I/O request blocks the thread 
+
+         * Higher priority thread preempts the thread
+   * - fifo
+     - The policy use First-in First-out (FIFO) scheduling These threads always immediately preempt any currently running other, batch or idle threads. The scheduler runs the highest priority thread until it completes, unless a higher priority thread preempts it, or it is blocked. This policy does not use time slicing.
```
venkatesh-prasad-v: Not an issue, but I would prefer
```suggestion
     - The policy uses First-in First-out (FIFO) scheduling. These threads always immediately preempt any currently running other, batch or idle threads. The threads are run in a FIFO manner until they are completed, unless a higher priority thread preempts it, or blocks it. This policy does not use time slicing.
```

venkatesh-prasad-v (approved):

doc/source/wsrep-provider-index.rst
patrickbirch: Rewritten.

PR #1461, opened 2021-07-02
PXC-3589 LOCK=NONE usage is not supported in PXC-5.7 and above (8.0)

patrickbirch:
Merged 5.7 version
Fixed conflicts in source/limitations

### DIFF
--- a/doc/source/limitation.rst
+++ b/doc/source/limitation.rst
@@ -74,3 +74,11 @@ The following limitations apply to |PXC|:
 
      Galera Documentation: Tables without Primary Keys
         http://galeracluster.com/documentation-webpages/limitations.html#tables-without-primary-keys
+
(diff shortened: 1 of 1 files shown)


doc/source/limitation.rst
```diff
@@ -82,3 +82,13 @@ Avoid reusing the names of persistent tables for temporary tables
 		<https://dev.mysql.com/doc/refman/8.0/en/temporary-table-problems.html>`_
 
 .. include:: .res/replace.opt.txt
+
+As of version 8.0.25, an INPLACE `ALTER TABLE <https://dev.mysql.com/doc/refman/5.7/en/alter-table.html>`__  query takes an internal shared lock on the table during the execution of the query. The ``LOCK=NONE`` clause is no longer allowed for all of the INPLACE ALTER TABLE queries due to this change.
```
venkatesh-prasad-v: We have this limitation 8.0.21. So, It is better we mention 8.0.21 instead of 8.0.25.

venkatesh-prasad-v: The link is pointing to 5.7 Docs. Shouldn't we change it to 8.0 Docs?

venkatesh-prasad-v (changes requested):

doc/source/limitation.rst
patrickbirch: Fixed both.

venkatesh-prasad-v (approved):

PR #1464, opened 2021-07-08
PXC-3628: Server allows mysql.wsrep_* tables to be converted to MyISAM

nitendra-bhosle:
https://jira.percona.com/browse/PXC-3628

Analysis

Server is allowing to change engine for alter table statements
because wsrep_cluster,wsrep_cluster_members and wsrep_streaming_log
tables are not registered as system tables.

Fix

Added wsrep_cluster,wsrep_cluster_members and wsrep_streaming_log
tables to system tables registry.

### DIFF
--- a/sql/dd/impl/system_registry.cc
+++ b/sql/dd/impl/system_registry.cc
@@ -246,6 +246,12 @@ void System_tables::add_remaining_dd_tables() {
   register_table("time_zone_transition_type", system);
   register_table("user", system);
 
+#ifdef WITH_WSREP
+  register_table("wsrep_cluster", system);
+  register_table("wsrep_cluster_members", system);
+  register_table("wsrep_streaming_log", system);
+#endif /* WITH_WSREP */
+
   /*
     MTR tests expects following tables to be created in the 'mysql' tablespace.
     So tables are listed here.
--- a/mysql-test/collections/disabled.def
+++ b/mysql-test/collections/disabled.def
@@ -283,9 +283,7 @@ galera.galera_fk_lock_parent_update_child : BUG#0 Few cases have timing issues(P
 galera.galera_toi_ddl_fk_insert : BUG#0 CODERSHIP qa#39 test fails sporadically (PXC-3431)
 
 # Bugs failing due to PXC-3548
-main.system_tables_myisam                           : BUG#3548 / PS-6837 Timestamp is not set to CURRENT_TIMESTAMP in mysql.tables_priv
 binlog.binlog_stm_unsafe_read_acl_tables_in_dml_ddl : BUG#3548 / PS-6837 Timestamp is not set to CURRENT_TIMESTAMP in mysql.tables_priv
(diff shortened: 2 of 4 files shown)


mysql-test/t/alter_table.test
```diff
@@ -3673,3 +3673,13 @@ JOIN t2;
...
+--echo #
+--echo # PXC-3628 : Server allows mysql.wsrep_* tables to be converted to MyISAM
+--echo #
+--error ER_UNSUPPORTED_ENGINE
+ALTER TABLE mysql.wsrep_cluster ENGINE=myisam;
+--error ER_UNSUPPORTED_ENGINE
+ALTER TABLE mysql.wsrep_cluster_members ENGINE=myisam;
+--error ER_UNSUPPORTED_ENGINE
+ALTER TABLE mysql.wsrep_streaming_log ENGINE=myisam;
+
```
venkatesh-prasad-v: There is no need to add a new test case for this. There are already some test cases which validate this behavior, but at present, they have been disabled now. So, we just need to enable the below mentioned tests in `mysql-test/collections/disabled.def`

1. main.system_tables_myisam_lctn_1
2. main.system_tables_myisam

sql/dd/impl/system_registry.cc
```diff
@@ -246,6 +246,12 @@ void System_tables::add_remaining_dd_tables() {
   register_table("time_zone_transition_type", system);
   register_table("user", system);
 
+#ifdef WITH_WSREP
+  register_table("wsrep_cluster", system);
+  register_table("wsrep_cluster_members", system);
+  register_table("wsrep_streaming_log", system);
```
venkatesh-prasad-v, line 252: Though I see that it is working as expected and this is how other system tables are added to the system registry, but I am not sure if this needs needs a bump in the DD version.

@satya-bodapati  What do you say?

venkatesh-prasad-v (changes requested):

sql/dd/impl/system_registry.cc
satya-bodapati: These are non-core DD tables and called as 'system tables'. So DD version bump is not necessary.
A side note:
As a downstream, we cannot bump the DD version at all because it will cause upgrade incompatibilities (upgrade from MySQL server).

We could have added another version number but a bit trickier to maintain and keep in sync.

venkatesh-prasad-v: Thank you for checking.

nitendra-bhosle:
Jenkins : https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/283/

venkatesh-prasad-v (approved):
Thank you for addressing the review comments. Approved!

satya-bodapati (changes requested):
Ready for approval, Please squash

nitendra-bhosle:
> Ready for approval, Please squash

Can I do Squash and Merge from github only?

venkatesh-prasad-v (approved):

satya-bodapati (approved):

PR #1465, opened 2021-07-08

kamil-holubicki:

This is the follow up of the commit
bd6f682f64da1aa463bd0e9ee37d1542a1b55673.

Problem:
It was discovered that the patch merged from the Codership is not
comprehensive, in particular in addition to OPTIMIZE, REPAIR, ALTER and
TRUNCATE, CREATE TABLE and DROP TABLE are affected by not including
parent table keys in certification set as well.

Solution:
1. CREATE TABLE and DROP TABLE add parent tables keys into the
certification set
2. WSREP_TO_ISOLATION_BEGIN_FK_TABLES renamed to
WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF as it hides 'if' inside
3. wsrep_append_child_tables() and wsrep_append_fk_parent_table()
functions reworked to use only Data Dictionary. This is mostly due to
being able to handle temporary tables.
4. MTR test extended to cover CREATE/DROP TABLE

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -4117,7 +4117,15 @@ int mysql_execute_command(THD *thd, bool first_level) {
         if (!lex->drop_temporary &&
             (!thd->is_current_stmt_binlog_format_row() ||
              !find_temporary_table(thd, table))) {
-          WSREP_TO_ISOLATION_BEGIN(NULL, NULL, all_tables);
+          // We cannot use WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF, because here
+          // lex->no_write_to_binlog is uninitialized
+          wsrep::key_array keys;
+          wsrep_append_fk_parent_table(thd, all_tables, &keys);
+          if (WSREP(thd) &&
+              wsrep_to_isolation_begin(thd, NULL, NULL, all_tables, NULL, NULL,
+                                       &keys)) {
+            goto error;
+          }
           break;
         }
       }
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -18,6 +18,7 @@
 #include "mysqld.h"
 #include "rpl_slave.h"
 #include "sql/dd/cache/dictionary_client.h"  // dd::cache::Dictionary_client
+#include "sql/dd/dictionary.h"
 #include "sql/dd/types/table.h"
 #include "sql/key_spec.h"
 #include "sql/sql_alter.h"
@@ -1644,7 +1645,7 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
   }
   if (alter_info && (alter_info->flags & Alter_info::ADD_FOREIGN_KEY)) {
     wsrep::key_array fk(wsrep_prepare_keys_for_alter_add_fk(
-        table_list->get_db_name(), alter_info));
+        (table_list ? table_list->get_db_name() : db), alter_info));
     if (!fk.empty()) {
       ret.insert(ret.end(), fk.begin(), fk.end());
     }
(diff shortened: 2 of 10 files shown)


sql/wsrep_mysqld.cc
```diff
@@ -1665,86 +1666,82 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
-  /* close the table and release MDL locks */
-  close_thread_tables(thd);
-  thd->mdl_context.rollback_to_savepoint(mdl_savepoint);
-  for (table = tables; table; table = table->next_local) {
-    table->table = NULL;
-    table->mdl_request.ticket = NULL;
+/*
+ * Collect wsrep keys corresponding to tables referenced by the given table
+ */
+static void collect_fk_parrents(const dd::Table *table_def,
```
venkatesh-prasad-v: Here are in other places.
```suggestion
static void collect_fk_parents(const dd::Table *table_def,
```

sql/wsrep_mysqld.cc
```diff
@@ -1665,86 +1666,82 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
 /*
  * Collect wsrep keys corresponding to tables that are referencing given table
  */
-static bool collect_fk_children(const dd::Table *table_def,
+static void collect_fk_children(const dd::Table *table_def,
                                 wsrep::key_array *keys) {
-  for (const dd::Foreign_key_parent *fk : table_def->foreign_key_parents()) {
+  for (auto fk : table_def->foreign_key_parents()) {
```
venkatesh-prasad-v: Please make it `const auto`

sql/wsrep_mysqld.cc
```diff
@@ -1665,86 +1666,82 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
-    if (!is_temporary_table(table) && table->table) {
-      for (TABLE_SHARE_FOREIGN_KEY_INFO *fk = table->table->s->foreign_key;
-           fk < table->table->s->foreign_key + table->table->s->foreign_keys;
-           ++fk) {
-        WSREP_INFO("(p) appended fkey %s", fk->referenced_table_name.str);
-        keys->push_back(wsrep_prepare_key_for_toi(fk->referenced_table_db.str,
-                                                  fk->referenced_table_name.str,
-                                                  wsrep::key::shared));
+
+  for (auto *table = tables; table; table = table->next_local) {
```
venkatesh-prasad-v: In `collect_fk_children()` function, the same loop is like
```
for (auto table = tables; table; table = table->next_local) {
```
Can we make it look consistent in both the functions?

```suggestion
  for (auto table = tables; table; table = table->next_local) {
```

sql/wsrep_mysqld.cc
```diff
@@ -1665,86 +1666,82 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
+      dd::cache::Dictionary_client::Auto_releaser releaser(thd->dd_client());
+      if (thd->dd_client()->acquire(dd::String_type(table->db),
+                                    dd::String_type(table->table_name),
+                                    &table_obj)) {
+        assert(0);
+        dd::release_mdl(thd, mdl_ticket);
+        continue;
+      }
+
+      collect_fk_parrents(table_obj, keys);
```
venkatesh-prasad-v: Not an issue, but an improvement.

The implementation of functions `wsrep_append_fk_parent_table` and `wsrep_append_child_tables` are almost the same except the fact that the former calls `collect_fk_parrents()` and the latter calls `collect_fk_children()` in the end of the function.

I wonder if we could have a single function which accepts the callback (func_ptr or an instance of std::function), so that duplication of code can be avoided.

sql/wsrep_mysqld.cc
```diff
@@ -1665,86 +1666,82 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
-              dd::String_type(table->table->s->db.str),
-              dd::String_type(table->table->s->table_name.str), &table_obj)) {
+      if (thd->dd_client()->acquire(dd::String_type(table->db),
+                                    dd::String_type(table->table_name),
+                                    &table_obj)) {
         assert(0);
+        dd::release_mdl(thd, mdl_ticket);
+        continue;
       }
       collect_fk_children(table_obj, keys);
```
venkatesh-prasad-v: Can we add an assert that `table_obj` ptr is valid here?
```suggestion
      }
      assert(tab_obj);
      collect_fk_children(table_obj, keys);
```

sql/wsrep_mysqld.cc
kamil-holubicki: fixed

sql/wsrep_mysqld.cc

sql/wsrep_mysqld.cc

sql/wsrep_mysqld.cc
kamil-holubicki: No, because it can be nullptr if the table does not exist. That's why we check if it is not nullptr (inside collect_fk* before refactoring, and right before calling action() after refactoring.

sql/wsrep_mysqld.cc
kamil-holubicki: Refactored using C++11 approach.

venkatesh-prasad-v (approved):
Thank you for addressing the review comments. Approved from my side!

kamil-holubicki:
While testing, I discovered that the change in sql_parse.cc causes problems, because lex->no_write_to_binlog is uninitialized there (it comes from upstream). I decided not to initialize it in LEX::LEX constructor, but instead use wsrep_to_isolation_begin() call.

sql/sql_parse.cc
```diff
@@ -4117,7 +4117,15 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
             (!thd->is_current_stmt_binlog_format_row() ||
              !find_temporary_table(thd, table))) {
-          WSREP_TO_ISOLATION_BEGIN(NULL, NULL, all_tables);
+          // We cannot use WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF, because here
+          // lex->no_write_to_binlog is uninitialized
+          wsrep::key_array keys;
+          wsrep_append_fk_parent_table(thd, all_tables, &keys);
+          if (WSREP(thd) &&
+              wsrep_to_isolation_begin(thd, NULL, NULL, all_tables, NULL, NULL,
+                                       &keys)) {
```
venkatesh-prasad-v, line 4126: Just a thought. Shouldn't we also check for the below condition?
```
thd->wsrep_cs().state() != wsrep::client_state::s_none
```
If we were using `WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF`, then we would have done the above check. But it looks we don't do it now. 

Can you check this?

kamil-holubicki: As we discussed on Slack, as it was not checked before as well, let's leave it for now as it is.

satya-bodapati: Can you guys please point to me to slack conversation on dev-server?

venkatesh-prasad-v: The conversation happened over IM and the summary was that, Kamil had indeed done some test with the condition we had proposed, and saw no difference in behavior. So, it was hard to judge the correct fix.

Also, considering the fact that the previous condition didn't have the check for the wsrep::client_state, we found it to be safer to revert to previous implementation and not push an unsure patch.

satya-bodapati: Thanks for the explanation!

PR #1468, opened 2021-07-12
PXC-3706 - Make sure to be the first thread in the group commit FIFO …

KSauter:
…queue in any case. Before, it could happen that the unregister in `wsrep_after_commit` is too fast, and remove another thread in the queue which would result in a deadlock situation and a innodb crash after 600 seconds because of the stuck semaphore.

See https://jira.percona.com/browse/PXC-3706

### DIFF
--- a/sql/wsrep_binlog.cc
+++ b/sql/wsrep_binlog.cc
@@ -26,6 +26,7 @@
 #include "service_wsrep.h"
 #include "transaction.h"
 #include "wsrep_applier.h"
+#include "mutex_lock.h"
 
 /*
   Write the contents of a cache to a memory buffer.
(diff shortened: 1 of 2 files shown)


sql/wsrep_trans_observer.h
```diff
@@ -373,6 +373,7 @@ static inline int wsrep_after_commit(THD *thd, bool all) {
     thread handler to register in wsrep group commit queue but since storage
     engine commit is not done it would fail to unregister the said thread
     handler as part of storage engine commit. Handle unregistration here. */
+    wsrep_wait_for_turn_in_group_commit(thd);
```
venkatesh-prasad-v, line 376: Now that we are touching this part of the code, I think it would better to fix the other issues found in this code as well.

1. In https://github.com/percona/percona-xtradb-cluster/blob/8.0/sql/wsrep_binlog.cc#L474-L475, 
  the order of the calls to `mysql_mutex_unlock(&LOCK_wsrep_group_commit)` and `mysql_cond_broadcast(&COND_wsrep_group_commit)` seem to be wrong. The correct order is 
```
  mysql_cond_broadcast(&COND_wsrep_group_commit);
  mysql_mutex_unlock(&LOCK_wsrep_group_commit);
```


2. Can you please add an assert that the current thread is in the front of the queue when it is trying to unregister from the queue.
https://github.com/percona/percona-xtradb-cluster/blob/8.0/sql/wsrep_binlog.cc#L469
Suggestion:
```
  thd->wsrep_enforce_group_commit = false;
  assert(wsrep_group_commit_queue.front() == thd);
  wsrep_group_commit_queue.pop();
```

3. Can we use mutex lock guards in all occurences where we acquire `LOCK_wsrep_group_commit`
Suggestion:
```
#include "mutex_lock.h" and use
MUTEX_LOCK (lock_guard, &LOCK_wsrep_group_commit)
instead of
mysql_mutex_lock(&LOCK_wsrep_group_commit);
```

4. This is optional, but it would be better to have `DBUG_TRACE` and `return` statements instead of combinations of `DBUG_ENTER` and `DBUG_VOID_RETURN`

venkatesh-prasad-v (commented):
Hi @KSauter! Thank you very much for the contribution and we appreciate your efforts. The patch looks good and fixes the problem.

If you don't mind, have a few suggestions below. Can you please check the below comment?

Best regards,

sql/wsrep_trans_observer.h
KSauter: Hi @venkatesh-prasad-v ,

thanks for your review. I had look on your suggestions and did a cleanup like you suggested.
Please have a look again.

Best,
Kevin

sql/wsrep_binlog.cc
```diff
@@ -434,44 +430,40 @@ void wsrep_wait_for_turn_in_group_commit(THD *thd) {
...
+       during recovery using XID. */
+    return;
   }
 
-  mysql_mutex_lock(&LOCK_wsrep_group_commit);
+  MUTEX_LOCK(guard, &LOCK_wsrep_group_commit);
 
   if (!thd->wsrep_enforce_group_commit) {
     /* Said handler was not register for wsrep group commit */
     mysql_mutex_unlock(&LOCK_wsrep_group_commit);
```
venkatesh-prasad-v: This `mysql_mutex_unlock(&LOCK_wsrep_group_commit);` also needs to be removed as this is no longer needed.

venkatesh-prasad-v:
Testing link: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/292/console (In progress)

venkatesh-prasad-v (changes requested):

sql/wsrep_binlog.cc
KSauter: Fixed in bef8fc0d66d2d8caf6caed762cafb3dd16cbe9b6

venkatesh-prasad-v:
Thank you for addressing the review comments. Your fix is currently being tested on our internal testing framework and will update you if there are any failures.

KSauter:
Hi @venkatesh-prasad-v , any update on this? Did the tests run fine?

dutow (approved):

venkatesh-prasad-v:
Hi @KSauter! Can you please squash all commits into one single commit?

KSauter:
Hi @venkatesh-prasad-v, sure. Please have a look again :)

venkatesh-prasad-v (approved):

PR #1472, opened 2021-07-19
PXC-3181 Shutdown signal does not shutdown the node used to bootstrap the cluster. (8.0)

patrickbirch:
Merged 5.7 version


nastena1606 (approved):
LGTM

PR #1473, opened 2021-07-21
[5.7] PXC-3710: PXC server causes the client libraries to crash

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3710

This is a follow-up of commit 0dc85026.

Problem
-------
PXC node can send malformed packets to client and can cause client to fail with
asserion

  `!check_buffer || (vio_pending(net->vio) <= 1)' in net_clear().

Background
----------
For any query that sends result set to the client, the server does the following
things as per classic protocol.

1. Result metadata specifying the number of fields in the result set
   i.e, `protocol->start_result_metadata()`.
2. The field metadata specifying the types of fields in the result set
   i.e, `protocol->send_field_metadata()` followed by
   `protocol->end_result_metadata()`.
3. Actual row data (any data sent between `protocol->start_row()` and
   `protocol->end_row()`
4. An OK/EOF packet indicating the end of result set.
   `protocol->send_eof()` / `protocol->send_ok()` usually called from
   `THD::send_statement_status()` in the end of `dispatch_command()`

Analysis
--------
When the server is running in autocommit mode and when it is executing a query
that sends some result set to client programs and it was BF aborted by TOI or
high priority transactions, the `wsrep_retry_autocommit` mechanism comes into
effect and the server retries the autocommit query withouy returning the error
to the client.

However, when the server is retrying the query, it is possible that the client
program may have already received partial result from server and may have been
already waiting for the OK/EOF packet to report it to the user (i.e, Steps 1-3
are over and waiting for the step 4 to happen).

In such a scenario, when a retry is performed, the server shall start executing
from Step-1 to Step-4 and if the query execution is successful, it sends OK/EOF
packet in the end to indicate that the query is complete. But on the client
side, this causes the client program to receive unexepected result metadata in
place of an OK/EOF packet and thus causes the client to error out with Malformed
packet error.

This issue was earlier seen with CHECK TABLE query and was fixed as part
of commit 0dc85026. However, didn't fix it for ALTER TABLE..CHECK
PARTITION, as both commands have same path of excution. This commit adds
additional check for ALTER TABLE CHECK PARTITION in
`wsrep_should_retry_in_autocommit()` to avoid retry after BF-abort.

Note to reviewers
---
This is not a GCA based commit as the dependent commit 0dc85026 doesn't exist in the GCA branch.

Testing Done
---
Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/178/console
Failing tests: None

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -7969,7 +7969,7 @@ static uint kill_one_thread(THD *thd, my_thread_id id, bool only_kill_query)
 }
 
 #ifdef WITH_WSREP
-static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command)
+static bool wsrep_should_retry_in_autocommit(const THD* thd)
 {
   /*
     We are here could mean that the query resulted in a cluster-wide
@@ -7992,11 +7992,17 @@ static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command)
     same symptom is found for other commands, then please add it to the
     below list.
   */
-  switch (sql_command)
+
+  switch (thd->lex->sql_command)
   {
     case SQLCOM_CHECK:
     case SQLCOM_SELECT:
       return false;
+    case SQLCOM_ALTER_TABLE:
+    {
+      return (thd->lex->alter_info.flags & Alter_info::ALTER_ADMIN_PARTITION ?
+              false : true);
+    }
     default:
       return true;
   }
@@ -8105,7 +8111,7 @@ static void wsrep_mysql_parse(THD *thd, const char *rawbuf, uint length,
         mysql_reset_thd_for_next_command(thd);
         thd->killed= THD::NOT_KILLED;
         if (is_autocommit &&
-            wsrep_should_retry_in_autocommit(thd->lex->sql_command) &&
+            wsrep_should_retry_in_autocommit(thd) &&
            (thd->wsrep_retry_counter < thd->variables.wsrep_retry_autocommit))
         {
           WSREP_DEBUG("Retrying auto-commit query (on abort): %s", WSREP_QUERY(thd));
--- a/mysql-test/suite/galera/r/galera_disable_retry_for_check_table.result
+++ b/mysql-test/suite/galera/r/galera_disable_retry_for_check_table.result
@@ -6,7 +6,7 @@ CREATE TABLE t1(i INT PRIMARY KEY);
 # 2. Execute CHECK TABLE query on node_1 and halt the query in
 #    ha_innobase::check() function.
 SET DEBUG_SYNC="ha_innobase_check SIGNAL reached WAIT_FOR continue";
-CHECK TABLE t1;;
+CHECK TABLE t1;
 [node_1b]
 SET DEBUG_SYNC="now WAIT_FOR reached";
 #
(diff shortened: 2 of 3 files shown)


sql/sql_parse.cc
```diff
@@ -7992,11 +7992,22 @@ static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command)
...
       return false;
+    case SQLCOM_ALTER_TABLE:
+    {
+      if (thd->lex->alter_info.flags & Alter_info::ALTER_ADMIN_PARTITION)
+      {
+        return false;
+      }
+      else
+        return true;
+    }
```
satya-bodapati: CHECK TABLE doesn't return rows to the client, right?

mysql> check table t1;
+---------+-------+----------+----------+
| Table   | Op    | Msg_type | Msg_text |
+---------+-------+----------+----------+
| test.t1 | check | status   | OK       |
+---------+-------+----------+----------+

why is CHECK TABLE /ALTER CHECK special?

Also, instead of aborting retry, cant we reset (or end the current data sent to client) and retry transaction?

venkatesh-prasad-v: >CHECK TABLE doesn't return rows to the client, right?

It does send a row and the row itself is
```
+---------+-------+----------+----------+
| Table | Op | Msg_type | Msg_text |
+---------+-------+----------+----------+
| test.t1 | check | status | OK |
+---------+-------+----------+----------+
```
> why is CHECK TABLE /ALTER CHECK special?

This is because CHECK TABLE is not replicated and any High Priority Transaction can bf-abort CHECK TABLE/ALTER CHECK command.


>Also, instead of aborting retry, cant we reset (or end the current data sent to client) and retry transaction?

I checked the implementation of protocol_classic, and I didn't see any functions for clearing the net buffers. May be it was by design that there should not be a way to clear them (probably to catch bugs like this). I'm not sure.

IMO, implementing it from scratch may have unforeseen effects on protocol/client (we will never know), and may also need lot of testing. So, let us leave it as is for now.

satya-bodapati: ok, this (OK output) is considered as a single row. Instead of relying on thd->lex->sql_command, is there any other way of knowing we have sent at least a single row? this would be generic

venkatesh-prasad-v: I see `Protocol_classic::get_output_pkt_nr()` function to return the number of packets written to net buffers, but this function is not used anywhere in the code to check our assumption is valid or not.

venkatesh-prasad-v: Also, since `net.pkt_nr` is incremented on every `my_net_write()`, we cannot ensure that we have sent the actual rows (they are incremented even while sending metadata).

satya-bodapati: Thanks for checking. It does look a bit complicated. Lets go with sql_command way for now

sql/sql_parse.cc
```diff
@@ -7992,11 +7992,22 @@ static bool wsrep_should_retry_in_autocommit(enum_sql_command &sql_command)
...
     case SQLCOM_SELECT:
       return false;
+    case SQLCOM_ALTER_TABLE:
+    {
+      if (thd->lex->alter_info.flags & Alter_info::ALTER_ADMIN_PARTITION)
+      {
+        return false;
+      }
+      else
+        return true;
```
satya-bodapati: optional: return (thd->lex.. & ALTER_ADMIN ? false : true)

satya-bodapati (approved):
Please ensure the testcase passes on Jenkins and then merge

sql/sql_parse.cc
venkatesh-prasad-v: Done.

venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/178/testReport/
Failing tests: None

PR #1481, opened 2021-08-16
PXC-3265: Implement NBO OSU mode

dutow:
This commit implements non blocking updates for ALTER TABLE, CREATE
INDEX and DROP INDEX.

After this change, it'll be possible to set wsrep_osu_method to NBO for
a session, and execute these statements in non blocking mode.

Compared to TOI, NBO mode uses more relaxed synchronization: it takes
MDL locks and executes the transactions on all nodes, and then commits
them after all completes successfully.


sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+                client_state.toi_meta().seqno().get(), WSREP_QUERY(thd));
+    WSREP_DEBUG("NBO END start: %lld: %s",
+                client_state.transaction().ws_meta().seqno().get(),
+                WSREP_QUERY(thd));
+    // assert(strstr(WSREP_QUERY(thd), "f1") == 0);
+    int ret = client_state.begin_nbo_phase_two(thd->wsrep_nbo.keys);
+
+    if (ret) {
+      assert(client_state.current_error());
+      // HERE's the cnonconflicting issue
```
```suggestion
      // HERE's the conflicting issue
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+        WSREP_QUERY(thd), (long long)wsrep_thd_trx_seqno(thd),
+        wsrep_thd_client_mode_str(thd));
+
+    snprintf(thd->wsrep_info, sizeof(thd->wsrep_info),
+             "wsrep: TO isolation initiated for write set (%lld)",
+             (long long)wsrep_thd_trx_seqno(thd));
+    WSREP_DEBUG("%s", thd->wsrep_info);
+    thd_proc_info(thd, thd->wsrep_info);
+
+    /* DDL transaction are now atomic so append wsrep xid.
```
```suggestion
    /* DDL transactions are now atomic, so append wsrep xid.
```
OR
```suggestion
    /* DDL transaction is now atomic, so append wsrep xid.
```

sql/sql_show.cc
```diff
@@ -2998,7 +2998,8 @@ class Fill_process_list : public Do_THD_Impl {
             : client_priv_user;
     ulonglong now_utime = my_micro_time();
 
-    if ((!inspect_thd->get_protocol()->connection_alive() &&
+    if ((inspect_thd->get_protocol() != nullptr &&
```
venkatesh-prasad-v: This is a PXC only change. So, I think this should be inside `#ifdef WITH_WSREP` block

sql/sql_show.cc
```diff
@@ -4417,13 +4419,11 @@ static int get_schema_tmp_table_columns_record(THD *thd, TABLE_LIST *tables,
...
-            : field->is_flag_set(UNIQUE_KEY_FLAG)
-                  ? "UNI"
-                  : (field->is_flag_set(MULTIPLE_KEY_FLAG))
-                        ? "MUL"
-                        : (field->is_flag_set(CLUSTERING_FLAG)) ? "CLU" : "");
+        field->is_flag_set(PRI_KEY_FLAG)          ? "PRI"
+        : field->is_flag_set(UNIQUE_KEY_FLAG)     ? "UNI"
+        : (field->is_flag_set(MULTIPLE_KEY_FLAG)) ? "MUL"
+        : (field->is_flag_set(CLUSTERING_FLAG))   ? "CLU"
+                                                  : "");
```
venkatesh-prasad-v: Is this change due to clang-format-10? 

I think it is better to have this code look like PS https://github.com/percona/percona-server/blob/release-8.0.26-16/sql/sql_show.cc#L4426-L4432.

sql/sql_table.cc
```diff
@@ -17416,7 +17430,8 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
     //
     // i.e, Upgrade to SHARED LOCK when thread is a WSREP thd, and exec mode is
     // either TOI or replicated and lock acquisition is a variant of NO_LOCK.
-    if ((wsrep_thd_is_toi(thd) || wsrep_thd_is_in_rsu(thd)) &&
+    if ((wsrep_thd_is_toi(thd) ||
+         wsrep_thd_is_in_rsu(thd) /* || wsrep_thd_is_in_nbo(thd) */) &&
```
venkatesh-prasad-v: This block was added while fixing PXC-3418, and I don't think this extensive locking is required for NBO. Please revert this change.

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+int wsrep_nbo_isolation_end_begin(THD *thd) {
+  WSREP_DEBUG("NBO END maybe start");
+  wsrep::client_state &client_state(thd->wsrep_cs());
+  if (wsrep_thd_is_in_nbo(thd)) {
+    WSREP_DEBUG("NBO END start: %lld: %s",
+                client_state.toi_meta().seqno().get(), WSREP_QUERY(thd));
+    WSREP_DEBUG("NBO END start: %lld: %s",
+                client_state.transaction().ws_meta().seqno().get(),
+                WSREP_QUERY(thd));
+    // assert(strstr(WSREP_QUERY(thd), "f1") == 0);
```
venkatesh-prasad-v: Dead code?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+      }
+
+      return 1;
+    }
+  }
+  return 0;
+}
+
+static void wsrep_NBO_end(THD *thd) {
+  wsrep_to_isolation--;
```
venkatesh-prasad-v: Can we use pre-decrement here?
```suggestion
  --wsrep_to_isolation;
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+    wsrep::mutable_buffer err;
+    if (thd->is_error() && !wsrep_must_ignore_error(thd)) {
+      // why does this initiale a vote??? :(
+      wsrep_store_error(thd, err);
+    }
+    int ret = client_state.end_nbo_phase_two(err);
+
+    if (!ret) {
+      WSREP_DEBUG("TO END: %lld", client_state.nbo_meta().seqno().get());
```
```suggestion
      WSREP_DEBUG("NBO END: %lld", client_state.nbo_meta().seqno().get());
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+      /* Reset XID on completion of DDL transactions */
+      bool atomic_ddl = is_atomic_ddl(thd, true);
+      if (atomic_ddl) {
+        thd->get_transaction()->xid_state()->get_xid()->reset();
+      }
+    } else {
+      WSREP_WARN("TO isolation end failed for: %d, schema: %s, sql: %s", ret,
```
```suggestion
      WSREP_WARN("NBO end failed for: %d, schema: %s, sql: %s", ret,
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+      /* jump to error handler in mysql_execute_command() */
+      switch (client_state.current_error()) {
+        default:
+          WSREP_WARN(
+              "NBO isolation failed for: %d, schema: %s, sql: %s. "
+              "Check wsrep connection state and retry the query.",
+              ret, (thd->db().str ? thd->db().str : "(null)"),
+              WSREP_QUERY(thd));
+          if (!thd->is_error()) {
+            WSREP_WARN("Adjusting error to deadlock");
```
venkatesh-prasad-v: I don't think this message needs to be logged as a warning. Can we change it to
```suggestion
            WSREP_DEBUG("Adjusting error to deadlock");
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+      // HERE's the cnonconflicting issue
+      WSREP_DEBUG("begin_nbo_phase_two() failed for %u: %s, seqno: %lld",
+                  thd->thread_id(), WSREP_QUERY(thd),
+                  (long long)wsrep_thd_trx_seqno(thd));
+
+      /* jump to error handler in mysql_execute_command() */
+      switch (client_state.current_error()) {
+        default:
+          WSREP_WARN(
+              "NBO isolation failed for: %d, schema: %s, sql: %s. "
```
venkatesh-prasad-v: Here and at other places. Do we need to mention the word `isolation` here?

sql/wsrep_high_priority_service.cc
```diff
@@ -636,11 +645,182 @@ int Wsrep_applier_service::apply_write_set(const wsrep::ws_meta &ws_meta,
...
-  return 0;
+PSI_thread_key key_nbo_thread;
+
+static PSI_thread_info nbo_threads[] = {
+    {&key_nbo_thread, "NBO update thread", 0, 0, PSI_DOCUMENT_ME}};
+
+int Wsrep_applier_service::apply_nbo_begin(const wsrep::ws_meta &ws_meta,
+                                           const wsrep::const_buffer &data,
+                                           wsrep::mutable_buffer & /*err*/) {
+  DBUG_ENTER("Wsrep_applier_service::apply_nbo_begin");
```
venkatesh-prasad-v: `DBUG_ENTER` is not being used anywhere in MySQL. Can we use `DBUG_TRACE` instead?

sql/wsrep_high_priority_service.cc
```diff
@@ -636,11 +645,182 @@ int Wsrep_applier_service::apply_write_set(const wsrep::ws_meta &ws_meta,
...
+    client_state.before_command();
+    client_state.before_statement();
+    int ret = client_state.enter_nbo_mode(ws_meta);
+
+    entered_nbo_mode = true;
+    cv.notify_one();
+
+    assert(ret == 0);
+    if (ret != 0) {
+    }
```
venkatesh-prasad-v: Dead code?

sql/wsrep_high_priority_service.cc
```diff
@@ -636,11 +645,182 @@ int Wsrep_applier_service::apply_write_set(const wsrep::ws_meta &ws_meta,
...
+
+  Wsrep_non_trans_mode non_trans_mode(m_thd, ws_meta);
+
+  const char *category = "sql";
+  mysql_thread_register(category, nbo_threads, 1);
+
+  snprintf(m_thd->wsrep_info, sizeof(m_thd->wsrep_info),
+           "wsrep: applying NBO write-set (%lld)",
+           (long long)wsrep_thd_trx_seqno(m_thd));
+  WSREP_DEBUG("%s", m_thd->wsrep_info);
```
venkatesh-prasad-v: Not an issue.

Here, the same message gets logged from both the parent and the child thread. Wouldn't this be redundant?
If this is required, then does it make sense to log it whether it is parent or child thread that is logging the message?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+  }
+  return 0;
+}
+
+static void wsrep_NBO_end(THD *thd) {
+  wsrep_to_isolation--;
+
+  wsrep::client_state &client_state(thd->wsrep_cs());
+  WSREP_DEBUG("NBO END: %lld: %s", client_state.nbo_meta().seqno().get(),
+              WSREP_QUERY(thd));
```
venkatesh-prasad-v: Compilation fails here
```
[ 79%] Building CXX object sql/CMakeFiles/sql_main.dir/wsrep_var.cc.o
make[2]: *** [sql/CMakeFiles/sql_main.dir/build.make:141: sql/CMakeFiles/sql_main.dir/wsrep_high_priority_service.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from /home/venki/work/pxc/review/nbo/sql/sql_class.h:165,
                 from /home/venki/work/pxc/review/nbo/sql/rpl_commit_stage_manager.h:32,
                 from /home/venki/work/pxc/review/nbo/sql/binlog.h:52,
                 from /home/venki/work/pxc/review/nbo/sql/wsrep_mysqld.cc:16:
/home/venki/work/pxc/review/nbo/sql/wsrep_mysqld.cc: In function ‘void wsrep_NBO_end(THD*)’:
/home/venki/work/pxc/review/nbo/sql/wsrep_mysqld.cc:2484:49: error: ‘class wsrep::client_state’ has no member named ‘nbo_meta’; did you mean ‘wsrep::ws_meta wsrep::client_state::nbo_meta_’? (not accessible from this context)
 2484 |   WSREP_DEBUG("NBO END: %lld: %s", client_state.nbo_meta().seqno().get(),
      |                                                 ^~~~~~~~
/home/venki/work/pxc/review/nbo/sql/wsrep_mysqld.h:283:25: note: in definition of macro ‘WSREP_LOG’
  283 |         .message(fmt, ##__VA_ARGS__);                              \
      |                         ^~~~~~~~~~~
/home/venki/work/pxc/review/nbo/sql/wsrep_mysqld.cc:2484:3: note: in expansion of macro ‘WSREP_DEBUG’
 2484 |   WSREP_DEBUG("NBO END: %lld: %s", client_state.nbo_meta().seqno().get(),
      |   ^~~~~~~~~~~
```
Can you check this?

venkatesh-prasad-v (changes requested):
Hi Zsolt! Thank you for working on this. Here are my first round of review comments.

sql/wsrep_mysqld.cc
dutow: https://github.com/percona/wsrep-lib/pull/7 

Looks like I didn't add this function in my wsrep-lib change PR, but I had it in my local NBO checkout.

venkatesh-prasad-v: Can you please merge the wsrep-lib PR and update this branch?

sql/wsrep_mysqld.cc
dutow: this was a debugging comment for me, removed.

sql/wsrep_mysqld.cc
dutow: fixed (also in the TOI function, I copy pasted it from there)

sql/sql_show.cc
dutow: This change is no longer required, I fixed this issue since then, I reverted it.

sql/sql_show.cc
dutow: I accidentally ran clang-format-12 on the file, fixed.

sql/sql_table.cc
dutow: removed.

sql/wsrep_mysqld.cc
dutow: removed.

sql/wsrep_mysqld.cc
dutow: fixed, and adjusted the TOI method similarly.

sql/wsrep_mysqld.cc
dutow: fixed

sql/wsrep_mysqld.cc
dutow: fixed

sql/wsrep_mysqld.cc
dutow: fixed, and removed the commented block below.

sql/wsrep_mysqld.cc
dutow: removed.

sql/wsrep_high_priority_service.cc
dutow: changed.

sql/wsrep_high_priority_service.cc
dutow: yes, I used it earlier for debugging, removed.

sql/wsrep_high_priority_service.cc
dutow: Removed WSREP_DEBUG

sql/wsrep_mysqld.cc
dutow: done.

sql/wsrep_mysqld.cc
```diff
@@ -2494,20 +2488,19 @@ static void wsrep_NBO_end(THD *thd) {
...
-      WSREP_DEBUG("TO END: %lld", client_state.nbo_meta().seqno().get());
+      WSREP_DEBUG("NBO END: %lld", client_state.nbo_meta().seqno().get());
       /* Reset XID on completion of DDL transactions */
       bool atomic_ddl = is_atomic_ddl(thd, true);
       if (atomic_ddl) {
         thd->get_transaction()->xid_state()->get_xid()->reset();
       }
     } else {
-      WSREP_WARN("TO isolation end failed for: %d, schema: %s, sql: %s", ret,
+      WSREP_WARN("NBO isolation end failed for: %d, schema: %s, sql: %s", ret,
```
venkatesh-prasad-v: Remove the word `isolation` here.

mysql-test/suite/galera_ee/t/GCF-329A.test
```diff
@@ -40,4 +40,4 @@ CALL mtr.add_suppression("Aborting");
 CALL mtr.add_suppression("Plugin 'InnoDB' will be forced to shutdown");
 CALL mtr.add_suppression("Slave SQL: Error 'Got error 7 during COMMIT'");
 CALL mtr.add_suppression("Adjusting error to deadlock");
```
venkatesh-prasad-v: This is no more a warning after it has been made to use `WSREP_DEBUG`. Can you please remove this suppression?

sql/sql_show.cc
```diff
@@ -4345,8 +4344,7 @@ static int fill_temporary_tables(THD *thd, TABLE_LIST *tables, Item *cond) {
 
   for (tmp = thd->temporary_tables; tmp; tmp = tmp->next) {
     if (store_temporary_table_record(thd, tables->table, tmp,
-                                     thd->lex->query_block->db,
-                                     thd->mem_root)) {
+                                     thd->lex->query_block->db, thd->mem_root)) {
```
venkatesh-prasad-v: Running clang-format-10 on this file produces a different output for this block. Can you check it?

```
diff --git a/sql/sql_show.cc b/sql/sql_show.cc
index c5a68592db4..65d7aabe8b4 100644
--- a/sql/sql_show.cc
+++ b/sql/sql_show.cc
@@ -4344,7 +4344,8 @@ static int fill_temporary_tables(THD *thd, TABLE_LIST *tables, Item *cond) {
 
   for (tmp = thd->temporary_tables; tmp; tmp = tmp->next) {
     if (store_temporary_table_record(thd, tables->table, tmp,
-                                     thd->lex->query_block->db, thd->mem_root)) {
+                                     thd->lex->query_block->db,
+                                     thd->mem_root)) {
       DBUG_RETURN(1);
     }
   }
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+int wsrep_nbo_isolation_end_begin(THD *thd) {
+  WSREP_DEBUG("NBO END maybe start");
+  wsrep::client_state &client_state(thd->wsrep_cs());
+  if (wsrep_thd_is_in_nbo(thd)) {
+    WSREP_DEBUG("NBO END start: %lld: %s",
+                client_state.toi_meta().seqno().get(), WSREP_QUERY(thd));
+    WSREP_DEBUG("NBO END start: %lld: %s",
+                client_state.transaction().ws_meta().seqno().get(),
+                WSREP_QUERY(thd));
```
venkatesh-prasad-v: Looks that both these WSREP_DEBUG print same value.
```
2021-08-18T08:09:37.904049Z 17 [Note] [MY-000000] [WSREP] NBO END start: -1: alter table t1 engine=innodb
2021-08-18T08:09:37.904076Z 17 [Note] [MY-000000] [WSREP] NBO END start: -1: alter table t1 engine=innodb
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+  if (wsrep_can_run_in_nbo(thd, db_, table_, table_list) == false) {
+    WSREP_DEBUG("Can't execute %s in NBO mode", WSREP_QUERY(thd));
+    my_error(ER_NOT_SUPPORTED_YET, MYF(0), "Can't execute yet in NBO mode");
+    return -1;
+  }
+
+  WSREP_DEBUG(
+      "Executing Query (%s) with write-set (%lld) and exec_mode: %s"
+      " in NBO Isolation mode",
```
venkatesh-prasad-v: Remove the word `Isolation` here.
```
2021-08-18T08:09:37.860245Z 17 [Note] [MY-000000] [WSREP] Executing Query (alter table t1 engine=innodb) with write-set (-1) and exec_mode: local in NBO Isolation mode
```

sql/sql_parse.h
```diff
@@ -146,10 +146,17 @@ extern const LEX_CSTRING command_name[];
...
 
-#define WSREP_TO_ISOLATION_END                                                 \
-  if ((WSREP(thd) && wsrep_thd_is_local_toi(thd)) || wsrep_thd_is_in_rsu(thd)) \
+#define WSREP_TO_ISOLATION_END                                 \
+  if ((WSREP(thd) && wsrep_thd_is_local_toi(thd)) ||           \
+      wsrep_thd_is_in_rsu(thd) || wsrep_thd_is_local_nbo(thd)) \
     wsrep_to_isolation_end(thd);
 
+#define WSREP_TO_ISOLATION_BEFORE_MDL_RELEASE \
+  if (WSREP(thd) && wsrep_nbo_isolation_end_begin(thd)) goto error;
```
venkatesh-prasad-v: `wsrep_nbo_isolation_end_begin()` is called unconditionally even in TOI mode. Can we check for `wsrep_thd_is_in_nbo()` before calling `wsrep_nbo_isolation_end_begin()`?

As of now, `NBO END maybe start` gets into the log for every command that is executed on the server with wsrep_debug enabled.
```
2021-08-18T08:13:04.534013Z 17 [Note] [MY-000000] [WSREP] Executing Query (alter table t1 engine=innodb) with write-set (-1) and exec_mode: local in TO Isolation mode
2021-08-18T08:13:04.534060Z 17 [Note] [MY-000000] [WSREP] wsrep: initiating TOI for write set (-1)
2021-08-18T08:13:04.534568Z 17 [Note] [MY-000000] [WSREP] Query (alter table t1 engine=innodb) with write-set (9) and exec_mode: toi replicated in TO Isolation mode
2021-08-18T08:13:04.534580Z 17 [Note] [MY-000000] [WSREP] wsrep: TO isolation initiated for write set (9)
2021-08-18T08:13:04.568648Z 17 [Note] [MY-000000] [WSREP] Updating WSREPXid: 2b3633ae-fffb-11eb-b6b0-efed34410649:9
2021-08-18T08:13:04.576848Z 17 [Note] [MY-000000] [WSREP] NBO END maybe start
2021-08-18T08:13:04.576945Z 17 [Note] [MY-000000] [WSREP] TO END: 9: alter table t1 engine=innodb

```

sql/sql_parse.h
```diff
@@ -146,10 +146,17 @@ extern const LEX_CSTRING command_name[];
...
+
+#define WSREP_TO_ISOLATION_AFTER_MDL_ACQUIRED \
```
venkatesh-prasad-v: Can we rename this to `WSREP_NBO_AFTER_MDL_ACQUIRED`?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
   }
 }
 
+static int wsrep_NBO_begin(THD *thd, const char *db_, const char *table_,
```
venkatesh-prasad-v: We are having too many confusing functions like `wsrep_NBO_begin`,`wsrep_nbo_isolation_begin_end()`, `wsrep_nbo_isolation_end_begin()` and `wsrep_NBO_end`.

Can we try to avoid them and rename them to be aligned with `wsrep::client_state` member functions

Here is my suggestion:
```
+-------------------------------+---------------------------+
|            Existing           |          Proposed         |
+-------------------------------+---------------------------+
| wsrep_NBO_begin               | wsrep_NBO_begin_phase_one |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_begin_end | wsrep_NBO_end_phase_one   |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_end_begin | wsrep_NBO_begin_phase_two |
+-------------------------------+---------------------------+
| wsrep_NBO_end                 | wsrep_NBO_end_phase_two   |
+-------------------------------+---------------------------+
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+  thd->wsrep_gtid_event_buf_len = 0;
+  thd->wsrep_gtid_event_buf = NULL;
+
+  if (rc) wsrep_NBO_begin_failed(thd, NULL);
+
+  return rc;
+}
+
+void wsrep_nbo_isolation_begin_end(THD *thd) {
```
venkatesh-prasad-v: We are having too many confusing functions like `wsrep_NBO_begin`,`wsrep_nbo_isolation_begin_end()`, `wsrep_nbo_isolation_end_begin()` and `wsrep_NBO_end`.

Can we try to avoid them and rename them to be aligned with `wsrep::client_state` member functions

Here is my suggestion:
```
+-------------------------------+---------------------------+
|            Existing           |          Proposed         |
+-------------------------------+---------------------------+
| wsrep_NBO_begin               | wsrep_NBO_begin_phase_one |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_begin_end | wsrep_NBO_end_phase_one   |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_end_begin | wsrep_NBO_begin_phase_two |
+-------------------------------+---------------------------+
| wsrep_NBO_end                 | wsrep_NBO_end_phase_two   |
+-------------------------------+---------------------------+
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+  WSREP_DEBUG("NBO phase one ended");
+  if (ret) {
+    assert(thd->wsrep_cs().current_error());
+    WSREP_DEBUG("end_nbo_phase_one() failed for %u: %s, seqno: %lld",
+                thd->thread_id(), WSREP_QUERY(thd),
+                (long long)wsrep_thd_trx_seqno(thd));
+  }
+}
+
+int wsrep_nbo_isolation_end_begin(THD *thd) {
```
venkatesh-prasad-v: We are having too many confusing functions like `wsrep_NBO_begin`,`wsrep_nbo_isolation_begin_end()`, `wsrep_nbo_isolation_end_begin()` and `wsrep_NBO_end`.

Can we try to avoid them and rename them to be aligned with `wsrep::client_state` member functions

Here is my suggestion:
```
+-------------------------------+---------------------------+
|            Existing           |          Proposed         |
+-------------------------------+---------------------------+
| wsrep_NBO_begin               | wsrep_NBO_begin_phase_one |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_begin_end | wsrep_NBO_end_phase_one   |
+-------------------------------+---------------------------+
| wsrep_nbo_isolation_end_begin | wsrep_NBO_begin_phase_two |
+-------------------------------+---------------------------+
| wsrep_NBO_end                 | wsrep_NBO_end_phase_two   |
+-------------------------------+---------------------------+
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+  WSREP_DEBUG("NBO END maybe start");
```
venkatesh-prasad-v: Can we instead print "NBO Phase 2 started" or something like that?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,216 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+  WSREP_DEBUG("Starting NBO phase one");
+
+  mysql_mutex_lock(&thd->LOCK_thd_data);
+
+  bool is_thd_killed = thd->is_killed();
+
+  mysql_mutex_unlock(&thd->LOCK_thd_data);
+
+  if (is_thd_killed) {
```
venkatesh-prasad-v, line 2328: Not related to this issue. But when the NBO query is killed in between the execution, it seems that the cluster enters into `Disconnected` state.

```
Session 1:
mysql> SET GLOBAL debug = 'd,sync.alter_locked_tables_inplace';

Session 2:
mysql> show processlist;
+----+-----------------+-----------------+------+---------+------+--------------------------+-----------------------------------------+---------+-----------+---------------+
| Id | User            | Host            | db   | Command | Time | State                    | Info                                    | Time_ms | Rows_sent | Rows_examined |
+----+-----------------+-----------------+------+---------+------+--------------------------+-----------------------------------------+---------+-----------+---------------+
|  1 | system user     |                 | NULL | Sleep   | 7086 | wsrep aborter idle       | NULL                                    | 7086340 |         0 |             0 |
|  2 | system user     |                 | NULL | Sleep   | 6193 | innobase_commit_low (-1) | NULL                                    | 6192907 |         0 |             0 |
|  7 | event_scheduler | localhost       | NULL | Daemon  | 7084 | Waiting on empty queue   | NULL                                    | 7084327 |         0 |             0 |
| 18 | root            | localhost:38330 | test | Query   |    6 | debug sync point: now    | alter table t3 add column k varchar(10) |    6345 |         0 |             0 |
| 20 | root            | localhost:55980 | test | Query   |    0 | init                     | show processlist                        |       0 |         0 |             0 |
+----+-----------------+-----------------+------+---------+------+--------------------------+-----------------------------------------+---------+-----------+---------------+
5 rows in set (0.01 sec)

mysql> kill 18;

Session 1:
mysql> alter table t3 add column l varchar(10);
ERROR 1047 (08S01): WSREP has not yet prepared node for application use

mysql> show status like 'wsrep_cluster_status%';
+----------------------+--------------+
| Variable_name        | Value        |
+----------------------+--------------+
| wsrep_cluster_status | Disconnected |
+----------------------+--------------+
1 row in set (0.00 sec)
```

venkatesh-prasad-v (changes requested):
Hi Zsolt! Thank you for addressing the review comments. 
While I'm still trying to understand the entire feature which needs some more time, I thought of publishing my second round of comments. Please find them below.

sql/wsrep_mysqld.cc
dutow: I'll remove this line, it was only there for debugging purposes, we don't need it now.

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2296,223 @@ static void wsrep_TOI_end(THD *thd) {
...
+    }
+
+    ++wsrep_to_isolation;
+    rc = 0;
+  }
+
+  if (buf) my_free(buf);
+
+  thd->wsrep_gtid_event_buf_len = 0;
+  thd->wsrep_gtid_event_buf = NULL;
```
venkatesh-prasad-v: Can we just call `free_gtid_event_buf(thd)` as `buf` will be same as `thd->wsrep_gtid_event_buf`? (See `wsrep_to_buf_helper` for more details)

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+        }
+    }
+    rc = -1;
+    thd->wsrep_skip_wsrep_hton = false;
+  } else {
+    thd->wsrep_nbo.keys = key_array;
+
+    WSREP_DEBUG(
+        "Query (%s) with write-set (%lld) and exec_mode: %s"
+        " replicated in TO Isolation mode",
```
```suggestion
        " replicated in NBO mode",
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+        WSREP_QUERY(thd), (long long)wsrep_thd_trx_seqno(thd),
+        wsrep_thd_client_mode_str(thd));
+
+    snprintf(thd->wsrep_info, sizeof(thd->wsrep_info),
+             "wsrep: TO isolation initiated for write set (%lld)",
```
```suggestion
             "wsrep: NBO initiated for write set (%lld)",
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+                thd->thread_id(), WSREP_QUERY(thd),
+                (long long)wsrep_thd_trx_seqno(thd));
+    thd->wsrep_skip_wsrep_hton = false;
+  }
+}
+
+int wsrep_NBO_begin_phase_two(THD *thd) {
+  wsrep::client_state &client_state(thd->wsrep_cs());
+  if (wsrep_thd_is_in_nbo(thd)) {
+    WSREP_DEBUG("NBO END start: %lld: %s",
```
```suggestion
    WSREP_DEBUG("NBO Phase Two start: %lld: %s",
```

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+      assert(client_state.current_error());
+      WSREP_DEBUG("begin_nbo_phase_two() failed for %u: %s, seqno: %lld",
+                  thd->thread_id(), WSREP_QUERY(thd),
+                  (long long)wsrep_thd_trx_seqno(thd));
+
+      /* jump to error handler in mysql_execute_command() */
+      switch (client_state.current_error()) {
+        default:
+          WSREP_WARN(
+              "NBO failed for: %d, schema: %s, sql: %s. "
```
```suggestion
              "NBO Phase Two failed for: %d, schema: %s, sql: %s. "
```

sql/wsrep_mysqld.cc
```diff
@@ -2593,7 +2880,6 @@ int wsrep_must_ignore_error(THD *thd) {
   const uint flags = sql_command_flags[thd->lex->sql_command];
 
   assert(error);
-  assert(wsrep_thd_is_toi(thd));
```
venkatesh-prasad-v, line 2602: Instead of removing this assert, can we change the condition to 
```
assert(wsrep_thd_is_toi(thd) || wsrep_thd_is_in_nbo(thd));
```

storage/innobase/trx/trx0sys.cc
```diff
@@ -449,7 +449,7 @@ void trx_sys_update_wsrep_checkpoint(
         the same xid_seqno.
         Allow current and new values be the same, without introducing new flags
         and logic to prevent double storing of the same value */
-        ut_ad(xid_seqno >= trx_sys_cur_xid_seqno);
+        //ut_ad(xid_seqno >= trx_sys_cur_xid_seqno);
```
venkatesh-prasad-v: Is this change necessarily required? This assertion has helped us in finding the cases which causes multiple requests to InnoDB to persist the xid_seqno. See https://jira.percona.com/browse/PXC-3442.

If it is possible, can we retain this assertion and fix the underlying issue?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+  WSREP_DEBUG("Starting NBO phase one");
+
+  mysql_mutex_lock(&thd->LOCK_thd_data);
+
+  bool is_thd_killed = thd->is_killed();
+
+  mysql_mutex_unlock(&thd->LOCK_thd_data);
+
+  thd->wsrep_skip_wsrep_hton = true;
```
venkatesh-prasad-v: In other places, `wsrep_skip_wsrep_hton` is set under the protection of `thd->LOCK_thd_data`.
See `wsrep_to_isolation_end()` in `sql/wsrep_mysqld.cc`
``` 
  2694   DEBUG_SYNC(thd, "wsrep_to_isolation_end_before_wsrep_skip_wsrep_hton");


  2695   mysql_mutex_lock(&thd->LOCK_thd_data);
  2696 
  2697   thd->wsrep_skip_wsrep_hton = false;
  2698 
  2699   mysql_mutex_unlock(&thd->LOCK_thd_data);
  2700   DEBUG_SYNC(thd, "wsrep_to_isolation_end_after_wsrep_skip_wsrep_hton");
```
```suggestion
  thd->wsrep_skip_wsrep_hton = true;
  mysql_mutex_unlock(&thd->LOCK_thd_data);
```

venkatesh-prasad-v:
Also please extend the `galera_bf_bf*` testcases introduced as part of PXC-3449 with NBO for better coverage.

sql/sql_alter.cc
```diff
@@ -593,6 +594,8 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
   /* This could be looked upon as too restrictive given it is taking
   a global mutex but anyway being TOI if there is alter tablespace
   operation active in parallel TOI would streamline it. */
+  // TODO: the above comment is no longer true with NBO, this could be
```
kamil-holubicki, line 603: If not true anymore, why not update it?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+      /* jump to error handler in mysql_execute_command() */
+      switch (client_state.current_error()) {
+        default:
+          WSREP_WARN(
+              "NBO failed for: %d, schema: %s, sql: %s. "
+              "Check wsrep connection state and retry the query.",
+              ret, (thd->db().str ? thd->db().str : "(null)"),
+              WSREP_QUERY(thd));
+          // NOT adjusting: for whatever reason is_error is unset,
```
kamil-holubicki: dead code

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
   }
 }
 
+static int wsrep_NBO_begin_phase_one(THD *thd, const char *db_,
```
kamil-holubicki, line 2306: This function and `wsrep_TOI_begin()` seem to have exactly the same logic with the difference that they this one calls another validator (`wsrep_can_run_in_nbo` instead of `wsrep_can_run_in_toi` - but both have the same prototype) and this one calls `begin_nbo_phase_one` instead of `enter_toi_local` - again the same prototype). 
Wouldn't it be better to have the common code only once and call the different parts via delegates?

sql/wsrep_mysqld.cc
```diff
@@ -2242,6 +2295,232 @@ static void wsrep_TOI_end(THD *thd) {
...
+
+static void wsrep_NBO_end_phase_two(THD *thd) {
+  --wsrep_to_isolation;
+
+  wsrep::client_state &client_state(thd->wsrep_cs());
+  WSREP_DEBUG("NBO phase two END: %lld: %s",
+              client_state.nbo_meta().seqno().get(), WSREP_QUERY(thd));
+
+  if (wsrep_thd_is_in_nbo(thd)) {
+    THD_STAGE_INFO(thd, stage_wsrep_completed_TO_isolation);
```
kamil-holubicki, line 2518: `wsrep_set_SE_checkpoint()` not needed? See TOI end.

sql/wsrep_mysqld.cc
```diff
@@ -2463,15 +2749,16 @@ bool wsrep_handle_mdl_conflict(const MDL_context *requestor_ctx,
   mysql_mutex_lock(&request_thd->LOCK_wsrep_thd);
 
   if (wsrep_thd_is_toi(request_thd) || wsrep_thd_is_in_rsu(request_thd) ||
-      wsrep_thd_is_applying(request_thd)) {
+      wsrep_thd_is_applying(request_thd) || wsrep_thd_is_in_nbo(request_thd)) {
```
kamil-holubicki, line 2777: Just  a side note: The logic in this  function becomes more and more obfuscated :(

sql/sql_parse.h
```diff
@@ -146,10 +146,17 @@ extern const LEX_CSTRING command_name[];
...
+#define WSREP_TO_ISOLATION_END                                 \
+  if ((WSREP(thd) && wsrep_thd_is_local_toi(thd)) ||           \
+      wsrep_thd_is_in_rsu(thd) || wsrep_thd_is_local_nbo(thd)) \
     wsrep_to_isolation_end(thd);
 
+#define WSREP_NBO_BEFORE_MDL_RELEASE \
+  if (WSREP(thd) && wsrep_NBO_begin_phase_two(thd)) goto error;
+
+#define WSREP_NBO_AFTER_MDL_ACQUIRED \
+  if (wsrep_thd_is_local_nbo(thd)) wsrep_NBO_end_phase_one(thd);
```
kamil-holubicki: Shouldn't these 2 macros be similar? I mean in 1st case we have check wif in NBO mode inside `wsrep_NBO_begin_phase_two()`, but for the 2nd one the check is directly in the macro - it looks like it could be moved inside `wsrep_NBO_end_phase_one()`. 
No need to check WSREP(thd) in 2nd macro?

sql/sql_parse.h
```diff
@@ -146,10 +146,17 @@ extern const LEX_CSTRING command_name[];
...
                                alter_info_, fk_tables_))
 
-#define WSREP_TO_ISOLATION_END                                                 \
-  if ((WSREP(thd) && wsrep_thd_is_local_toi(thd)) || wsrep_thd_is_in_rsu(thd)) \
+#define WSREP_TO_ISOLATION_END                                 \
+  if ((WSREP(thd) && wsrep_thd_is_local_toi(thd)) ||           \
+      wsrep_thd_is_in_rsu(thd) || wsrep_thd_is_local_nbo(thd)) \
     wsrep_to_isolation_end(thd);
 
+#define WSREP_NBO_BEFORE_MDL_RELEASE \
```
kamil-holubicki: I would name these macros to reflect what they really do, like
WSREP_NBO_2ND_PAHASE_BEGIN

sql/wsrep_mysqld.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Fixed.

storage/innobase/trx/trx0sys.cc
dutow: Fixed.

sql/wsrep_mysqld.cc
dutow: Removed.

sql/wsrep_mysqld.cc
dutow: There are many small differeneces, I'm not sure if this would make it readable or easier to maintain.

sql/wsrep_mysqld.cc
dutow: No, I had there originally, but it causes test failures.

kamil-holubicki (commented):
Hi @dutow , As we've been chatting on Slack, please remove test.cpp and wm.bak files that were accidentally added.
I'm fine with everything else.

kamil-holubicki (approved):
lgtm

venkatesh-prasad-v:
OK to squash. Please run Jenkins once and add the link as a comment.

venkatesh-prasad-v (approved):
Thank you for addressing all the review comments. Great work on this! Approved.

PR #1482, opened 2021-08-17
Pxc 3731 5.7 po c

kamil-holubicki:


sql/wsrep_applier.cc
```diff
@@ -483,6 +493,7 @@ wsrep_cb_status_t wsrep_commit_cb(void*         const     ctx,
     thd->wsrep_apply_toi= false;
   }
 
+  thd->variables.option_bits = option_bits_save;
```
dutow: Shouldn't this go before the above if block? Seems like it's possible that we set the OPTION_BEGIN bit there, and then revert that change here.

dutow: Or instead of restoring the entire bitset, we could just remember if we changed OPTION_BIN_LOG or not, and restore that bit?

kamil-holubicki: Yeah, you are right. What is funny is that no test failed because of that, so above 'if' is either not covered or not needed.
I will change as suggested (remember bit state before, then restore)

PR #1485, opened 2021-08-26
PXC-3731: Fix behavior for sql_log_bin=0

kamil-holubicki:
Problem:
1. Fixing PXC-3464 (commit 39bf1f9b408cb163c15cd1d352464d48c5cc0851)
introduced the side effect that even with sql_log_bin=0, events are
logged into the binlog.

2. In the meantime the analysis showed another problem that queries
executed on PXC source with sql_log_bin=0 are replicated using Galera
replication, which is proper behavior, but replica logs these queries
into its binlog if log_slave_updates=1 which is wrong behavior.

Cause:
1. The fix for PXC-3464 was not proper. It was missing the cut-off of
events writing into the binlog

2. The problem is that the information about sql_log_bin state is not
through PXC replication channel, so there is no way to make a decision
on the replica side.

Solution:
1. Proper handling of the case when sql_log_bin=0 was added.

2. Two approaches were considered. The 1st one which propagates new flag
along with replication writeset on Galera replication layer (together
with already existing WSREP_FLAG_* -like flags). It needs changes in
server, wsrep-api, wsrep-lib, Galera layers.
The 2nd approach is to inject additional event at the beginning of
the replication writeset, controlling binlogging on replica side. It
needs changes in server layer only.

The 2nd approach was chosen mainly because its lower cost of further
maintenance and its simplicity.
Intvar_log_event was extended to pass the information about
sql_log_bin=0. Old versions of the server will skip this event and
behave in the old way.

This work is not based on GCA commit, because currently, GCA is too old
and does not contain all necessary changes (code and MTR suite fixes)

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -1734,7 +1734,17 @@ bool MYSQL_BIN_LOG::write_transaction(THD *thd, binlog_cache_data *cache_data,
   DBUG_PRINT("info",
              ("transaction_length= %llu", gtid_event.transaction_length));
 
+#ifdef WITH_WSREP
+  bool ret = 0;
+
+  if (!(thd->variables.option_bits & OPTION_BIN_LOG_INTERNAL_OFF) &&
+      !(thd->variables.option_bits & OPTION_BIN_LOG))
+    goto end;
+
+  ret = gtid_event.write(writer);
+#else
   bool ret = gtid_event.write(writer);
+#endif
   if (ret) goto end;
 
   /*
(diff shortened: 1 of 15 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-param/313/

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/316/

sql/binlog.cc
```diff
@@ -12382,6 +12392,44 @@ int THD::binlog_query(THD::enum_binlog_query_type qtype, const char *query_arg,
 }
 
 #ifdef WITH_WSREP
+int prepend_binlog_control_event(THD *const thd) {
+  unsigned char *read_pos = NULL;
```
venkatesh-prasad-v: Can we use `nullptr` here?

sql/log_event.cc
```diff
@@ -6065,11 +6078,12 @@ Log_event::enum_skip_reason Intvar_log_event::do_shall_skip(
   */
   return continue_group(rli);
 }
+#endif  // MYSQL_SERVER
 
 /**************************************************************************
   Rand_log_event methods
 **************************************************************************/
-
+#ifdef MYSQL_SERVER
```
venkatesh-prasad-v: Unrelated change?

sql/wsrep_binlog.cc
```diff
@@ -22,8 +22,10 @@
 #include "binlog.h"
 #include "log.h"
 #include "log_event.h"  // Log_event_writer
+#include "mutex_lock.h"
 #include "mysql/psi/mysql_file.h"
 #include "service_wsrep.h"
+#include "sql_base.h"
```
venkatesh-prasad-v: Probably not required anymore.

sql/wsrep_high_priority_service.cc
```diff
@@ -416,6 +427,9 @@ int Wsrep_high_priority_service::apply_toi(const wsrep::ws_meta &ws_meta,
                                            wsrep::mutable_buffer &err) {
   DBUG_TRACE;
   THD *thd = m_thd;
+
+  ulonglong option_bin_log_save = thd->variables.option_bits & OPTION_BIN_LOG;
+
```
venkatesh-prasad-v: I guess this is not necessay as `wsrep_apply_events()` may have handled this in a higher level. Can you check it?

venkatesh-prasad-v (changes requested):

sql/binlog.cc
kamil-holubicki: done

sql/log_event.cc
kamil-holubicki: fixed

sql/wsrep_binlog.cc

sql/wsrep_high_priority_service.cc
kamil-holubicki: fixed and added detailed explanation what happens here

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/322/

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/324/
no failures related to the change

venkatesh-prasad-v (approved):
Thanks for working on this issue. Approved!

PR #1486, opened 2021-08-26
PXC-3731: Fix behavior for sql_log_bin=0

kamil-holubicki:
Problem:
1. Fixing PXC-3464 (commit 4f83f7217782ec1c7541046d5458d380e7242150)
introduced the side effect that even with sql_log_bin=0, events are
logged into the binlog.

2. In the meantime the analysis showed another problem that queries
executed on PXC source with sql_log_bin=0 are replicated using Galera
replication, which is proper behavior, but replica logs these queries
into its binlog if log_slave_updates=1 which is wrong behavior.

Cause:
1. The fix for PXC-3464 was not proper. It was missing the cut-off of
events writing into the binlog

2. The problem is that the information about sql_log_bin state is not
through PXC replication channel, so there is no way to make a decision
on the replica side.

Solution:
1. Proper handling of the case when sql_log_bin=0 was added.

2. Two approaches were considered. The 1st one which propagates new flag
along with replication writeset on Galera replication layer (together
with already existing WSREP_FLAG_* -like flags). It needs changes in
server, wsrep-api, wsrep-lib, Galera layers.
The 2nd approach is to inject additional event at the beginning of
the replication writeset, controlling binlogging on replica side. It
needs changes in server layer only.

The 2nd approach was chosen mainly because its lower cost of further
maintenance and its simplicity.
Intvar_log_event was extended to pass the information about
sql_log_bin=0. Old versions of the server will skip this event and
behave in the old way.

This work is not based on GCA commit, because currently, GCA is too old
and does not contain all necessary changes (code and MTR suite fixes)

### DIFF
--- a/mysql-test/suite/galera/t/galera_log_bin.test
+++ b/mysql-test/suite/galera/t/galera_log_bin.test
@@ -50,5 +50,116 @@ SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
 DROP TABLE t1;
 DROP TABLE t2;
 
+--echo #
+--echo # Test that disabling binlog does not prevent Galera replication,
+--echo # but events are not recorded in the binlog
+--echo #
+--connection node_2
+RESET MASTER;
+
+--connection node_1
+RESET MASTER;
+
+--let $master_pos_before= query_get_value(SHOW MASTER STATUS,Position,1)
+SET SESSION sql_log_bin = OFF;
+CREATE TABLE t1 (a int primary key);
+INSERT INTO t1 VALUES (1);
+BEGIN;
+INSERT INTO t1 VALUES (10);
+INSERT INTO t1 VALUES (11);
+INSERT INTO t1 VALUES (12);
+COMMIT;
+
+# Assert that the above statements are not written to binlog.
+--let $master_pos_after= query_get_value(SHOW MASTER STATUS,Position,1)
+--let $assert_text= assert that the above events are not written to binlog
+--let $assert_cond= $master_pos_before = $master_pos_after
+--source include/assert.inc
+
+# check that turning it back to ON works
+SET SESSION sql_log_bin = ON;
+CREATE TABLE t2 (a int primary key);
+INSERT INTO t2 VALUES (1);
+BEGIN;
+INSERT INTO t2 VALUES (10);
+INSERT INTO t2 VALUES (11);
+INSERT INTO t2 VALUES (12);
+COMMIT;
+
+# we expect only t2 operations to be logged
+--replace_regex /table_id: [0-9]+/table_id: #/ /xid=[0-9]+/xid=#/
+SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
+
+# Check node_2
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't1'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't2'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 4 FROM t1
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 4 FROM t2
+--source include/wait_condition.inc
+
+# # we expect only t2 operations to be logged
+--replace_regex /table_id: [0-9]+/table_id: #/ /xid=[0-9]+/xid=#/
+SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
+
+DROP TABLE t1;
+DROP TABLE t2;
+
+--echo #
+--echo # Check that log-slave-updates=OFF block all PXC replicated binlogging on node_2
+--echo #
+--let $restart_parameters = "restart:--log-slave-updates=OFF --log-bin"
+--source include/restart_mysqld.inc
+
+--connection node_2
+RESET MASTER;
+--let $master_pos_before= query_get_value(SHOW MASTER STATUS,Position,1)
+
+--connection node_1
+RESET MASTER;
+
+SET SESSION sql_log_bin = OFF;
+CREATE TABLE t1 (a int primary key);
+INSERT INTO t1 VALUES (1);
+
+# check that turning it back to ON works
+SET SESSION sql_log_bin = ON;
+CREATE TABLE t2 (a int primary key);
+INSERT INTO t2 VALUES (1);
+
+# we expect only t2 operations to be logged
+--replace_regex /table_id: [0-9]+/table_id: #/ /xid=[0-9]+/xid=#/
+SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
+
+# Check node_2
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't1'
+--source include/wait_condition.inc
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't2'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t1
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t2
+--source include/wait_condition.inc
+
+# Assert that the above statements are not written to binlog.
+--let $master_pos_after= query_get_value(SHOW MASTER STATUS,Position,1)
+--let $assert_text= assert that the above events are not written to binlog
+--let $assert_cond= $master_pos_before = $master_pos_after
+--source include/assert.inc
+
+# we expect nothing in binlog
+SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
+
+DROP TABLE t1;
+DROP TABLE t2;
+
 --connection node_1
 RESET MASTER;
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -1599,7 +1599,17 @@ bool MYSQL_BIN_LOG::write_gtid(THD *thd, binlog_cache_data *cache_data,
   */
   thd->get_transaction()->last_committed= SEQ_UNINIT;
 
-
+#ifdef WITH_WSREP
+  /*
+    If binlog is disabled for this session, skip actual writing to the file.
+    But do not skip if it was disabled internally, and we still got here.
+   */
+  if (!(thd->variables.option_bits & OPTION_BIN_LOG_INTERNAL_OFF)
+     && !(thd->variables.option_bits & OPTION_BIN_LOG))
+  {
+      DBUG_RETURN(0);
+  }
+#endif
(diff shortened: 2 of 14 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.7-param/201

sql/sql_class.h
```diff
@@ -3550,6 +3550,7 @@ class THD :public MDL_context_owner,
   } binlog_evt_union;
 
 #ifdef WITH_WSREP
+  ulonglong                 wsrep_bin_log_flag_save;
```
venkatesh-prasad-v, line 3553: If this flag is used only in the context of the applier thread, then can we move it under `thd->wsrep_rli`? i.e, under `Relay_log_info` class.

sql/wsrep_binlog.cc
```diff
@@ -16,6 +16,7 @@
 #include "wsrep_binlog.h"
 #include "wsrep_priv.h"
 #include "log_event.h"
+#include "sql_base.h"
```
venkatesh-prasad-v: It would be better to add why this file was included. A convention which is used in Server code.
```suggestion
#include "sql_base.h"  // TEMP_PREFIX
```

sql/wsrep_binlog.cc
```diff
@@ -380,6 +381,48 @@ static int wsrep_write_cache_inc(wsrep_t*  const wsrep,
     return err;
 }
 
+static int prepend_binlog_control_event(wsrep_t* const wsrep, THD* const thd)
+{
+    if ((thd->variables.option_bits & OPTION_BIN_LOG) != 0)
+      return 0;
+
+    IO_CACHE tmp_io_cache;
+    if (open_cached_file(&tmp_io_cache, mysql_tmpdir, TEMP_PREFIX,
```
venkatesh-prasad-v: I think we must fill the cache with zeros to be on the safer side. 
```suggestion
    IO_CACHE tmp_io_cache;
    memset(&tmp_io_cache, 0, sizeof(stmt_cache_log));
    if (open_cached_file(&tmp_io_cache, mysql_tmpdir, TEMP_PREFIX,
```
The same has been done in `THD::binlog_setup_trx_data()` as well.

sql/wsrep_binlog.cc
```diff
@@ -380,6 +381,48 @@ static int wsrep_write_cache_inc(wsrep_t*  const wsrep,
...
+                        65536, MYF(MY_WME)))
```
venkatesh-prasad-v: I don't think there is a need to create a 64k buffer as we only write `val=0` and `magic='0xAABBCCDDDDCCBBAA'`.

It only requires 35 bytes to write the event.
```
(gdb) p length
$21 = 35
(gdb)  p  tmp_io_cache.write_pos - tmp_io_cache.write_buffer 
$22 = 35
```
Probably setting it to 64/128/256 should be fine for writing header, data and footer. Can you check this?

venkatesh-prasad-v (commented):
Posting my first set of comments so that we don't miss them if at all we choose this approach.

sql/sql_class.h
kamil-holubicki: I'm not sure. Relay_log_info class is not touched by WITH_WSREP. Technically we can do it, but for now, we keep everything in thd.

sql/wsrep_binlog.cc

sql/wsrep_binlog.cc
kamil-holubicki: This was just for PoC. Removed.

sql/wsrep_binlog.cc
kamil-holubicki: I don't think so. Please check `wsrep_to_buf_helper`.

kamil-holubicki: OK, after the discussion, let's initialize it in both places :)

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.7-param/204/

sql/binlog.cc
```diff
@@ -8827,6 +8837,14 @@ bool MYSQL_BIN_LOG::write_cache(THD *thd, binlog_cache_data *cache_data,
   DBUG_ENTER("MYSQL_BIN_LOG::write_cache(THD *, binlog_cache_data *, bool)");
 #ifdef WITH_WSREP
   if (WSREP_EMULATE_BINLOG(thd)) DBUG_RETURN(0);
+
+  /*
+    If binlog is disabled for this session, skip actual writing to the file.
+    But do not skip if it was disabled internally, and we still got here.
+   */
+  if(!(thd->variables.option_bits & OPTION_BIN_LOG_INTERNAL_OFF)
```
venkatesh-prasad-v: Space between `if` and `(`.

sql/binlog.cc
```diff
@@ -1599,7 +1599,17 @@ bool MYSQL_BIN_LOG::write_gtid(THD *thd, binlog_cache_data *cache_data,
   */
   thd->get_transaction()->last_committed= SEQ_UNINIT;
 
-
+#if WITH_WSREP
+  /*
+    If binlog is disabled for this session, skip actual writing to the file.
+    But do not skip if it was disabled internally, and we still got here.
+   */
+  if(!(thd->variables.option_bits & OPTION_BIN_LOG_INTERNAL_OFF)
```
venkatesh-prasad-v: Likewise

sql/log_event.cc
```diff
@@ -7223,6 +7228,13 @@ int Intvar_log_event::do_apply_event(Relay_log_info const *rli)
   case INSERT_ID_EVENT:
     thd->force_one_auto_inc_interval(val);
     break;
+#ifdef WITH_WSREP
+  case BINLOG_CONTROL_EVENT:
+   if (val == 0){
```
venkatesh-prasad-v:
```suggestion
   if (val == 0)
   {
```

sql/wsrep_binlog.cc
```diff
@@ -391,6 +426,9 @@ int wsrep_write_cache(wsrep_t*  const wsrep,
                       IO_CACHE* const cache,
                       size_t*   const len)
 {
+    if (int res = prepend_binlog_control_event(wsrep, thd)) {
```
venkatesh-prasad-v:
```suggestion
    if (int res = prepend_binlog_control_event(wsrep, thd))
    {
```

sql/wsrep_mysqld.cc
```diff
@@ -1754,6 +1754,11 @@ int wsrep_to_buf_helper(
     *buf_len = thd->wsrep_gtid_event_buf_len;
   }
 
+  if ((thd->variables.option_bits & OPTION_BIN_LOG) == 0) {
```
```suggestion
  if ((thd->variables.option_bits & OPTION_BIN_LOG) == 0)
  {
```

sql/sql_class.h
venkatesh-prasad-v, line 3553: Can we initialize it to 0?
```suggestion
  ulonglong                 wsrep_bin_log_flag_save{0};
```

sql/binlog.cc
```diff
@@ -1599,7 +1599,17 @@ bool MYSQL_BIN_LOG::write_gtid(THD *thd, binlog_cache_data *cache_data,
   */
   thd->get_transaction()->last_committed= SEQ_UNINIT;
 
-
+#if WITH_WSREP
```
venkatesh-prasad-v:
```suggestion
#ifdef WITH_WSREP
```

sql/log_event.cc
```diff
@@ -7187,6 +7187,11 @@ void Intvar_log_event::print(FILE* file, PRINT_EVENT_INFO* print_event_info)
   case INSERT_ID_EVENT:
     msg="INSERT_ID";
     break;
+#ifdef WITH_WSREP
+  case BINLOG_CONTROL_EVENT:
+    msg="BINLOG_CONTROL";
+    break;
```
venkatesh-prasad-v: Can we add an assert here since BINLOG_CONTROL Intvar_event is not written to the binlog and this code will never be hit when running mysqlbinlog?

mysql-test/suite/galera/t/galera_log_bin.test
```diff
@@ -50,5 +50,102 @@ SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
...
+RESET MASTER;
+
+SET SESSION sql_log_bin = OFF;
+CREATE TABLE t1 (a int primary key);
+INSERT INTO t1 VALUES (1);
+BEGIN;
+INSERT INTO t1 VALUES (10);
+INSERT INTO t1 VALUES (11);
+INSERT INTO t1 VALUES (12);
+COMMIT;
```
venkatesh-prasad-v, line 71: Can we add an assert that nothing is logged into the binlog?
```
--let $master_pos_before= query_get_value(SHOW MASTER STATUS,Position,1)
SET SESSION sql_log_bin = OFF;
CREATE TABLE t1 (a int primary key);
INSERT INTO t1 VALUES (1);
BEGIN;
INSERT INTO t1 VALUES (10);
INSERT INTO t1 VALUES (11);
INSERT INTO t1 VALUES (12);
COMMIT;

# Assert that the above statements are not written to binlog.
--let $master_pos_after= query_get_value(SHOW MASTER STATUS,Position,1)
--let $assert_text= assert that the above events are not written to binlog
--let $assert_cond= $master_pos_before = $master_pos_after
--source include/assert.inc
```

mysql-test/suite/galera/t/galera_log_bin.test
```diff
@@ -50,5 +50,102 @@ SHOW BINLOG EVENTS IN 'mysqld-bin.000001' FROM 123;
...
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't2'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t1
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t2
+--source include/wait_condition.inc
+
+# we expect nothing in binlog
```
venkatesh-prasad-v, line 158: We can use the above mentioned assertion here as well.

venkatesh-prasad-v (changes requested):

sql/binlog.cc
kamil-holubicki: done

sql/binlog.cc
kamil-holubicki: done

sql/log_event.cc
kamil-holubicki: done

sql/wsrep_mysqld.cc

sql/sql_class.h
kamil-holubicki: done

sql/binlog.cc
kamil-holubicki: done

sql/log_event.cc
kamil-holubicki: done

mysql-test/suite/galera/t/galera_log_bin.test
kamil-holubicki: done

mysql-test/suite/galera/t/galera_log_bin.test
kamil-holubicki: done

sql/wsrep_binlog.cc

venkatesh-prasad-v (approved):
Thank you for addressing all the comments. Great work on this! Approved from my side.

venkatesh-prasad-v:
Please squash before merging.

PR #1495, opened 2021-09-30
PXC-3741: IST hangs forever if nodes suffer network disconnection

dutow:
Backported PXC-3580 for 5.7

Changed submodule pointer and rerecorded tests with new warnings

https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/223/


kamil-holubicki (approved):
lgtm

PR #1496, opened 2021-10-04
PXC-3733: Multiple procedure calls in the same

kamil-holubicki:
session causes deadlock errors when using the same
table name created by SP execution.

https://jira.percona.com/browse/PXC-3733

Problem:
When table is created by the stored procedure, its share is added to
the table definition cache. If the procedure is executed again and
creates new table with the same name, the corresponding share is being
found in the cache. However, it does not match newly created table, so
item tree objects recreation is requested through setting DA error to
ER_NEED_REPREPARE and thd's stmt arena state to STMT_INITIALIZED_FOR_SP.
WSREP transaction state is shifted to aborted.
Then execution is retried (up to MAX_REPREPARE_ATTEMPTS (3) times).
Retry is done without cleanup of WSREP transaction state, which is
normally done after SP execution.

Cause:
During the above-mentioned retry we try to prepare WSREP transaction,
but we detect that it is already aborted.

Solution:
Clean up the WSREP transaction state if it was requested to be
reprepared.


venkatesh-prasad-v (approved):
lgtm

PR #1505, opened 2021-10-21
PXC-3720: Assertion failure `assert(0)` in `wsrep_do_action_for_tables()`

kamil-holubicki:
https://jira.percona.com/browse/PXC-3720

This is the follow up of commit bd6f682f64da1aa463bd0e9ee37d1542a1b55673

The fix for BF-BF collision involves querying Data Dictionary to obtain
parent and child foreign keys.
We need to acquire a table MDL lock to be able to access DD. As we do it
before entering into TOI, another thread executing TOI in parallel may
abort the current thread.
In such a case we abort the current execution and rely on
wsrep_retry_autocommit functionality.

### DIFF
--- a/mysql-test/suite/galera/t/galera_bf_bf.test
+++ b/mysql-test/suite/galera/t/galera_bf_bf.test
@@ -213,6 +213,74 @@ SHOW TABLES;
 DROP TABLE p1;
 
 
+#
+# The fix involves querying Data Dictionary to obtain parent and child foreign
+# keys. We need to acquire table MDL lock to be able to access DD.
+# As we do it before entering into TOI, another thread executing TOI in paralel
+# may abort the current thread.
+# In such a case we rely on wsrep_retry_autocommit functionality.
+#
+--connection node_1
+--let $wsrep_retry_autocommit_saved = `SELECT @@wsrep_retry_autocommit`
+CALL mtr.add_suppression(".*Unable to collect FKs metadata.*");
+
+SET wsrep_retry_autocommit = 1;
+
+CREATE TABLE t1 (a INT PRIMARY KEY);
+SET DEBUG_SYNC='wsrep_do_action_for_tables_after_lock SIGNAL reached WAIT_FOR continue';
+--send ALTER TABLE t1 ADD (b INT)
+
+--connection node_1a
+SET DEBUG_SYNC='now WAIT_FOR reached';
+ALTER TABLE t1 ADD (c INT);
+
+--connection node_1
+--reap
+
+SHOW CREATE TABLE t1;
+DROP TABLE t1;
+
+#
+# Now the same, but with wsrep_retry_autocommit = 0
+#
+SET wsrep_retry_autocommit = 0;
+
+CREATE TABLE t1 (a INT PRIMARY KEY);
+SET DEBUG_SYNC='wsrep_do_action_for_tables_after_lock SIGNAL reached WAIT_FOR continue';
+--send ALTER TABLE t1 ADD (b INT)
+
+--connection node_1a
+SET DEBUG_SYNC='now WAIT_FOR reached';
+ALTER TABLE t1 ADD (c INT);
+
+--connection node_1
+--error ER_LOCK_DEADLOCK
+--reap
+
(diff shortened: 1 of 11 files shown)


dutow (approved):

sql/wsrep_mysqld.cc
```diff
@@ -1669,34 +1669,63 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
   if (!WSREP(thd) || !WSREP_CLIENT(thd)) return false;
 
+  // If we fail to collect FK meta data, we set the error and return true.
+  // That will trigger retry logic inside wsrep_dispatch_sql_command()
   for (auto table = tables; table; table = table->next_local) {
     if (!is_temporary_table(table)) {
+      DBUG_EXECUTE_IF("wsrep_do_action_for_tables_lock_fail", {
+        my_error(ER_LOCK_DEADLOCK, MYF(0));
+        return true;
+      });
```
venkatesh-prasad-v, line 1684: Unused DBUG block. Can we remove this if we are not using it anywhere?

sql/wsrep_mysqld.cc
```diff
@@ -1669,34 +1669,63 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
 
-      if (table_obj != nullptr) {
-        action(table_obj);
+      {
+        const dd::Table *table_obj = nullptr;
+        dd::cache::Dictionary_client::Auto_releaser releaser(thd->dd_client());
+        if (thd->dd_client()->acquire(dd::String_type(table->db),
+                                      dd::String_type(table->table_name),
+                                      &table_obj)) {
+          dd::release_mdl(thd, mdl_ticket);
```
venkatesh-prasad-v: I don't see any function calling `dd::release_mdl()` when there is any failure in `thd->dd_client()->acquire()`. Looks like it may be unsafe to call `dd:release_mdl()` in case of a failure. Can you check this?

sql/wsrep_mysqld.cc
```diff
@@ -1669,34 +1669,63 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
+        }
+
+        if (table_obj != nullptr) {
+          action(table_obj);
+        }
       }
 
       dd::release_mdl(thd, mdl_ticket);
+
+      // We maight have been aborted in the mantime by some TOI
```
```suggestion
      // We might have been aborted in the meantime by some other TOI
```

mysql-test/suite/galera/t/galera_bf_bf.test
```diff
@@ -213,6 +213,57 @@ SHOW TABLES;
...
+#
+# The fix involves querying Data Dictionary to obtain parent and child foreign
+# keys. We need to acquire table MDL lock to be able to access DD.
+# As we do it before entering into TOI, another thread executing TOI in paralel
+# may abort the current thread.
+# In such a case we rely on wsrep_retry_autocommit functionality.
+#
+--connection node_1
+--let $wsrep_retry_autocommit_saved = `SELECT @@wsrep_retry_autocommit`
+call mtr.add_suppression(".*Unable to collect FKs metadata.*");
```
venkatesh-prasad-v: Please use uppercase for all SQL commands.

sql/wsrep_mysqld.cc
kamil-holubicki: `dd::release_mdl()` is for `dd::acquire_shared_table_mdl()`
These are just wrappers around `MDL_context::mdl_context.acquire_locks()` and `MDL_context::release_lock()`

mysql-test/suite/galera/t/galera_bf_bf.test
kamil-holubicki: fixed

sql/wsrep_mysqld.cc
kamil-holubicki: I added MTR test using this

venkatesh-prasad-v (approved):
LGTM

venkatesh-prasad-v (approved):

PR #1513, opened 2021-10-27
PXC-3779: SST script treats wsrep_debug>1 as non debug

dutow:
Issue: a boolean check in the SST script only treatead "1" as a true
value, instead of any positive value.

Fix: modified the boolean check to allow any positive numbers


kamil-holubicki (approved):
LGTM

venkatesh-prasad-v (approved):

PR #1520, opened 2021-11-03
PXC-3758 [DOC] missing SST option: sst-idle-timeout (5.7)

patrickbirch:
source/manual/xtrabackup_sst.rst


altmannmarcelo (approved):
LGTM

PR #1531, opened 2021-11-17
PXC-3810: Implement BiDi scan for PXC via Azure pipelines (8.0)

oleksandr-kachan:
https://jira.percona.com/browse/PXC-3810


inikep (approved):
LGTM

PR #1540, opened 2021-12-01
PXC-3776 : Server hangs due to a deadlock, all queries stuck in proce…

satya-bodapati:
…sslist

PXC-3724 : PXC crashes with long semaphore wait

https://jira.percona.com/browse/PXC-3776
https://jira.percona.com/browse/PXC-3724

Problem:
Under heavy R/W parallel workload, after some time all client sessions
are blocked and the server eventually fails with
[FATAL] Semaphore wait has lasted > 600 seconds.

Cause:

    Client session 1 (TH1): Execute COMMIT (transaction 1).
    During prepare phase (before certification) it is BF aborted by HP
    thread. was_chosen_as_wsrep_victim flag is set to true.
    It does not get into the commit phase and ha_rollback_trans() is called,
    but was_chosen_as_wsrep_victim flag is not cleaned up.
    Transaction is finished.

    Another scenario: The transaction in TH1 is waiting to acquire table MDL.
    This waiting trx is chosen as victim and BF Aborted by High Priority
    transaction. In such case, we set was_chosen_as_wsrep_victim to true.
    Since the transaction is not started by InnoDB, Server will neither
    call commit nor rollback. The flag is left over and the future transaction
    in the same session will 'carry' this flag.

    Client session 1 (TH1): Execute COMMIT (transaction 2).
    It has was_chosen_as_wsrep_victim==true (as not cleaned up before).
    It certifies and then enters commit phase. It adds itself to
    the wsrep_group_commit_queue and to the mysql group commit queue.
    Other threads are added to simultaneously these queues as well
    (before and after TH1)

    Leader starts processing mysql group commit queue. It gets to TH1.
    Transaction is committed, but as was_chosen_as_wsrep_victim==true
    we skip the call to trx_sys_update_wsrep_checkpoint() (and internal
    call to wsrep_unregister_from_group_commit() for this thread).
    So TH1 is still at the beginning of wsrep_group_commit_queue

    Leader goes to the next thread in mysql group commit queue.
    It calls trx_sys_update_wsrep_checkpoint() which internally calls
    wsrep_wait_for_turn_in_group_commit(). This blocks, as TH1 is still
    in front of wsrep_group_commit_queue and the processing thread is the
    2nd one in the wsrep_group_commit_queue.

    The result is that Leader thread waits forever blocking all client
    threads

Solution:
1. Clean was_chosen_as_wsrep_victim flag when the transaction is being
   rolled back.

2. For the scenario, the victim is chosen before table opening or
   before transaction start, we set this victim flag only if transaction
   is started

MTR test not provided as the scenario needs multiple client
sessions and applier thread executing group commit logic simultaneously
which is problematic to model in MTR test.

### DIFF
--- a/storage/innobase/trx/trx0roll.cc
+++ b/storage/innobase/trx/trx0roll.cc
@@ -175,6 +175,9 @@ static dberr_t trx_rollback_for_mysql_low(
   transactions. */
 
   trx_rollback_to_savepoint_low(trx, nullptr);
+#ifdef WITH_WSREP
+  ut_ad(!trx->lock.was_chosen_as_wsrep_victim);
+#endif /* WITH_WSREP */
 
   trx->op_info = "";
 
@@ -197,6 +200,15 @@ static dberr_t trx_rollback_low(trx_t *trx) {
     case TRX_STATE_NOT_STARTED:
       trx->will_lock = 0;
       ut_ad(trx->in_mysql_trx_list);
+#ifdef WITH_WSREP
+      /* If transaction is BF-Aborted (rolledback) and
+      if transaction state is TRX_STATE_NOT_STARTED, the victim
+      flag has to be cleaned here because we do not call trx
+      commit. For all other states, the flag is cleaned at
+      trx_commit_in_memory(). */
+
+      trx->lock.was_chosen_as_wsrep_victim = false;
+#endif /* WITH_WSREP */
       return (DB_SUCCESS);
 
     case TRX_STATE_ACTIVE:
--- a/storage/innobase/handler/ha_innodb.cc
+++ b/storage/innobase/handler/ha_innodb.cc
@@ -24766,7 +24766,10 @@ int wsrep_innobase_kill_one_trx(void *const bf_thd_ptr,
    * which is already marked as BF victim
    * lock_sys is held until this vicitm has aborted
    */
-  victim_trx->lock.was_chosen_as_wsrep_victim = true;
+  if (victim_trx->state != TRX_STATE_NOT_STARTED) {
+    victim_trx->lock.was_chosen_as_wsrep_victim = true;
+  }
+
 
   if (wsrep_thd_set_wsrep_aborter(bf_thd, thd)) {
     WSREP_DEBUG("innodb kill transaction skipped due to wsrep_aborter set");
--- a/storage/innobase/trx/trx0trx.cc
+++ b/storage/innobase/trx/trx0trx.cc
@@ -474,6 +474,9 @@ static trx_t *trx_create_low() {
 
   /* We just got trx from pool, it should be non locking */
   ut_ad(trx->will_lock == 0);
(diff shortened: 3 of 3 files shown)


storage/innobase/trx/trx0roll.cc
```diff
@@ -197,6 +200,12 @@ static dberr_t trx_rollback_low(trx_t *trx) {
     case TRX_STATE_NOT_STARTED:
       trx->will_lock = 0;
       ut_ad(trx->in_mysql_trx_list);
+#ifdef WITH_WSREP
+      /* If the transaction is aborted, clean this flag.
+      Otherwise, a new transaction will use this trx
+      object and will be treated as a victim */
+      trx->lock.was_chosen_as_wsrep_victim = false;
```
kamil-holubicki, line 210: The flow is no so obvious here....
Maybe worth mentioning that killing transaction calls rollback() which calls  commit(), which detects that transaction was aborted and sets its state to TRX_STATE_FORCED_ROLLBACK. 
Then, when rollback is called again after commit prepare failure, we are detecting this flag here.

kamil-holubicki (approved):
LGTM

venkatesh-prasad-v (approved):
Great work on this @kamil-holubicki and @satya-bodapati ! 

Approved!

storage/innobase/trx/trx0roll.cc
satya-bodapati: updated comment

meyerder:
Can you clarify as I have had issues attempting to track this issue… and noticed in the .26 version this..

All links from this issue "appear" to be closed...

https://jira.percona.com/browse/PXC-3706 

Which has been marked as fixed and really looks like this is the same issue?

satya-bodapati:
> Can you clarify as I have had issues attempting to track this issue… and noticed in the .26 version this..
> 
> All links from this issue "appear" to be closed...
> 
> https://jira.percona.com/browse/PXC-3706
> 
> Which has been marked as fixed and really looks like this is the same issue?

@meyerder PXC-3706 is different issue and is merged to PXC 8026
https://github.com/percona/percona-xtradb-cluster/pull/1468/commits/ca6049a80b9c5cb7adead511f37dab348991e721

meyerder:
I cant seem to find a 8026 link..  Can you confirm for me please?

venkatesh-prasad-v:
@meyerder I agree with @satya-bodapati. PXC-3706 is different from PXC-3776 and both have been fixed in 8.0.26.

Commits:
PXC-3776: https://github.com/percona/percona-xtradb-cluster/commit/66d942032dfbc94d63d99318d755c60f1ab04ab8
PXC-3706: https://github.com/percona/percona-xtradb-cluster/commit/ca6049a80b9c5cb7adead511f37dab348991e721

The reason you are not seeing PXC-3776 in the relase notes is that it was reported during the QA and thus did not make it to the release notes.

PR #1542, opened 2021-12-06
PXC-3777 : Server crashes due to a deadlock detection of mutexes

satya-bodapati:
https://jira.percona.com/browse/PXC-3777

Problem:
--------

This is because of deadlock of mutexes trx->mutex and lock_sys mutex

1. aborter thread: acquires trx->mutex and wants to acquire lock_sys in X mode
2. Other thread:   acquires lock_sys in S mode and wants to acquire trx->mutex

trx-t1 does DDL tries to acquire X lock MDL and as part of this, we abort all
the trxs with conflicting MDL locks (ie do wsrep_abort_thd) However, the
conflicting trx (t2) is a DML on the same table that is in waiting state.
This trx t2 is in waiting state because another thread t3 is also a DML in
committing state.

t3 is in state of releasing locks: acquires lock_sys in Shared mode and
(requests)tries to acquire trx->mutex.

t1 acquires trx->mutex and requests for lock_sys in X mode (to abort t2)

Fix:
----

When doing BF Abort, if the victim trx is suspended state, we need not explicitly
release locks. The lock_wait_timeout thread will release locks of the
interrupted trx and wakes up the victim that was in the suspended state.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_kill_victim_waiting.result
+++ b/mysql-test/suite/galera/r/pxc_kill_victim_waiting.result
@@ -1,0 +1,38 @@
+#
+# PXC-3777 : Server crashes due to a deadlock detection of mutexes
+#
+CREATE TABLE t1(a INT PRIMARY KEY, b VARCHAR(10));
+INSERT INTO t1 VALUES (1, 'hi');
+# connection node_1a
+# Lock the record in this session node_1a
+BEGIN;
+SELECT * FROM t1 WHERE a = 1 FOR UPDATE;
+a	b
+1	hi
+# connection node_1b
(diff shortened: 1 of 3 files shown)


kamil-holubicki (commented):
The fix and the corresponding explanation look fine.
As discussed on Slack, it would be nice to introduce MTR for this.

satya-bodapati:
> The fix and the corresponding explanation look fine. As discussed on Slack, it would be nice to introduce MTR for this.

added testcase and submitted to Jenkins: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/368/

kamil-holubicki (approved):
lgtm

PR #1547, opened 2021-12-14
PXC-3743: Make wsrep session variables updatable through SET_VAR() hint

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3743

Allow users to utilize SET_VAR optimizer hint to temporarily set the wsrep
session variables.

This patch allows the following session variables to be updatable through
SET_VAR() extension.

1. wsrep_sync_wait
2. wsrep_dirty_reads
3. wsrep_trx_fragment_size
4. wsrep_trx_fragment_unit
```
Example:

mysql> SELECT @@wsrep_sync_wait; SELECT /*+ SET_VAR(wsrep_sync_wait=4 )*/ @@wsrep_sync_wait; SELECT @@wsrep_sync_wait;
+-------------------+
| @@wsrep_sync_wait |
+-------------------+
|                15 |
+-------------------+
1 row in set (0.00 sec)
+-------------------+
| @@wsrep_sync_wait |
+-------------------+
|                 4 |
+-------------------+
1 row in set (0.00 sec)
+-------------------+
| @@wsrep_sync_wait |
+-------------------+
|                15 |
+-------------------+
1 row in set (0.00 sec)
```
Note: The following session variables cannot be updatable through
SET_VAR

1. wsrep_on: Because it may cause inconsistent read/writes.
2. wsrep_causal_reads: Deprecated.

In addition to the above change, the `wsrep` suite has been added to the
default MTR suite as well.

Testing Done
---
Galera suite: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/364/testReport/
Failing tests: 

### DIFF
--- a/mysql-test/suite/wsrep/my.cnf
+++ b/mysql-test/suite/wsrep/my.cnf
@@ -2,10 +2,16 @@
 !include include/default_mysqld.cnf
 
 [mysqld.1]
+binlog-format=row
+innodb_autoinc_lock_mode=2
+innodb_flush_log_at_trx_commit=2
+
+# wsrep variables
 wsrep_provider=@ENV.WSREP_PROVIDER
+wsrep_provider_options='base_port=@mysqld.1.#galera_port;gcache.size=10M'
 wsrep_cluster_address='gcomm://'
 wsrep_node_address=127.0.0.1
-binlog-format=row
+wsrep_node_name=node1
 
 [ENV]
 GALERA_BASE_PORT=@mysqld.1.#galera_port
--- a/mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
+++ b/mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
@@ -1,0 +1,133 @@
+# === Purpose ===
+#
+# This test verifies that the following wsrep session variables are hint updatable.
+#
+# 1. wsrep_sync_wait
+# 2. wsrep_dirty_reads
+# 3. wsrep_trx_fragment_size
+# 4. wsrep_trx_fragment_unit
+#
+# === References ===
+#
+# PXC-3743: Make wsrep session variables updatable through SET_VAR() hint
+
+--source include/have_wsrep_provider.inc
+
+# Auxiliary SP which returns variable value before hint, with hint and after hint applying
+DELIMITER \\; 
+
+CREATE PROCEDURE test_hint (hint_str VARCHAR(255), var_str VARCHAR(64))
+BEGIN
+
+SET @orig_q= CONCAT("SELECT VARIABLE_VALUE INTO @before_val FROM performance_schema.session_variables where VARIABLE_NAME = '",  var_str, "'");
+
+SET @hint_q= CONCAT("SELECT /*+ ", hint_str,
+                    "*/ VARIABLE_VALUE" ,
+                    " INTO @hint_val FROM performance_schema.session_variables where VARIABLE_NAME = '",  var_str, "'");
+
+SET @after_q= CONCAT("SELECT VARIABLE_VALUE INTO @after_val FROM performance_schema.session_variables where VARIABLE_NAME = '",  var_str, "'");
+
+PREPARE orig_stmt FROM @orig_q;
+PREPARE hint_stmt FROM @hint_q;
+PREPARE after_stmt FROM @after_q;
+
+EXECUTE orig_stmt;
+EXECUTE hint_stmt;
+EXECUTE after_stmt;
+
+SELECT  hint_str;
+SELECT @before_val, @hint_val, @after_val;
+
+DEALLOCATE PREPARE orig_stmt;
+DEALLOCATE PREPARE hint_stmt;
+DEALLOCATE PREPARE after_stmt;
+
+END\\
+
+DELIMITER ;\\
+
+--echo #########################################
+--echo # Test SET_VAR hint for wsrep variables #
+--echo #########################################
+
+--echo
+--echo # Test 1: wsrep_sync_wait
+CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+--assert (`SELECT @hint_val = 4`)
+
+SET wsrep_sync_wait=2;
+CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
+
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+--assert (`SELECT @hint_val = 4`)
+
+SET wsrep_sync_wait=default;
+
+--echo
+--echo # Test 2: wsrep_dirty_reads
+CALL test_hint("SET_VAR(wsrep_dirty_reads=ON)", "wsrep_dirty_reads");
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+--assert (`SELECT @hint_val = 'ON'`)
+
+SET wsrep_dirty_reads=1;
+CALL test_hint("SET_VAR(wsrep_dirty_reads=OFF)", "wsrep_dirty_reads");
+
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+--assert (`SELECT @hint_val = 'OFF'`)
+
+SET wsrep_dirty_reads=default;
+
+--echo
+--echo # Test 3: wsrep_trx_fragment_unit
+CALL test_hint("SET_VAR(wsrep_trx_fragment_unit='statements')", "wsrep_trx_fragment_unit");
+
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+--assert (`SELECT @hint_val = 'statements'`)
+
+SET wsrep_trx_fragment_unit='statements';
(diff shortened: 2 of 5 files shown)


mysql-test/suite/wsrep/my.cnf
```diff
@@ -2,10 +2,17 @@
...
+innodb_flush_log_at_trx_commit=2
+
+# wsrep variables
+wsrep_causal_reads=ON
 wsrep_cluster_address='gcomm://'
 wsrep_node_address=127.0.0.1
-binlog-format=row
+wsrep_node_name=node1
+wsrep_provider=@ENV.WSREP_PROVIDER
+wsrep_sync_wait = 15
```
kamil-holubicki: Inconsistent formating

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
```diff
@@ -0,0 +1,130 @@
...
+
+DELIMITER ;\\
+
+--echo #########################################
+--echo # Test SET_VAR hint for wsrep variables #
+--echo #########################################
+
+--echo
+--echo # Test 1: wsrep_sync_wait
+CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
```
kamil-holubicki, line 55: What is the purpose of this call?

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
```diff
@@ -0,0 +1,130 @@
...
+
+SET wsrep_sync_wait=2;
+CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
+
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+
+--assert (`SELECT /*+ SET_VAR(wsrep_sync_wait=10)*/ @@wsrep_sync_wait = 10`)
```
kamil-holubicki: Wouldn't it be more clear to add 1 more assert to 3 above asserts like
`--assert (SELECT @hint_val = 4)`
?

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
```diff
@@ -0,0 +1,130 @@
...
+SET wsrep_sync_wait=default;
+
+--echo
+--echo # Test 2: wsrep_dirty_reads
+CALL test_hint("SET_VAR(wsrep_dirty_reads=1)", "wsrep_dirty_reads");
```
kamil-holubicki: As above, what is the purpose of this first CALL?

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
```diff
@@ -0,0 +1,130 @@
...
+--assert (`SELECT @before_val != @hint_val`)
+--assert (`SELECT @hint_val != @after_val`)
+--assert (`SELECT @before_val = @after_val`)
+
+--assert (`SELECT /*+ SET_VAR(wsrep_dirty_reads=1)*/ @@wsrep_dirty_reads = 1`)
+SET wsrep_dirty_reads=default;
+
+
+--echo
+--echo # Test 2: wsrep_dirty_reads
```
kamil-holubicki: Does this check/test differ from the one above?

kamil-holubicki (commented):
The change looks fine.
I left a couple of questions, suggestions.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: The intention here is to first test with the MTR's default value for the variables, then changing the value and testing again.

mysql-test/suite/wsrep/my.cnf
venkatesh-prasad-v: Fixed.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: Thanks for catching this. This is a duplicate. Removed now.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: Added.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
kamil-holubicki: OK, but shouldn't we do asserts as below? The stored procedure does not do any validation.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
kamil-holubicki: My intention was to add this assert and remove this:
`--assert (SELECT /*+ SET_VAR(wsrep_sync_wait=10)*/ @@wsrep_sync_wait = 10)`
Makes sense?

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: We want to stop executing the test immediately our expectations are not met and this is not possible with the stored routines. So, the stored procedure only sets the uservars and they will be later used for the validation.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: Done. Removed.

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: I think I got your point now. You meant to add asserts even after the first call right?

mysql-test/suite/wsrep/my.cnf
```diff
@@ -2,10 +2,17 @@
...
+innodb_flush_log_at_trx_commit=2
+
+# wsrep variables
+wsrep_causal_reads=ON
 wsrep_cluster_address='gcomm://'
 wsrep_node_address=127.0.0.1
-binlog-format=row
+wsrep_node_name=node1
+wsrep_provider=@ENV.WSREP_PROVIDER
+wsrep_sync_wait=15
```
satya-bodapati: why did you add the wsrep_sync_wait change to suite cnf? wouldn't it affect entire suite?

venkatesh-prasad-v: This is the first test to go into the `wsrep` suite in 8.0. So, we are safe :)

venkatesh-prasad-v: The reason for adding it to the cnf is to get consistent reads in the test cases so that it becomes easier to write tests  without thinking about synchronization of previous writes.
```
--connection node_1
INSERT INTO t1 VALUES(1);

--connection node_2
SELECT * FROM t1;
```
With wsrep_sync_wait set to 15, we can always ensure that `SELECT` on `node_2` is executed only after `INSERT` is replicated and applied on `node_2`

satya-bodapati: ok, 15 comes from this. So we do cluster-sync check on these SQLs.
```
enum enum_wsrep_sync_wait {
    WSREP_SYNC_WAIT_NONE = 0x0,
    // select, begin
    WSREP_SYNC_WAIT_BEFORE_READ = 0x1,
    WSREP_SYNC_WAIT_BEFORE_UPDATE_DELETE = 0x2,
    WSREP_SYNC_WAIT_BEFORE_INSERT_REPLACE = 0x4,
    WSREP_SYNC_WAIT_BEFORE_SHOW = 0x8,
    WSREP_SYNC_WAIT_MAX = 0xF
  };
```

satya-bodapati: @venkatesh-prasad-v if this is a new suite, I would rename it to wsrep_one_node or wsrep_1node

satya-bodapati: @venkatesh-prasad-v do you know why we are different from upstream for this variable? upstream supports an additional value of 3.
https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-sync-wait

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
kamil-holubicki: Exactly

mysql-test/suite/wsrep/my.cnf
venkatesh-prasad-v: @satya-bodapati 
> if this is a new suite, I would rename it to wsrep_one_node or wsrep_1node

wsrep suite is actually meant for one node only. So it is implicit and there is no need to add any extension.


>do you know why we are different from upstream for this variable? upstream supports an additional value of 3.

We also support value 3 (since it is bitmask internally) . It has also been documented well in https://www.percona.com/doc/percona-xtradb-cluster/8.0/wsrep-system-index.html#wsrep_sync_wait

```
mysql> set wsrep_sync_wait=3;

mysql> select @@wsrep_sync_wait;
+-------------------+
| @@wsrep_sync_wait |
+-------------------+
|                 3 |
+-------------------+
1 row in set (0.00 sec)
```

mysql-test/suite/wsrep/t/wsrep_set_var_extension.test
venkatesh-prasad-v: Fixed now.

mysql-test/suite/wsrep/my.cnf
kamil-holubicki: >With wsrep_sync_wait set to 15, we can always ensure that SELECT on node_2 is executed only after INSERT is replicated and applied on node_2

Wasn't the intention of this suite to use only 1 node?

>if this is a new suite, I would rename it to wsrep_one_node or wsrep_1node

or galera_1node :)

satya-bodapati (approved):
LGTM!

PR #1549, opened 2021-12-15
Add SASL dependences

adivinho:


EvgeniyPatlan (changes requested):
what about installing them in build script?

PR #1556, opened 2021-12-21
Fixed PXC-3766 - Xtrabackup based SST always runs version-check proce…

altmannmarcelo:
…dure

https://jira.percona.com/browse/PXC-3766

Problem:
Half way thorught the life of xtrabackup/innobackupex, --version-check
functionality was added. This made SST script to validate if pxb
had support for --version-check, if true we were adding
--no-version-check parameter to skip it. By default, SST was using
innobackupex script. The script got deprecated and during the process of
migrating to xtrabackup binary we left the check passing what was
suppose to be the --target-dir without properly setting it as parameter,
causing xtrabackup command to fail and the check never resolving to
true.
This particular check is no longer necessary since current version of
PXC will always be running on a version of xtrabackup that has
--version-check support.

Fix:
Removed the check for the said functionality and always run xtrabackup
with --no-version-check.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1614,7 +1614,7 @@ if [[ "$pxc_encrypt_cluster_traffic" == "on" ]]; then
     wsrep_log_debug "with encrypt=4  ssl_ca=$ssl_ca  ssl_cert=$ssl_cert  ssl_key=$ssl_key"
 fi
 
-if ${INNOBACKUPEX_BIN} /tmp --help 2>/dev/null | grep -q -- '--version-check'; then
+if ${INNOBACKUPEX_BIN} --help 2>/dev/null | grep -q -- '--version-check'; then
     disver="--no-version-check"
 fi
 


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1661,15 +1656,15 @@ if [[ $ssyslog -eq 1 ]]; then
...
-    INNOBACKUP="${INNOBACKUPEX_BIN} --defaults-file=${WSREP_SST_OPT_CONF}  --defaults-group=mysqld${WSREP_SST_OPT_CONF_SUFFIX} $disver $iopts \$ieopts \$INNOEXTRA \$keyringbackupopt --lock-ddl --backup --galera-info --binlog-info=ON \$encrypt_backup_options --stream=\$sfmt --target-dir=\$itmpdir 2>\${DATA}/innobackup.backup.log"
+    INNOPREPARE="${INNOBACKUPEX_BIN} --no-version-check $iapts --prepare --binlog-info=ON \$rebuildcmd \$keyringapplyopt \$encrypt_prepare_options --target-dir=\${DATA} &>\${DATA}/innobackup.prepare.log"
+    INNOMOVE="${INNOBACKUPEX_BIN} --defaults-file=${WSREP_SST_OPT_CONF}  --defaults-group=mysqld${WSREP_SST_OPT_CONF_SUFFIX} --datadir=\${TDATA} --no-version-check $impts  --move-back --binlog-info=ON --force-non-empty-directories \$encrypt_move_options --target-dir=\${DATA} &>\${DATA}/innobackup.move.log"
+    INNOBACKUP="${INNOBACKUPEX_BIN} --defaults-file=${WSREP_SST_OPT_CONF}  --defaults-group=mysqld${WSREP_SST_OPT_CONF_SUFFIX} --no-version-check $iopts \$ieopts \$INNOEXTRA \$keyringbackupopt --lock-ddl --backup --galera-info --binlog-info=ON \$encrypt_backup_options --stream=\$sfmt --target-dir=\$itmpdir 2>\${DATA}/innobackup.backup.log"
```
satya-bodapati: does $INNOBACKUPEX_BIN  point to xtrabackup binary?

altmannmarcelo: yes, at https://github.com/percona/percona-xtradb-cluster/blob/Percona-XtraDB-Cluster-5.7.35-31.53/scripts/wsrep_sst_xtrabackup-v2.sh#L149

satya-bodapati (approved):
LGTM! I hope all occurrences of $disver are removed.

kamil-holubicki (approved):
lgtm

altmannmarcelo:
@satya-bodapati @kamil-holubicki - I've changed the fix a bit. After further investigation, turns out that version check (the feature) can be disabled at PXB compilation time bypassing `-D HAVE_VERSION_CHECK=OFF` - https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/src/xtrabackup.cc#L1070 .  So in an odd case, a customer could be using their own version of xtrabackup with that functionality disabled. Instead of removing the check completelly, I've fixed the routine to detect if version-check is compiled on pxb and adding --no-version-check to SST script.

satya-bodapati:
> @satya-bodapati @kamil-holubicki - I've changed the fix a bit. After further investigation, turns out that version check (the feature) can be disabled at PXB compilation time bypassing `-D HAVE_VERSION_CHECK=OFF` - https://github.com/percona/percona-xtrabackup/blob/8.0/storage/innobase/xtrabackup/src/xtrabackup.cc#L1070 . So in an odd case, a customer could be using their own version of xtrabackup with that functionality disabled. Instead of removing the check completelly, I've fixed the routine to detect if version-check is compiled on pxb and adding --no-version-check to SST script.

@altmannmarcelo am sure if this is necessary for pxb shipped with PXC.

satya-bodapati (approved):

PR #1562, opened 2021-12-27
PXC-3704: wsrep_max_ws_size not linked to repl.max_ws_size when defined in configuration file

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3704

Problem
=======
Server variable `wsrep_max_ws_size` and provider option `repl.max_size` have
been linked together from a long time. However, passing `wsrep_max_ws_size` in
the cnf file has no effect on the provider's repl.max_size option and the
server shows different values post startup.

Solution
========
1. When only `wsrep_max_ws_size` is passed during startup, then the same value
   used as the maximum writeset size.
2. When only `repl.max_ws_size` is passed during startup, then the same value
   used as the maximum writeset size.
3. When both `repl.max_ws_size` and `wsrep_max_ws_size` are passed during
   startup, then `repl.max_ws_size` takes precedence over `wsrep_max_ws_size`.

Testing Done
---
galera* suite: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/377/console
Failing tests:

### DIFF
--- a/mysql-test/suite/galera/t/galera_repl_max_ws_size.test
+++ b/mysql-test/suite/galera/t/galera_repl_max_ws_size.test
@@ -28,3 +28,223 @@ DROP TABLE t1;
 #CALL mtr.add_suppression("rbr write fail");
 CALL mtr.add_suppression("Transaction/Write-set size limit.*");
 CALL mtr.add_suppression("Failed to create write-set from binlog.*");
+
+
+# === Purpose ===
+#
+# Test that setting repl.max_ws_size provider option also updates server's
+# wsrep_max_ws_size and vice versa.
+#
+# === References ====
+#
+# PXC-3704: wsrep_max_ws_size not linked to repl.max_ws_size when defined in
+#           configuration file
+
+--connection node_2
+
+# Save the provider options before testing
+--let $wsrep_provider_options_orig = `SELECT @@wsrep_provider_options`
+
+# Check that the cluster started correctly.
+SET SESSION wsrep_sync_wait = 0;
+
+--let $members=2
+--source include/galera_wait_until_synced.inc
+--source include/wsrep_wait_membership.inc
+
+--echo
+--echo # Test-1: Verify that value of wsrep_max_ws_size picked from cnf file is
+--echo #         also updated in the provider on server startup.
+--echo #
+
+# Create a temporary cnf file for testing with wsrep_max_ws_size
+--let $CONFIG_FILE = $MYSQLTEST_VARDIR/tmp/galera_repl_max_ws_size.cnf
+--copy_file $MYSQLTEST_VARDIR/my.cnf $CONFIG_FILE
+--append_file $CONFIG_FILE
+[mysqld.2]
+wsrep-max-ws-size= 8192
+wsrep-debug=1
+EOF
+
+--echo # Shutdown node2
+--connection node_2
+--echo [connection node_2]
+--echo # Shutting down node2
+--source include/shutdown_mysqld.inc
+
+--echo # restarting node2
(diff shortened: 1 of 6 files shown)


satya-bodapati (commented):
@venkatesh-prasad-v why have two variables for the same configuration max_write_set_size? Is this a mistake by upstream?

venkatesh-prasad-v:
@satya-bodapati Yes. This exists on upstream as well.

https://galeracluster.com/library/documentation/galera-parameters.html#repl-max-ws-size
https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-max-ws-size

satya-bodapati:
> @satya-bodapati Yes. This exists on upstream as well.
> 
> https://galeracluster.com/library/documentation/galera-parameters.html#repl-max-ws-size https://galeracluster.com/library/documentation/mysql-wsrep-options.html#wsrep-max-ws-size

Hmm, ideally the right fix would be to deprecate the wsrep variable wsrep_max_ws_size? :) what do you think?

mysql-test/suite/galera/t/galera_repl_max_ws_size.test
```diff
@@ -28,3 +28,231 @@ DROP TABLE t1;
...
+# Save the provider options before testing
+--let $wsrep_provider_options_orig = `SELECT @@wsrep_provider_options`
+
+# Check that the cluster started correctly.
+SET SESSION wsrep_sync_wait = 0;
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment';
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size';
+--source include/wait_condition.inc
```
satya-bodapati: may be create a sync_cluster.inc?

sql/wsrep_mysqld.cc
```diff
@@ -945,6 +945,39 @@ static void wsrep_init_provider_status_variables() {
           sizeof(provider_vendor) - 1);
 }
 
+/*
```
satya-bodapati: use doxygen style /** and @return ..

satya-bodapati (changes requested):
LGTM, minor comments

mysql-test/suite/galera/t/galera_repl_max_ws_size.test
venkatesh-prasad-v: I had already created https://jira.percona.com/browse/PXC-3210 for the same. Looks like it is the time now :)

venkatesh-prasad-v: Looks like we already have `wsrep_wait_membership.inc` for waiting on the number of nodes. I'll reuse them.

satya-bodapati: Rename it to wsrep_sync_cluster.inc

venkatesh-prasad-v: IMO, `wsrep_sync_cluster` sounds a bit ambiguous as it doesn't sync the cluster, but only waits until the cluster has the desied number of members. So, I think the current name is enough.

satya-bodapati: ok fine with me

satya-bodapati (approved):
LGTM! Please squash and merge

PR #1571, opened 2022-01-07
PXC-3855: Assertion `!thd->in_active_multi_stmt_transaction() || thd->in_multi_stmt_transaction_mode()'

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3855

Problem
=======
Server hits the assertion

    !thd->in_active_multi_stmt_transaction() || thd->in_multi_stmt_transaction_mode()

when AUTOCOMMIT=ON and there is a binlog cache error when binlog is disabled
(i.e, when in binlog emulation mode)

Background
========
1. When a server is in the binlog emulation mode and if there is an error
during the binlog cache write stage, we have a special handling in PXC which
makes the transaction to register with the binlog handlerton by calling
`trans_register_ha()` with binlog_hton, so that WSREP layer can can rollback
the transaction.

2. At the end of every statement, in `mysql_execute_command()` the server
   expects that

  - Session should not have SERVER_STATUS_IN_TRANS flag set.
  - If session has SERVER_STATUS_IN_TRANS flag set, then the session must be in
    multi statement transaction mode (either AUTOCOMMIT = OFF or inside a
    transaction started with BEGIN).

Analysis
========
When the server calls `trans_register_ha()` during binlog cache error, it sets
the `SERVER_STATUS_IN_TRANS` flag indicating that there was a beginning of a
multi-statement transaction, even when there was no multi statements involved
(because AUTOCOMMIT=ON).

This caused the assertion

    !thd->in_active_multi_stmt_transaction() || thd->in_multi_stmt_transaction_mode()

in `mysql_execute_command()` to fail, since we have SERVER_STATUS_IN_TRANS flag
set for the session and the session is not in a multi statement transaction mode
(since AUTOCOMMIT=ON), resulting in server abort.

Solution
========
In case of a binlog cache write error, register the transaction scope
binlog handler (call `trans_register_ha()` with all = true) only when
the server is in multi statement transaction mode, just like what is
being done in `binlog_start_trans_and_stmt()`.

Testing Done
---
galera* suite: https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/385/testReport/
Failing tests:
Failing tests:
5. galera.galera_sst_xtrabackup-v2_pxc_encrypt - Sporadic in nature
6. galera_3nodes_sr.galera_sr_kill_slave_before_apply - Sporadic in nature

### DIFF
--- a/mysql-test/suite/wsrep/r/wsrep_set_var_extension.result
+++ b/mysql-test/suite/wsrep/r/wsrep_set_var_extension.result
@@ -26,7 +26,7 @@ CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
 hint_str
 SET_VAR(wsrep_sync_wait=4)
 @before_val	@hint_val	@after_val
-15	4	15
+0	4	0
 SET wsrep_sync_wait=2;
 CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
 hint_str
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -5577,13 +5577,19 @@ void MYSQL_BIN_LOG::report_cache_write_error(THD *thd, bool is_transactional) {
   }
 
 #ifdef WITH_WSREP
-  /* If wsrep transaction is active and binlog emulation is on,
-     binlog write error may leave transaction without any registered
-     htons. This makes wsrep rollback hooks to be skipped and the
-     transaction will remain alive in wsrep world after rollback.
-     Register binlog hton here to ensure that rollback happens in full. */
+  /*
+     If wsrep transaction is active and binlog emulation is on, binlog
+     write error may leave transaction without any registered htons.
+     This makes wsrep rollback hooks to be skipped and the transaction
+     will remain alive in wsrep world after rollback.  Register binlog
+     hton here to ensure that rollback happens in full only when the
+     session is in multi stmt transaction mode (i.e, either
+     autocommit=OFF or when the transaction is started with explicit
+     BEGIN).
+  */
   if (WSREP_EMULATE_BINLOG(thd)) {
-    if (is_transactional) trans_register_ha(thd, true, binlog_hton, NULL);
+    if (is_transactional && thd->in_multi_stmt_transaction_mode())
+      trans_register_ha(thd, true, binlog_hton, NULL);
     trans_register_ha(thd, false, binlog_hton, NULL);
   }
 #endif /* WITH_WSREP */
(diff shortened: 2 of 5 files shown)


sql/binlog.cc
```diff
@@ -5577,13 +5577,19 @@ void MYSQL_BIN_LOG::report_cache_write_error(THD *thd, bool is_transactional) {
...
+     hton here to ensure that rollback happens in full only when the
+     session is in multi stmt transaction mode (i.e, either
+     autocommit=OFF or when the transaction is started with explicit
+     BEGIN).
+  */
   if (WSREP_EMULATE_BINLOG(thd)) {
-    if (is_transactional) trans_register_ha(thd, true, binlog_hton, NULL);
+    if (is_transactional && thd->in_multi_stmt_transaction_mode())
+      trans_register_ha(thd, true, binlog_hton, NULL);
     trans_register_ha(thd, false, binlog_hton, NULL);
```
satya-bodapati, line 5593: should there be else here
if (is_transactional && multi) {
   trans_register(.. true ..) 
} else {
  trans_register_ha(..flase);
}

satya-bodapati (changes requested):
LGTM! minor comments

Also plz fix two typos in commit message
cashe -> cache
Backgound -> Background

sql/binlog.cc
venkatesh-prasad-v: In case of a muli-stmt transaction, both statement and transaction scoped binlog handlers should be registered.
Having (if..else) only registers only one scope and may cause undefined behavior. 

https://github.com/percona/percona-xtradb-cluster/blob/8.0/sql/binlog.cc#L10136-L10137 also does the same.

satya-bodapati: Right venky, I remember now

mysql-test/suite/wsrep/r/wsrep_set_var_extension.result
```diff
@@ -26,7 +26,7 @@ CALL test_hint("SET_VAR(wsrep_sync_wait=4)", "wsrep_sync_wait");
 hint_str
 SET_VAR(wsrep_sync_wait=4)
 @before_val	@hint_val	@after_val
-15	4	15
+0	4	0
 SET wsrep_sync_wait=2;
```
satya-bodapati, line 30: @venkatesh-prasad-v why did this test result change?

venkatesh-prasad-v: @satya-bodapati  I recorded this test while working on https://jira.percona.com/browse/PXC-3743. Back then `wsrep_sync_wait=15` was in the `mysql-test/suite/wsrep/my.cnf`. But I forgot to re-record it we I removed the line. So this change was necessary to make the test suite clean.

satya-bodapati (approved):

PR #1584, opened 2022-01-28
PXC 3878 - Merge PS 5.6 backport security fixes

kamil-holubicki:
https://jira.percona.com/browse/PXC-3878

This is PS 5.6 -> PXC 5.6 merge.

It contains security fixes backported to PS in the following PRs;
https://github.com/percona/percona-server/pull/4422
https://github.com/percona/percona-server/pull/4426
https://github.com/percona/percona-server/pull/4541


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.6-param/134

Permanent failures (fail on pxc 5.6 trunk as well):
rpl.rpl_stm_mixed_mts_rec_crash_safe_checksum rpl.rpl_stm_mixed_mts_rec_crash_safe rpl.rpl_stm_mixed_mts_rec_crash_safe_small rpl.rpl_stm_mixed_mts_crash_safe rpl.rpl_stm_mixed_crash_safe rpl.rpl_crash_safe_master rpl.rpl_row_mts_rec_crash_safe rpl.rpl_row_mts_crash_safe rpl.rpl_row_crash_safe main.alter_table innodb.innodb_bug59641 main.tc-heuristic-recover

All other tests pass locally.

sql/sys_vars.cc
```diff
@@ -3730,6 +3731,44 @@ static Sys_var_charptr Sys_time_format(
...
+    if (trans_commit_stmt(thd) || trans_commit(thd))
+    {
+      thd->mdl_context.release_transactional_locks();
+      WSREP_DEBUG("autocommit, MDL TRX lock released: %lu", thd->thread_id);
+      return true;
+    }
+#else
+    if (trans_commit_stmt(thd) || trans_commit(thd))
+      return true;
+#endif
```
venkatesh-prasad-v: Here we have a common `if (trans_commit_stmt(thd) || trans_commit(thd))` condition in both `#ifdef` and `#else` part, which I think can be put outside the `#ifdef` blocks. Can you try to make both look alike?

Note: In 5.7, we have this as
https://github.com/percona/percona-xtradb-cluster/blob/Percona-XtraDB-Cluster-5.7.35-31.53/sql/sys_vars.cc#L5054-L5065

kamil-holubicki: That was my 1st thought as well, but because the above mentioned `if` does not use curly brackets in original version, the result will be hard to read like:
```
    if (trans_commit_stmt(thd) || trans_commit(thd))
#ifdef WITH_WSREP
    {
      thd->mdl_context.release_transactional_locks();
      WSREP_DEBUG("autocommit, MDL TRX lock released: %lu", thd->thread_id);
#endif
      return true;
#ifdef WITH_WSREP
    }
#endif
```

venkatesh-prasad-v: How about this 
```C++
    if (trans_commit_stmt(thd) || trans_commit(thd))
#ifdef WITH_WSREP
    {
      thd->mdl_context.release_transactional_locks();
      WSREP_DEBUG("Transaction commit failed while toggling autocommit."
                  " Release MDL trx lock for thread: %u", thd->thread_id());
      return true;
    }
#else
      return true;
#endif /* WITH_WSREP */
```

kamil-holubicki: Agree.

venkatesh-prasad-v (approved):
lgtm

PR #1590, opened 2022-02-21
PXC-3831: Nodes fail with two applier threads conflict

kamil-holubicki:
https://jira.percona.com/browse/PXC-3831

1. lock_rec_has_to_wait() function unified with Codership's version
(returns false in case of detection of 2 BF threads attempting to wait.
Our implementation was definitely wrong as we always called
lock_rec_has_to_wait() function with for_locking flag equal 'false'

2. Added upstream MTR tests related to BF abort

3. Cleaned up unused wsrep_UK_scan flag

### DIFF
--- a/storage/innobase/lock/lock0lock.cc
+++ b/storage/innobase/lock/lock0lock.cc
@@ -491,9 +491,6 @@ static inline ulint lock_rec_get_insert_intention(
 /** Checks if a lock request for a new lock has to wait for request lock2.
  @return true if new lock has to wait for lock2 to be removed */
 static inline bool lock_rec_has_to_wait(
-#ifdef WITH_WSREP
-    ibool for_locking,   /*!< is caller locking or releasing */
-#endif                   /* WITH_WSREP */
     const trx_t *trx,    /*!< in: trx of new lock */
     ulint type_mode,     /*!< in: precise mode of the new lock
                        to set: LOCK_S or LOCK_X, possibly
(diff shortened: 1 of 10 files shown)


storage/innobase/lock/lock0lock.cc
```diff
@@ -919,6 +889,20 @@ static const lock_t *lock_rec_other_has_conflicting(
     ulint heap_no,            /*!< in: heap number of the record */
     const trx_t *trx)         /*!< in: our transaction */
 {
+#ifdef WITH_WSREP
+  std::shared_ptr<bool> unlocker;
+  if (wsrep_log_conflicts) {
+    mutex_enter(&trx_sys->mutex);
+    // we are not using raii::Sentry from sql space to not introduce
+    // the dependency
```
satya-bodapati: if the entire code is entire header, including in .cc file is OK

satya-bodapati (commented):
Can you please explain how this wsrep_UK_scan code helps?

satya-bodapati (commented):
Maybe upstream also didn't care to explain how this bool is used (or what it achieves)

```
  #ifdef WITH_WSREP
    bool wsrep_UK_scan;
  #endif /* WITH_WSREP */
  
````

File: storage/innobase/lock/lock0lock.cc [hunk 2]
```diff
@@ -513,6 +510,22 @@ request is really for a 'gap' type lock */
...
+
+#ifdef WITH_WSREP
+  if (is_hp && trx != lock2->trx && trx_is_high_priority(lock2->trx)) {
+    /* conflicting high priority locks are supposed to be false positives
+       i.e. locks not on actual certified pay load data or GAP locks.
+       these should not block applier execution, hence returning false here */
+    if (wsrep_debug)
+      ib::info() << "BF-BF lock conflict skipped, \n"
+                 << " trx-1 lock request: " << trx->id
+                 << " trx-2 lock holder: " << lock2->trx->id;
```
satya-bodapati: I think we can print some lock info as well.  there is lock_print() or something like that. 
Would be good to know for debugging purposes.
Also maybe the THD id, other WSREP info

storage/innobase/lock/lock0lock.cc
```diff
@@ -513,6 +510,22 @@ request is really for a 'gap' type lock */
...
+    return (false);
+  }
+
+  if (!is_hp && trx_is_high_priority(lock2->trx)) return (true);
```
satya-bodapati: This piece and above lock check (that returns false), both should first lock_mode compatible first. It will reduce  false "lock conflict skipped" messages. From my analysis, I see these messages even for non-conflicting locks.

So better to check for lock_mode_compatible first.

satya-bodapati (changes requested):

storage/innobase/lock/lock0lock.cc
kamil-holubicki: done

storage/innobase/lock/lock0lock.cc
kamil-holubicki: Good catch @satya-bodapati !. This is because I merged it in the wrong way. Now its location is aligned with the upstream.

satya-bodapati: Looks good now

satya-bodapati (changes requested):
Please squash commit now
And please provide Jenkins link

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%208.0/job/pxc-8.0-param/424/

satya-bodapati (approved):

PR #1596, opened 2022-03-21
PXC-3599 SST fails with gtid=on: Move file mysqld-bin.index to /other…

patrickbirch:
…fs/mysqld-bin.log.index failed: Destination file exists (5.7)

 M source/limitation.rst


altmannmarcelo (approved):
Text looks good but seems like you have conflicts.

PR #1597, opened 2022-03-22
PXC-3901: BF-BF conflict with the dict stats thread

dutow:
Issue: a few galera_nbo MTR tests fail with BF-BF conflict when ran with
the --mem options. These failures are all caused by the NBO applier
conflicting with the innodb dict stats thread.

Fix: as there is no way to detect if a THD is the dict stats thread,
this patch introduces a flag background threads can set to prevent BF-BF
abort caused by them.

Since the dict thread only holds this lock for a short time, and there
is no way to kill it anyway, this is the safest and easiest solution.

### DIFF
--- a/sql/sql_thd_internal_api.cc
+++ b/sql/sql_thd_internal_api.cc
@@ -64,10 +64,17 @@
 struct mysql_cond_t;
 struct mysql_mutex_t;
 
+#ifdef WITH_WSREP
+THD *create_internal_thd(bool allow_mdl_conflict) {
+#else
 THD *create_internal_thd() {
+#endif /* WITH_WSREP */
   /* For internal threads, use enabled_plugins = false. */
   THD *thd = new THD(false);
   thd->system_thread = SYSTEM_THREAD_BACKGROUND;
+#ifdef WITH_WSREP
+  thd->wsrep_allow_mdl_conflict = allow_mdl_conflict;
+#endif /* WITH_WSREP */
   // Skip grants and set the system_user flag in THD.
   thd->security_context()->skip_grants();
   thd->thread_stack = reinterpret_cast<char *>(&thd);
--- a/sql/sql_thd_internal_api.h
+++ b/sql/sql_thd_internal_api.h
@@ -42,7 +42,11 @@ class THD;
 class partition_info;
 struct fragmentation_stats_t;
 
+#ifdef WITH_WSREP
+THD *create_internal_thd(bool allow_mdl_conflict = false);
+#else
 THD *create_internal_thd();
+#endif /* WITH_WSREP */
 void destroy_internal_thd(THD *thd);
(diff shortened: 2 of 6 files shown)


storage/innobase/dict/dict0stats_bg.cc
```diff
@@ -358,6 +359,7 @@ statistics. */
 void dict_stats_thread() {
   ut_a(!srv_read_only_mode);
   THD *thd = create_internal_thd();
+  thd->wsrep_allow_mdl_conflict = true;
```
kamil-holubicki: ```create_internal_thd()``` sets ```thd->system_thread = SYSTEM_THREAD_BACKGROUND;```.
Isn't it enough? or do we need to distinguish between background threads?

dutow: There are many background threads: lots of innodb threads, the wsrep applier, rollbacker, nbo applier threads. I don't think ignoring all of them would be safe, especially that would mean any other new background thread in the future.

So far the conflict only seems to happen because the dict stat thread, nothing else, so it seemed safer to ignore just that one.

sql/sql_thd_internal_api.cc
```diff
@@ -64,10 +64,17 @@
...
+#ifdef WITH_WSREP
+THD *create_internal_thd(bool allow_mdl_conflict) {
+#else
 THD *create_internal_thd() {
+#endif /* WITH_WSREP */
   /* For internal threads, use enabled_plugins = false. */
   THD *thd = new THD(false);
   thd->system_thread = SYSTEM_THREAD_BACKGROUND;
+#ifdef WITH_WSREP
+  thd->wsrep_allow_mdl_conflict = true;
```
kamil-holubicki: ```thd->wsrep_allow_mdl_conflict = allow_mdl_conflict```?

sql/sql_thd_internal_api.h
```diff
@@ -42,7 +42,11 @@ class THD;
 class partition_info;
 struct fragmentation_stats_t;
 
+#ifdef WITH_WSREP
+THD *create_internal_thd(bool allow_mdl_conflict=false);
```
kamil-holubicki: Is this formating clang compliant? (spaces around =)

kamil-holubicki (changes requested):

sql/sql_thd_internal_api.cc
dutow: :facepalm: I looked at that line at least a few times and I didn't notice it. Fixed.

sql/sql_thd_internal_api.h
dutow: Fixed.

kamil-holubicki (approved):
lgtm

PR #1598, opened 2022-03-22
(5.7) PXC-3848: Cluster node crashes when using

kamil-holubicki:
ALTER USER CURRENT_USER()

https://jira.percona.com/browse/PXC-3848

Problem:
1. PXC does not allow using CURRENT_USER() function when operating in
cluster mode.
2. If AuditLog plugin is installed, the error is being logged.
3. As the query contains plain text password, this causes query being
rewritten.
4. Rewrite logic calls get_current_user() function which reports
the error and we end up in point 1 which causes recurrent calls.

Solution:
Allow normal behavior of get_current_user() when the function is called
for query rewrite purpose.


venkatesh-prasad-v (approved):

dutow:
`mysql_rewrite_grant` (GRANT .. TO user IDENTIFIED BY) in the same file doesn't have the same issue?

kamil-holubicki:
You are right @dutow. ```GRANT  ... TO CURRENT_USER IDENTIFIED BY ...``` was affected as well. Fixed and updated the testcase.

sql/sql_rewrite.cc
```diff
@@ -385,7 +385,7 @@ void mysql_rewrite_create_alter_user(THD *thd, String *rlb,
 
   while ((tmp_user_name= user_list++))
   {
-    if ((user_name= get_current_user(thd, tmp_user_name)))
+    if ((user_name= get_current_user(thd, tmp_user_name, true)))
```
dutow, line 397: Looks good, just one more thing I noticed now: these changes should be in ifdef wsrep blocks.

kamil-holubicki: Yup, exactly the same thing I noticed in the meantime :) Fixed.

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.7-param/280/

dutow (approved):

PR #1599, opened 2022-03-23
(8.0) PXC-3848: Cluster node crashes when using ALTER USER CURRENT_USER()

kamil-holubicki:
https://jira.percona.com/browse/PXC-3848

Problem:
1. PXC does not allow using CURRENT_USER() function when operating in
cluster mode.
2. If AuditLog plugin is installed, the error is being logged.
3. As the query contains plain text password, this causes query being
rewritten.
4. Rewrite logic calls get_current_user() function which reports
the error and we end up in point 1 which causes recurrent calls.

Note that for 8.0 the syntax GRANT ... ON ... TO ... IDENTIFIED BY ...
is invalid, so we can be sure that Rewriter_grant::rewrite() will never
be reached with pain text password.

Solution:
Allow normal behavior of get_current_user() when the function is called
for query rewrite purpose.


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/28/

dutow (approved):

venkatesh-prasad-v (approved):

PR #1615, opened 2022-04-12
[5.7] PXC-3609: Do not update binlog status vars when binlog is disabled

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-3609

Problem: In cluster mode, binlog related status vars are updated even
when binlog is disabled.

Solution: Do not register the status variables in the cache manager when
in binlog emulation mode (i.e, binlog is disabled)

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -729,6 +729,10 @@ private:
   */
   void compute_statistics()
   {
+#ifdef WITH_WSREP
+    if (wsrep_emulate_bin_log)
+      return;
+#endif /* WITH_WSREP */
     if (!is_binlog_empty())
     {
       (*ptr_binlog_cache_use)++;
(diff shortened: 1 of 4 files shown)


venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-param/284/console

nitendra-bhosle (approved):
Change looks good to me.

sql/binlog.cc
```diff
@@ -731,9 +731,17 @@ class binlog_cache_data
   {
     if (!is_binlog_empty())
     {
+#ifdef WITH_WSREP
```
kamil-holubicki: Wouldn't it be easier to check here
```WSREP_EMULATE_BINLOG```
and update variables or not?
Then we would not need all the stuff from ```THD::binlog_setup_trx_data()```

venkatesh-prasad-v: Yes. It is easier. But what stopped me from doing it is that `THD` is not available in this part of the code, thereby making us depend on the `current_thd` to get the thread pointer. As `current_thd` has had problems with performance in the past, I though it is better to not register the status vars in `binlog_setup_trx_data()` is the best solution at hand.

venkatesh-prasad-v: Fixed! Used `wsrep_emulate_bin_log` instead of `WSREP_EMULATE_BINLOG`.

kamil-holubicki (approved):
lgtm

PR #1620, opened 2022-04-25
PXC-3923: ANALYZE TABLE crashes node when [super_]read_only is set

kamil-holubicki:
https://jira.percona.com/browse/PXC-3923

Problem:
When super_read_only or read_only is set, ANALYZE TABLE fails
(as expected) which triggers consistency voting mechanism and causes
node to leave the cluster

Cause:
In original flow the validation if admin action can be done in readonly
mode is done on transaction commit. As the statement is replicated as
TOI it is too late.

Solution:
Check if action can be done before starting TOI.

### DIFF
--- a/mysql-test/suite/galera/t/galera_admin.test
+++ b/mysql-test/suite/galera/t/galera_admin.test
@@ -82,6 +82,34 @@ SELECT COUNT(*) = 10 FROM x1;
 SELECT COUNT(*) = 10000 FROM t2;
 SELECT COUNT(*) = 10 FROM x2;
 
+
+# Check that super_read_only and read_only are handled properly (PXC-3923)
+--connection node_1
+--let super_read_only_saved = `SELECT @@super_read_only`
+--let read_only_saved = `SELECT @@read_only`
+
+# super_read_only
+SET GLOBAL super_read_only=1;
+--error ER_OPTION_PREVENTS_STATEMENT
+ANALYZE TABLE t1;
+
+# read_only
+SET GLOBAL super_read_only=0;
+SET GLOBAL read_only=1;
+
+CREATE USER 'testuser'@'%' IDENTIFIED WITH 'mysql_native_password' BY 'testuser';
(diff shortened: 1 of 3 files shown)


mysql-test/suite/galera/t/galera_admin.test
```diff
@@ -82,6 +82,34 @@ SELECT COUNT(*) = 10 FROM x1;
...
+SET GLOBAL super_read_only=1;
+--error ER_OPTION_PREVENTS_STATEMENT
+ANALYZE TABLE t1;
+
+# read_only
+SET GLOBAL super_read_only=0;
+SET GLOBAL read_only=1;
+
+CREATE USER 'testuser'@'%' IDENTIFIED WITH 'mysql_native_password' BY 'testuser';
+GRANT select, update, insert, delete ON *.* to 'testuser'@'%';
```
venkatesh-prasad-v: Not a strong opinion though, could we have the SQL commands in uppercase?

mysql-test/suite/galera/t/galera_admin.test
```diff
@@ -82,6 +82,34 @@ SELECT COUNT(*) = 10 FROM x1;
...
+--connect node_1a, 127.0.0.1, testuser, testuser, test, $NODE_MYPORT_1
+--connection node_1a
+--error ER_OPTION_PREVENTS_STATEMENT
+ANALYZE TABLE t1;
+
+--connection node_1
+--eval set global read_only = $read_only_saved
+--eval set global super_read_only = $super_read_only_saved
```
venkatesh-prasad-v: likewise

venkatesh-prasad-v (approved):

mysql-test/suite/galera/t/galera_admin.test
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_admin.test
kamil-holubicki: fixed

PR #1623, opened 2022-04-27
PXC-3924: TRUNCATE TABLE X; INSERT INTO X; results in

kamil-holubicki:
HA_ERR_FOUND_DUPP_KEY on slave node while foreign keys are disabled
and violated

https://jira.percona.com/browse/PXC-3924

Problem:
It occurred only if the table referenced the parent table by FK.
In such a case it happened randomly that after truncation of the table
on node_1 and insertion of the row with a previously existing primary
key, the replicated writeset caused HA_ERR_FOUND_DUPP_KEY error on
node_2.

Cause:
TRUNCATE TABLE is replicated as TOI. For triggering TOI there was used
WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF macro which checked if associated
LEX::no_write_to_binlog was set. As TRUNCATE TABLE does not support
NO_WRITE_TO_BINLOG the value was random causing TRUNCATE TABLE to be not
replicated. Subsequent INSERT replicated, but the record with the given
PK was already present on node_2 causing the error.

Solution:
Do not check no_write_to_binlog flag when initiating TOI for
TRUNCATE_TABLE.
1. WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF was renamed to
WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF_WRTCHCK to have consistent naming
to WSREP_TO_ISOLATION_BEGIN_WRTCHK.
2. WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF does not check
no_write_to_binlog now
It is done this way to avoid setting no_write_to_binlog flag in
PT_truncate_table_stmt which would make the flow not clear.

Additionally:
1. galera.galera_truncate MTR test enabled
2. compilation warning in wsrep_mysqld.cc suppressed


sql/wsrep_mysqld.cc
```diff
@@ -1096,9 +1096,12 @@ int wsrep_init() {
 
     char ssl_opts[4096];
     server_main_callback.populate_wsrep_ssl_options(ssl_opts, sizeof(ssl_opts));
+#pragma GCC diagnostic push
```
dutow: This should be MY_COMPILER_DIAGNOSTIC_PUSH (and the similar macros for the other changes too)

sql/sql_parse.h
```diff
@@ -172,13 +172,23 @@ extern uint sql_command_flags[];
       wsrep_to_isolation_begin(thd, db_, table_, table_list_))                \
     goto wsrep_error_label;
 
-#define WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF(db_, table_, table_list_,       \
+/* Checks if lex->no_write_to_binlog is set for statements that use
+  LOCAL or NO_WRITE_TO_BINLOG
+*/
+#define WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF_WRTCHK(db_, table_, table_list_,       \
```
dutow: Why the code duplication for the condition, it could reuse the other macro?

Also, shouldn't the formatting of the lines below change?

dutow (commented):
Also, there's a comment in sql_parse.cc explaining that we can't use this macro because of this reason - maybe after this change that could also be changed?

sql/sql_parse.h
kamil-holubicki: `IF_WRTCHK` versions add `thd->lex->no_write_to_binlog` condition in `if`. This condition has to be checked after `WSREP(thd)` which validates `thd` pointer, so no easy/clean way to reuse `WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF` in `WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF_WRTCHK'. 
That would need introducing some intermediate macros making things more obfuscated.

Another way of doing it would be doing something like:
```
#define WSREP_TO_ISOLATION_BEGIN_FK_TABLES_IF_WRTCHK if (thd != nullptr && !thd->lex->no_write_to_binlog && WSREP(thd) ...
``
This way `thd` would be validated against nullptr twice, which should be optimized by the compiler.
Additionally, I see `WSREP_TO_ISOLATION_BEGIN_(WRTCHK)` has the same code duplication and exactly the same reasoning as above.
Let's leave it as it is.

Formatting fixed.

sql/wsrep_mysqld.cc

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/107/

dutow (approved):

PR #1625, opened 2022-04-28
PXC-2687: BF abort of a SP causes a write-after-free error

voodam:
https://jira.percona.com/browse/PXC-2687

Extend PXC-3226: disable wsrep_retry_autocommit mechanism for all required commands inside CALL statement.

Bug is not reproduced in 8.0.

### DIFF
--- a/sql/sp_instr.h
+++ b/sql/sp_instr.h
@@ -246,6 +246,10 @@ public:
       return m_lex ? m_lex->sql_command : -1;
   }
 
+#ifdef WITH_WSREP
+  const LEX *get_lex() const { return m_lex; }
+#endif
+
   const LEX_CSTRING *get_prepared_stmt_name() const {
     return m_lex ? &m_lex->prepared_stmt_name : NULL;
   }
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -6407,7 +6407,7 @@ static Sys_var_charptr Sys_wsrep_data_home_dir(
 
 static Sys_var_charptr Sys_wsrep_cluster_name(
        "wsrep_cluster_name", "Name for the cluster",
-       READ_ONLY GLOBAL_VAR(wsrep_cluster_name), CMD_LINE(REQUIRED_ARG),
+       READ_ONLY PREALLOCATED GLOBAL_VAR(wsrep_cluster_name), CMD_LINE(REQUIRED_ARG),
        IN_FS_CHARSET, DEFAULT(WSREP_CLUSTER_NAME),
        NO_MUTEX_GUARD, NOT_IN_BINLOG,
        ON_CHECK(wsrep_cluster_name_check),
--- a/mysql-test/suite/galera/r/galera_disable_retry_for_check_table.result
+++ b/mysql-test/suite/galera/r/galera_disable_retry_for_check_table.result
@@ -55,3 +55,43 @@ ERROR 40001: WSREP detected deadlock/conflict and aborted the transaction. Try r
 #
 # 5. Cleanup
 DROP TABLE t1;
+#
+# 1. Create a table on node_1.
+[node_1a]
+CREATE TABLE t1(i INT PRIMARY KEY)
+PARTITION BY RANGE (i)
+(
+PARTITION p0 VALUES LESS THAN (2),
+PARTITION p1 VALUES LESS THAN (4)
+);
+CREATE PROCEDURE p1()
+BEGIN
+ALTER TABLE t1 CHECK PARTITION p0;
+END|
+#
+# 2. Execute CHECK TABLE query on node_1 and halt the query in
+#    ha_innobase::check() function.
+SET DEBUG_SYNC="ha_innobase_check SIGNAL reached WAIT_FOR continue";
+CALL p1();;
+[node_1b]
+SET DEBUG_SYNC="now WAIT_FOR reached";
+#
+# 3. Execute an ALTER TABLE query on node_2 and wait for it to be
+#    replicated to node_1.
+[node_2]
+ALTER TABLE t1 ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1);
+SET DEBUG_SYNC="now SIGNAL continue";
+SET DEBUG_SYNC="RESET";
+#
+# 4. ALTER TABLE, being replicated as TOI, aborts the CHECK TABLE
+#    query. Verify that CHECK TABLE query fails with ER_LOCK_DEADLOCK error.
+Table	Op	Msg_type	Msg_text
+test.t1	check	error	Partition p0 returned error
+test.t1	check	Error	Query execution was interrupted
+test.t1	check	error	Unknown - internal error 168 during operation
+ERROR 40001: WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction
+#
+# 5. Cleanup
+DROP TABLE t1;
+DROP PROCEDURE p1;
--- a/mysql-test/suite/galera/r/galera_sp_bf_abort2.result
+++ b/mysql-test/suite/galera/r/galera_sp_bf_abort2.result
@@ -8,6 +8,14 @@ UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
 INSERT INTO t1 VALUES (10, "from sp");
 COMMIT;
 END|
+CREATE PROCEDURE proc_with_transaction_and_select()
+BEGIN
+START TRANSACTION;
+SELECT 1;
+UPDATE t1 SET f2 = "from sp" WHERE f1 > 1;
+INSERT INTO t1 VALUES (10, "from sp");
+COMMIT;
+END|
 CREATE PROCEDURE proc_with_transaction_continue_handler()
 BEGIN
 DECLARE CONTINUE HANDLER FOR SQLEXCEPTION BEGIN END;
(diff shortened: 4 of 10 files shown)


sql/sys_vars.cc
```diff
@@ -6397,7 +6397,7 @@ static Sys_var_charptr Sys_wsrep_data_home_dir(
 
 static Sys_var_charptr Sys_wsrep_cluster_name(
        "wsrep_cluster_name", "Name for the cluster",
-       READ_ONLY GLOBAL_VAR(wsrep_cluster_name), CMD_LINE(REQUIRED_ARG),
+       READ_ONLY PREALLOCATED GLOBAL_VAR(wsrep_cluster_name), CMD_LINE(REQUIRED_ARG),
```
kamil-holubicki, line 6410: Why this change is needed?

kamil-holubicki (commented):
Is it possible to create MTR test for this?

sql/sql_class.cc
```diff
@@ -3308,6 +3308,12 @@ bool Query_result_send::send_result_set_metadata(List<Item> &list, uint flags)
     /* Metadata is already send during first try that failed so avoid
     resending it. Just plan to send the result. */
     WSREP_DEBUG("Skip resending metadata if query is being re-tried");
+
+    /* Initialization is required for proper asserting */
+    if (thd->get_protocol()->start_result_metadata(
```
kamil-holubicki: Is this OK? `start_result_metadata` returns false in case of success.
So `is_result_set_started` will not be set in success path.

sql/sys_vars.cc
voodam: For preventing memory leak.

sql/sql_class.cc
voodam: No, it's wrong, fixed.

sql/sys_vars.cc
voodam: If `PREALLOCATED` is set, variable will be freed, this is configured in that way for other `wsrep_*` string variables.

venkatesh-prasad-v:
Hi! Is there any way to reproduce the `write-after-free` error? Can you please create an MTR test for this?

voodam:
> Hi! Is there any way to reproduce the `write-after-free` error? Can you please create an MTR test for this?

Hi! Probably yes, I can see some testcase examples, where ASAN is mentioned, will try soon.

voodam:
Finally added test.

scripts/wsrep_sst_common.sh
```diff
@@ -18,6 +18,11 @@
 
 set -u
 
+# When MySQL is built with ASAN, libasan will be propagated to child processes via LD_PRELOAD.
+# This can be annoying because Unix utils can have memory errors itself.
+# It seems that it's safe to unset LD_PRELOAD here - no other libs should be preloaded.
+unset LD_PRELOAD
```
venkatesh-prasad-v: This is still not guaranteed to work, as asan can report memory errors in bash itself.
```
2022-06-29T04:10:27.351633Z 0 [Note] WSREP: Initiating SST/IST transfer on JOINER side (wsrep_sst_xtrabackup-v2 --role 'joiner' --address '127.0.0.1:13009' --datadir '/home/venki/work/pxc/57/asan/mysql-test/var/mysqld.2/data/' --defaults-file '/home/venki/work/pxc/57/asan/mysql-test/var/my.cnf' --defaults-group-suffix '.2' --parent '988760' --mysqld-version '5.7.38-41'   '' )

=================================================================
==988775==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2 byte(s) in 1 object(s) allocated from:
    #0 0x7f9b53747808 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x555fa03974f3 in xmalloc (/usr/bin/bash+0x904f3)

SUMMARY: AddressSanitizer: 2 byte(s) leaked in 1 allocation(s).

=================================================================
==988779==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2 byte(s) in 1 object(s) allocated from:
    #0 0x7f9b53747808 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cc:144
    #1 0x555fa03974f3 in xmalloc (/usr/bin/bash+0x904f3)

SUMMARY: AddressSanitizer: 2 byte(s) leaked in 1 allocation(s).
```

In my opinion, it is better to move the unsetting logic to server. The below patch worked well locally.
```diff
diff --git a/sql/wsrep_sst.cc b/sql/wsrep_sst.cc
index 3089984e40c..c9731a89acb 100644
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -780,6 +780,16 @@ static void* sst_joiner_thread (void* a)
 
 #define WSREP_SST_AUTH_ENV "WSREP_SST_OPT_AUTH"
 
+#if defined(HAVE_ASAN)
+/*
+ * When MySQL is built with ASAN, libasan will be propagated to child processes
+ * via LD_PRELOAD. This can be annoying because unix utils including bash can
+ * have memory errors itself. Let's unset LD_PRELOAD on ASAN builds.
+ */
+static void reset_ld_preload(wsp::env& env) {
+  env.append("LD_PRELOAD=");
+}
+#endif
 static int sst_append_auth_env(wsp::env& env, const char* sst_auth)
 {
   int const sst_auth_size= strlen(WSREP_SST_AUTH_ENV) + 1 /* = */
@@ -857,6 +867,10 @@ static ssize_t sst_prepare_other (const char*  method,
     return -env.error();
   }
 
+#if defined(HAVE_ASAN)
+  reset_ld_preload(env);
+#endif
+
   if ((ret= sst_append_auth_env(env, sst_auth)))
   {
     WSREP_ERROR("sst_prepare_other(): appending auth failed: %d", ret);
@@ -1595,6 +1609,10 @@ wsrep_cb_status_t wsrep_sst_donate_cb (void* app_ctx, void* recv_ctx,
     return WSREP_CB_FAILURE;
   }
 
+#if defined(HAVE_ASAN)
+  reset_ld_preload(env);
+#endif
+
   /* Wait for wsrep-SE to initialize that also signals
   completion of init_server_component which is important before we initiate
   any meangiful action especially DONOR action from this node. */
```

mysql-test/suite/galera/t/bug2687.test
```diff
@@ -0,0 +1,97 @@
+# PXC-2687: BF abort of a SP causes a write-after-free error
+# Test mainly taken from PXC-2684.
+# But for some reason waits and sleeps are not working, so SELECT is used instead.
+
+--source include/have_innodb.inc
+--source include/galera_cluster.inc
+--source include/asan.inc
```
venkatesh-prasad-v: Considering the test case coverage, I would prefer running this test irrespective of asan builds. Can you remove this `include/asan` line and other related require files?

mysql-test/suite/galera/t/bug2687.test
```diff
@@ -0,0 +1,97 @@
+# PXC-2687: BF abort of a SP causes a write-after-free error
+# Test mainly taken from PXC-2684.
+# But for some reason waits and sleeps are not working, so SELECT is used instead.
```
venkatesh-prasad-v: Is it not posisble to simulate BF abort with debug syncs?

venkatesh-prasad-v (changes requested):

scripts/wsrep_sst_common.sh
kamil-holubicki: This is a good idea.
We have a similar workaround in 8.0, but in Jenkins script
https://github.com/kamil-holubicki/jenkins-pipelines/blob/parallel-mtr/pxc/local/test-binary-pxc-parallel-mtr#L148
I would say, the one proposed by @venkatesh-prasad-v is better, it is compiled-in only in test builds, and works locally without the magic we currently do in 8.0 Jenkins. My proposal is to propagate it to 8.0 as well. Then I will remove the workaround from Jenkins script.

mysql-test/suite/galera/t/bug2687.test
voodam: I don't know the point in code where to set debug sync, so it can take some time. I'll try to do this quickly.

voodam:
@venkatesh-prasad-v, 

I mentioned that "malformed packet" error (`--error 2027`) is produced exactly because of not sending metadata. It produced in another valid case, here: https://github.com/percona/percona-xtradb-cluster/blob/5.7/mysql-test/suite/galera/t/galera_sp_bf_abort2.test#L22 if we add `SELECT f1 FROM t1`, "malformed packet" error will occur too if `wsrep_retry_autocommit = 1`.

So, why we don't need to send metadata? Git blame didn't give much: https://github.com/percona/percona-xtradb-cluster/commit/facd915a4673#diff-65c68b03ca220f14939f817d3c9cfe7767580750f9ef47876b383780a0fd82c3R2979. I tried to run `galera`, `galera_3nodes` and `wsrep` test suites (are these all cluster-specific tests?) and all of them wasn't affected by deleting `return FALSE` line (and hence sending metadata).

kamil-holubicki (commented):
As this is 5.7 fix, It has to be done according to GCA merge process

voodam:
> As this is 5.7 fix, It has to be done according to GCA merge process

Done

voodam:
@kamil-holubicki, this PR includes patch from [PS-8327](https://github.com/percona/percona-server/pull/4839). What is better: full copy-paste PS-8327 to PXC-2687 or leave as is until merging percona-server to PXC?

kamil-holubicki:
> @kamil-holubicki, this PR includes patch from [PS-8327](https://github.com/percona/percona-server/pull/4839). What is better: full copy-paste PS-8327 to PXC-2687 or leave as is until merging percona-server to PXC?

I see... As there were no complains about PXC-2687 I would wait for PS to be merged into PXC and then add what is necessary from PXC point of view. If we do the other way, there will be conflicts during the merge.

kamil-holubicki:
Hi @voodam , we merged PS 5.7.40 into the PXC, so the above-mentioned PS fix is already there. Could you please check this PR? (it shows some conflicts)

voodam:
> Hi @voodam , we merged PS 5.7.40 into the PXC, so the above-mentioned PS fix is already there. Could you please check this PR? (it shows some conflicts)

Hi! Now it should be OK.

UPD. Probably some problems after the merge: 
```
storage/innobase/handler/ha_innodb.cc:9508:12: error: unused variable 'err' [-Werror,-Wunused-variable]
                        dberr_t err = row_lock_table_for_mysql(
```

voodam:
Catched a bug due the broken galera_sp_bf_abort2 test. We need to require that all instructions should retry in autocommit to retry CALL statement, so I fixed that.

voodam:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5149/#showFailuresLink
Now all is OK, `merge` and `merge_mmap` were broken somewhere in the previous commits.

@kamil-holubicki I can't to sign CLA for some reason: it redirects me to GitHub, but stays pending.

kamil-holubicki (approved):
Looks fine. Please squash commits and write descriptive commit message (what was the problem, why was the problem, how it was solved)

voodam:
Done and also rewinded to GCA.
UPD. Some problems with branches, I'll fix it a little bit later.
UPD2. I have a trouble with following GCA process, waiting for a responce.

voodam:
Rewinded back to 5.7, GCA process was cancelled.

voodam:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5167/

sql/sp_instr.h
```diff
@@ -242,9 +242,9 @@ class sp_lex_instr : public sp_instr
   virtual SQL_I_List<Item_trigger_field>* get_instr_trig_field_list()
   { return &m_trig_field_list; }
 
-  int get_command() const {
-      return m_lex ? m_lex->sql_command : -1;
-  }
+  int get_command() const { return m_lex ? m_lex->sql_command : -1; }
```
venkatesh-prasad-v: This must be inside `#ifdef WITH_WSREP`

venkatesh-prasad-v (changes requested):

sql/sp_instr.h
kamil-holubicki: Why? I would rather say: "please revert formating change" :)
This code is from PS

venkatesh-prasad-v: Sorry for mistakenly commenting on `get_command()`. I intended to make the below `get_lex()` function be enclosed under `#ifdef WITH_WSREP` as this change is a PXC specific change.

voodam: Done (get_lex I suppose, get_command is just reformatted).

venkatesh-prasad-v: Can you please revert the formatting changes done for the `get_command()` function? This will cause merge conflict in future when the nearby lines are changed in PS.

```suggestion
  int get_command() const
  { 
    return m_lex ? m_lex->sql_command : -1; 
  }
```

voodam: Done

venkatesh-prasad-v (approved):

PR #1627, opened 2022-05-04
PXC-3872: Normal user dropping system user

kamil-holubicki:
https://jira.percona.com/browse/PXC-3872

Problem:
When user with SYSTEM_USER privilege is dropped by the user without
SYSTEM_USER privilege on node_1, node_2 initiates consistency voting
mechanism and self-leaves the cluster.
During the investigation I've found that not only DROP USER is affected,
but ALTER USER, RENAME USER, REVOKE ALL PRIVILEGES, REVOKE ... ON,
GRANT ... ON PROCEDURE, REVOKE ... ON PROCEDURE, GRANT <role> TO,
REVOKE <role> FROM, DROP PROCEDURE.

Cause:
Affected statements are replicated as TOI. Replication was initated
before checking if SYSTEM_USER privileges for both users allow to
proceed. As the applier thread on node_2 works in superuser context,
it is allowed to proceed with affected statement, which leads to cluster
inconsistency.

Solution:
Move initiation of TOI replication after SYSTEM_USER privileges
validation.

### DIFF
--- a/mysql-test/suite/galera/t/pxc_drop_user.test
+++ b/mysql-test/suite/galera/t/pxc_drop_user.test
@@ -1,0 +1,116 @@
+# Test that DROP USER executed on the user with SYSTEM_USER privilege
+# is not replicated when executing user has no SYSTEM_USER
+
+--source include/have_debug_sync.inc
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+CREATE USER 'x'@'localhost' IDENTIFIED BY 'x';
+GRANT CREATE USER, EXECUTE, ALTER ROUTINE, DROP ROLE, ROLE_ADMIN, SELECT, UPDATE, SYSTEM_VARIABLES_ADMIN ON *.* TO 'x'@'localhost' WITH GRANT OPTION;
+
+CREATE USER 'y'@'localhost' IDENTIFIED BY 'y';
+GRANT ALL ON *.* TO 'y'@'localhost' WITH GRANT OPTION;
+
+DELIMITER //;
+
+CREATE PROCEDURE p1()
+BEGIN
+  INSERT INTO t1 (a) VALUES (1);
+END//
+
+DELIMITER ;//
+
+CREATE ROLE 'test_role';
+GRANT ALL ON *.* TO 'test_role';
+
+CREATE TABLE t1 (a INT PRIMARY KEY);
+
+--connect node_1a, 127.0.0.1, x, x, test, $NODE_MYPORT_1
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+DROP USER 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+ALTER USER 'y'@'localhost' PASSWORD EXPIRE NEVER;
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+REVOKE EXECUTE ON *.* FROM 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+GRANT EXECUTE ON PROCEDURE test.p1 TO 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+REVOKE EXECUTE ON PROCEDURE test.p1 FROM 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+GRANT 'test_role' TO 'y'@'localhost';
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+REVOKE 'test_role' FROM 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+GRANT SELECT ON test.t1 TO 'y'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+ALTER USER 'y'@'localhost' DEFAULT ROLE NONE;
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+DROP PROCEDURE test.p1;
+
+#
+# not affected, but let's test them anyway
+#
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+RENAME USER 'y'@'localhost' TO 'z'@'localhost';
+
+--error ER_SPECIFIC_ACCESS_DENIED_ERROR
+SET PASSWORD FOR 'y'@'localhost' TO RANDOM;
+
+#
+# utility user drop
+#
+--error ER_CANNOT_USER
+DROP USER 'utilityuser'@'localhost';
+
+
+#
+# Two nodes interaction
+#
+--connection node_1
+CALL mtr.add_suppression(".*Operation DROP USER failed.*");
+CALL mtr.add_suppression(".*Fail to replicate: DROP USER.*");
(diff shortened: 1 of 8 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/106/

sql/auth/auth_internal.h
```diff
@@ -204,12 +204,11 @@ int replace_routine_table(THD *thd, GRANT_NAME *grant_name, TABLE *table,
...
-#ifdef WITH_WSREP
-int open_grant_tables(THD *thd, TABLE_LIST *tables, bool *transactional_tables,
-                      const char *db = WSREP_MYSQL_DB,
-                      const char *table = NULL);
-#else
 int open_grant_tables(THD *thd, TABLE_LIST *tables, bool *transactional_tables);
+#ifdef WITH_WSREP
+bool start_toi_after_open_grant_tables(THD *thd,
+                                       const char *db = WSREP_MYSQL_DB,
+                                       const char *table = NULL);
```
venkatesh-prasad-v: Can we use `nullptr` here?

sql/sql_parse.cc
```diff
@@ -5340,15 +5340,24 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
-          sp_revoke_privileges(thd, db, name,
-                               lex->sql_command == SQLCOM_DROP_PROCEDURE)) {
-        push_warning(thd, Sql_condition::SL_WARNING, ER_PROC_AUTO_REVOKE_FAIL,
-                     ER_THD(thd, ER_PROC_AUTO_REVOKE_FAIL));
-        /* If this happens, an error should have been reported. */
-        goto error;
+#ifdef WITH_WSREP
+      // sp_drop_routine() returns SP_INTERNAL_ERROR in case of system user
+      // privileges check failure. In such a case we don't want to proceed with
+      // sp_revoke_privieges which will fail anyway.
```
venkatesh-prasad-v:
```suggestion
      // sp_revoke_privileges which will fail anyway.
```

sql/sql_parse.cc
```diff
@@ -5535,8 +5544,7 @@ int mysql_execute_command(THD *thd, bool first_level) {
         WSREP_SYNC_WAIT(thd, WSREP_SYNC_WAIT_BEFORE_SHOW);
       }
 
-      if (lex->sql_command == SQLCOM_ALTER_USER_DEFAULT_ROLE ||
```
venkatesh-prasad-v, line 5538: Why this change?

sql/auth/sql_user_table.cc
```diff
@@ -2040,6 +2014,30 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
   return 0;
 }
 
+#ifdef WITH_WSREP
+bool start_toi_after_open_grant_tables(THD *thd, const char *db,
+                                       const char *table) {
+  /* CREATE/DROP function/procedure implicitly grant priviliges.
```
venkatesh-prasad-v:
```suggestion
  /* CREATE/DROP function/procedure implicitly grant privileges.
```

sql/auth/sql_user_table.cc
```diff
@@ -2040,6 +2014,30 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
...
+  Check for detail comment in respected switch handler in sql_parse.cc
+  Since the original statement is already replicated using TOI
+  sub-action of grant/revoke privilges doesn't need to get replicated. */
```
venkatesh-prasad-v:
```suggestion
  sub-action of grant/revoke privileges doesn't need to get replicated. */
```

sql/auth/sql_user.cc
```diff
@@ -3070,6 +3087,13 @@ bool mysql_drop_user(THD *thd, List<LEX_USER> &list, bool if_exists,
       return true;
     }
 
+#ifdef WITH_WSREP
+    if (start_toi_after_open_grant_tables(thd)) {
+      commit_and_close_mysql_tables(thd);
+      return true;
+    }
+#endif
```
venkatesh-prasad-v, line 3089: Can we move this block after the validation of `utility_user`?. i.e, around line no 3120 in this file.

venkatesh-prasad-v (commented):
In commit message, s/initated/initiated/

sql/auth/sql_user_table.cc
kamil-holubicki: done

sql/auth/sql_user_table.cc
kamil-holubicki: done

sql/sql_parse.cc
kamil-holubicki: done

sql/auth/auth_internal.h
kamil-holubicki: done

sql/auth/sql_user.cc
kamil-holubicki: I think there is no need for doing this. We check if the user we want to drop is the utility user, so it will behave the same on source and replica regardless of the user executing the query.

sql/auth/sql_user.cc
```diff
@@ -2038,6 +2041,13 @@ bool change_password(THD *thd, LEX_USER *lex_user, const char *new_password,
       return true;
     }
 
+#ifdef WITH_WSREP
```
dutow: Shouldn't this be in the `if(WSREP(thd))` block? The start_toi function itself also checks it, so it should work the same - but the current diff was confusing because db_name and table_name aren't set in the else block.

sql/sql_parse.cc
kamil-holubicki: It is not necessary. TOI was started here and then `mysql_alter_or_clear_default_roles()` was called where TOI start attempt was made again. That 2nd attempt was skipped because already in TOI.

venkatesh-prasad-v: Okay. Got it. Thanks for the explanation.

sql/auth/sql_user_table.cc
```diff
@@ -2040,6 +2014,30 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
   return 0;
 }
 
+#ifdef WITH_WSREP
+bool start_toi_after_open_grant_tables(THD *thd, const char *db,
```
dutow, line 2018: Can't this result in a deadlock with unlucky timing? (both nodes lock at the same time, and try to start TOI after that)

kamil-holubicki: The sequence is:
1. Lock tables
2. Start TOI

If node_1 is between (1) and (2), and node_2 executes TOI, the operation on node_1 will fail (BF-abort).
I've added the test case for this.

sql/auth/sql_user.cc
kamil-holubicki: You are right. It will do the same, but the flow will be easier to analyze. Good point.

mysql-test/suite/galera/t/pxc_drop_user.test
```diff
@@ -0,0 +1,115 @@
+# Test that DROP USER executed on the user with SYSTEM_USER privilege
+# is not replicated when executing user has no SYSTEM_USER
+
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
```
dutow, line 6: This also needs have_debug_sync with the latest changes.

dutow (commented):
I added one more comment to the test case, otherwise looks good.

mysql-test/suite/galera/t/pxc_drop_user.test
kamil-holubicki: added

dutow (approved):

PR #1641, opened 2022-05-19
PXC 8.0.28 refresh - Q1 2022

kamil-holubicki:
https://jira.percona.com/browse/PXC-3907

PS 8.0.28 -> PXC merge

Please note that all submodules are taken from my forks for now to allow testing. Once submodules PRs are approved and merged I will adjust pointers to official repos.


sql/sql_class.cc
```diff
@@ -701,7 +701,6 @@ THD::THD(bool enable_plugins)
       wsrep_applier(is_applier),
       wsrep_applier_closing(false),
       wsrep_client_thread(false),
-      wsrep_allow_mdl_conflict(false),
```
dutow: Why do we need to remove these?

storage/tokudb/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2017, Percona and/or its affiliates. All rights reserved.
+ # Copyright (c) 2017, Percona and/or its affiliates. All rights reserved.
```
dutow: Accidental change?

mysql-test/suite/galera/t/galera_alter_table_mdl.test
```diff
@@ -188,11 +188,9 @@ ALTER TABLE t1 ENGINE=InnoDB, ALGORITHM= INPLACE, LOCK= NONE;
 ALTER TABLE t1 ENGINE=InnoDB, ALGORITHM= INPLACE, LOCK= SHARED;
 ALTER TABLE t1 ENGINE=InnoDB, ALGORITHM= INPLACE, LOCK= EXCLUSIVE;
 
-# Tests for inplace rename column (LOCK=NONE is not supported)
---error ER_ALTER_OPERATION_NOT_SUPPORTED
 ALTER TABLE t1 CHANGE COLUMN b c INT, ALGORITHM= INPLACE, LOCK= NONE;
-ALTER TABLE t1 CHANGE COLUMN b c INT, ALGORITHM= INPLACE, LOCK= SHARED;
-ALTER TABLE t1 CHANGE COLUMN c b INT, ALGORITHM= INPLACE, LOCK= EXCLUSIVE;
+ALTER TABLE t1 CHANGE COLUMN c b INT, ALGORITHM= INPLACE, LOCK= SHARED;
+ALTER TABLE t1 CHANGE COLUMN b c INT, ALGORITHM= INPLACE, LOCK= EXCLUSIVE;
```
dutow, line 193: Is this correct? The query says `INPLACE`, not `INSTANT`, and `LOCK=NONE` queries were disabled because of some deadlock issue with concurrent loads.

mysql-test/std_data/sst_garbd_test.sh
```diff
@@ -1,33 +1,30 @@
...
-  if (( $FIRST_RECEIVED == 0 )); then
-    echo "SST request failed"
-    SST_FAILED=1
-    kill $(ps -s $$ -o pid=)
+  if [[ -n "${socat_pid}" ]]; then
+      echo "SST request failed"
+      echo "killing ${socat_pid}"    
+      kill -9 ${socat_pid} ||:
   fi
+  exit 1
```
dutow: This will kill the bash script even if we receive a sigint after the first socat completed. We do not want to do that, as in that case, the sst definitely started correctly.

While this is just a dummy script, the real script in the operator is based on this logic, they should stay as close as possible (since we are testing a simplified version of that): https://github.com/percona/percona-docker/blob/main/percona-xtradb-cluster-8.0-backup/run_backup.sh

garbd will always send a sigint to the script at the end of every sst transfer, when the DONOR switches back to SYNCED. But the script only wants to interpret it as a kill signal if it didn't receive the first socat request yet. Otherwise, it's possible that the second socat data is still in transfer, or that the script is doing some postprocessing on the data received, which should finish.

Also, before making any logic changes to this script, please discuss it with Slava so that the operator team can also test the changes.

sql/sql_class.cc
```diff
@@ -701,6 +701,7 @@ THD::THD(bool enable_plugins)
       wsrep_applier(is_applier),
       wsrep_applier_closing(false),
       wsrep_client_thread(false),
+      wsrep_allow_mdl_conflict(false),
```
dutow: Please see my previous comment about the deletions in sql_class.cc. wsrep_allow_mdl_conflict was set to false at another place in this function too, that should also be restored.

sql/sql_class.cc
kamil-holubicki: fixed. It was wrong conflict resolution

storage/tokudb/CMakeLists.txt
kamil-holubicki: fixed

mysql-test/std_data/sst_garbd_test.sh
kamil-holubicki: fixed as we discussed on Slack

sql/sql_class.cc
kamil-holubicki: wrong conflict resolution. Fixed

mysql-test/suite/galera/t/galera_alter_table_mdl.test
kamil-holubicki: Yes, this is correct. The mentioned commit changed handler0alter.cc implementation. This causes `ha_innobase::check_if_supported_inplace_alter()` returning `HA_ALTER_INPLACE_INSTANT`  here:
```
    // Ask storage engine whether to use copy or in-place
    enum_alter_inplace_result inplace_supported =
        table->file->check_if_supported_inplace_alter(altered_table,
                                                      &ha_alter_info);
```
and then is not changed by following `#ifdef WITH_WSREP` block to `HA_ALTER_INPLACE_SHARED_LOCK_AFTER_PREPARE`.
As the result we do not end up with the error here:
```
      case HA_ALTER_INPLACE_SHARED_LOCK_AFTER_PREPARE:
      case HA_ALTER_INPLACE_SHARED_LOCK:
        // If weaker lock was requested, report errror.
        if (alter_info->requested_lock == Alter_info::ALTER_TABLE_LOCK_NONE) {
          ha_alter_info.report_unsupported_error("LOCK=NONE", "LOCK=SHARED");
          close_temporary_table(thd, altered_table, true, false);
          goto err_new_table_cleanup;
        }
        break;
```

Previously `ha_innobase::check_if_supported_inplace_alter()` returned `HA_ALTER_INPLACE_NO_LOCK_AFTER_PREPARE` which at the end resulted with above mentioned error.

kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/164/

dutow (approved):

PR #1642, opened 2022-05-24
PXC-3909: Update submodule pointer and MTR test

dutow:


kamil-holubicki (approved):
Before merging please remember about merging galera part first and then updating submodule pointer + .gitmodules file.

mysql-test/std_data/sst_garbd_test.sh
```diff
@@ -27,4 +28,8 @@ echo 2
 socat TCP-LISTEN:$1,reuseaddr - > /dev/null
 echo 3
 
+sleep 5
```
kamil-holubicki, line 31: This is some stability hack, or just to make SST process longer?

dutow: This is just for testing: this way the MTR test always triggers the terminate code path.

PR #1648, opened 2022-06-13
PXC-3948: Improve initial socat retry logic

dutow:
Issue: socat retry + timeout doesn't work together on localhost or on
kubernetes, it exits immediately instead of retrying.

Fix: this commit backports the manual nc retry logic from 8.0, which
instead of relying on the socat internal retry implements it manually,
with 1 second sleeps in between.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -947,9 +947,12 @@ read_cnf()
         done
     fi
 
-    # Retry the connection 30 times (at 1-second intervals)
-    if [[ ! "$sockopt" =~ retry= ]]; then
-        sockopt+=",retry=30"
+    if [[ "$WSREP_SST_OPT_ROLE"  == "joiner" ]]; then
+      # only add retry to joiners
+      # donor uses a custom retry logic to avoid a socat bug with retry+timeout on localhost
+      if [[ ! "$sockopt" =~ retry= ]]; then
+       sockopt+=",retry=30"
+      fi
     fi
 
     xbstream_opts=$(parse_cnf sst xbstream-opts "")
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -947,11 +947,6 @@ read_cnf()
         done
     fi
 
-    # Retry the connection 30 times (at 1-second intervals)
```
kamil-holubicki, line 950: I'm not sure if we should just remove it. This is used globally, so on the joiner (server) side as well.
As socat man page says:
```
retry=<num>
Number of retries before the connection or listen attempt is aborted. Default is 0, which means just one attempt.
```

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1476,11 +1471,24 @@ send_data_from_donor_to_joiner()
 
     pushd ${dir} 1>/dev/null
     set +e
-    timeit "$msg" "$strmcmd | $tcmd; RC=( "\${PIPESTATUS[@]}" )"
+
+
+    # Implement retry logic ourselves, as nc needs it, and socat has a bug
```
kamil-holubicki, line 1480: Here we've got hardcoded 30. Previously it was configurable via `retry` config parameter. Maybe it would be better to detect the value of `retry` in `read_cnf()` and use it here in the loop?

dutow: It was configurable, but not documented, and I didn't see any practical use in our uses, so I'm not sure if that's really needed.

Also this is how the 8.0 code works for nc, with a hardcoded loop up to 30, ignoring the retry value.

scripts/wsrep_sst_xtrabackup-v2.sh
dutow: I readded the retry parameter for joiners.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1476,7 +1475,24 @@ send_data_from_donor_to_joiner()
...
+    # when retry is used together with timeout
+    local rc=1
+    local counter=1
+
+    # remove the retry parameter from the command
+    local ltcmd=$(echo $tcmd | sed s/,sockopt=[0-9]*//)
```
kamil-holubicki: Are we going to remove all sockopt? not only retry=?
Something like:
`local ltcmd=${tcmd//,*retry=[0-9]*/""}`

dutow: :facepalm: I wanted to write retry there instead of sockopt, fixed.

kamil-holubicki (approved):

PR #1650, opened 2022-06-15
[5.7] PXC-3951: Current computation of disk used with sst-idle-timeout could misidentify SST as idle

venkatesh-prasad-v:

Problem
---

SST is in progress.

Solution
---

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -306,7 +306,7 @@ monitor_sst_progress() {
       break;
     fi
 
-    current_size=$(du --max-depth=1 ${tmpsstdir} | tail -n 1 | awk '{print $1}')
+    current_size=$(du -b -s ${tmpsstdir} | awk '{print $1}')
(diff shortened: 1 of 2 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -306,7 +306,7 @@ monitor_sst_progress() {
       break;
     fi
 
-    current_size=$(du --max-depth=1 ${tmpsstdir} | tail -n 1 | awk '{print $1}')
+    current_size=$(du -b -s ${tmpsstdir} | tail -n 1 | awk '{print $1}')
```
altmannmarcelo: do we need `| tail -1` if we use --summarize and only ${tmpsstdir} as target ?

altmannmarcelo (approved):
apart from tail, PR looks good. You can decide what to do (keep or remove) and it's fine to me either way.

scripts/wsrep_sst_xtrabackup-v2.sh
venkatesh-prasad-v: Removed `| tail -n 1`

PR #1653, opened 2022-06-16
[8.0] PXC-3951: Current computation of disk used with sst-idle-timeout could misidentify SST as idle

venkatesh-prasad-v:
    
Merge remote-tracking branch 'origin/5.7-PXC-3951' into 8.0-PXC-3951
    
Problem
=======
    
SST is in progress.

Solution
========

Jenkins:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/314/console


altmannmarcelo (approved):
LGTM

PR #1654, opened 2022-06-20
PXC-3951: Current computation of disk used with sst-idle-timeout could misidentify SST as idle

venkatesh-prasad-v:
**Note: This is a PR to 8.0.28 release branch**

Merge remote-tracking branch 'origin/5.7-PXC-3951' into 8.0-PXC-3951

Problem
=======

SST is in progress.

Solution
========

(cherry picked from commit c387b056334bdc3d6124a20fb01aca254930b187)


altmannmarcelo (approved):
LGTM

PR #1655, opened 2022-06-21
PXC-3948: Improve initial socat retry logic (8.0)

dutow:
Issue: socat retry + timeout doesn't work together on localhost or on
kubernetes, it exits immediately instead of retrying.

Fix: instead of relying on the internal socat retry logic, the script
now uses the same manual retry logic that was already present for nc

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1371,24 +1371,22 @@ send_data_from_donor_to_joiner()
 
     set +e
 
-    if [[ $tfmt == "nc" ]]; then
-        # if using nc as a transfer method, then implement the retry
-        # logic ourselves
-        local rc=1
-        local counter=1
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1371,24 +1371,22 @@ send_data_from_donor_to_joiner()
...
+    local rc=1
+    local counter=1
+
+    # remove the retry parameter from the command
+    local ltcmd=$(echo $tcmd | sed s/,retry=[0-9]*//)
+
+    while [[ $rc -ne 0 && counter -le 30 ]]; do
+        timeit "$msg" "$strmcmd | $ltcmd; RC=( "\${PIPESTATUS[@]}" )"
+
+        # Retry if nc returns an error
```
kamil-holubicki: nc or socat

kamil-holubicki (approved):
lgtm

PR #1658, opened 2022-06-24
Update garbd_howto.rst

patrickbirch:
Flag socket.ssl=YES; was missing from options. Without this flag, ssl would not be considered and garbd start will fail in case SSL is enabled (which is by defualt on 8.0)


alina-derkach-oaza (approved):

utdrmac:
@alina-derkach-oaza What's holding this merge up? Almost 3 weeks old.

PR #1660, opened 2022-07-06
PXC-702: GCache encryption

kamil-holubicki:
PXC-3063: Record Set cache data at rest encryption

https://jira.percona.com/browse/PXC-702
https://jira.percona.com/browse/PXC-3063

Server part implementation.
1. Needed changes around keyring component and Galera initialization
2. Master key rotation


plugin/keyring/common/keys_container.cc
```diff
@@ -183,8 +206,11 @@ bool Keys_container::load_keys_from_keyring_storage() {
       if (serialized_keys->get_next_key(&key_loaded) || key_loaded == nullptr ||
           key_loaded->is_key_valid() == false ||
           store_key_in_hash(key_loaded)) {
-        was_error = true;
-        delete key_loaded;
+#ifdef WITH_WSREP
+        if (forceKeysFetch) {
```
dutow: Shouldn't `else` keep existing behavior?

plugin/keyring/common/keys_container.cc
```diff
@@ -88,6 +92,12 @@ bool Keys_container::store_key_in_hash(IKey *key) {
 }
 
 bool Keys_container::store_key(IKey *key) {
+#ifdef WITH_WSREP
+  if (forceKeysFetch) {
+    load_keys_from_keyring_storage();
```
dutow: These hacks ignore the return value of load_keys_from_keyring_storage, which might fail, especially with a remote keyring.

dutow (commented):
This seems too much of a hack to me.

Can't we explicitly reload the keyring cache after SST instead?

plugin/keyring/common/keys_container.cc
kamil-holubicki: The part related to this hack has been completely reworked. Now Keyring plugin is reloaded after SST.

plugin/keyring/common/keys_container.cc
kamil-holubicki: The part related to this hack has been completely reworked. Now Keyring plugin is reloaded after SST.

PR #1668, opened 2022-08-02
PXC-3982: Slave SQL thread fails with WSREP not ready yet error

kamil-holubicki:
https://jira.percona.com/browse/PXC-3982

Problem:
It occurs when PXC node is the async replica to the PS.
When PXC node starts it kicks-off Galera/WSREP initialization which does
SST/IST. While it is happening async replication infrastructure is
initialized in parallel.
It has been observed that sometimes async replica's SQL thread reports
'WSREP has not yet prepared node for application use' upon server start
and async replication stops.

Cause:
PXC node blocks processing of application events until the node reaches
'Synced' state. During the initialization sometimes it happens that
async replication is ready before Galera part finishes initialization
getting to the 'Synced' state. In such a case async-replicated event is
going to be applied, but it is blocked, causing the error, because PXC node
is not in 'Synced' state yet.

Solution:
If replica node is PXC node, delay applying of async-replicated events
(SQL thread) until node reaches 'Synced' state.

### DIFF
--- a/mysql-test/suite/galera/t/galera_as_slave_restart.test
+++ b/mysql-test/suite/galera/t/galera_as_slave_restart.test
@@ -1,0 +1,83 @@
+#
+# Test Galera as a replica to a MySQL source being restarted.
+# While restarting Galera replica, new events appear on source.
+# As soon as async replica is started, they are replicated
+# but if WSREP not ready yet (Galera replica node is not in sync state)
+# applying them should be delayed to not cause
+# 'WSREP has not yet prepared node for application use' error.
+#
+# The galera/galera_2node_slave.cnf describes the setup of the nodes
+#
+
+--source include/have_debug.inc
+--source include/have_log_bin.inc
+--source include/force_restart.inc
+
+# As node #1 is not a Galera node, we connect to node #3 in order to run include/galera_cluster.inc
+--connect node_3, 127.0.0.1, root, , test, $NODE_MYPORT_3
+--source include/galera_cluster_master_slave.inc
+
+--connection node_3
+--echo #node-3 (async slave)
+call mtr.add_suppression("Recovery from master pos");
+call mtr.add_suppression("when attempting to connect");
+--disable_query_log
+--eval CHANGE REPLICATION SOURCE TO SOURCE_HOST='127.0.0.1',SOURCE_USER='root',SOURCE_PORT=$NODE_MYPORT_1;
+--enable_query_log
+START REPLICA;
+
+#
+# Test if replication is working fine.
+--connection node_1
+--echo #node-1 (source)
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1);
+
+--connection node_3
+--echo #node-3 (async replica)
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't1';
+--source include/wait_condition.inc
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t1;
+--source include/wait_condition.inc
+
+#
+# shutdown node-3
+--echo #Shutting down node-3 (async replica) ...
+--source include/shutdown_mysqld.inc
+
+#
+# Some new events to be replicated when node-3 restarts.
+--connection node_1
+--echo #node-1 (source)
+INSERT INTO t1 VALUES(2);
+
+#
+# Start node-3 but force async replication to start before WSREP reports being ready.
+--connection node_3
+--echo #node-3 (async replica)
+--let $start_mysqld_params="--debug='+d,pause_before_wsrep_ready'"
+--source include/start_mysqld.inc
+
+#
+# re-test if replication is working fine.
+--connection node_1
+--echo #node-1 (source)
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/485/

sql/wsrep_server_service.cc
```diff
@@ -342,6 +346,28 @@ void Wsrep_server_service::log_state_change(
...
+      /* This happens when the server starts. We are in wsrep applier context.
+         The intention here is to simulate delays in setting wsrep_ready = true,
+         to check how things initializing asynchronously in parallel react to
+         wsrep_ready == false.
+         We can't block here on DEBUG_SYNC because:
+         1. Whoever needs wsrep_ready == true, needs to wait calling
+         Wsrep_server_state::instance().wait_until_state(Wsrep_server_state::s_synced)
+         Such a call is unblocked by current context, just after returning
+         from this method, so we've got the situation when a client would need to
+         a) wait on wait_until_state() but that unbocks only after signalling
```
venkatesh-prasad-v:
```suggestion
         a) wait on wait_until_state() but that unblocks only after signalling
```

mysql-test/suite/galera/t/galera_as_slave_restart.test
```diff
@@ -0,0 +1,83 @@
...
+# 'WSREP has not yet prepared node for application use' error.
+#
+# The galera/galera_2node_slave.cnf describes the setup of the nodes
+#
+
+--source include/have_debug.inc
+--source include/have_log_bin.inc
+
+# As node #1 is not a Galera node, we connect to node #2 in order to run include/galera_cluster.inc
+--connect node_3, 127.0.0.1, root, , test, $NODE_MYPORT_3
```
venkatesh-prasad-v: Comment says we are using node#2, but the connection is done on node_3. Shouldn't it be node_2?

sql/rpl_replica.cc
```diff
@@ -6099,6 +6099,20 @@ static void *handle_slave_worker(void *arg) {
...
+     Allow async replication infrastructure to init up to this point, but
+     wait with applying async-replicated events until WSREP infrastructure
+     is fully initialized (wsrep_ready == true). wsrep_ready is set when
+     server state shifts to s_synced.
+     Before this state, events are rejected with error
+     'WSREP has not yet prepared node for application use' */
+  if (WSREP(thd)) {
+    Wsrep_server_state::instance().wait_until_state(
+        Wsrep_server_state::s_synced);
+  }
```
venkatesh-prasad-v: This code gets executed only when the server is configured with `--replica-parallel-workers>1` i.e, in  multi-threaded replica mode. If there is only one async applier thread running on replica, then we may still get the issue. I think moving this wait to `handle_slave_sql()` before we call `slave_start_workers()` would handle both single-threaded and multi-threaded replicas.

mysql-test/suite/galera/t/galera_as_slave_restart.test
```diff
@@ -0,0 +1,83 @@
...
+# Some new events to be replicated when node-3 restarts.
+--connection node_1
+--echo #node-1 (source)
+INSERT INTO t1 VALUES(2);
+
+#
+# Start node-3 but force async replication to start before WSREP reports being ready.
+--connection node_3
+--echo #node-3 (async replica)
+--let $start_mysqld_params="--debug='+d,pause_before_wsrep_ready'"
```
venkatesh-prasad-v, line 58: Don't we need to clear this debug simulation? May be `--source include/force_restart.inc` in the beginning of the file.

venkatesh-prasad-v (changes requested):

sql/wsrep_server_service.cc
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_as_slave_restart.test
kamil-holubicki: fixed

mysql-test/suite/galera/t/galera_as_slave_restart.test
kamil-holubicki: added

sql/rpl_replica.cc
kamil-holubicki: Right! Fixed.

mysql-test/suite/galera/t/galera_as_slave_restart.test
```diff
@@ -0,0 +1,84 @@
...
+--connection node_3
+--echo #node-3 (async replica)
+--let $wait_condition = SELECT COUNT(*) = 3 FROM t1;
+--source include/wait_condition.inc
+
+--connection node_1
+DROP TABLE t1;
+
+--connection node_3
+--sleep 1
```
venkatesh-prasad-v: This sleep is redundant as we will be waiting anyways in the next step.

venkatesh-prasad-v (approved):
Approved with an optional comment.

mysql-test/suite/galera/t/galera_as_slave_restart.test
kamil-holubicki: comment addressed

PR #1669, opened 2022-08-08

kamil-holubicki:
(post merge fix)

Problem:
It has been discovered that executing DROP INDEX in parallel to DML can
cause BF-BF abort and node shutdown.

Cause:
The problem is similar to already merged fixes to BF-BF in scope of the
same ticket. When we have 2 tables, parent and child, DROP INDEX
executed in parallel to DML can cause abort of already certified and
replicated DML. In such a case DML is replayed locally, but can cause
BF-BF abort on replica. This is because the certification process has
not enough information about tables dependencies, which leads to faulty
certification process/transactions ordering.

Solution:
Add parent/child tables to certification keys set when executing
DROP/CREATE INDEX. This causes local DML to fail certification.

### DIFF
--- a/mysql-test/suite/galera_nbo/t/galera_bf_bf.test
+++ b/mysql-test/suite/galera_nbo/t/galera_bf_bf.test
@@ -83,6 +83,77 @@ SET SESSION wsrep_osu_method=TOI;
 DROP TABLE c1;
 DROP TABLE p1;
 
+#
+# Similar to the above, but for DROP INDEX
+#
+# Test that DROP INDEX on parent table is properly ordered with simultaneous
+# transactions that perform DML on child table.
+#
+# UPDATE should fail certification as its certification keys conflict
+# with DROP INDEX certification keys.
+#
+# Without the fix, UPDATE is certified, then aborted by already replicated TOI
+# transaction which causes its replay. When replaying it conflicts with TOI
+# causing BF-BF conflict.
+#
+# This test checks if parent table is included into the certification keys
+# of the writeset related to TOI done on child table.
+#
+
+--connection node_1_toi
+CREATE TABLE p1 (pk INTEGER PRIMARY KEY, f2 INTEGER, KEY f2_key (f2));
+CREATE TABLE c1 (pk INTEGER PRIMARY KEY, data INTEGER, fk INTEGER, FOREIGN KEY (fk) REFERENCES p1(pk));
+
+INSERT INTO p1 VALUES (0, 0);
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/520/

sql/wsrep_server_service.cc
```diff
@@ -342,6 +342,28 @@ void Wsrep_server_service::log_state_change(
...
+         from this method, so we've got the situation when a client would need
+         to a) wait on wait_until_state() but that unbocks only after signalling
+         the DEBUG_SYNC and b) signal this debug sync after the wait, which
+         situations are mutually exclusive.
+         2. If we blocked here with DEBUG_SYNC's WAIT_FOR signal, there would be
+         no way to signal it from user's connection, because we are owning
+         wsrep::server_state::mutex_, which in case of wsrep_ready==false is
+         acquired by the connection as well in
+         block_write_while_in_rolling_upgrade(), so we would have a deadlock.
+         Introducing a delay is enough for testing. */
```
venkatesh-prasad-v: Looks a leftover from PXC-3982 bugfix. Please remove these changes.

mysql-test/suite/galera_nbo/t/galera_bf_bf.test
```diff
@@ -83,6 +83,88 @@ SET SESSION wsrep_osu_method=TOI;
...
+#set global wsrep_min_log_verbosity=3;
+#set global wsrep_debug=SERVER;
+#set global wsrep_provider_options="cert.log_conflicts=yes";
+#set global wsrep_log_conflicts=1;
+
+--connection node_1_toi
+#set global wsrep_min_log_verbosity=3;
+#set global wsrep_debug=SERVER;
+#set global wsrep_provider_options="cert.log_conflicts=yes";
+#set global wsrep_log_conflicts=1;
```
venkatesh-prasad-v: Unused?

venkatesh-prasad-v (changes requested):

sql/wsrep_server_service.cc
kamil-holubicki: removed

mysql-test/suite/galera_nbo/t/galera_bf_bf.test
kamil-holubicki: removed

venkatesh-prasad-v (approved):

PR #1675, opened 2022-08-11
PXC-3999: Cluster stuck on "Waiting for acl cache lock" with concurrent User DDL commands

kamil-holubicki:
https://jira.percona.com/browse/PXC-3999

Problem:
When multiple CREATE USER statements are executed in parallel the node
can stuck with deadlock.

Cause:
This is the regression caused by commit
6cc5a28b7c4a72239b35804fc704c2ee9a32abad.
In that commit the order of acquiring acl_cache_lock and starting TOI
was changed (now acl_cache_lock is acquired first).
When CREATE USER is executed, acl_cache_lock is acquired, then TOI is
started. So far so good, but then acl_cache_lock is released and
acquired again. This may lead to the situation when another thread
acquires acl_cache_lock and tries to start TOI, which will block as the
1st thread has active TOI. Then, if the 1st thread tries to acquire
acl_cache_lock it will end up with a deadlock.

Additionally, it has been found that CHANGE PASSWORD has wrong
acl_cache_lock and start TOI order, which could lead to the deadlock
between session changing the password and another session which executes
user alter related statement (e.g. RENAME USER)

Other paths are not affected.

Solution:
1. Extended acl_cache_lock critical session scope in mysql_create_user()
function. Now the acl_cache_lock is not released in the middle of the
function keeping it in write mode. As previously it was reacquired in
read mode, this potentially affects concurrency, but it should not be
a big deal.
2. Fixed locking order in change_password() function.

### DIFF
--- a/sql/auth/sql_user.cc
+++ b/sql/auth/sql_user.cc
@@ -2004,11 +2004,6 @@ bool change_password(THD *thd, LEX_USER *lex_user, const char *new_password,
       thd->set_query(query_save);
       return (ret != 1);
     }
-    // we start TOI only for WSREP threads
-    if (start_toi_after_open_grant_tables(thd, WSREP_MYSQL_DB, "user")) {
-      commit_and_close_mysql_tables(thd);
-      return true;
-    }
   } else {
     if ((ret = open_grant_tables(thd, tables, &transactional_tables)))
       return (ret != 1);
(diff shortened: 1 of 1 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/522/
failures not related to the fix (unstable tests, or not fixed yet CI_FS tests)

sql/auth/sql_user.cc
```diff
@@ -2978,7 +2985,15 @@ bool mysql_create_user(THD *thd, List<LEX_USER> &list, bool if_not_exists,
     User_params user_params(&extra_users);
     result = log_and_commit_acl_ddl(thd, transactional_tables, &extra_users,
                                     &user_params);
+#ifndef WITH_WSREP
+    // clang-format off
+    /* We already started TOI. Below we will acquire ACL lock again.
```
venkatesh-prasad-v, line 2990: Does this problem exist with only CREATE TABLE? i.e, are there any other ACL statements which release the ACL lock temporarily and reacquire it later?

kamil-holubicki: I reviewed all cases where `acl_cache_lock` is used together with `start_toi_after_open_grant_tables` and found that besides `CREATE USER`, there is `ALTER USER` which suffers the same problem (exiting and re-entering critical section). I can't explain how I missed it providing the fix.
Fixed.

kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/540/

venkatesh-prasad-v (approved):

venkatesh-prasad-v (approved):

PR #1683, opened 2022-08-31
PXC-3943: PXC 8.0.29 refresh - Q2 2022 (Fix MTR tests)

venkatesh-prasad-v:


sql/persisted_variable.cc
```diff
@@ -2094,6 +2115,27 @@ bool Persisted_variables_cache::reset_persisted_variables(THD *thd,
...
+       * wsrep_slave_UK_checks, wsrep_applier_UK_checks
+       *
+       * SET PERSIST of these variables can result in multiple entries in the
+       * performance_schema.persisted_variables. After server restart, any
+       * attempt to RESET PERSIST of these variables may fail the operation and
+       * result in an inconsistent state.
+       *
+       * So, temporarily we skip converting "wsrep_" variables to lowercase
+       * before persisting them. */
+      const char wsrep_prefix[]{"wsrep_"};
```
kamil-holubicki: static constexpr?

Or even better (name_str is std::string) something like:
```
static const std::string wsrep_prefix("wsrep_");
if (!name_str.rfind(wsrep_prefix, 0)) {
  // name starts with wsrep_
}
```
or
```
static const std::string wsrep_prefix("wsrep_");
if (!name_str.compare(o, wsrep_prefix.size(), wsrep_prefix)) {
  // name starts with wsrep_
}

```

venkatesh-prasad-v: Changed to 
```
+      static const std::string wsrep_prefix("wsrep_");
+      if (!name_str.rfind(wsrep_prefix, 0)) {
+        // name starts with wsrep_
+        name_str.assign(name_cptr);
+      }
```

kamil-holubicki (approved):
lgtm

PR #1688, opened 2022-09-07
PXC-4033: Assertion during execution of prepared statement

kamil-holubicki:
https://jira.percona.com/browse/PXC-4033

Problem:
When the prepared statement is executed in parallel to the DDL modifying the table which prepared statement uses the server fails with assertion saying that prepared statement transaction was aborted, so it cannot be commited.

Cause:
When the session executing prepared statement is opening the table it detects that table's metadata changed. Execution is aborted, internal error is propagated up the callstack and as the transaction ended with error, wsrep part marks it as aborted.
Error is internal, its purpose is to trigger prepared statement re-execution, which happens, but wsrep transaction state is in aborted state. Later when the transaction is commited, it is detected that transaction is aborted and the assert triggers.

During the investigation I've found that for multi statement transactions the state is also changed to aborted which causes skipping certification (and replication) of such transaction, but local commit happens which leads to the cluster inconsistency.

Solution:
Before re-execution of aborted prepared statement do the cleanup (call wsrep_after_statement()).
Added assertions in wsrep-lib.

### DIFF
--- a/sql/sql_prepare.cc
+++ b/sql/sql_prepare.cc
@@ -179,6 +179,7 @@ When one supplies long data for a placeholder:
 #include "sql/window.h"
 #include "sql_string.h"
 #include "violite.h"
+#include "wsrep_trans_observer.h"
 
 namespace resourcegroups {
 class Resource_group;
@@ -3164,9 +3165,16 @@ reexecute:
       }
     }
 
-    if (!error) /* Success */
+    if (!error) { /* Success */
+#ifdef WITH_WSREP
+      // We are going to retry the statement, so clean up first.
+      wsrep_after_statement(thd);
+#endif
       goto reexecute;
+    }
+#ifdef WITH_WSREP
   }
+#endif
   reset_stmt_parameters(this);
 
   // Reenable the general log if it was temporarily disabled while repreparing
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/673/

sql/sql_prepare.cc
```diff
@@ -3156,9 +3157,16 @@ bool Prepared_statement::execute_loop(String *expanded_query,
       }
     }
 
-    if (!error) /* Success */
+    if (!error) { /* Success */
+#ifdef WITH_WSREP
+      // We are going to retry the statement, so clean up first.
+      wsrep_after_statement(thd);
+#endif
```
venkatesh-prasad-v, line 3172: Can we do it this way please?
```diff
     }
 
     if (!error) /* Success */
+#ifdef WITH_WSREP
+    // We are going to retry the statement, so clean up first.
+    {
+      wsrep_after_statement(thd);
+#endif /* WITH_WSREP */
       goto reexecute;
+#ifdef WITH_WSREP
+    }
+#endif /* WITH_WSREP */
   }
   reset_stmt_parameters(this);
```

kamil-holubicki: Right! That's the only way.
Added also missing WITH_WSREP around the include file.

venkatesh-prasad-v (approved):

PR #1694, opened 2022-10-05
PXC-4048: Empty GRA_x_y_v2.log files

kamil-holubicki:
https://jira.percona.com/browse/PXC-4048

Problem:
GRA_* files created in case of applier failures are empty.

Cause:
wsrep_dump_rbr_buf_with_header() which creates GRA_* files used Binlog_cache_storage class object a the auxiliary cache. The case was created by providing paging file used by cache which was target GRA_* file. The idea behing the implementation was that Binlog_cache_storage::flush() should write all the content of the cache kept in the RAM memory into the file. Unfortunately Binlog_cache_storage::flush() method is the dummy one.

Solution:
Used IO_CACHE_binlog_cache_storage which is also used in wsrep_to_buf_helper() during TOI writeset creation. Introduced wsrep_write_cache_file() function which dumps the cache into the provided file.
As wsrep_write_cache_file() and wsrep_write_cache_buf() functions have a lot in common, the common part was extracted.

MTR test has been fixed. It passed before because of wrong output recorded.

### DIFF
--- a/sql/wsrep_binlog.cc
+++ b/sql/wsrep_binlog.cc
@@ -22,11 +22,57 @@
 #include "binlog.h"
 #include "log.h"
 #include "log_event.h"  // Log_event_writer
+#include "mutex_lock.h"
 #include "mysql/psi/mysql_file.h"
 #include "service_wsrep.h"
+#include "sql/sql_base.h"  // TEMP_PREFIX
 #include "transaction.h"
 #include "wsrep_applier.h"
-#include "mutex_lock.h"
+
+static bool wsrep_dump_cache_common(
+    IO_CACHE_binlog_cache_storage *cache,
+    std::function<bool(unsigned char *, my_off_t)> dumpFn) {
+  my_off_t const saved_pos(cache->position());
+
+  unsigned char *read_pos = NULL;
+  my_off_t read_len = 0;
+
+  if (cache->begin(&read_pos, &read_len)) {
+    WSREP_ERROR("Failed to initialize io-cache");
+    return true;
+  }
+
+  if (read_len == 0 && cache->next(&read_pos, &read_len)) {
+    WSREP_ERROR("Failed to read from io-cache");
+    return true;
+  }
+
+  bool res = false;
+  while (read_len > 0) {
+    if (dumpFn(read_pos, read_len)) {
+      res = true;
+      break;
+    }
+    cache->next(&read_pos, &read_len);
+  }
+
+  if (cache->truncate(saved_pos)) {
+    WSREP_WARN("Failed to reinitialize io-cache");
+    res = true;
+    ;
+  }
+
+  return res;
+}
+
+bool wsrep_write_cache_file(IO_CACHE_binlog_cache_storage *cache, File file) {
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/720/

sql/wsrep_binlog.cc
```diff
@@ -85,33 +118,22 @@ int wsrep_write_cache_buf(IO_CACHE_binlog_cache_storage *cache, uchar **buf,
...
-  }
+    return false;
+  });
 
-  if (cache->truncate(saved_pos)) {
-    WSREP_WARN("Failed to reinitialize io-cache");
-    goto cleanup;
+  if (res) {
+    my_free(*buf);
+    *buf = NULL;
```
dutow: nullptr

sql/wsrep_binlog.cc
```diff
@@ -22,11 +22,57 @@
...
+      res = true;
+      break;
+    }
+    cache->next(&read_pos, &read_len);
+  }
+
+  if (cache->truncate(saved_pos)) {
+    WSREP_WARN("Failed to reinitialize io-cache");
+    res = true;
+    ;
```
dutow: A break or something missing here?

sql/wsrep_binlog.cc
```diff
@@ -22,11 +22,57 @@
...
+
+  if (read_len == 0 && cache->next(&read_pos, &read_len)) {
+    WSREP_ERROR("Failed to read from io-cache");
+    return true;
+  }
+
+  bool res = false;
+  while (read_len > 0) {
+    if (dumpFn(read_pos, read_len)) {
+      res = true;
```
dutow, line 53: res = true; break; doesn't seem to do anything differently than a simple "return true;"?

sql/wsrep_binlog.cc
```diff
@@ -22,11 +22,57 @@
...
 #include "transaction.h"
 #include "wsrep_applier.h"
-#include "mutex_lock.h"
+
+static bool wsrep_dump_cache_common(
+    IO_CACHE_binlog_cache_storage *cache,
+    std::function<bool(unsigned char *, my_off_t)> dumpFn) {
+  my_off_t const saved_pos(cache->position());
+
+  unsigned char *read_pos = NULL;
```
dutow: nullptr

sql/wsrep_binlog.cc

sql/wsrep_binlog.cc

sql/wsrep_binlog.cc
kamil-holubicki: No, this is just a not needed semicolon.

sql/wsrep_binlog.cc
kamil-holubicki: It truncates the cache after the loop.

dutow (approved):

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/724/

PR #1697, opened 2022-10-21
(5.7) PXC-3821: Truncating performance_schema's table in one node, replicating the truncate statement in other nodes also

kamil-holubicki:
https://jira.percona.com/browse/PXC-3821

Problem:
Truncation of P_S table is replicated across the cluster

Cause:
Truncation of P_S tables is done exactly the same as for any other table, so the query is replicated as TOI.

Solution:
1. Filter out P_S tables truncation and do not initiate TOI for them.
2. For the case when there is source node without this fix, filter out truncation on the replica node in the same way as it was introduced in commit 44c56d37dfe767c31b5ca80812229881c32140f3 (is_executed_by_slave() function fixed to be aware of wsrep applier thread).

### DIFF
--- a/mysql-test/suite/galera/r/galera_truncate.result
+++ b/mysql-test/suite/galera/r/galera_truncate.result
@@ -27,3 +27,5 @@ DROP TABLE t1;
 DROP TABLE t2;
 DROP TABLE t3;
 DROP TABLE t4;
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-5.7-pipeline/4924/

venkatesh-prasad-v (approved):

PR #1698, opened 2022-10-24
(8.0) PXC-3821: Truncating performance_schema's table in one node, replicating the truncate statement in other nodes also

kamil-holubicki:
https://jira.percona.com/browse/PXC-3821

Problem:
Truncation of P_S table is replicated across the cluster

Cause:
Truncation of P_S tables is done exactly the same as for any other table, so the query is replicated as TOI.

Solution:

Filter out P_S tables truncation and do not initiate TOI for them.
For the case when there is source node without this fix, filter out truncation on the replica node in the same way as it was introduced in commit https://github.com/percona/percona-xtradb-cluster/commit/44c56d37dfe767c31b5ca80812229881c32140f3 (is_executed_by_slave() function fixed to be aware of wsrep applier thread).

### DIFF
--- a/mysql-test/suite/galera/t/galera_truncate.test
+++ b/mysql-test/suite/galera/t/galera_truncate.test
@@ -55,3 +55,34 @@ DROP TABLE t1;
 DROP TABLE t2;
 DROP TABLE t3;
 DROP TABLE t4;
+
+#
+# TRUNCATE TABLE performance_schema.<table> should not be replicated
+#
+--disable_query_log
+--disable_result_log
+
+--connection node_2
+# We need to execute SELECT COUNT before getting the actual result to have it already in events_statements_summary_by_digest table 
+SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest;
+--let $node_2_count_before = `SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest`
+
+--connection node_1
+TRUNCATE TABLE performance_schema.events_statements_summary_by_digest;
+
+# It should be 1 at the moment. We will not use assert.inc directly, as it sends more queries, and the result is not deterministic.
+--let $node_1_count = `SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest`
+--let $assert_text = "performance_schema.events_statements_summary_by_digest table should be truncated"
+--let $assert_cond = $node_1_count = 1
+--source include/assert.inc
+
+--connection node_2
+--let $node_2_count_after = `SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest`
+
+# Check if node_2 is not affected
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/751/

mysql-test/suite/galera/t/galera_truncate.test
```diff
@@ -103,3 +103,34 @@ TRUNCATE TABLE t1;
...
 DROP TABLE t1;
+
+#
+# TRUNCATE TABLE performance_schema.<table> should not be replicated
+#
+--disable_query_log
+--disable_result_log
+
+--connection node_2
+# We need to execute SELECT COUNT before geting the actual result to have it already in events_statements_summary_by_digest table
```
venkatesh-prasad-v, line 114:
```suggestion
# We need to execute SELECT COUNT before getting the actual result to have it already in events_statements_summary_by_digest table 
```

mysql-test/suite/galera/t/galera_truncate.test
```diff
@@ -103,3 +103,34 @@ TRUNCATE TABLE t1;
...
+--let $node_2_count_before = `SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest`
+
+--connection node_1
+TRUNCATE TABLE performance_schema.events_statements_summary_by_digest;
+
+# It should be 1 at the moment. We will not use assert.inc directly, as it sends more queries, and the result is not deterministic.
+--let $node_1_count = `SELECT COUNT(*) FROM performance_schema.events_statements_summary_by_digest`
+--let $assert_text = "performance_schema.events_statements_summary_by_digest table should be truncated"
+--let $assert_cond = $node_1_count = 1
+--source include/assert.inc
```
venkatesh-prasad-v, line 125: Here we say that 
> We will not use assert.inc directly

But we are using assert.inc. Is this expected?

kamil-holubicki: Yes, this is fine. The point is to not use it like
`--let $assert_cond = COUNT(*) = 1 FROM performance_schema.events_statements_summary_by_digest`

venkatesh-prasad-v: Ok. Got it. Thanks for the clarification.

mysql-test/suite/galera/t/galera_truncate.test
kamil-holubicki: Fixed in 5.7 PR. 
Once https://github.com/percona/percona-xtradb-cluster/pull/1697 is approved, I will merge it to this PR according to GCA.

venkatesh-prasad-v (approved):

kamil-holubicki:
https://ps80.cd.percona.com/view/Parallel%20MTR/job/percona-server-8.0-pipeline-parallel-mtr/289/

PR #1703, opened 2022-11-07
PXC-512: Replication slave from PXC can crash if wsrep_forced_binlog_format=ROW

voodam:
https://jira.percona.com/browse/PXC-512

Problem:
Duplication of procedure on replication if `wsrep_forced_binlog_format=ROW`.

Cause:
`THD::set_current_stmt_binlog_format_row` and `THD::clear_current_stmt_binlog_format_row` methods inconsistent with `THD::is_current_stmt_binlog_format_row` (when `wsrep_forced_binlog_format` is enabled): `THD::is_current_stmt_binlog_format_row` returns value of `wsrep_forced_binlog_format`, but modifying methods update `THD::current_stmt_binlog_format` variable. So, when we temporary clear binlog format it continues to be forced set and this breaks replication logic.

Solution:
Update also `wsrep_forced_binlog_format` variable. For simplicity, global variable is updated for all sessions. Because `wsrep_forced_binlog_format` is deprecated this is considered enough.

Comment: probably this is dirty solution, but I tried for the first time assuming this can be enough.

Alternative approach, may be it will be better:
Create `THD::wsrep_forced_binlog_format_save` variable and modify it instead of global variable. Also it requires to pass `thd` into `WSREP_BINLOG_FORMAT`, because we need to check not only `wsrep_forced_binlog_format` but also `THD::wsrep_forced_binlog_format_save`.

### DIFF
--- a/sql/sql_class.h
+++ b/sql/sql_class.h
@@ -270,11 +270,11 @@ enum enum_binlog_format {
 };
 
 #ifdef WITH_WSREP
-#define WSREP_BINLOG_FORMAT(my_format)                         \
-   ((wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC) ?     \
-   wsrep_forced_binlog_format : my_format)
-#else
-#define WSREP_BINLOG_FORMAT(my_format) my_format
+#define WSREP_BINLOG_FORMAT(thd, my_format)                                 \
+   ((thd->wsrep_forced_binlog_format_override() != BINLOG_FORMAT_UNSPEC) ?  \
+     thd->wsrep_forced_binlog_format_override() :                           \
+     ((wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC) ?                \
+       wsrep_forced_binlog_format : my_format))
 #endif /* WITH_WSREP */
 
 /* Bits for different SQL modes modes (including ANSI mode) */
@@ -2557,7 +2557,7 @@ public:
     assert(current_stmt_binlog_format == BINLOG_FORMAT_STMT ||
            current_stmt_binlog_format == BINLOG_FORMAT_ROW);
 #ifdef WITH_WSREP
-    return (WSREP_BINLOG_FORMAT((ulong)current_stmt_binlog_format) ==
+    return (WSREP_BINLOG_FORMAT(this, (ulong)current_stmt_binlog_format) ==
             BINLOG_FORMAT_ROW);
 #else
     return current_stmt_binlog_format == BINLOG_FORMAT_ROW;
@@ -3661,6 +3661,11 @@ public:
     return m_wsrep_next_trx_id;
   }
 
+  ulong wsrep_forced_binlog_format_override() const
+  {
+    return m_wsrep_forced_binlog_format_override;
+  }
+
 private:
   /* Imagine this be a query-id that is assigned to all statements
   including non-replicating statement like SELECT/SET.
@@ -3668,6 +3673,24 @@ private:
   For data-changing DML statement wsrep_ws_handle trx_id is set
   to real-transaction-id. */
   wsrep_trx_id_t m_wsrep_next_trx_id; /* cast from query_id_t */
+
+  /**
+    Used for overriding wsrep_forced_binlog_format variable per session.
+
+    Main idea: stay unspecified most part of time (BINLOG_FORMAT_UNSPEC),
+    but after clear_current_stmt_binlog_format_row become BINLOG_FORMAT_STMT
+    to override global variable. And after set_current_stmt_binlog_format_row
+    become unspeficied again.
+
+    Because wsrep_forced_binlog_format is also used for overriding original
+    binlog format, this is pretty tricky logic. Probably it should be refactored,
+    but because wsrep_forced_binlog_format is deprecated, that's considered OK.
+
+    @see WSREP_BINLOG_FORMAT
+    @see set_current_stmt_binlog_format_row
+    @see clear_current_stmt_binlog_format_row
+  */
+  ulong m_wsrep_forced_binlog_format_override;
 public:
 #endif /* WITH_WSREP */
   /**
@@ -4405,7 +4428,7 @@ public:
       lex->binlog_row_based_if_mixed upwards to the caller.
     */
 #ifdef WITH_WSREP
-    if ((WSREP_BINLOG_FORMAT(variables.binlog_format) == BINLOG_FORMAT_MIXED) &&
+    if ((WSREP_BINLOG_FORMAT(this, variables.binlog_format) == BINLOG_FORMAT_MIXED) &&
 #else
     if ((variables.binlog_format == BINLOG_FORMAT_MIXED) &&
 #endif /* WITH_WSREP */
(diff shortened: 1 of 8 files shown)


sql/sql_class.h
```diff
@@ -4421,12 +4421,28 @@ class THD :public MDL_context_owner,
...
+    }
+#endif
     current_stmt_binlog_format= BINLOG_FORMAT_ROW;
     DBUG_VOID_RETURN;
   }
   inline void clear_current_stmt_binlog_format_row()
   {
     DBUG_ENTER("clear_current_stmt_binlog_format_row");
+#ifdef WITH_WSREP
+    if (wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC)
```
kamil-holubicki: This will force setting `wsrep_forced_binlog_format` to `BINLOG_FORMAT_STMT`. Is it OK? Shouldn't we restore the original value?

kamil-holubicki (changes requested):
I think it will be better to go with the "alternative option".
`wsrep_forced_binlog_format` is the global variable, so it affects all sessions. What if we change it in a session that causes problems, but another parallel session would use it?
I would consider introducing a new session-bound member of THD (like `current_stmt_binlog_format`, e.g. THD::wsrep_forced_binlog_format_override). Then set it accordingly in `set_current_stmt_binlog_format_row()/clear_current_stmt_binlog_format_row()` to indicate that we internally forced the format regardless of `wsrep_forced_binlog_format`.
Of course, we will need to change the implementation of WSREP_BINLOG_FORMAT macro to be aware of `THD::wsrep_forced_binlog_format_override` (log.h and sql_class.h - BTW, would it be possible to get rid of one of them and have implementation only in one place?)

sql/sql_class.h
voodam: `clear_...` may be is not so good naming. Purpose of that method is set STMT format temporary. Original format is restored in `set_current_stmt_binlog_format_row`  method.

But it's right that there will be bug on restoring if `wsrep_forced_binlog_format != BINLOG_FORMAT_ROW`, I'll fix it soon.

sql/sql_class.h
```diff
@@ -2638,6 +2640,24 @@ class THD :public MDL_context_owner,
   */
   enum_binlog_format current_stmt_binlog_format;
 
+  /**
```
kamil-holubicki: This should go into #ifdef WITH_WSREP. Could we move it down to the existing section of WSREP-related members? (That will cause initialization order change in constructor)

Please consider changing the type of it to `enum_binlog_format`. 
Maybe changing the type of `wsrep_forced_binlog_format` is not a bad idea as well?

sql/sql_class.h
```diff
@@ -2701,6 +2721,10 @@ class THD :public MDL_context_owner,
       skip_readonly_check= false;
   }
 
+  ulong wsrep_forced_binlog_format_override() const {
```
kamil-holubicki: #ifdef WITH_WSREP

sql/sql_class.h
```diff
@@ -270,11 +270,13 @@ enum enum_binlog_format {
...
-   ((wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC) ?     \
-   wsrep_forced_binlog_format : my_format)
+#define WSREP_BINLOG_FORMAT(thd, my_format)                                 \
+   ((thd->wsrep_forced_binlog_format_override() != BINLOG_FORMAT_UNSPEC) ?  \
+     thd->wsrep_forced_binlog_format_override() :                           \
+     ((wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC) ?                \
+       wsrep_forced_binlog_format : my_format))
 #else
-#define WSREP_BINLOG_FORMAT(my_format) my_format
+#define WSREP_BINLOG_FORMAT(thd, my_format) my_format
```
kamil-holubicki: There is no WSREP_BINLOG_FORMAT if there is no WITH_WSREP. I know it was there before, but it is not necessary. Please remove 'else' part.

kamil-holubicki (changes requested):
Looks fine, just please consider refactoring as requested.

kamil-holubicki:
Please also rebase considering GCA process.

sql/sql_class.h
voodam: Done

voodam: Done

voodam:
> Please also rebase considering GCA process.

Probably changes are not required in 8.0, there are no `wsrep_forced_binlog_format` variable in that branch.

sql/sql_class.h
voodam: Done

sql/sql_class.h
voodam: 1. Done #ifdef.
2. Changing type: all variables passed inside `WSREP_BINLOG_FORMAT` are ulong, so compiler throw warnings about mixed types. Is it worth to change the type in this case?

sql/sql_class.h
```diff
@@ -4421,12 +4444,20 @@ class THD :public MDL_context_owner,
...
+    m_wsrep_forced_binlog_format_override= BINLOG_FORMAT_UNSPEC;
+#endif
     current_stmt_binlog_format= BINLOG_FORMAT_ROW;
     DBUG_VOID_RETURN;
   }
   inline void clear_current_stmt_binlog_format_row()
   {
     DBUG_ENTER("clear_current_stmt_binlog_format_row");
+#ifdef WITH_WSREP
+    if (wsrep_forced_binlog_format != BINLOG_FORMAT_UNSPEC &&
```
kamil-holubicki, line 4453: One more question. Is this `if` needed at all? what if we just set
`m_wsrep_forced_binlog_format_override= BINLOG_FORMAT_STMT;` ?
It will work the same, right?

voodam: The problem with `m_wsrep_forced_binlog_format_override` is that `wsrep_forced_binlog_format` can change and overrideing variable will become outdated. So I tried to set it to `BINLOG_FORMAT_UNSPEC` as soon as posible and don't change it without need.

Another problem, for example: `reset_current_stmt_binlog_format_row` is used in `THD::init`, so if `wsrep_forced_binlog_format` is not set and this isn't checked in if, override variable can be set in wrong value.

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/4962/

voodam:
@kamil-holubicki, There is some problem in pipeline:
`[tempesta-tech/percona-xtradb-cluster/am-PXC-512-5.7/VERSION](https://raw.githubusercontent.com/tempesta-tech/percona-xtradb-cluster/am-PXC-512-5.7/VERSION)`
There is no such repository `tempesta-tech/percona-xtradb-cluster`, only `tempesta-tech/percona-server` exists.

kamil-holubicki:
> > Please also rebase considering GCA process.
> 
> Probably changes are not required in 8.0, there are no `wsrep_forced_binlog_format` variable in that branch.

Yes, but we still need to merge it to 8.0 (it will be empty merge)

sql/sql_class.h
kamil-holubicki: > 1. Done #ifdef.
> 2. Changing type: all variables passed inside `WSREP_BINLOG_FORMAT` are ulong, so compiler throw warnings about mixed types. Is it worth to change the type in this case?

OK, let's leave it as it is

kamil-holubicki (approved):
LGTM. Please create 8.0 PR

voodam:
> LGTM. Please create 8.0 PR

https://github.com/percona/percona-xtradb-cluster/pull/1704

kamil-holubicki (approved):
lgtm

PR #1705, opened 2022-11-23
PXC 8.0.30 refresh - Q3 2022

kamil-holubicki:
All tests pass on ubuntu:focal
Will continue (check and fix if necessary) if Jenkins build images are ready for other platforms.

For now, it is more than enough for merging into the release branch to unblock the packaging team.


sql/rpl_replica.cc
```diff
@@ -7065,22 +7065,10 @@ wsrep_restart_point :
...
-  thd->thread_stack = (char *)&thd;  // remember where our stack is
-  mysql_mutex_lock(&rli->info_thd_lock);
-  rli->info_thd = thd;
-||||||| merged common ancestors
-    thd = new THD;  // note that contructor of THD uses DBUG_ !
-    thd->thread_stack = (char *)&thd;  // remember where our stack is
-    mysql_mutex_lock(&rli->info_thd_lock);
-    rli->info_thd = thd;
-=======
     thd = new THD;  // note that constructor of THD uses DBUG_ !
```
dutow: The head version seems to be the correct one for indentation?

sql/rpl_replica.cc
```diff
@@ -7519,41 +7501,6 @@ wsrep_restart_point :
...
-    THD_CHECK_SENTRY(thd);
-    mysql_mutex_lock(&rli->info_thd_lock);
-    rli->info_thd = nullptr;
-    if (commit_order_mngr) {
-      rli->set_commit_order_manager(nullptr);
-      delete commit_order_mngr;
-    }
-=======
     /*
       TODO: see if we can do this conditionally in next_event() instead
```
dutow: Also indentation

storage/innobase/trx/trx0trx.cc
```diff
@@ -3798,6 +3798,10 @@ int wsrep_signal_replicator(trx_t *victim_trx, trx_t *bf_trx) {
     victim_trx->lock.was_chosen_as_wsrep_victim = true;
   }
   wsrep_thd_UNLOCK(thd);
+// KH: TODO: we should have it wrapped with
```
dutow: Accidentally left/forgotten todo?

sql/rpl_replica.cc
kamil-holubicki: This part is interesting because upstream has indentation here, but PXC has a label before this block, which causes clang to think the indentation should be smaller.
Fixed.

sql/rpl_replica.cc
kamil-holubicki: fixed

storage/innobase/trx/trx0trx.cc
kamil-holubicki: Fixed. I've found this part by the way of merging and left a comment for further investigation. Now I see that the caller acquires the mutex, co converted it into a proper comment and assert

dutow (approved):

PR #1712, opened 2022-12-12
PXC-4077: Skipping GTID events introduce GTID

kamil-holubicki:
inconsistencies between PXC nodes

https://jira.percona.com/browse/PXC-4077

Problem:
When PXC cluster acts as the async replica to some async source, it may be the case when we want to skip some events on async replica by commiting empty transactions with GTID. Doing so on one of PXC nodes is not replicated among PXC cluster which leads to GTID inconsistencies between PXC cluster nodes.

Cause:
GTID event is not generated and replicated if the transaction is empty. It is only replicated if it is received from async source and processed by replica thread.

Solution:
Allow processing replication of GTID event for all threads besides wsrep applier threads.


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5047/

dutow (approved):

PR #1715, opened 2022-12-14
PXC-4087: BF abort for MDL conflict causes a crash

kamil-holubicki:
when the victim thread is executing a rollback and thread pool is enabled

https://jira.percona.com/browse/PXC-4087

Problem:
The connection on node-1 having big uncommited transaction is killed. This causes transaction on node-1 being rolled back. If during the rollback node-2 issures another transaction which confilcts with the one from node-1, the server asserts (debug) or crashes (release). It only happens if thread_handling is 'pool-of-threads'.

Cause:
The proper flow is:
1. node-1 transaction (T1) is being rolled back as the consequence of connection KILL. THD::wsrep_query_state is QUERY_EXEC, which indicates ongoing query.
2. Transaction (T2) from node-2 is replicated. It sees that there is conflicting T1, so T1 has to be BF-aborted.
3. T1 abort is triggered, however applier thread (T2) waits for T1's rollback
4. Applier thread continues with T2

The problem was that in case of 'pool-of-threads' usage, THD::wsrep_query_state was set to QUERY_IDLE instead of QUERY_EXEC for the time of rollback. This caused logic of applier to not wait for T1, but to delegate the rollback to the wsrep rollbacker thread.

ha_rollback_low() is executed as a consequence of connection KILL resets/clears ha_info of rolled back SE.
In the same time rollbacker thread tries to register wsrep handlerton in wsrep_register_hton(). It iterates over ha_infos of transaction and tries to dereference already cleared m_ht.

Solution:
Changed how THD::wsrep_query_state is set to QUERY_EXEC/QUERY_IDLE to include the case when the connection is killed and does rollback as the part of error handling.

### DIFF
--- a/sql/threadpool_unix.cc
+++ b/sql/threadpool_unix.cc
@@ -1580,6 +1580,31 @@ static void handle_event(connection_t *connection)
   }
   else
   {
+#ifdef WITH_WSREP
+    /* This is the new request to be handled.
+     It can be connection KILL request. In such a case if there is
+     an open transaction in this connection, we will start rollback
+     process in connection_abort() above.
+     We need to keep query_state marked as QUERY_EXEC for all the
+     rollback time to be able to properly handle BF abort which may
+     happen in the meantime from the applier thread.
+     In such a case we want applier thread to wait until KILL-triggered
+     abort is over than see our thd as being idle and delegate the rollback
+     to the wsrep rollbacker thread. In such a case we would have two
+     parallel rollback requests for the same transaction which is asking
+     for troubles.
+     In other words, we want our thd to be QUERY_EXEC during the rollback
+     which will cause applier thread to execute the flow as the transaction
+     was killed by applier (force its rollback, wait till it is finished
+     and continue) */
+    if (WSREP_ON)
+    {
+      mysql_mutex_lock(&connection->thd->LOCK_wsrep_thd);
+      wsrep_thd_set_query_state(connection->thd, QUERY_EXEC);
+      mysql_mutex_unlock(&connection->thd->LOCK_wsrep_thd);
+    }
+#endif /* WITH_WSREP */
+
     err= threadpool_process_request(connection->thd);
   }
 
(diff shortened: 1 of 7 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5092/

sql/threadpool_unix.cc
```diff
@@ -1580,6 +1580,32 @@ static void handle_event(connection_t *connection)
...
+     parallel rollback requests for the same transaction which is asking
+     for troubles.
+     In other words, we want our thd to be QUERY_EXEC during the rollback
+     which will cause applier thread to execute the flow as the transaction
+     was killed by applier (force its rollback, wait till it is finished
+     and continue) */
+    if (WSREP_ON)
+    {
+      mysql_mutex_lock(&connection->thd->LOCK_wsrep_thd);
+      fprintf(stderr, "KH: threadpool_process_request setting QUERY_IDLE\n");
```
dutow: Leftover debug print?

sql/threadpool_unix.cc
```diff
@@ -1593,6 +1619,17 @@ static void handle_event(connection_t *connection)
   if (err)
     connection_abort(connection);
 
+#ifdef WITH_WSREP
+  /* Set the thd->wsrep_query_state back to the QUERY_IDLE state. */
+  if (WSREP_ON)
+  {
+    mysql_mutex_lock(&connection->thd->LOCK_wsrep_thd);
+    fprintf(stderr, "KH: threadpool_process_request setting QUERY_IDLE\n");
```
dutow: Leftover debug print?

sql/threadpool_unix.cc
kamil-holubicki: yup :)

sql/threadpool_unix.cc
kamil-holubicki: fixed

dutow (approved):

PR #1720, opened 2022-12-19
PXC-3990: Build galera with CMake

dutow:


mysql-test/suite/galera/t/galera_garbd_backup.test
```diff
@@ -38,23 +38,22 @@ CALL mtr.add_suppression("Allocated tablespace ID");
 
 --let $ssl_opts=cafile=$MYSQLTEST_VARDIR/std_data/cacert.pem,cert=$MYSQLTEST_VARDIR/std_data/server-cert.pem,key=$MYSQLTEST_VARDIR/std_data/server-key.pem
 
---exec `dirname $WSREP_PROVIDER`/../bin/garbd --sst="xtrabackup-v2:127.0.0.1:9999/xtrabackup_sst//1" --recv-script="$MYSQLTEST_VARDIR/std_data/sst_garbd_test.sh 9999 $ssl_opts" --address "gcomm://127.0.0.1:$NODE_GALERAPORT_1" --group my_wsrep_cluster --options 'evs.inactive_timeout=PT6S;base_port=$NODE_GALERAPORT_3' > $MYSQLTEST_VARDIR/tmp/garbd.log 2>&1 
-SELECT VARIABLE_VALUE FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size';
```
kamil-holubicki, line 42: We don't need this SELECT anymore?

cmake/dev-helper.cmake
```diff
@@ -3,46 +3,14 @@ OPTION(WITH_GALERA_DEV "Automatically build & copy & install debug galera librar
...
-
-  INSTALL(PROGRAMS
-    "${CMAKE_CURRENT_SOURCE_DIR}/percona-xtradb-cluster-galera/garb/garbd"
-    DESTINATION bin)
-  INSTALL(PROGRAMS
-    "${CMAKE_CURRENT_SOURCE_DIR}/percona-xtradb-cluster-galera/libgalera_smm.so"
-    DESTINATION lib)
 endif(WITH_GALERA_DEV)
-
+  MESSAGE(WARNING WITH_GALERA_DEV is no longer required, galera is built as part of the normal PXC build)
```
kamil-holubicki, line 6: Shouldn't it be withing if-endif (WITH_GALERA_DEV) block?. Now if-endif is empty

mysql-test/suite/galera/t/galera_garbd_backup.test
dutow: Accidental change, I readded it.

cmake/dev-helper.cmake
dutow: You are right, I didn't notice this when I ran cmake. fixed.

kamil-holubicki (approved):

PR #1725, opened 2023-01-09
PXC-702: GCache encryption PXC-3063: Record Set cache data at rest encryption

kamil-holubicki:
https://jira.percona.com/browse/PXC-702
https://jira.percona.com/browse/PXC-3063

Server part implementation.
1. Needed changes around the keyring component and Galera initialization
2. Master key rotation
3. Master key replication

Integrated existing MTR test suites in the way that they can be executed
with encryption enabled or disabled.

Introduced new Galera encryption MTR tests.


mysql-test/suite/galera/t/galera_gcache_recover_manytrx.test
```diff
@@ -128,6 +128,9 @@ DELIMITER ;|
 --connection node_2
 SET SESSION wsrep_sync_wait = 0;
 
+# we need to give it a chance to put something in gcache and in case of encryption for some enc cache pages to be flushed
+--sleep 18
```
dutow, line 132: Will this also work on jenkins, isn't there a better check?

mysql-test/suite/galera_encryption/t/gcache_mk_rotation.test
```diff
@@ -0,0 +1,90 @@
...
+--source include/assert_grep.inc
+
+# Check that GCache ring buffer file is using MK id 1
+--let $assert_text = Assert that GCache uses MK id 1
+--let $assert_file = $NODE2_DATA_DIR/galera.cache
+--let $assert_select = enc_mk_id: 1
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+ALTER INSTANCE ROTATE GCACHE MASTER KEY;
```
dutow, line 69: What if alter instance ... is executed without wsrep_on, should that work, and rotate the key only on one server?

Is it possible that one node uses gcache encryption, the other doesn't?

sql/sql_plugin.cc
```diff
@@ -1581,25 +1584,45 @@ bool update_persisted_plugin_sysvars(const char *name) {
...
   MEM_ROOT tmp_root(key_memory_plugin_init_tmp, 4096);
 
   mysql_rwlock_wrlock(&LOCK_system_variables_hash);
   mysql_mutex_lock(&LOCK_plugin);
+
   initialized = true;
 
   /* First we register the builtin mandatory and optional plugins */
-  for (struct st_mysql_plugin **builtins = mysql_mandatory_plugins;
+  for (struct st_mysql_plugin **builtins = plugins /*mysql_mandatory_plugins*/;
```
dutow: missing ifdef wsrep - and with this change, the above comment is also incorrect.

sql/sys_vars.cc
```diff
@@ -8708,6 +8708,20 @@ static Sys_var_bool Sys_pxc_encrypt_cluster_traffic(
     READ_ONLY GLOBAL_VAR(pxc_encrypt_cluster_traffic), CMD_LINE(OPT_ARG),
     DEFAULT(true), NO_MUTEX_GUARD, NOT_IN_BINLOG);
 
+static const char *wsrep_encrypt_modes[] = {"OFF", "ON", "NONE",
+                                        NullS};
+static Sys_var_enum Sys_wsrep_gcache_encrypt(
+    "wsrep_gcache_encrypt", "Encrypt GCache. This variable is solely for testing purposes. "
```
dutow, line 8714: Do we plan to keep this variable when we release it? Shouldn't it be behind some testing/internal ifdef?

sql/wsrep_master_key_manager.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  /* Keyring error */
+  if (retval == true) {
+    logger_("Failed to initialize keyring reader service");
+    return std::string();
+  }
+  /* Key absent */
+  if (reader_object == nullptr) return std::string();
+
+  auto cleanup_guard = create_scope_guard([&] {
```
dutow, line 118: Shouldn't this guard be somewhere earlier, like around the definition of reader_object?

dutow (commented):
Is this pull request a duplicate of #1660?

kamil-holubicki:
>Is this pull request a duplicate of https://github.com/percona/percona-xtradb-cluster/pull/1660?
Not exactly. This is reworked version (based on that one), but a lot of squashing, reordering... I decided to go with new one as it was easier/quicker. 
Closing the other one.

sql/sys_vars.cc
kamil-holubicki: I think it is OK to have it here. I would even not document it. It will not do any harm and in worst-case scenario will act as the override of wsrep_provider_params' gcache.encrypt parameter.
Having it in release allows us to run standard MTR suites

sql/wsrep_master_key_manager.cc
kamil-holubicki: This guard is to deinit the `keyring_reader_service` which was properly inited. 
We know that it was properly inited if 
1. retval != true (1st `if`, if retval == true, it is not said that keyring_reader_service is untouched.
2. `keyring_reader_service` is not null (2nd if)

So I would say we can move it before 2nd if, but for sure not before the 1st one.
I would also say, that current implementation is fine :). The implementation is the same as keyring_operations_helper.cc::read_secret()

Do you agree?

mysql-test/suite/galera/t/galera_gcache_recover_manytrx.test
kamil-holubicki: Yes, it works on Jenkins.

This magic --sleep 18 is here a few lines below in the original test.
The flow of the original test is:
1. Start transactions on node_1
2. Wait on node_2 just for CREATE TABLE to be replicated (and probably some inserts, but do not care how much of them)
3. kill node_2
4. wait for 18 secs. node_1 still processes inserts
5. start node_2 - it should do gcache recovery and join with SST.

There is a problem with points 2 and 3 when we use gcache encryption. As encryption uses in-memory cache which is being flushed only when there are no more cache pages available, if we kill immediately in point 3, there is nothing in GCache file. So the recovery will not take place.
We need to wait for some time for GCache encryption cache to be flushed internally (and this will be done if we receive more transactions from node_1). Then, after node_2 restart it will see that there is "something" in gcache file and the recovery will take place.
Why 18 secs? well, because the original test used 18 as well :).  There is no way to say that the encryption cache has been flushed (the only way is to examine gcache file itself, but we still would need to wait for some arbitrary time, check and then wait again if check fails...) but I didn't want to change this test too much.

mysql-test/suite/galera_encryption/t/gcache_mk_rotation.test
kamil-holubicki: This is not wsrep_on -aware. It works the same way as ALTER INSTANCE ROTATE INNODB MASTER KEY works.

If GCache encryption is not enabled, it does not rotate the key (because enabling gcache encryption needs server restart and gcache reset, so no point in rotating the key if it will not be used anyway), but request for rotation is replicated, because another servers can have encryption enabled.

I will add test scenario for this.

sql/sql_plugin.cc
kamil-holubicki: fixed

sql/wsrep_master_key_manager.cc
dutow: Yes, it seems to be fine with the current components/plugins we have. 

My thought was: what if a keyring sets the reader object, but returns  an error, shouldn't we deinit it in that case? Of course we don't have anything like that, so this isn't an actual issue for PXC.

sql/wsrep_mysqld.cc
```diff
@@ -3122,3 +3180,29 @@ bool wsrep_consistency_check(THD *thd) {
...
+  return masterKeyManager->Init();
+}
+
+static void wsrep_deinit_master_key() {
+  masterKeyManager->DeInit();
+  masterKeyManager.reset();
+}
+
+std::string wsrep_get_master_key(const std::string &keyId) {
+  std::string mk = masterKeyManager->GetKey(keyId);
```
dutow: Is this variable needed, wouldn't a simple return work?

sql/wsrep_master_key_manager.cc
```diff
@@ -0,0 +1,153 @@
...
+
+  keyring_reader_service_ = nullptr;
+  keyring_writer_service_ = nullptr;
+  keyring_generator_service_ = nullptr;
+
+  mysql_plugin_registry_release(reg_svc);
+}
+
+bool MasterKeyManager::GenerateKey(const std::string &keyId) {
+  if (keyring_generator_service_->generate(keyId.c_str(), nullptr, key_type,
```
dutow: Could be return ... instead of if?

sql/wsrep_master_key_manager.cc
```diff
@@ -0,0 +1,153 @@
...
+    logger_("Failed to fetch secret length from keyring reader service");
+    return std::string();
+  }
+
+  if (secret_length == 0 || secret_type_length == 0) return std::string();
+
+  /* Allocate requried memory for key and secret_type */
+  std::vector<unsigned char> secret_v;
+  std::vector<char> secret_type_v;
+  secret_v.reserve(secret_length);
```
dutow: Why reserve and not resize? That would also take care of the memset.

sql/wsrep_mysqld.cc
```diff
@@ -1066,6 +1073,50 @@ void wsrep_deinit_server() {
...
+  if (wsrep_gcache_encrypt != WSREP_ENCRYPT_MODE_NONE) {
+    const std::string& val = wsrep_gcache_encrypt == WSREP_ENCRYPT_MODE_ON ? yes : no;
+    override_galera_option("gcache.encryption", val, options);
+  }
+  if (wsrep_disk_pages_encrypt != WSREP_ENCRYPT_MODE_NONE) {
+    const std::string& val = wsrep_disk_pages_encrypt == WSREP_ENCRYPT_MODE_ON ? yes : no;
+    override_galera_option("allocator.disk_pages_encryption", val, options);
+  }
+
+  // todo: strncpy, buffer size, etc
```
dutow: leftover todo

mysql-test/suite/galera_encryption/t/gcache_mk_rotation.test
kamil-holubicki: MTR test added

sql/wsrep_mysqld.cc

sql/wsrep_master_key_manager.cc
kamil-holubicki: done

sql/wsrep_master_key_manager.cc
kamil-holubicki: done

sql/wsrep_mysqld.cc

mysql-test/suite/galera_encryption/t/gcache_mk_rotation.test
kamil-holubicki: Please forget my above explanation. I messed things up :D. 
WSREP_ON = OFF prevents replication. rotation is done on local node and is not replicated. The same as for InnoDB MK. Added MTR for this.

dutow (approved):

PR #1728, opened 2023-01-20
PXC-4126: Create user is forbidden when stream replication is active

kamil-holubicki:
https://jira.percona.com/browse/PXC-4126

Problem:
When SR is enabled, CREATE USER statement fails

Cause:
This issue is the regression caused by fix for PXC-3872 (commit 487328e781d5acd44cd94e526969874372080b51). It moved starting TOI after open_grant_tables(). Before the fix, TOI was started at the beginning of open_grant_tables() (so effectively, 'before' open_grant_tables()). open_grant_tables() calls decide_logging_format(), where we make a check if we are in local mode and doing streaming replication (SR). wsrep_thd_is_local() is used as a part of the decision input. If we are in TOI mode, wsrep_thd_is_local() returns 'false' (before above mentioned commit), but if we didn't start TOI yet, it returns 'true', so the error is returned, as local + STMT + SR is not allowed.

Solution:
As SR and TOI are independent things, instead of disabling SR for the time TOI is executed, inform the logic in decide_logging_format() that TOI is about to be started (THD::wsrep_TOI_preparation flag introduced)

1. pxc_drop_user.test MTR test moved to pxc_drop_user.inc
2. pxc_drop_user.test which includes pxc_drop_user.inc introduced. It tests standard replication case
3. pxc_drop_user_sr.test which includes pxc_drop_user.inc introduced. It tests SR case


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1090/

VarunNagaraju (approved):
LGTM!

sql/auth/auth_internal.h
```diff
@@ -92,6 +92,41 @@ class Table_access_map {
...
+  but if we didn't start TOI yet, it returns 'true', so the error is returned,
+  as local + STMT + SR is not allowed.
+  We need to inform decide_logging_format() logic that our intention is to
+  start TOI soon.
+
+  Another working solution would be disabling SR for the time of
+  query execution, but this way does not follow the conception, that TOI can be
+  executed while SR is enabled (these are two separate things).
+*/
+class After_open_grant_tables_start_TOI_guard {
```
venkatesh-prasad-v: Can we rename this guard as `Enable_TOI_preparation_guard` just like other server guards defined in `sql/thd_raii.h` (most of the guards there start with a verb) so that it is easier to understand what the guad is doing internally without looking into its implementation?

kamil-holubicki: Done

venkatesh-prasad-v (approved):

PR #1729, opened 2023-01-31
PXC-3211 REPLACES GALERA VERSION NUMBERS WITH ENUM VALUES

VarunNagaraju:
https://jira.percona.com/browse/PXC-3211

The checks used in PXC source code to to write version specific code were using hard coded version numbers in numerous places. Such hard coded version numbers are replaced with a typesafe and meaningful enum variable to improve the quality of the source code.

### DIFF
--- a/include/service_wsrep.h
+++ b/include/service_wsrep.h
@@ -27,12 +27,16 @@ enum Wsrep_service_key_type {
   WSREP_SERVICE_KEY_EXCLUSIVE
 };
 
+namespace wsrep_version {
+enum wsrep_version : long { UNSPECIFIED = 0, V1 = 1, V2 = 2, V3 = 3, V4 = 4 };
+}
+typedef wsrep_version::wsrep_version WsrepVersion;
+
 extern ulong wsrep_debug;
 extern bool wsrep_log_conflicts;
 extern bool wsrep_certify_nonPK;
 extern bool wsrep_load_data_splitting;
 extern bool wsrep_recovery;
-extern long wsrep_protocol_version;
 
 /* Must match to definition in sql/mysqld.h */
 typedef int64 query_id_t;
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1732,7 +1732,8 @@ static bool block_write_while_in_rolling_upgrade(THD *thd) {
   bool block = false;
   LEX *lex = thd->lex;
   if (sql_command_flags[lex->sql_command] & CF_CHANGES_DATA) {
-    bool multi_version_cluster = wsrep_protocol_version < 4;
+    bool multi_version_cluster =
+        wsrep_protocol_version < WsrepVersion::V4;
     if (multi_version_cluster ||
         DBUG_EVALUATE_IF("simulate_wsrep_multiple_major_versions", true,
                          false)) {
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -101,8 +101,9 @@ ulong wsrep_max_ws_rows = 65536;             // max number of rows in ws
 int wsrep_to_isolation = 0;                  // # of active TO isolation threads
 bool wsrep_certify_nonPK = 1;  // certify, even when no primary key
 ulong wsrep_certification_rules = WSREP_CERTIFICATION_RULES_STRICT;
-long wsrep_max_protocol_version = 4;  // maximum protocol version to use
-long int wsrep_protocol_version = wsrep_max_protocol_version;
+static constexpr WsrepVersion wsrep_max_protocol_version =
+    WsrepVersion::V4;  // maximum protocol version to use
+WsrepVersion wsrep_protocol_version = wsrep_max_protocol_version;
 ulong wsrep_trx_fragment_unit = WSREP_FRAG_BYTES;
 // unit for fragment size
 ulong wsrep_SR_store_type = WSREP_SR_STORE_TABLE;
@@ -636,7 +637,7 @@ void wsrep_init_sidno(const wsrep::id &uuid) {
     uuid.
   */
   rpl_sid sid;
-  if (wsrep_protocol_version >= 4) {
+  if (wsrep_protocol_version >= WsrepVersion::V4) {
     memcpy((void *)&sid, (const uchar *)uuid.data(), 16);
   } else {
     wsrep_uuid_t ltid_uuid;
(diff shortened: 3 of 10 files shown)


include/service_wsrep.h
```diff
@@ -27,12 +27,19 @@ enum Wsrep_service_key_type {
   WSREP_SERVICE_KEY_EXCLUSIVE
 };
 
+enum enum_wsrep_version {
```
dutow: Why do we need "enum_" in the name?

Also, if we are using the typename as a prefix, it could be an `enum class`.

Enums / enums classes also allow to specify the underlying type by writing  to `enum class wsrep_version : long`, to keep it the same as before.

If using the type as a prefix, and an enum class, I would also leave the `WSREP` 
 and `VERSION` out from the version names, as it's redudant. `wsrep_version::v1`

include/service_wsrep.h
```diff
@@ -27,12 +27,19 @@ enum Wsrep_service_key_type {
...
+  WSREP_PLACEHOLDER = 0,
```
dutow: Maybe something like `UNSPECIFIED` or `UNKNOWN`?

VarunNagaraju: Done

include/service_wsrep.h
VarunNagaraju: Done

sql/wsrep_mysqld.cc
```diff
@@ -101,8 +101,9 @@ ulong wsrep_max_ws_rows = 65536;             // max number of rows in ws
 int wsrep_to_isolation = 0;                  // # of active TO isolation threads
 bool wsrep_certify_nonPK = 1;  // certify, even when no primary key
 ulong wsrep_certification_rules = WSREP_CERTIFICATION_RULES_STRICT;
-long wsrep_max_protocol_version = 4;  // maximum protocol version to use
-long int wsrep_protocol_version = wsrep_max_protocol_version;
+wsrep_version wsrep_max_protocol_version =
```
dutow: This could be `constexpr`

kamil-holubicki: We can add static as well as it is used only in this file.

sql/wsrep_server_state.cc
```diff
@@ -50,14 +50,15 @@ void Wsrep_server_state::init_once(const std::string &name,
                                    const std::string &address,
                                    const std::string &working_dir,
                                    const wsrep::gtid &initial_position,
-                                   int max_protocol_version) {
+                                   const wsrep_version &max_protocol_version) {
```
kamil-holubicki: I think there is no advantage in passing it as the referece

sql/wsrep_server_state.h
```diff
@@ -32,7 +32,7 @@ class Wsrep_server_state : public wsrep::server_state {
                         const std::string &address,
                         const std::string &working_dir,
                         const wsrep::gtid &initial_position,
-                        int max_protocol_version);
+                        const wsrep_version &max_protocol_version);
```
kamil-holubicki: Likewise

sql/wsrep_mysqld.cc
```diff
@@ -1558,7 +1559,9 @@ static bool wsrep_prepare_key_for_isolation(const char *db, const char *table,
     }
     default:
       assert(0);
-      WSREP_ERROR("Unsupported protocol version: %ld", wsrep_protocol_version);
+      WSREP_ERROR("Unsupported protocol version: %ld",
+                  static_cast<std::underlying_type<wsrep_version>::type>(
```
kamil-holubicki: This static_cast is really needed?

sql/wsrep_server_service.cc
```diff
@@ -212,7 +214,9 @@ void Wsrep_server_service::log_view(
     if (wsrep_pxc_maint_mode_forced &&
         pxc_maint_mode != PXC_MAINT_MODE_SHUTDOWN) {
       std::ostringstream os;
-      os << "Detected Protocol version: " << wsrep_protocol_version
+      os << "Detected Protocol version: "
+         << static_cast<std::underlying_type<wsrep_version>::type>(
```
kamil-holubicki: This static_cast is really needed?

sql/wsrep_server_state.cc
VarunNagaraju: Agree. Will update it.

sql/wsrep_mysqld.cc
VarunNagaraju: Since enum classes are NOT implicitly convertible to int, this seemed like a good solution to fetch the underlying type.
Also, if the type changes in the future, this wouldn't require a change.

kamil-holubicki: You are right, enum class adds explicit scoping but removes implicit conversion, which results in the need for such an unpleasant casting...

@dutow @VarunNagaraju What about the following approach?
```
#include <iostream>

namespace MyEnum {
enum MyEnum : long {
  one = 1,
  two = 2
};
}

int main() {
  std::cout << MyEnum::one << "\n";
  return 0;
}
```

VarunNagaraju: It removes the unpleasant casting and the members are scoped. Seems good to me.

sql/sql_parse.cc
```diff
@@ -1732,7 +1732,8 @@ static bool block_write_while_in_rolling_upgrade(THD *thd) {
   bool block = false;
   LEX *lex = thd->lex;
   if (sql_command_flags[lex->sql_command] & CF_CHANGES_DATA) {
-    bool multi_version_cluster = wsrep_protocol_version < 4;
+    bool multi_version_cluster =
+        wsrep_protocol_version < WsrepVersion::wsrep_version::V4;
```
kamil-holubicki: Multiple occurences:
I think WsrepVersion::V4 is enough and is what we need :)

include/service_wsrep.h
```diff
@@ -27,12 +27,15 @@ enum Wsrep_service_key_type {
   WSREP_SERVICE_KEY_EXCLUSIVE
 };
 
+namespace WsrepVersion {
+enum wsrep_version : long { UNSPECIFIED = 0, V1 = 1, V2 = 2, V3 = 3, V4 = 4 };
```
kamil-holubicki, line 31: Does it conform clang formating rules?

kamil-holubicki (changes requested):

kamil-holubicki:
I think we can achieve the state of the art with using typedef :)
```
#include <iostream>

namespace my_enum {
enum my_enum : long {
  one = 1,
  two = 2
};
}

typedef my_enum::my_enum MyEnum;

void foo(MyEnum e) {
  std::cout << e << "\n";
}

int main() {
  MyEnum value = MyEnum::two;
  std::cout << MyEnum::one << "\n";
  foo(value);  
  return 0;
}
```

dutow:
There's a typo in the commit message: "GLAERA".

kamil-holubicki (approved):
LGTM

PR #1730, opened 2023-01-31
PXC-4116: PXC stalls with parallel replication workers

dutow:
Issue: MySQL parallel replication workers and galera both order commits in a well defined order, and use locks/signals to enforce it, but they end up generating different sequences.

Galera assigns monotonic sequence numbers in wsrep_before_prepare, at the prepare stage of the commit, and waits for earlier commits in wsrep_before_commit, before starting the commit.

MySQL uses the commit order in the relay log, and waits for earlier commits later before flushing. Up to that point, workers process commits in parallel.

If the parallel workers reach wsrep_before_prepare in a different order than the order in the relay log, wsrep assigns sequence numbers in a different order than MySQL, and the two logics will end up in a deadlock:

Commit A has the higher galera seqno and waits for commit B in the galera code, but Commit B has the higher MySQL seqno and waits for commit A in teh server code.

Additionally, one seqno for Galera means one commit, while for MySQL it can mean multiple commits.

Fix: we disable parallel execution of commits before calling wsrep_before_prepare. Only the worker that is next in line to flush the commit can start preparing the commit.

This strict exclusion is required. A more relaxed implementation, wihch only requires workers to enter wsrep_before_prepare in the same order as their seqnos doesn't work because of the issue described at the end:

As we can process multiple commits with the same server seqno, we can't let other threads proceed until one worker flushed all of its commits. If we allow that, we could assign galera seqno N+1 to the next in line worker, and after that galera seqno N+2 to the second commit of the current worker, resulting in the same deadlock.

### DIFF
--- a/mysql-test/suite/galera/t/galera_parallel_slave.test
+++ b/mysql-test/suite/galera/t/galera_parallel_slave.test
@@ -1,0 +1,85 @@
+--source include/have_debug.inc
+--source include/have_log_bin.inc
+--source include/force_restart.inc
+
+# As node #1 is not a Galera node, we connect to node #2 in order to run include/galera_cluster.inc
+--connect node_2a, 127.0.0.1, root, , test, $NODE_MYPORT_2
+--source include/galera_cluster_master_slave.inc
+
+--connection node_2
+--source include/only_mts_replica_parallel_workers.inc
+
+--disable_query_log
+--eval CHANGE REPLICATION SOURCE TO SOURCE_HOST='127.0.0.1', SOURCE_PORT=$NODE_MYPORT_1;
+--enable_query_log
+START REPLICA USER='root';
+
+--connection node_1
+# Initialize
+CREATE TABLE t1(c1 INT PRIMARY KEY, c2 INT, INDEX(c2)) ENGINE = InnoDB;
+INSERT INTO t1 VALUES(1, NULL),(2, 2), (3, NULL), (4, 4), (5, NULL), (6, 6);
+
+CREATE TABLE t2(c1 INT PRIMARY KEY, c2 INT, INDEX(c2)) ENGINE = InnoDB;
+INSERT INTO t2 VALUES(1, NULL),(2, 2), (3, NULL), (4, 4), (5, NULL), (6, 6);
+
+--connection node_2
+--sleep 1
+--let $wait_condition = SELECT COUNT(*) = 1 FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = 't2';
+--source include/wait_condition.inc
+
+STOP REPLICA;
+
+SET GLOBAL replica_transaction_retries = 3;
+SET GLOBAL replica_parallel_workers = 5;
+
+--connection node_1
+
+INSERT INTO t1 VALUES(10, NULL);
+
+--let $debug_point= d,set_commit_parent_100
+--source include/add_debug_point.inc
+INSERT INTO t2 VALUES(11, NULL);
+INSERT INTO t1 VALUES(11, NULL);
+INSERT INTO t2 VALUES(12, NULL);
+INSERT INTO t1 VALUES(12, NULL);
+INSERT INTO t2 VALUES(13, NULL);
+INSERT INTO t1 VALUES(13, NULL);
+
+--connection node_2
+# It blocks above INSERT statement
+FLUSH TABLES t2 WITH READ LOCK;
+
+--connection node_2a
+START REPLICA USER='root';
+
+# This sleep is not ideal, but information_schema.processlist selects fail at this point with lock wait timeout
+--sleep 10
+#--let $wait_condition= SELECT count(*) > 0 FROM information_schema.processlist WHERE INFO LIKE "%dependent transaction to commit%"
+#--source include/wait_condition.inc
+
+--connection node_2
+UNLOCK TABLES;
+
+--let $wait_condition = SELECT COUNT(*) = 10 FROM t1;
+--source include/wait_condition.inc
+
+SELECT * FROM t1;
+SELECT * FROM t2;
+
+--echo #
+--echo # Deinitialize
+--echo #
+--connection node_1
+--let $debug_point= d,set_commit_parent_100
+--source include/remove_debug_point.inc
+DROP TABLE t1;
+DROP TABLE t2;
+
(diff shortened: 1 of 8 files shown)


sql/rpl_replica_commit_order_manager.cc
```diff
@@ -197,6 +205,19 @@ bool Commit_order_manager::wait(Slave_worker *worker) {
...
+#ifdef WITH_WSREP
+void Commit_order_manager::wait_for_galera_turn(THD* thd) {
+  Slave_worker *worker = dynamic_cast<Slave_worker *>(thd->rli_slave);
+  if(!worker) return;
+
+  mysql_mutex_lock(&m_galera_seqno_mtx);
+  while (m_workers.front() != worker->id) {
+    mysql_cond_wait(&m_galera_seqno_cond, &m_galera_seqno_mtx);
+  }
+  mysql_mutex_unlock(&m_galera_seqno_mtx);
```
venkatesh-prasad-v: Consider a scenario like this

The source(master) server executes the below 2 queries in parallel and INSERT is binlogged before DELETE. On replica, they get assigned to w1 and w2 respectively to execute parallely


w1:    INSERT INTO t1 (11, NULL);
w2:    DELETE FROM t1 WHERE c2 <= 3;

    On replica server, they are applied parallel by w1 and w2. And the deadlock is:
      w1                                        w2
    -------------------------                 --------------------------
     Applying INSERT                          Applying DELETE
                                              Gets the row locks first.
     Waits for w2 to release
     the row lock.
                                              Waits for w1 to prepare in wait_for_galera_turn().

In such a case, w1 will wait for w2 to release row locks and w2 will wait for w1 to prepare and we will have deadlocks.

venkatesh-prasad-v: I debugged the code further, it looks `Commit_order_manager::wait_on_graph()` uses MDL's `timed_wait()` to wait for its turn. This is done because it can provide protection against ACL and FTWRL queries. The `mysql_cond_wait()` here does not provide any protection against ACL and FTWRL and it may result in deadlock.

Can you please evaluate the effects of calling `Commit_order_manager::wait_on_graph()`  here as well (it will be called again in `ha_commit_low()`) ?

dutow: Fixed.

sql/rpl_replica_commit_order_manager.h
```diff
@@ -229,6 +228,8 @@ class Commit_order_manager {
     @return false if the worker is ready to commit, true if not.
    */
   bool wait_on_graph(Slave_worker *worker);
+
+ private:
```
kamil-holubicki, line 237: WITH_WSREP

sql/rpl_replica_commit_order_manager.h
```diff
@@ -218,7 +218,6 @@ class Commit_order_manager {
   */
   void register_trx(Slave_worker *worker);
 
- private:
```
kamil-holubicki, line 221: WITH_WSREP

mysql-test/suite/galera/t/galera_parallel_slave.test
```diff
@@ -0,0 +1,39 @@
...
+--copy_file $MYSQLTEST_VARDIR/std_data/pxc4116.zip $MYSQL_TMP_DIR/pxc4116.zip
+--file_exists $MYSQL_TMP_DIR/pxc4116.zip
+--exec unzip -qo $MYSQL_TMP_DIR/pxc4116.zip -d $MYSQL_TMP_DIR/pxc4116z
+
+--echo # Set different datadir path
+--let $MYSQLD_DATADIR= $MYSQL_TMP_DIR/pxc4116z/data2
+--let $MYSQLD_LOG= $MYSQLTEST_VARDIR/log/pxc4116.log
+
+
+use mysql;
```
kamil-holubicki: Why do we need this?

mysql-test/suite/galera/t/galera_parallel_slave.test
```diff
@@ -0,0 +1,39 @@
+
+
```
kamil-holubicki, line 8: 2 empty lines at the beginning

mysql-test/suite/galera/t/galera_parallel_slave.test
```diff
@@ -0,0 +1,39 @@
...
+--source include/wait_condition.inc
+
+--echo # Restart the server with default options
+--let $restart_parameters= restart;
+--let $wait_counter= 10000
+--source include/restart_mysqld.inc
+
+--remove_file $MYSQL_TMP_DIR/pxc4116.zip
+--force-rmdir $MYSQL_TMP_DIR/pxc4116z
+
```
kamil-holubicki, line 83: empty line

kamil-holubicki (commented):
Looks fine. Just a few cosmetic things.

mysql-test/suite/galera/t/galera_parallel_slave.test
dutow: The data.zip I attached doesn't have a "test" database in it. I thought it was easier to use a database that exists in both than to recreate the zip again.

mysql-test/suite/galera/t/galera_parallel_slave.test
dutow: removed.

mysql-test/suite/galera/t/galera_parallel_slave.test
dutow: removed.

kamil-holubicki (approved):

sql/handler.cc
```diff
@@ -2396,6 +2396,14 @@ int ha_prepare_low(THD *thd, bool all) {
       const bool run_wsrep_hooks = wsrep_run_commit_hook(thd, all);
       int err;
 
+      auto rpl_com = get_commit_order_manager(thd);
+      if (rpl_com != nullptr) {
+        if (rpl_com->wait_on_graph((Slave_worker*)thd->rli_slave)) {
```
venkatesh-prasad-v, line 2417: Can we use `dynamic_cast` here?
```suggestion
        if (rpl_com->wait_on_graph(dynamic_cast<Slave_worker *>(thd->rli_slave)) {
```

PR #1731, opened 2023-02-05
PXC-4120: Error log gets spammed with wsrep_commit_empty() when wsrep-debug is enabled

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4120

Problem:
In PXC, the error log gets spammed with messages like

    [WSREP] wsrep_commit_empty()

when wsrep-debug is enabled since the message is written to the log everytime a query is executed.

Solution:
Don't write the above message to error log when wsrep-debug is enabled.


kamil-holubicki (approved):
lgtm

PR #1732, opened 2023-02-07
PXC-924: LP #1250380: Handle full tables more gracefully

voodam:
https://jira.percona.com/browse/PXC-924

Problem:
When you run out of disk space (or innodb_data_file_path max value exceed) on one of the nodes, the storage engine returns HA_ERR_RECORD_FILE_FULL and then Galera became aborted. This is considered inconvenient for the administrator who will fix this later.

Cause:
The storage engine returns HA_ERR_RECORD_FILE_FULL, applies callback and returns error code to Galera, then (after retrying) Galera became aborted.

Solution:
Put the node in desync state and abort the current transaction to make node available for causal reads.


sql/log_event.cc
```diff
@@ -11845,6 +11845,23 @@ int Rows_log_event::do_apply_event(Relay_log_info const *rli)
...
+  if (WSREP(thd) && error == HA_ERR_RECORD_FILE_FULL)
+  {
+    /*
+      Prevent a node to be aborted to make it easier for an administrator
+      to solve the problem later.
+
+      Aborting the commit is required to release Galera monitor from the
+      transaction.
+    */
+    wsrep->desync(wsrep);
```
kamil-holubicki, line 11858: I think we should desync and pause. Pause will prevent applying/committing writesets.
Desync only turns off the flow controll, applying still takes place, so I assume the result of the current implementation will be constant error logs that commit failed because of out-of-space. Is it what we need or rather we want to pause writeset applying?
Please check FTWRL implementation. It first desycs (prevent cluster to stall) and then pauses the provider (prevents writeset applying).

Moreover, I see that applying does not stop immediately when we pause the provider. It first drains commit monitors. Maybe not a big problem, but this is what I noticed.

It would be nice to test it in MTR. I mean: 
1. reach out-of-space on node_2
2. node_2 should desync and pause
3. node_1 should continue producing writesets. As the result, node_2 should never be stopped (no flow control kicks-in) and node_1 should silently wait for administrator's intervention ;)

Additionally: maybe worth adding explicit message/error log here that we stopped the node because of the problem and it needs the administrator intervention?

kamil-holubicki (changes requested):

sql/log_event.cc
voodam: > Desync only turns off the flow controll, applying still takes place, so I assume the result of the current implementation will be constant error logs that commit failed because of out-of-space.

True

> Is it what we need or rather we want to pause writeset applying?

I don't know, I assumed that's very rare situation and log spamming can be acceptable. But pause probably is more accurate.

> It would be nice to test it in MTR
> Additionally: maybe worth adding explicit message/error log

Yes, I think it's good ideas, I'll do this.

voodam:
This issue considered shouldn't be fixed. Galera doesn't allow to rollback the transaction after certification and stay non-aborted without wsrep API modification. But checking transaction before every certification seem to be a big overehead. On the other side, it seems it's not too unhandy to fix the problem after server aborting.

We need to:
* Add max size in innodb_data_file_path.
* gcache.recover should be enabled to recover state from gcache.
* Run the node with --wsrep-recover option, get the commited transation uuid+seqno and write them to grastate.dat.
* Start the node.

PR #1733, opened 2023-02-08
PXC-3936: state transfer with SSL disabled in wsrep_provider_options …

voodam:
…crashes Receiver and Donor node

https://jira.percona.com/browse/PXC-3936

Problem:
When setting socket.ssl = NO in wsrep_provider_options without explicitly specifying if SSL is used in the [SST] part of my.cnf, not only will the state transfer fail, but it will crash the donor node as well as the receiver node.

Cause:
IST_fix_addr_scheme only relies on socket.ssl_key config property and doesn't take into account socket.ssl (use_ssl in code).

Solution:
Take socket.ssl into account in IST_fix_addr_scheme.


kamil-holubicki (approved):
Please remember about updating galera repo pointer afer merging https://github.com/percona/galera/pull/249
and before merging this PR into the trunk.

voodam:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1193/
Fixed wscache_encryption_on_off, other tests should be OK.

PR #1735, opened 2023-02-15
PXC-4137: WSREP applier threads fail to modify read only schemas

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4137

Problem
-------
When a PXC cluster acts as the async replica to a MySQL source, and if the cluster has a schema has marked as read only in the PXC cluster, then updates on the schema from async source server are not handled incorrectly leading to inconsistencies between PXC cluster nodes.

Analysis
--------
Consider the scenario where a schema has been marked as read only in the PXC Cluster, while the schema on source server is open for writes.

Topology:

               async           PXC
        node1 -------> node2 <-----> node3

1. n1> CREATE DATABASE testdb;
2. n2> ALTER SCHEMA testdb READ ONLY = 1; -> schema will be read only on all nodes of the cluster
3. n1> CREATE TABLE testdb.t1 ..

In such a case, the CREATE TABLE query in step 3 will be applied on pxc replica by the replication applier thread on node2 and the same will also be replicated to node3 as TOI.

Since the replication applier threads are allowed to continue even when schema is read only, the CREATE TABLE will be successful on node2.

However, the same operation will error out on node3 as wsrep applier threads are not allowed to operate on read only schemas, thereby resulting in inconsistency voting to kick-in and thereby resulting in the cluster shutdown.

NOTE:
This problem can be easily fixed by allowing the wsrep applier threads to operate on read only schemas. However this could still result in inconsistency if the user tries to create database objects directly on PXC nodes (node2 and node3) as the the queries are replicated as TOI before the checks for read only schemas are carried out.  As a result, the query can still succeed on the replicated node while it fails on the originating node, thereby again resulting in inconsistency followed by cluster shutdown.

node2> create table testdb.t1(i int);
ERROR 3989 (HY000): Schema 'testdb' is in read only mode.

Solution
--------
1. Allow wsrep applier threads to operate on read only schemas.
2. Add checks to disallow creation of database objects inside read only schemas before the query is replicated as TOI.

Testing Done
---
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/96/console
Result: test run in progress


sql/wsrep_mysqld.cc
```diff
@@ -1099,17 +1105,20 @@ static void override_galera_option(const std::string& option, const std::string&
   options = modified_options;
 }
 
-static void setup_galera_encryption_params(char* provider_options, size_t buf_size) {
+static void setup_galera_encryption_params(char *provider_options,
```
dutow: The changes in this file are mostly unrelated formatting changes. Do we really want to add them as part of this PR?

sql/wsrep_mysqld.cc
```diff
@@ -2130,15 +2139,13 @@ static bool wsrep_can_run_in_toi(THD *thd, const char *db, const char *table,
 
   switch (lex->sql_command) {
     case SQLCOM_CREATE_TABLE:
-      assert(!table_list);
```
dutow, line 2133: `#ifndef WITH_WSREP`?

sql/wsrep_mysqld.cc
```diff
@@ -2285,6 +2292,124 @@ static void wsrep_NBO_begin_failed(THD *thd,
...
 
+/*
+  This function decide whether a DB operation can continue when it is using a
+  read only schema.
+
+  If a schema is read only, then this function shall disallow TOI operations on
+  all database objects(tables, triggers, functions, views, procedures, events)
+  inside the read only schemas except ALTER SCHEMA <schema> READ ONLY = 0.
+
+  @param thd          Thread context.
```
dutow, line 2300: The param list here is different, copy-paste error?

sql/wsrep_mysqld.cc
```diff
@@ -2285,6 +2292,124 @@ static void wsrep_NBO_begin_failed(THD *thd,
...
+  @param schema_name  Name of schema to check.
+
+  @return false if the DB operation can continue, true if not.
+*/
+static bool handle_read_only_schema(THD *thd, const char *db_,
```
kamil-holubicki: Can we rename it to something like `is_operation_allowed_on_db()`
and return true in case of success?

sql/wsrep_mysqld.cc
```diff
@@ -2285,6 +2292,124 @@ static void wsrep_NBO_begin_failed(THD *thd,
...
+
+     - query is not ALTER DB
+     - query is ALTER DB but
+          - HA_CREATE_USED_READ_ONLY is not set.
+          - Or if we have set other fields as well and set READ ONLY to true.
+
+  */
+  LEX *lex = thd->lex;
+  HA_CREATE_INFO *create_info = lex->create_info;
+  bool some_tables_in_read_only_schema{false};
```
kamil-holubicki: Let's  keep consistent variables initialization way across the file `bool my_bool = false`

sql/wsrep_mysqld.cc
```diff
@@ -2285,6 +2292,124 @@ static void wsrep_NBO_begin_failed(THD *thd,
...
+  std::string first_read_only_schema{};
```
kamil-holubicki: curly brackets are not needed

sql/sp.cc
```diff
@@ -964,7 +964,7 @@ enum_sp_return_code sp_drop_routine(THD *thd, enum_sp_type type,
     return SP_INTERNAL_ERROR;
 
 #ifdef WITH_WSREP
-  if (WSREP(thd) && wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL, NULL)) {
+  if (WSREP(thd) && wsrep_to_isolation_begin(thd, name->m_db.str, NULL, NULL)) {
```
kamil-holubicki, line 967: Why this change?

sql/sql_admin.cc
```diff
@@ -1760,7 +1760,7 @@ bool Sql_cmd_analyze_table::execute(THD *thd) {
   });
 
   if (WSREP(thd) && !thd->lex->no_write_to_binlog &&
-      wsrep_to_isolation_begin(thd, NULL, NULL, first_table)) {
+      wsrep_to_isolation_begin(thd, first_table->db, NULL, first_table)) {
```
kamil-holubicki, line 1763: why this change?

mysql-test/suite/galera/t/pxc_read_only_schema.test
```diff
@@ -0,0 +1,749 @@
...
+write_file $INIT_FILE;
+  CREATE SCHEMA schema_read_only;
+  ALTER SCHEMA schema_read_only READ ONLY=1;
+  ALTER SCHEMA schema_read_only COLLATE utf8_bin;
+  CREATE TABLE schema_read_only.t(i INT) TABLESPACE innodb_system;
+  DROP SCHEMA schema_read_only;
+  CREATE TABLE S.init_file(i INT);
+EOF
+
+let $wait_counter= 10000;
```
kamil-holubicki, line 107: please be consistent across the file
`--let` instead of `let`
`--eval` instead of `eval`

kamil-holubicki (changes requested):

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Ok. This is a side-effect of running clang-format-10 on the entire file. If we have plans to fix this as part of a different JIRA/PR, then I will revert this change and submit only those changes that are necessary for this bugfix.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: We are inside `wsrep_can_run_in_toi()` function, which is specific to WITH_WSREP builds. So there is no need.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Good catch. I will fix this.

sql/sp.cc
venkatesh-prasad-v: In order to check if the schema is read only, we need to first know the DB name. Here,  we were sending "mysql" as the DB name for all TOI operations on stored procedures and functions. This resulted in inconsistency when we execute create/drop operations on procedures and functions. So this change was required.

mysql-test/suite/galera/t/pxc_read_only_schema.test
venkatesh-prasad-v: Most of the test scenarios written in this file are from `./mysql-test/include/schema_read_only.inc` and was adjusted as per PXC's behavior. So, I have left them as-is.

sql/sql_admin.cc
venkatesh-prasad-v: As said in my previous comment, we need to know the db name of the table which are running ANALYZE TABLE on. So, it is better to send the db name if we already have that information.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Changed to `is_operating_on_read_only_schema()`.

venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1351/console

dutow (approved):

venkatesh-prasad-v:
Closing this PR as we decided to go with a different approach. This will be fixed as part of https://github.com/percona/percona-xtradb-cluster/pull/1752

PR #1740, opened 2023-02-22
PXC-4152 PXC-8.0.31 Refresh(Fixes compilation failures)

VarunNagaraju:
https://jira.percona.com/browse/PXC-4152

Fixes some array out of bound warnings upgraded to error for a successful compilation of code.

### DIFF
--- a/sql/wsrep_high_priority_service.cc
+++ b/sql/wsrep_high_priority_service.cc
@@ -588,7 +588,7 @@ int Wsrep_high_priority_service::log_dummy_write_set(
   return ret;
 }
 
-void Wsrep_high_priority_service::debug_crash(const char *crash_point) {
+void Wsrep_high_priority_service::debug_crash(const char *crash_point __attribute__((unused))) {
   assert(m_thd == current_thd);
   DBUG_EXECUTE_IF(crash_point, DBUG_SUICIDE(););
 }
--- a/sql/wsrep_mysqld.cc
+++ b/sql/wsrep_mysqld.cc
@@ -1115,7 +1115,7 @@ static void setup_galera_encryption_params(char* provider_options, size_t buf_si
   }
 
   snprintf(provider_options, buf_size, "%s", options.c_str());
-  provider_options[buf_size] = 0;
+  provider_options[buf_size - 1] = 0;
 }
 
 int wsrep_init() {
@@ -1175,7 +1175,7 @@ int wsrep_init() {
   } else {
     snprintf(buffer, buf_size, "%s", provider_options ? provider_options : "");
   }
-  buffer[buf_size] = 0;
+  buffer[buf_size - 1] = 0;
 
   /* Setup GCache and WSCache encryption according to
       wsrep_gcache_encrypt and wsrep_disk_pages_encrypt.


sql/partitioning/partition_base.cc
```diff
@@ -100,13 +100,13 @@ void part_name(char *out_buf, const char *path, const char *parent_elem_name,
       parent_elem_name ? parent_elem_name : elem_name, part_name, FN_REFLEN);
   if (parent_elem_name) {
     assert(part_name_length + sp_prefix_length < sizeof(part_name));
-    strncat(part_name, sp_prefix, sizeof(part_name) - part_name_length - 1);
+    strncat(part_name, sp_prefix, sizeof(part_name) - part_name_length);
```
dutow: Why? strncat can write a terminating null after the specified length, that -1 makes sense there.

sql/mysqld.cc
```diff
@@ -6237,7 +6237,7 @@ static int generate_server_uuid() {
 
   delete thd;
 
-  strncpy(server_uuid, uuid.c_ptr(), sizeof(server_uuid));
+  strncpy(server_uuid, uuid.c_ptr(), sizeof(server_uuid) - 1);
```
kamil-holubicki: I think the original is fine. We have 
```server_uuid[UUID_LENGTH] = '\0';```
2 lines below.
Does the compiler complain here?

sql/partitioning/partition_base.cc
VarunNagaraju: Right. But, It seems to have thrown the following warning

In function 'char* strncat(char*, const char*, size_t)',
    inlined from 'void native_part::part_name(char*, const char*, const char*, const char*)' at /home/venki/work/pxc/review/80/percona-xtradb-cluster/sql/partitioning/partition_base.cc:109:12:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:136:34: error: 'char* __builtin___strncat_chk(char*, const char*, long unsigned int, long unsigned int)' output may be truncated copying between 0 and 511 bytes from a string of length 511 [-Werror=stringop-truncation]
  136 |   return __builtin___strncat_chk (__dest, __src, __len, __bos (__dest));
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors

Also, it seems like a PS issue. If required, should be handled in a different ticket. I'll remove this change.

sql/mysqld.cc
VarunNagaraju: It did not complain for me in the ubuntu:focal docker image with gcc 9.4.0. It also have seemed to have complained only on native focal. The warnings in sql/wsrep_high_priority_service.cc and sql/wsrep_mysqld.cc reproduced didn't reproduce locally and they appeared only in focal docker image. 
But anyway, since there's a null character added later, I'll remove this change as well.

sql/partitioning/partition_base.cc
kamil-holubicki: This warning comes from line 109. This is the 2nd strncat. It is emitted, because the src buffer is FN_REFLEN long, so the compiler thinks it may be the case that it will be truncated.
But in my opinion, it is OK, that's why we use strncat.
The solution for this warning would be dynamically allocating subpart_name with the maximum size that would fit strncat, but I don't think we wan't dynamic allocation here. 
Long story short: leave as it is (maybe we should suppress this warning in upstream PS @percona-ysorokin )

kamil-holubicki (approved):

PR #1743, opened 2023-02-23
PXC-4050: 8.0.31 refresh

kamil-holubicki:
Server-side necessary adjustments.
Galera, wsrep_lib repo pointers updated.

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -202,17 +202,17 @@ bool wsrep_setup_allowed_sst_methods() {
 static bool sst_awaiting_callback = false;
 
 // Signal end of SST
-static void wsrep_sst_complete(THD *thd, int const rcode) {
+static int wsrep_sst_complete(THD *thd, int const rcode) {
   Wsrep_client_service client_service(thd, thd->wsrep_cs());
-  Wsrep_server_state::instance().sst_received(client_service, rcode,
-                                              &sst_awaiting_callback);
+  return Wsrep_server_state::instance().sst_received(client_service, rcode,
+                                                     &sst_awaiting_callback);
 }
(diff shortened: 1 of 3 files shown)


sql/wsrep_sst.cc
```diff
@@ -661,7 +663,13 @@ static void *sst_joiner_thread(void *a) {
     /* Read committed isolation to avoid gap locking */
     thd->variables.transaction_isolation = ISO_READ_COMMITTED;
 
-    wsrep_sst_complete(thd, -err);
+    if (wsrep_sst_complete(thd, -err)) {
+      WSREP_ERROR(
+          "wsrep_sst_complete() failed, can't signal end of SST. "
```
venkatesh-prasad-v: I am not sure if it is a good idea to put a function name in the error message. 

Can we use something like `Server encounterd a failure while signalling the completion of SST. Aborting` ? If we want to write the function name here, then we could consider `WSREP_DEBUG()` here.

sql/wsrep_var.cc
```diff
@@ -281,7 +283,9 @@ void wsrep_start_position_init(const char *val) {
     return;
   }
 
-  wsrep_set_local_position(NULL, val, strlen(val), false);
+  if (wsrep_set_local_position(NULL, val, strlen(val), false)) {
+    WSREP_ERROR("wsrep_set_local_position() failed");
```
venkatesh-prasad-v: I am not sure if it is a good idea to put a function name in the error message. 

Can we use something like `Server failed to set local position.` ? 
If we want to write the function name here, then we could consider `WSREP_DEBUG()` here.

venkatesh-prasad-v (approved):

PR #1752, opened 2023-03-03
PXC-4137: WSREP applier threads fail to modify read only schemas

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4137

    Problem
    -------
    When a PXC cluster acts as the async replica to a MySQL source, and if
    the cluster has a schema marked as read only in the PXC cluster, then
    updates on the schema from async source server are not handled correctly
    leading to inconsistencies between PXC cluster nodes.
    
    Analysis
    --------
    Consider the scenario where a schema has been marked as read only in the
    PXC Cluster, while the schema on source server is open for writes.
    
    Topology:
    
                   async           PXC
            node1 -------> node2 <-----> node3
    
    1. n1> CREATE DATABASE testdb;
    2. n2> ALTER SCHEMA testdb READ ONLY = 1; -> schema will be read only on all
                                                 nodes of the cluster
    3. n1> CREATE TABLE testdb.t1 ..
    
    In such a case, the CREATE TABLE query in step 3 will be applied on pxc
    replica by the replication applier thread on node2 and the same will
    also be replicated to node3 as TOI.
    
    Since the replication applier threads are allowed to continue even when
    schema is read only, the CREATE TABLE will be successful on node2.
    
    However, the same operation will error out on node3 as wsrep applier
    threads are not allowed to operate on read only schemas, thereby
    resulting in inconsistency voting to kick-in and thereby resulting in
    the cluster shutdown.
    
    NOTE:
    This problem could have been easily fixed by allowing the wsrep applier
    threads to operate on read only schemas. However this resulted in
    inconsistency if the user tries to create database objects directly on
    PXC nodes (node2 and node3) as the the queries are replicated as TOI
    before the checks for read only schemas are carried out.  As a result,
    the query can still succeed on the replicated node while it fails on the
    originating node, thereby again resulting in inconsistency followed by
    cluster shutdown.
    
    node2> create table testdb.t1(i int);
    ERROR 3989 (HY000): Schema 'testdb' is in read only mode.
    
    Solution
    --------
    - A new Query_event flag Q_WSREP_SKIP_READONLY_CHECKS has been added.
    - Server now enables this flag and sends it in the replication
      changestream only when the thread is an async/semisync applier thread.
    - On receiving node, server checks for Q_WSREP_SKIP_READONLY_CHECKS flag
      if enabled it sets thd->wsrep_applier_skip_readonly_checks.
    - Server ignores all readonly checks when
      thd->wsrep_applier_skip_readonly_checks is set.

### DIFF
--- a/libbinlogevents/include/binlog_event.h
+++ b/libbinlogevents/include/binlog_event.h
@@ -116,6 +116,7 @@
   packet (i.e. a query) sent from client to master;
   First, an auxiliary log_event status vars estimation:
 */
+#ifdef WITH_WSREP
 #define MAX_SIZE_LOG_EVENT_STATUS                                             \
   (1U + 4 /* type, flags2 */ + 1U + 8 /* type, sql_mode */ + 1U + 1 +         \
    255 /* type, length, catalog */ + 1U + 4 /* type, auto_increment */ + 1U + \
@@ -129,10 +130,27 @@
    1 + 255 /* host_len, host */ + 1U + 1 /* type, explicit_def..ts*/ + 1U +   \
    8 /* type, xid of DDL */ + 1U +                                            \
    2 /* type, default_collation_for_utf8mb4_number */ + 1U +                  \
-   1 /* type, sql_require_primary_key */ + 1U +                                     \
+   1 /* type, sql_require_primary_key */ + 1U +                               \
    1 /* type, default_table_encryption */ + 1U +                              \
-   1 /* type, binlog_ddl_skip_rewrite */)
-
+   1 /* type, binlog_ddl_skip_rewrite */ + 1U +                               \
+   1 /* type, wsrep_applier_skip_readonly_checks */)
+#else
+#define MAX_SIZE_LOG_EVENT_STATUS                                             \
+  (1U + 4 /* type, flags2 */ + 1U + 8 /* type, sql_mode */ + 1U + 1 +         \
+   255 /* type, length, catalog */ + 1U + 4 /* type, auto_increment */ + 1U + \
+   6 /* type, charset */ + 1U + 1 + 255 /* type, length, time_zone */ + 1U +  \
+   2 /* type, lc_time_names_number */ + 1U +                                  \
+   2 /* type, charset_database_number */ + 1U +                               \
+   8 /* type, table_map_for_update */ + 1U +                                  \
+   4 /* type, master_data_written */ + /* type, db_1, db_2, ... */            \
+   1U + (MAX_DBS_IN_EVENT_MTS * (1 + NAME_LEN)) + 3U +                        \
+   /* type, microseconds */ +1U + 32 * 3 + /* type, user_len, user */         \
+   1 + 255 /* host_len, host */ + 1U + 1 /* type, explicit_def..ts*/ + 1U +   \
+   8 /* type, xid of DDL */ + 1U +                                            \
+   2 /* type, default_collation_for_utf8mb4_number */ + 1U +                  \
+   1 /* type, sql_require_primary_key */ + 1U +                               \
+   1 /* type, default_table_encryption */)
+#endif /* WITH_WSREP */
 /**
    Uninitialized timestamp value (for either last committed or sequence number).
    Often carries meaning of the minimum value in the logical timestamp domain.
--- a/libbinlogevents/include/statement_events.h
+++ b/libbinlogevents/include/statement_events.h
@@ -670,8 +670,10 @@ class Query_event : public Binary_log_event {
   uint8_t sql_require_primary_key;
 
   uint8_t default_table_encryption;
+#ifdef WITH_WSREP
   uint8_t ddl_skip_rewrite;
   uint8_t wsrep_applier_skip_readonly_checks;
+#endif /* WITH_WSREP */
 
   /**
     The constructor will be used while creating a Query_event, to be
--- a/libbinlogevents/include/statement_events.h
+++ b/libbinlogevents/include/statement_events.h
@@ -552,7 +552,7 @@ class Query_event : public Binary_log_event {
     /*
       Replicate Q_WSREP_SKIP_READONLY_CHECKS.
     */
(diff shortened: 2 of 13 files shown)


venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/1358/console

libbinlogevents/include/binlog_event.h
```diff
@@ -129,10 +130,27 @@
    1 + 255 /* host_len, host */ + 1U + 1 /* type, explicit_def..ts*/ + 1U +   \
    8 /* type, xid of DDL */ + 1U +                                            \
    2 /* type, default_collation_for_utf8mb4_number */ + 1U +                  \
-   1 /* type, sql_require_primary_key */ + 1U +                                     \
+   1 /* type, sql_require_primary_key */ + 1U +                               \
    1 /* type, default_table_encryption */ + 1U +                              \
-   1 /* type, binlog_ddl_skip_rewrite */)
-
+   1 /* type, binlog_ddl_skip_rewrite */ + 1U +                               \
+   1 /* type, wsrep_applier_skip_readonly_checks */)
```
kamil-holubicki, line 136: Is this flag transparent for PXB and other binlog-related tools? In other words: If we add this flag, does it impact any other external tool?

libbinlogevents/include/statement_events.h
```diff
@@ -657,6 +671,7 @@ class Query_event : public Binary_log_event {
 
   uint8_t default_table_encryption;
   uint8_t ddl_skip_rewrite;
+  uint8_t wsrep_applier_skip_readonly_checks;
```
kamil-holubicki, line 675: WITH_WSREP (ddl_skip_rewrite and wsrep_applier_skip_readonly_checks)

sql/log_event.cc
```diff
@@ -3674,11 +3674,27 @@ bool Query_log_event::write(Basic_ostream *ostream) {
...
+    binlog_ddl_skip_rewrite set.
+  */
   if (thd && thd->variables.binlog_ddl_skip_rewrite) {
     *start++ = Q_DDL_SKIP_REWRITE;
     *start++ = thd->variables.binlog_ddl_skip_rewrite;
   }
 
+  /*
+    Replicate Q_WSREP_SKIP_READONLY_CHECKS only if it is a
+    replication applier thread.
```
kamil-holubicki: I suggest different wording to make it clear that we are talking about SQL replication applier thread, not WSREP replication applier

sql/sql_parse.cc
```diff
@@ -4573,6 +4573,10 @@ int mysql_execute_command(THD *thd, bool first_level) {
         break;
 
 #ifdef WITH_WSREP
+      if (thd->locked_tables_mode) {
```
kamil-holubicki, line 4576: What is it for?

kamil-holubicki (commented):
Looks fine. Just a few questions for my better understanding and some cosmetic changes suggestions.

Commit message:

`if the cluster has a schema has marked` - remove 2nd 'has'
`are not handled incorrectly leading` => s/incorrectly/correctly

`Server now enables this flag and sends it in the replication changestream only when the replication applier thread is preparing Query_log_event for TOI.`

I suggest clear wording, saying that replication applier thread is SQL async/semisync applier thread, not wsrep applier thread.

libbinlogevents/include/binlog_event.h
venkatesh-prasad-v: No. I don think this will impact any other tool unless any of the dependent tool has a conflicting enum value in `Query_event_status_vars` enum.

libbinlogevents/include/statement_events.h
venkatesh-prasad-v: Done

sql/log_event.cc
venkatesh-prasad-v: Done

sql/sql_parse.cc
venkatesh-prasad-v: This is needed if we want to block DB operation while are in locked_tables_mode.
```
LOCK TABLE test.p WRITE;
CREATE SCHEMA S; -> This query will fail on node1 with ER_LOCK_OR_ACTIVE_TRANSACTION, but will be successful on node2
```

libbinlogevents/include/binlog_event.h
dutow: What if upstream adds another flag to this list in a later version, won't that cause backwards compatibility issues for PXC? (I see that we already added a flag to this list earlier)

venkatesh-prasad-v: This will be a problem only when we replicate from a new upstream server to old pxc server. But this is not supported generally.

If upstream adds a new flag, then we must ensure that our changes are added in the end during our merges. This way there won't be any issues between upstream and PXC.

dutow: I was thinking about the following situation:
* new flag added
* we release new PXC version
* user starts to upgrade, upgrades nodes one-by-one
* for a short period, we can have a situation where different nodes in the cluster run different versions, and the flag values have different meanings

kamil-holubicki: @dutow In rolling update we block writes on updated (newer, having higher version number) nodes.
EDIT: No, my bad. This is the case only for 5.7 - 8.0 mixed version cluster, so this is a valid question.

dutow (approved):

kamil-holubicki (approved):

libbinlogevents/include/binlog_event.h
venkatesh-prasad-v: It was decided on slack to leave the old flags (added by us) as they are and to add the new flag after a reasonable gap. A value of 128 was chosen for the new flag to ensure we have enoug gaps for upstream flags.

PR #1764, opened 2023-03-15
PXC-4163: pxc_strict_mode does not detect version mismatch

kamil-holubicki:
https://jira.percona.com/browse/PXC-4163

Server part of the fix.

Problem:
When in mixed-version cluster, 8.0 node becomes a donor for another 8.0 node, for the time of being donor, the node accepts writes.

Cause:
block_write_while_in_rolling_upgrade() function skips check if the node is not in SYNCED state.

Solution:
We need to keep checking for write restriction in mixed-version cluster for all the time, regardless of the state of the node. The only time we have to accept writes is when the node joins the cluster and does the initialization. Once the node is initialized it has to reject writes even if it becomes a donor.

Galera repo pointer updated.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -1809,9 +1809,15 @@ static bool pxc_strict_mode_lock_check(THD *thd) {
  * Writing on a new node can crash nodes with lower version 8.0->5.7
  */
 static bool block_write_while_in_rolling_upgrade(THD *thd) {
-  /* Ignore check while server is not ready or for background wsrep applier too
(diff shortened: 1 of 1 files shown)


venkatesh-prasad-v (approved):

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1426
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1427

PR #1766, opened 2023-03-17
PXC-4041: Rotating InnoDB MK from a read_only node breaks the cluster

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4041

Problem
-------
When super_read_only or read_only is set, ALTER INSTANCE fails (as expected) which triggers consistency voting mechanism and causes node to leave the cluster

Analysis
--------
In original flow, the validation happens inside Sql_cmd_alter_instance::execute

Eg: For ALTER INSTACE ROTATE INNODB MASTER KEY, it happens at

Sql_cmd_alter_instance::execute
 -> Rotate_innodb_master_key::execute
  -> acquire_shared_global_read_lock
   -> check_readonly
    -> err_readonly

As the statement is replicated as TOI before validation, this causes the query to fail on originating node and succeed on the replicating nodes.

In addition to the above problem, the following problems were also discovered during investigation.

1. It was also found that server allows read_only variable to be updated while a TOI is in progress thereby causing cluster inconsistency when the read_only variable was changed after the TOI was replicated.

2. Discovered the code in wsrep_read_only_option() that silently modified the global opt_readonly variable and always evaluated to true. This function has been removed in 5.7 by commit f481ce88437, but was not removed in 8.0

Solution
--------
1. In the beginning of TOI, the server now takes an intention-exclusive lock on the global read lock to have protection against concurrent change of read_only option. If the server is already in read only mode, we error out with ER_OPTION_PREVENTS_STATEMENT.

2. The function wsrep_read_only_option() has been removed.

### DIFF
--- a/mysql-test/suite/galera/r/galera_kill_ddl_query.result
+++ b/mysql-test/suite/galera/r/galera_kill_ddl_query.result
@@ -1,25 +1,5 @@
 CALL mtr.add_suppression("WSREP: TO isolation failed");
-################ Test 1 ################
-[connection node_1]
-FLUSH TABLES WITH READ LOCK;
-[connection node_1a]
-SET DEBUG_SYNC = 'wsrep_TOI_begin_after_wsrep_skip_wsrep_hton SIGNAL reached';
-CREATE TABLE t1 (ID INT PRIMARY KEY);;
-[connection node_1b]
-SET DEBUG_SYNC = "now WAIT_FOR reached";
-# Killing query on node1a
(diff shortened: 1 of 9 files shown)


venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/110/

kamil-holubicki (approved):

PR #1770, opened 2023-03-28
PXC-3047: Server hangs up if cannot connect to members PXC 8.0

kamil-holubicki:
https://jira.percona.com/browse/PXC-3047

Problem:
The server with a view stored in gvwstate.dat file hangs up infinitely, when it starts, but can not connect to other nodes.

Cause:
When the node starts, it waits for primary component for pc.wait_prim_timeout seconds. If PC is not reached, the node exits. However, this timeout is not respected in case of the PC view restored from the file. In such a case node just joins the cluster and waits for PC infinitely.

Solution:
(The solution is in Galera library, server part contains only MTR test) pc.wait_prim_timeout defaults to 30s and it may be not enough for cluster recovery after failure. Introduce new configuration option pc.wait_restored_prim_timeout with default of 0s.
If set to default, there is the old behavior - the node waits infinitely if set to any other value, the node waits for requested time (sec) for PC and exits if not reached.


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/1536/

mysql-test/suite/galera/t/galera_restored_pc.test
```diff
@@ -0,0 +1,51 @@
...
+--exec echo "restart: --wsrep_provider_options=base_port=$NODE_GALERAPORT_2" > $MYSQLTEST_VARDIR/tmp/mysqld.2.expect
+--let $restart_parameters =
+
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/gvwstate.dat
+--remove_file $MYSQLTEST_VARDIR/mysqld.1/data/grastate.dat
+--remove_file $MYSQLTEST_VARDIR/mysqld.1/data/gvwstate.dat
+--connection node_1
+--source include/start_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: This is redundant as `start_mysqld.inc` sources `include/wait_until_connected_again.inc` in the end.

mysql-test/suite/galera/t/galera_restored_pc.test
```diff
@@ -0,0 +1,51 @@
...
+--source include/galera_wait_ready.inc
+
+--connection node_2
+--source include/start_mysqld.inc
+--source include/wait_until_connected_again.inc
```
venkatesh-prasad-v: This is redundant as start_mysqld.inc sources include/wait_until_connected_again.inc in the end.

venkatesh-prasad-v:
Also, 

1. please re-record `galera.galera_defaults` test.
2. please check the test failure of `galera_restored_pc` (though it passes on retry)
```
[ 50%] galera.galera_restored_pc 'enc_off'      w2  [ retry-fail ]
        Test ended at 2023-03-29 01:49:25

CURRENT_TEST: galera.galera_restored_pc
mysqltest: At line 16: Query 'let KILL_NODE_PIDFILE = `SELECT @@pid_file`' failed.
In included file ./include/kill_galera.inc: 17
included from /tmp/results/PXC/mysql-test/suite/galera/t/galera_restored_pc.test: 15

The result from queries just before the failure was:
Killing server ...
Killing server ...
safe_process[5398]: Child process: 5399, exit: 1

 - the logfile can be found in '/tmp/results/PXC/mysql-test/var/log/galera.galera_restored_pc-enc_off/galera_restored_pc.log'
```

kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/1575/

venkatesh-prasad-v (approved):
LGTM. Please revert the changes in .gitmodules before merging the PR.

PR #1776, opened 2023-04-04
PXC-667: Crash during BF-abort of active HANDLER <table> OPEN AS <alias>

kamil-holubicki:
https://jira.percona.com/browse/PXC-667

Problem:
When HANDLER <table> OPEN AS <alias> query is running and it gets BF-aborted by DELETE TABLE <table>, the server crashes.

Cause:
Original flow:
HANDLER ... OPEN AS ... opens the table, creates MDL explicit lock on it and stores it in its internal hash. The table is not closed  (is not returned to table cache) when the query finishes. When other thread tries to delete the table, it is blocked until HANDLER ... CLOSE is executed and tables are closed and all explicit locks are released.

PXC flow:
session 1: HANDLER ... OPEN AS ... opens the table, creates MDL explicit lock on it and stores it in its internal hash. The table is not closed (is not returned to table cache) when the query finishes. session 2: If DELETE TABLE ... is executed while session 1 is still executing, session 1 detects that it was aborted and executes Wsrep_client_service::bf_rollback() as the part of wsrep_after_statement(). This causes all explicit locks release, which allows session 2 to go on, but does not close handler tables. When Session 1 executes CLOSE HANDLER ... it asserts during closing of handler talbes because it expects tables to be still locked, but they are already unlocked during bf_rollback()

Solution:
1. When HANDLER ... OPEN AS ... does self-rollback, close all handler tables before releasing explicit locks.
2. Unfortunately HANDLER ... OPEN AS ... does not start the transaction, so if it is BF-aborted while not active, there is no way to kick-off background rollbacker. The only way we can learn about BF-abort is to execute the next statement in victim session. Up to this moment, BF session needs to wait on explicit lock. This is similar situation to session explicit locks done by LOCK TABLES ... which is prohibited when pxc_strict_mode=ENFORCING.
As the consequence HANDLER ... OPEN AS ... has been prohibited for pxc_strict_mode=ENFORCING.

### DIFF
--- a/sql/sql_handler.cc
+++ b/sql/sql_handler.cc
@@ -163,6 +163,43 @@ static void mysql_ha_close_table(THD *thd, TABLE_LIST *tables)
   tables->mdl_request.ticket= NULL;
 }
 
+#ifdef WITH_WSREP
+static bool pxc_strict_mode_check(THD *thd)
+{
+  /* HANDLER OPEN is not supported by galera due as it not compatible
+  with multi-master semantics. It acquires explicit table lock. */
+  bool block= false;
+
+  switch (pxc_strict_mode)
+  {
+    case PXC_STRICT_MODE_DISABLED:
+    case PXC_STRICT_MODE_MASTER:
+      /* Do nothing */
+      break;
+    case PXC_STRICT_MODE_PERMISSIVE:
+    {
+      const char *msg= "Percona-XtraDB-Cluster doesn't recommend use of"
+                       " HANDLER <table> OPEN AS <alias>"
+                       " with pxc_strict_mode = PERMISSIVE";
+      WSREP_WARN("%s", msg);
+      push_warning_printf(thd, Sql_condition::SL_WARNING, ER_UNKNOWN_ERROR,
+                          "%s", msg);
+    }
+      break;
+    case PXC_STRICT_MODE_ENFORCING:
+    default:
+      const char *msg= "Percona-XtraDB-Cluster prohibits use of"
+                       " HANDLER <table> OPEN AS <alias>"
+                       " with pxc_strict_mode = ENFORCING";
+      block= true;
+      WSREP_ERROR("%s", msg);
+      my_message(ER_UNKNOWN_ERROR, msg, MYF(0));
+      break;
+  }
+
+  return block;
+}
+#endif
 
 /**
   Execute a HANDLER OPEN statement.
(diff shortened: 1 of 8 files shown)


sql/sql_parse.cc
```diff
@@ -7990,12 +7990,19 @@ static bool wsrep_should_retry_in_autocommit(const THD* thd)
     As of now, we only do this check for CHECK TABLE and SELECT, and if the
     same symptom is found for other commands, then please add it to the
     below list.
+
+    We also do not retry for HANDLER OPEN AS statement. This is because
```
venkatesh-prasad-v: Can you rephrase this and move the line that says `if the same symptom is found for other commands, then  please add it to the below list.` to the end of the comment?

sql/sql_handler.cc
```diff
@@ -163,6 +163,43 @@ static void mysql_ha_close_table(THD *thd, TABLE_LIST *tables)
   tables->mdl_request.ticket= NULL;
 }
 
+#ifdef WITH_WSREP
+static bool pxc_strict_mode_check(THD *thd)
+{
+  /* HANDLER OPEN is not supported by galera due as it not compatible
+  with multi-master semantics. It acquires explicit table lock. */
+  bool block = false;
```
venkatesh-prasad-v: Here and other assignments in this file: No space before '='.

sql/wsrep_thd.cc
```diff
@@ -70,6 +72,12 @@ void wsrep_client_rollback(THD *thd)
   /* Release all user-locks. */
   mysql_ull_cleanup(thd);
 
+  /* If there are any tables opened by HANDLER OPEN statement
+     we need to close them before releasing explicit MDL locks.
+     If we don't do it, DDL which caused BF-abort, like DELETE TABLE
```
venkatesh-prasad-v: DELETE TABLE is a DML. Did you mean DROP TABLE?

venkatesh-prasad-v (changes requested):

venkatesh-prasad-v (approved):

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5426/
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5427/
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5428/

PR #1780, opened 2023-04-07
PXC-4146: Option to fail startup if IST is unavailable

kamil-holubicki:
https://jira.percona.com/browse/PXC-4146

Problem:
When we want to join the new node to the cluster, it is possible to avoid time consuming SST process by manually restoring the backup to the new node, manually editing grastate.dat file and start the node. In such a scenario, only IST is expected to be performed. However, it appeared to be a common case when creation/edition of grastate.dat file is forgotten. In such a case the joiner node will receive the full SST which causes unnecessary load on donor and joiner side.

Cause:
The joiner node always sends State Transfer request which consists of:
1. SST method + receiver endpoint
2. IST receiver endpoint + minimum seqno to satisfy IST to the donor node.
It is the donor node who decides (basing on passed IST seqno) if SST is to be sent, or if it is bypassed and only IST is sent. In the case of grastate.dat absence on the joiner node, IST information sent to donor (empty cluster UUID, invalid seqno) makes it to conclude that the full SST is needed.

Solution:
Introduce new possible value, 'ist_only' of 'wsrep_sst_method'. When wsrep_sst_method=ist_only, the joiner sends State Transfer request to the donor which consists empty SST request part and only IST part. If the request cannot be satisfied by the donor, the joiner will fail to join leaving data directory unmodified.

### DIFF
--- a/mysql-test/suite/galera_3nodes/t/galera_ist_only.test
+++ b/mysql-test/suite/galera_3nodes/t/galera_ist_only.test
@@ -1,0 +1,175 @@
+# This test checks the case when the node is joining the cluster
+# and it specifies that it is interested only in receiving IST,
+# and no full SST should be done.
+
+--source include/big_test.inc
+--source include/galera_cluster_3nodes.inc
+
+--let $ofile= $MYSQLTEST_VARDIR/tmp/node.2.err
+
+--connection node_1
+CALL mtr.add_suppression(".*State transfer to .* failed: -125.*");
+CALL mtr.add_suppression(".*SST request is null, SST canceled.*");
+
+--connection node_3
+CALL mtr.add_suppression(".*State transfer to .* failed: -125.*");
+CALL mtr.add_suppression(".*SST request is null, SST canceled.*");
+
+--connection node_1
+CREATE TABLE t1 (a INT PRIMARY KEY);
+INSERT INTO t1 VALUES (1);
+
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t1;
+--source include/wait_condition.inc
+
+# Case 1:
+# node_2 should catch up with the cluster after the restart
+# because both possible donors have all IST-needed information
+--source include/shutdown_mysqld.inc
+
+--connection node_1
+INSERT INTO t1 VALUES (2);
+INSERT INTO t1 VALUES (3);
+INSERT INTO t1 VALUES (4);
+
+--connection node_2
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2' --wsrep_sst_method=ist_only --log-error=$ofile"
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+--source include/galera_wait_ready.inc
+--let $assert_text = Table t1 should have 4 rows
+--let $assert_cond = COUNT(*) = 4 FROM t1;
+--source include/assert.inc
+
+--let $assert_text = Assert that node_2 joined without SST
+--let $assert_file = $ofile
+--let $assert_select = The node will try to join the cluster using only IST.
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+# Case 2:
+# node_2 should fail joining the cluster, because it doesn't have
+# grastate.dat file (cluster uuid is missing)
+--source include/shutdown_mysqld.inc
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+--connection node_1
+INSERT INTO t1 VALUES (5);
+INSERT INTO t1 VALUES (6);
+INSERT INTO t1 VALUES (7);
+
+--connection node_2
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2' --wsrep_sst_method=ist_only --log-error=$ofile"
+--let $expect_crash_on_start = 1
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+
+# Now start node_2 as standalone. t1 still should have 4 rows.
+--let $restart_parameters = "restart: --wsrep_provider=none"
+--let $expect_crash_on_start = 0
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+--let $assert_text = Table t1 still should have 4 rows
+--let $assert_cond = COUNT(*) = 4 FROM t1;
+--source include/assert.inc
+
+# node_2 can recover only via full SST
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2'"
+--let $expect_crash_on_start = 0
+--let $do_not_echo_parameters = 1
+--source include/restart_mysqld.inc
+--source include/galera_wait_ready.inc
+--let $assert_text = Table t1 should have 7 rows
+--let $assert_cond = COUNT(*) = 7 FROM t1;
+--source include/assert.inc
+
+--let $assert_text = Assert that node_2 failed to join without SST
+--let $assert_file = $ofile
+--let $assert_select = Local state UUID 00000000-0000-0000-0000-000000000000 is different from group state
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+# Case 3:
+# node_2 should fail joining the cluster, because the requested donor
+# cannot serve IST
+
+--connection node_2
+--source include/shutdown_mysqld.inc
+
+--connection node_1
+INSERT INTO t1 VALUES (8);
+INSERT INTO t1 VALUES (9);
+INSERT INTO t1 VALUES (10);
+
+# prepare the donor with empty GCache
+--connection node_3
+--source include/shutdown_mysqld.inc
+--remove_file $MYSQLTEST_VARDIR/mysqld.3/data/galera.cache
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_3;gcache.size=0;gcache.page_size=16k'"
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+--source include/galera_wait_ready.inc
+
+# Overflow GCache to disable IST possibility
+--let $cnt = 200
+--disable_query_log
+CREATE DATABASE test2;
+USE test2;
+while ($cnt)
+{
+  --eval CREATE TABLE tt_$cnt(a int primary key);
+  --dec $cnt
+}
+USE test;
+--enable_query_log
+
+# Desync node_1. Galera will not elect it as IST donor.
+--connection node_1
+SET GLOBAL wsrep_desync=1;
+
+# now start node_2, but pointing to node_3 as the donor
+--connection node_2
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2' --wsrep_sst_method=ist_only --log-error=$ofile"
+--let $_expect_file_name = $MYSQLTEST_VARDIR/tmp/mysqld.2.expect
+--let $expect_crash_on_start = 1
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1592/

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+
+--let $assert_text = Assert that node_2 joined without SST
+--let $assert_file = $ofile
+--let $assert_select = The node will try to join the cluster using only IST.
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+# Case 2:
+# node_2 should fail joining the cluster, because it doesn't have
+# grastate.dat file (cluster uuid is missing)
```
venkatesh-prasad-v: Remove the trailing whitespace here

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+
+--connection node_2
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2' --wsrep_sst_method=ist_only --log-error=$ofile"
+--let $expect_crash_on_start = 1
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+
+# node_2 can recover only via full SST
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2'"
+--let $expect_crash_on_start = 0
```
venkatesh-prasad-v: Remove the trailing whitespace here

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
+# This test checks the case when the node is joining the cluster
+# and it specifies that it is interested only in receiving IST,
+# and no full SST should be done.
+
+--source include/big_test.inc
+--source include/galera_cluster_3nodes.inc
+
+SET @wsrep_sst_donor_global_saved = @@global.wsrep_sst_donor;
```
venkatesh-prasad-v: This variable is not being altered anywhere in the testcase. Any reasons why this is required?

Also `wsrep_sst_donor_global_saved` session variable is saved on node1, but restored on node3 at the end of the test.

sql/wsrep_sst.cc
```diff
@@ -772,6 +773,20 @@ std::string wsrep_sst_prepare() {
...
+    /* Inform Galera that we are done with SST and it can proceed with IST.
+    In fact the following call will wait until the server is fully initialized
+    and then inform Galera about two things:
+    1. call sst_received() - so from Galera's point of view it looks like we are done with
+                             sst
+    2. return empty string from this function - informs Galera that only IST should
+       be processed.*/
+    WSREP_INFO("State Transfer via SST was prohibited by setting wsrep_sst_method=ist_only. "
+                "The node will try to join the cluster using only IST.");
+    wsrep_sst_complete(current_thd, 0);
```
venkatesh-prasad-v: Can you make this a warning instead of info message so that users get to know the reason for failure of full SST? Currently it looks like below
```
2023-04-10T07:46:03.880661Z 2 [Note] [MY-000000] [WSREP] Server status change connected -> joiner
2023-04-10T07:46:03.880673Z 2 [Note] [MY-000000] [WSREP] wsrep_notify_cmd is not defined, skipping notification.
2023-04-10T07:46:03.880690Z 2 [Note] [MY-000000] [WSREP] State Transfer via SST was prohibited by setting wsrep_sst_method=ist_only. The node will try to join the cluster using only IST.
2023-04-10T07:46:03.880707Z 2 [Note] [MY-000000] [Galera] Processing SST received
2023-04-10T07:46:03.880723Z 2 [Note] [MY-000000] [Galera] server n2 state change: joiner -> initializing
2023-04-10T07:46:03.880741Z 2 [Note] [MY-000000] [WSREP] Server status change joiner -> initializing
```

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+INSERT INTO t1 VALUES (4);
+
+--connection node_2
+--let $restart_parameters = "restart: --wsrep_provider_options='base_port=$NODE_GALERAPORT_2' --wsrep_sst_method=ist_only --log-error=$ofile"
+--let $do_not_echo_parameters = 1
+--source include/start_mysqld.inc
+--source include/galera_wait_ready.inc
+--let $assert_text = Table t1 should have 4 rows
+--let $assert_cond = COUNT(*) = 4 FROM t1;
+
```
venkatesh-prasad-v: Missing `assert.inc`?

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+--let $assert_text = Assert that node_2 failed to join without SST
+--let $assert_file = $ofile
+--let $assert_select = Will never receive state. Need to abort.
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+--let $assert_text = Table t1 should have 10 rows
+--let $assert_cond = COUNT(*) = 10 FROM t1;
+
```
venkatesh-prasad-v: Missing `assert.inc`?

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+--let $assert_file = $ofile
+--let $assert_select = Local state UUID 00000000-0000-0000-0000-000000000000 is different from group state
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+# Case 3:
+# node_2 should fail joining the cluster, because the requested donor
+# cannot serve IST
+--connection node_1
+SET @@global.wsrep_provider_options = "pc.weight=3";
```
venkatesh-prasad-v: Why is this required when we are setting `wsrep_desync=1` to make node_3 serve as donor?

mysql-test/suite/galera_3nodes/t/galera_ist_only.test
```diff
@@ -0,0 +1,164 @@
...
+--source include/galera_wait_ready.inc
+
+--let $assert_text = Assert that node_2 failed to join without SST
+--let $assert_file = $ofile
+--let $assert_select = Will never receive state. Need to abort.
+--let $assert_count = 1
+--source include/assert_grep.inc
+
+--let $assert_text = Table t1 should have 10 rows
+--let $assert_cond = COUNT(*) = 10 FROM t1;
```
venkatesh-prasad-v, line 153: Also please validate we have 200 tables in `test2` database.

venkatesh-prasad-v (changes requested):

sql/wsrep_sst.cc
utdrmac: +1 for warning level

utdrmac:
> the joiner will fail to join leaving data directory unmodified

This is the most important part. Is there a test case specifically for this to ensure that the datadir on the joiner is unmodified if the IST fails?

kamil-holubicki:
> Is there a test case specifically for this to ensure that the datadir on the joiner is unmodified if the IST fails?

I've extended 'Case 2' scenario.

sql/wsrep_sst.cc
```diff
@@ -772,6 +773,20 @@ std::string wsrep_sst_prepare() {
...
+    In fact the following call will wait until the server is fully initialized
+    and then inform Galera about two things:
+    1. call sst_received() - so from Galera's point of view it looks like we are done with
+                             sst
+    2. return empty string from this function - informs Galera that only IST should
+       be processed.*/
+    WSREP_WARN("State Transfer via SST was prohibited by setting wsrep_sst_method=ist_only. "
+               "The node will try to join the cluster using only IST.");
+    wsrep_sst_complete(current_thd, 0);
+    return "";
```
altmannmarcelo: I would have a `define` to represent this empty string, just like `WSREP_STATE_TRANSFER_TRIVIAL` or `WSREP_STATE_TRANSFER_NONE`

altmannmarcelo (changes requested):

venkatesh-prasad-v (approved):

venkatesh-prasad-v:
Please squash the commits before merging the PR.

sql/wsrep_sst.cc
```diff
@@ -71,11 +71,14 @@ extern const char *wsrep_defaults_group_suffix;
 #define WSREP_SST_RSYNC "rsync"
 #define WSREP_SST_MYSQLDUMP "mysqldump"
 #define WSREP_SST_SKIP "skip"
+#define WSREP_SST_ONLY_IST "ist_only"
 #define WSREP_SST_XTRABACKUP "xtrabackup"
 #define WSREP_SST_XTRABACKUP_V2 "xtrabackup-v2"
 #define WSREP_SST_DEFAULT WSREP_SST_XTRABACKUP_V2
 #define WSREP_SST_ADDRESS_AUTO "AUTO"
 
+#define WSREP_STATE_TRANSFER_NO_SST ""
```
altmannmarcelo, line 80: perhaps it makes more sense to declare it at wsrep_api.h where WSREP_STATE_TRANSFER_TRIVIAL and WSREP_STATE_TRANSFER_NONE are also declared.

Also, I believe we should use the define on the wsrep API code where we take action to not send SST request if the return from wsrep_sst_prepare is empty. 

Example, for WSREP_STATE_TRANSFER_TRIVIAL I see that on percona-xtradb-cluster-galera/galera/src/replicator.cpp we populate `Replicator::TRIVIAL_SST(WSREP_STATE_TRANSFER_TRIVIAL);`  we set TRIVIAL_SST variable, and that is used to validate if we are in trivial SST mode. Same should be done for WSREP_STATE_TRANSFER_NO_SST

And possibly the name `WSREP_STATE_TRANSFER_NO_SST` is clashing with Replicator::NO_SST(WSREP_STATE_TRANSFER_NONE);

kamil-holubicki: `WSREP_STATE_TRANSFER_TRIVIAL` and `wsrep_api.h` are 3rd party. `WSREP_STATE_TRANSFER_NO_SST` is our thing.
Please note that the functionality which allows us to omit SST, but still have IST is already there in Galera, It is triggered when `sst_request_cb` responsible for preparing SST request string returns an empty string. 
WSREP_STATE_TRANSFER_TRIVIAL - this tells the donor to skip sending SST and IST, but joiner is waiting for IST anyway and gets stuck in JOINED state. Actually, this is what the comment for this define says, but actually it doesn't make much sense.
WSREP_STATE_TRANSFER_NONE - The same as trivial, but according to the comment, the node is supposed to join the cluster. But again it ends up in JONER state waiting for IST.
In fact, there is a mess around these variables. Please check `ReplicatorSMM::process_state_req()` (donor) and `ReplicatorSMM::request_state_transfer` (joiner). SST logic is mixed with IST logic...
So no `Replicator::TRIVIAL_SST`-like things are needed.

altmannmarcelo: > Please note that the functionality which allows us to omit SST, but still have IST is already there in Galera, It is triggered when sst_request_cb responsible for preparing SST request string returns an empty string.

That was my point. To have a clear code and make it easier to maintain (in 6 months when you forget the logic of empty string means no SST)  instead of checking for empty string on this code, we should check for a define (this define will translate to empty string) in which we can track back to  WSREP_STATE_TRANSFER_NO_SST, rather than trying to find in the source code all the places where we return empty string.

If that does not make the code any easier to maintain/understand I am ok with accepting the way it is.

altmannmarcelo (approved):

PR #1781, opened 2023-04-11
(8.0)PXC-667: Crash during BF-abort of active HANDLER <table> OPEN AS <alias>

kamil-holubicki:
https://jira.percona.com/browse/PXC-667

Problem:
When HANDLER <table> OPEN AS <alias> query is running and it gets
BF-aborted by DELETE TABLE <table>, the server crashes.

Cause:
Original flow:
HANDLER ... OPEN AS ... opens the table, creates MDL explicit lock on it
and stores it in its internal hash. The table is not closed  (is not
returned to table cache) when the query finishes. When other thread
tries to delete the table, it is blocked until HANDLER ... CLOSE is
executed and tables are closed and all explicit locks are released.

PXC flow:
session 1: HANDLER ... OPEN AS ... opens the table, creates MDL explicit
lock on it and stores it in its internal hash. The table is not closed
(is not returned to table cache) when the query finishes.
session 2: If DELETE TABLE ... is executed while session 1 is still
executing, session 1 detects that it was aborted and executes
Wsrep_client_service::bf_rollback() as the part of
wsrep_after_statement(). This causes all explicit locks release, which
allows session 2 to go on, but does not close handler tables.
Session 1 detects that tables are still opened and asserts.

Solution:
1. When HANDLER ... OPEN AS ... does self-rollback, close all handler
tables before releasing explicit locks.
2. Unfortunately HANDLER ... OPEN AS ... does not start the transaction,
so if it is BF-aborted while not active, there is no way to kick-off
background rollbacker. The only way we can learn about BF-abort is to
execute the next statement in victim session. Up to this moment, BF
session needs to wait on explicit lock. This is similar situation to
session explicit locks done by LOCK TABLES ... which is prohibited when
pxc_strict_mode=ENFORCING.
As the consequence HANDLER ... OPEN AS ... has been prohibited for
pxc_strict_mode=ENFORCING.

### DIFF
--- a/mysql-test/suite/galera/r/handler_open_as.result
+++ b/mysql-test/suite/galera/r/handler_open_as.result
@@ -1,0 +1,12 @@
+SET GLOBAL pxc_strict_mode = 'DISABLED';
+SET GLOBAL pxc_strict_mode = 'DISABLED';
+CREATE TABLE t1 (id INT PRIMARY KEY) ENGINE=InnoDB;
+SET DEBUG_SYNC = 'innodb_delay_open_table SIGNAL reached WAIT_FOR continue';
+HANDLER t1 OPEN AS t1;;
+SET DEBUG_SYNC = 'now WAIT_FOR reached';
+DROP TABLE t1;
+ERROR 40001: WSREP detected deadlock/conflict and aborted the transaction. Try restarting the transaction
+HANDLER t1 CLOSE;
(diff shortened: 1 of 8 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1601/

venkatesh-prasad-v (approved):

PR #1783, opened 2023-04-14
PXC-3760: Implement a feature to evict a lagging node from PXC

voodam:
https://jira.percona.com/browse/PXC-3760

Problem:
We want to evict a node if it slows down a whole cluster due to the flow control.

Solution:
We introduce two variables: gcs.fc_auto_evict_window and gcs.fc_auto_evict_threshold. If the node pauses in FC more then "threshold * window" in the last "window" seconds frame, node will just abort.

The abortion can be problematic if we have 2-nodes cluster. From the side of other nodes this will be indistinguishable from the crash of the lagging node.


mysql-test/suite/galera/r/galera_restart_on_unknown_option.result
```diff
@@ -4,6 +4,8 @@ call mtr.add_suppression("The privilege system failed to initialize correctly");
 call mtr.add_suppression("SST failed/interrupted");
 call mtr.add_suppression("IST was interrupted");
 call mtr.add_suppression("IST didn't contain all write sets");
+call mtr.add_suppression("Attempting backtrace");
```
kamil-holubicki: I think this one was fixed on 8.0.32 release branch.
@venkatesh-prasad-v

mysql-test/suite/galera/t/galera_fc_auto_evict.test
```diff
@@ -0,0 +1,51 @@
...
+
+--connection node_2
+--source include/expect_crash.inc
+# Triggering flow control logic taken from here:
+# https://www.percona.com/blog/monitoring-flow-control-percona-xtradb-cluster-5-6/
+SET GLOBAL wsrep_provider_options="gcs.fc_limit=1;gcs.fc_auto_evict_window=0.1";
+FLUSH TABLES WITH READ LOCK;
+
+--connection node_1
+CREATE TABLE t1(a INT);
```
kamil-holubicki: Doesn't it complain about the lack of the explicit primary key?

kamil-holubicki: Please wrap inserting loop in `--disable_query_log` / `--enable_query_log` and add `--echo ...` informing about what is happening.

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1613/

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: No, where this logic is located? Or the name of error. I couldn't find anything quickly.

mysql-test/suite/galera/r/galera_restart_on_unknown_option.result
voodam: Revert this?

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: Done

mysql-test/suite/galera/r/galera_restart_on_unknown_option.result
venkatesh-prasad-v: Yes. This has been fixed by commit https://github.com/percona/percona-xtradb-cluster/commit/c01b796d8fc31cb8352817fb7e79bb22646db67c. So, please revert this change.

mysql-test/suite/galera/t/galera_fc_auto_evict.test
venkatesh-prasad-v: PXC allows creation of tables without primary keys, however it [complains](https://github.com/venkatesh-prasad-v/percona-xtradb-cluster/blob/8.0/sql/sql_base.cc#L6219-L6224) when we try to execute a DMLs  when pxc_strict_mode=ENFORCING.

mysql-test/suite/galera/t/galera_fc_auto_evict.test
```diff
@@ -0,0 +1,54 @@
...
+--connection node_2
+UNLOCK TABLES;
+
+--connection node_1
+--let $wait_condition=SELECT VARIABLE_VALUE = 1 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+--let SEARCH_FILE=$MYSQLTEST_VARDIR/log/mysqld.2.err
+--let SEARCH_PATTERN=FC auto evict: pause limit exceeded
+--source include/search_pattern.inc
```
venkatesh-prasad-v: Please consider using `assert_grep.inc` as it fails immediately when the required pattern is not found.

mysql-test/suite/galera/t/galera_fc_auto_evict.test
```diff
@@ -0,0 +1,54 @@
+# PXC-3760: Implement a feature to evict a lagging node from PXC
+#
+# Problem:
+# We want to evict a node if it slows down a whole cluster due to the flow control.
+#
+# Solution:
+# We introduce two variables: gcs.fc_auto_evict_window and gcs.fc_auto_evict_threshold.
+# If the node pauses in FC more then "threshold * window" in the last  window" seconds
```
venkatesh-prasad-v:
```suggestion
# If the node pauses in FC more than "threshold * window" in the last "window" seconds
```

mysql-test/suite/galera/r/galera_defaults.result
```diff
@@ -24,7 +25,6 @@ WSREP_CERTIFICATION_RULES	strict
 WSREP_CERTIFY_NONPK	ON
 WSREP_CLUSTER_NAME	my_wsrep_cluster
 WSREP_DBUG_OPTION	
-WSREP_DEBUG	NONE
```
venkatesh-prasad-v: Why this change?

voodam: Probably it isn't required, I'll revert this too. At first I thought this param depends on build type, not on CLI parameter.

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: Done

mysql-test/suite/galera/t/galera_fc_auto_evict.test
kamil-holubicki:
```
mysql> create table t1 (a int);
Query OK, 0 rows affected (0,19 sec)

mysql> insert into t1 values (0);
ERROR 1105 (HY000): Percona-XtraDB-Cluster prohibits use of DML command on a table (test.t1) without an explicit primary key with pxc_strict_mode = ENFORCING or MASTER
```
Aren't we in `ENFORCING` mode during the test?

voodam: In my case it's `DISABLED`

venkatesh-prasad-v: It has been set to DISABLED by default for all test cases. See https://github.com/venkatesh-prasad-v/percona-xtradb-cluster/blob/8.0/mysql-test/include/default_mysqld.cnf#L27

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: Done

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: Anyway, rewrited with primary key

mysql-test/suite/galera/t/galera_fc_auto_evict.test
```diff
@@ -0,0 +1,56 @@
...
+--connection node_2
+UNLOCK TABLES;
+
+--connection node_1
+--let $wait_condition=SELECT VARIABLE_VALUE = 1 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+--let $assert_text= FС auto eviction triggered
+--let $assert_file= $MYSQLTEST_VARDIR/log/mysqld.2.err
+--let $assert_select= FC auto evict: pause limit exceeded
```
venkatesh-prasad-v: Please update the test since the error message was changed.

kamil-holubicki:
Please solve the conflict after merging galera

kamil-holubicki (approved):

mysql-test/suite/galera/t/galera_fc_auto_evict.test
voodam: Yes, sure

venkatesh-prasad-v (approved):

kamil-holubicki:
Looks like the new test fails on Jenkins
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1617/

voodam:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1618/

kamil-holubicki (approved):

PR #1788, opened 2023-04-24
PXC-4200: Node does not start if wsrep_sst_method=skip

kamil-holubicki:
https://jira.percona.com/browse/PXC-4200

Problem:
When wsrep_sst_method=skip, the Joiner starts up to the point when it reaches JOINER state and stuck forever.

Cause:
'skip' is translated into 'trivial' SST method on Galera level. This is transferred to the Donor. Donor properly skips SST/IST sending however, Joiner starts IST receiver expecting IST to be received.

Solution:
wsrep_sst_method=skip never worked in 8.0 branch. There were never complains about it. Also there is no clear use case when it can be useful.
It would be possible to fix it to work the same as in 5.7, but it needs bigger changes around state transfer implmentation at Galera and server level.
Considering above, this commit removes 'skip' SST method. From now on 'skip' is considered as any other SST method
(service by wsrep_sst_\<method\> script)

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -70,7 +70,6 @@ extern const char *wsrep_defaults_group_suffix;
 
 #define WSREP_SST_RSYNC "rsync"
 #define WSREP_SST_MYSQLDUMP "mysqldump"
-#define WSREP_SST_SKIP "skip"
 #define WSREP_SST_ONLY_IST "ist_only"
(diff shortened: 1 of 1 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1626

venkatesh-prasad-v (approved):

PR #1789, opened 2023-04-24
(5.7) PXC-679: Undetected state gap discovery causes server hung on shutdown

kamil-holubicki:
PXC-679: Undetected state gap discovery causes server to hang on shutdown

https://jira.percona.com/browse/PXC-679

Problem:
When the server detects a state gap which was not detected at galera
level it aborts itself. It is OK however, during the shutdown the server
hangs.

Cause:
This is because we call unireg_abort() from wsrep_view_handler_cb
(view_cb_), which is done from wsrep applier context. The process is
waiting for wsrep applier thread to be finished and never exits.

Solution:
1. Return WSREP_CB_FAILURE from view callback instead of calling
unireg_abort(). The failure will be detected by galera and abort will
be called.
2. Prohibit setting wsrep_start_position when wsrep provider is loaded
and server is not starting.
Backported 8.0 commit a11616faa1b8781dd6d36e3c0d25097d2c5d52a0 idea.
However, note that the original logic implemented in this commit is
wrong and is in contradiction to the warning message displayed.
wsrep_start_position variable's purpose is to tell the Joiner its start
position after SST. It is used by mysqldump SST.
The condition which prohibits setting wsrep_start_position has been
fixed.
3. Fixed wsrep_start_position default value
(SET GLOBAL wsrep_start_position=default didn't work as nullptr was
the default value instead of 00000000-0000-0000-0000-000000000000:-1
4. wsrep_start_position_basic MTR test moved to wsrep suite and fixed.
5. Added wsrep suite to the list of default suites

### DIFF
--- a/sql/wsrep_var.cc
+++ b/sql/wsrep_var.cc
@@ -157,6 +157,19 @@ static int wsrep_start_position_verify (const char* start_str)
 @return false if no error encountered with check else return true. */
 bool wsrep_start_position_check (sys_var *self, THD* thd, set_var* var)
 {
+    /* Allow setting wsrep_start_position:
+    1. During SST - always (this is used by mysqldump sst)
+    2. After boot - only when wsrep_provider=none
+    In other words, prohibit wsrep_start_position setting
+    if we are after SST and wsrep provider is loaded. */
+  if (!wsrep_sst_in_progress() && (strcmp(wsrep_provider, WSREP_NONE)))
+  {
+    my_message(ER_UNKNOWN_ERROR,
+              "wsrep_start_position can be set during server boot"
+              " or before loading wsrep_provider", MYF(0));
+    return true;
+  }
+
   char start_pos_buf[FN_REFLEN];
 
   if ((! var->save_result.string_value.str) ||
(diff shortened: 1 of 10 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5463

kamil-holubicki:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5464
wsrep.wsrep_mysqld_safe is failing without this patch. Probably not detected before because wsrep suite was not added to default suites.

sql/wsrep_var.cc
```diff
@@ -157,6 +157,19 @@ static int wsrep_start_position_verify (const char* start_str)
 @return false if no error encountered with check else return true. */
 bool wsrep_start_position_check (sys_var *self, THD* thd, set_var* var)
 {
+    /* Allow setting wsrep_start_position:
+    1. During SST - always (this is used by mysqldump sst)
+    2. After boot - only when wsrep_provider=none
+    In other words, prohibit wsrep_start_position setting
+    if we are after SST and wsrep provider is loaded. */
+  if (!wsrep_sst_in_progress() && (strcmp(wsrep_provider, WSREP_NONE)))
```
venkatesh-prasad-v, line 165: Is it possible to have a testcase for setting wsrep_start_position when sst is in progress?

venkatesh-prasad-v: Or do we have it covered in the wsrep_sst_mysqldump tests?

kamil-holubicki: Yes, it is covered by wsrep_sst_mysqldump tests. Without being it writable during the SST, mysqldump-sst related tests fail.

venkatesh-prasad-v (approved):

PR #1792, opened 2023-05-04
PXC-3546: [MTR] Fix failing keyring_vault sute tests

voodam:
https://jira.percona.com/browse/PXC-3546

All tests had the same cause of failure: keyring_vault plugin installation requires keyring_vault_config variable with valid data in the file. Setting this variable at the time of the test run is not possible as the file is generated within the test itself. Restarting the server solved the problem.

This breaking change has been made intentionally in PXC-2135 (commit https://github.com/percona/percona-xtradb-cluster/commit/b58c1d13dfe056ec2e5b659a5a0f75bfa0258197). Before the change plugin was loaded without valid or any keyring_vault_config file, so we could SET that plugin variable.

Reasons for deletion:
- keyring_vault_timeout: relies on keyring_vault_config with invalid data, which is now impossible.


voodam:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1645/

voodam:
Jenkins is not configured to be used with the vault probably. I used Hashicorp on the local machine.

plugin/keyring_vault/tests/mtr/innodb_online_alter_encryption.test
```diff
@@ -20,10 +20,7 @@ call mtr.add_suppression("keyring_vault initialization failure");
 call mtr.add_suppression("A message intended for a client cannot be sent there as no client-session is attached");
 --enable_query_log
 
---let $keyring_restart_param = restart: $KEYRING_VAULT_PLUGIN_OPT $KEYRING_VAULT_PLUGIN_EARLY_LOAD --loose-keyring_vault_config=$vault_conf_file --innodb-sort-buffer-size=64k --innodb_encrypt_online_alter_logs=ON
-
---replace_result $MYSQLTEST_VARDIR MYSQLTEST_VARDIR
-eval SET @@global.keyring_vault_config="$vault_conf_file";
+--let $keyring_restart_param = restart: $KEYRING_VAULT_PLUGIN_OPT $KEYRING_VAULT_PLUGIN_EARLY_LOAD --keyring_vault_config=$vault_conf_file --innodb-sort-buffer-size=64k --innodb_encrypt_online_alter_logs=ON
```
kamil-holubicki, line 23: This test is inherited from PS. How it is that it works for PS?
https://github.com/percona/percona-server/blob/c676f75d020dd0a56716f9dd6c7526c15818cf74/plugin/keyring_vault/tests/mtr/innodb_online_alter_encryption.test#L25

voodam: Updated the commit message that explains, why the suite works in Percona Server.

PR #1794, opened 2023-05-08
(8.0) PXC-679: Undetected state gap discovery causes server hung on shutdown

kamil-holubicki:
https://jira.percona.com/browse/PXC-679

Problem:
The original problem which exists in 5.7 does not show up in 8.0, but
during the work on 5.7 fix several things have been found which has
to be adjusted.

1. Prohibit setting wsrep_start_position when wsrep provider is loaded
and server is not starting.
Commit https://github.com/percona/percona-xtradb-cluster/commit/a11616faa1b8781dd6d36e3c0d25097d2c5d52a0 introduced the idea.
However, note that the original logic implemented in this commit is
wrong and is in contradiction to the warning message displayed.
wsrep_start_position variable's purpose is to tell the Joiner its start
position after SST. It is used by mysqldump SST.
The condition which prohibits setting wsrep_start_position has been
fixed.
3. Fixed wsrep_start_position default value
(SET GLOBAL wsrep_start_position=default didn't work as nullptr was
the default value instead of 00000000-0000-0000-0000-000000000000:-1
4. wsrep_start_position_basic MTR test fixed and enabled.

### DIFF
--- a/mysql-test/collections/disabled.def
+++ b/mysql-test/collections/disabled.def
@@ -235,7 +235,6 @@ sys_vars.sql_mode_func                       : BUG#0 Differences in result likel
 #sys_vars.innodb_support_xa_func              : BUG#0 XA is not supported by PXC
 
 # wsrep
-wsrep.wsrep_start_position_basic          : BUG#0 This needs server to start in cluster mode with wsrep initialized.
 wsrep.wsrep_preordered_basic              : BUG#0 TC pending.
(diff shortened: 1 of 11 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1647/

Failures are present also on 8.0

venkatesh-prasad-v (approved):

PR #1796, opened 2023-05-09
PXC-4211: Server aborts on binary log rotation

kamil-holubicki:
https://jira.percona.com/browse/PXC-4211

Problem:
When the binlog is rotated because of reaching its maximum size the server aborts with assertion.

Cause:
Upstream commit 9b4809f2fc3bcdfb90c77a463912e133d36acc53 introduced the optimization which skips blocking operation of compressing GTID sets during storing them into GTIDs table during binlog rotation. Now GTIDs are stored as well, but compression is delegated to compress_gtid_table thread.
Before the upstream fix, compression and store were triggered by connection context, but actual work was done in context of clone_gtid_thread which is not wsrep_thread so wsrep commit hooks were not executed.
After the upstream fix GTIDs are stored in table from connection context. When this implicit transaction commits, commit hooks are executed causing assertion.

Solution:
Skip wsrep commit hooks execution for implicit internal transaction of storing GTIDS.

### DIFF
--- a/mysql-test/suite/wsrep/t/binlog_rotation.test
+++ b/mysql-test/suite/wsrep/t/binlog_rotation.test
@@ -1,0 +1,23 @@
+# The purpose of this test is to check of binlog rotation works
+# when WSREP is enabled.
+
+--source include/have_wsrep_provider.inc
+
+--let $statement = SHOW BINARY LOGS
+--let $column = Log_name
+--source include/get_row_count.inc
+--let $binlogs_count_start = $row_count
+
+CREATE TABLE t1(a INT PRIMARY KEY, c1 varchar(8192));
+INSERT INTO t1 VALUES (0, repeat('a', 8192));
+DROP TABLE t1;
+
+# Check that binlog rotation occured
+--let $statement = SHOW BINARY LOGS
+--let $column = Log_name
+--source include/get_row_count.inc
+--let $binlogs_count_end = $row_count
+
+--let $assert_text = Binlog rotation should have occured
+--let $assert_cond = "[SELECT $binlogs_count_start = $binlogs_count_end]" = 0
+--source include/assert.inc
--- a/mysql-test/suite/wsrep/r/binlog_rotation.result
+++ b/mysql-test/suite/wsrep/r/binlog_rotation.result
@@ -1,0 +1,4 @@
+CREATE TABLE t1(a INT PRIMARY KEY, c1 varchar(8192));
+INSERT INTO t1 VALUES (0, repeat('a', 8192));
+DROP TABLE t1;
+include/assert.inc [Binlog rotation should have occured]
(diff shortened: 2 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1657/

sql/rpl_gtid_persist.cc
```diff
@@ -411,8 +411,19 @@ int Gtid_table_persistor::save(const Gtid_set *gtid_set, bool compress) {
   ret = error = save(table, gtid_set);
 
 end:
+#ifdef WITH_WSREP
+  /* The following deinit() call causes gtid table close (and transaction
+   commit) As it is internal, implicit transaction, so we skip wsrep hooks. */
+  bool run_wsrep_commit_hooks_save = thd->run_wsrep_commit_hooks;
+  thd->run_wsrep_commit_hooks = false;
```
venkatesh-prasad-v: `table_access_ctx.deinit()` is called from multiple places in this file. We should probably skip running wsrep commit hooks in all places. Can we introduce a RAII class in `sql/thd_raii.h` to do the cleanup and use it inside `Gtid_table_access_context::deinit()` so that we cover all cases.

mysql-test/suite/wsrep/t/binlog_rotation.test
```diff
@@ -0,0 +1,35 @@
...
+CREATE TABLE t1 (a INT PRIMARY KEY AUTO_INCREMENT, b INT);
+INSERT INTO t1(b) VALUES (0);
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
+INSERT INTO t1(b) SELECT (b) FROM t1;
```
venkatesh-prasad-v: Can we just to the following to force binlog rotation?
```
CREATE TABLE t1(c1 varchar(8192));
SET GLOBAL binlog_cache_size = 4096;
--connect(con1,localhost,root,,)
BEGIN;
INSERT INTO t1 VALUES (repeat('a', 8192));
COMMIT;

venkatesh-prasad-v (changes requested):

sql/rpl_gtid_persist.cc
kamil-holubicki: I have reworked the fix. Now it relies on `is_operating_gtid_table_implicitly` variable.

mysql-test/suite/wsrep/t/binlog_rotation.test
kamil-holubicki: `binlog_cache_size` does not cause binlog rotation. I've replaced a series of `insert into ... select` with your proposal.

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1663/

venkatesh-prasad-v (approved):

PR #1797, opened 2023-05-09
PXC-4218: Truncated output of SHOW VARIABLES LIKE 'wsrep_provider_opt…

voodam:
…ions'

https://jira.percona.com/browse/PXC-4218

Problem:

Commands show truncated output:
- SHOW VARIABLES LIKE 'wsrep_provider_options';
- SELECT VARIABLE_VALUE FROM performance_schema.global/session/persisted_variables WHERE VARIABLE_NAME = 'wsrep_provider_options';

Cause:
Internal buffers and VARIABLE_VALUE size are insufficient for long strings like wsrep_provider_options.

Solution:
- Increase buffers size (4096 bytes).
- Put it in a separate constant.
- Fix session_variables and persisted_variables tables similar to global_variables.

Logic of the fix repeats the old commit:

977327de82e6,
GCF-775: extended VARIABLE_VALUE column in performance_schema.global_variables to have 2048 character length.


voodam:
https://pxc.cd.percona.com/view/PXC%205.7/job/pxc-5.7-pipeline/5516/

venkatesh-prasad-v (approved):

cmake/wsrep.cmake
```diff
@@ -29,6 +29,8 @@ ENDIF()
 # is moved to its own wsrep_version.cmake file.
 
 SET(WSREP_C_FLAGS   "-DWITH_WSREP -DWSREP_PROC_INFO -DMYSQL_MAX_VARIABLE_VALUE_LEN=4096")
+ADD_DEFINITIONS(-DWSREP_PS_VAR_VALUE_BUFF_SIZE=4096)
```
kamil-holubicki, line 35: Could you please add a comment here which sql files (and where) we need to change if we change the value here? (we've got back-link from sql to here, but the other way may be useful as well I think)

kamil-holubicki (approved):

cmake/wsrep.cmake
voodam: Done (I'll rebase 8.0 PR after approving this one)

kamil-holubicki (approved):

kamil-holubicki:
Let's wait for 5.7.43, 8.0.34 merges to have this

PR #1799, opened 2023-05-19
PXC-4225: Incorrect COMMAND value in I_S.PROCESSLIST for wsrep applie…

kamil-holubicki:
…r threads

https://jira.percona.com/browse/PXC-4225

Problem:
After DML executon or DDL execution after previouw DML, the COMMAND column in I_S.PROCESSLIST shows incorrect value
(Query, instead of Sleep)

Cause:
In normal flow thread's command state is maintained in sql_parse.cc::dispatch_command() functions. PXC flow for replicated events do not use this function, however command state is set to Query in Rows_log_event::do_apply_event() and it remains so.

Solution:
Set proper state of thread's command in Wsrep_applier_service methods called by Galera provider when DDL and DML events are applied.

### DIFF
--- a/sql/wsrep_high_priority_service.cc
+++ b/sql/wsrep_high_priority_service.cc
@@ -448,6 +448,7 @@ int Wsrep_high_priority_service::apply_toi(const wsrep::ws_meta &ws_meta,
            (long long)wsrep_thd_trx_seqno(m_thd));
   WSREP_DEBUG("%s", m_thd->wsrep_info);
   thd_proc_info(thd, m_thd->wsrep_info);
+  thd->set_command(COM_QUERY);
 
   WSREP_DEBUG("Wsrep_high_priority_service::apply_toi: %lld",
               client_state.toi_meta().seqno().get());
@@ -483,6 +484,7 @@ int Wsrep_high_priority_service::apply_toi(const wsrep::ws_meta &ws_meta,
            (long long)wsrep_thd_trx_seqno(m_thd));
   WSREP_DEBUG("%s", m_thd->wsrep_info);
   thd_proc_info(thd, m_thd->wsrep_info);
+  thd->set_command(COM_SLEEP);
 
   wsrep_wait_rollback_complete_and_acquire_ownership(m_thd);
   wsrep_set_SE_checkpoint(client_state.toi_meta().gtid());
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1677/

sql/wsrep_high_priority_service.cc
```diff
@@ -483,6 +484,7 @@ int Wsrep_high_priority_service::apply_toi(const wsrep::ws_meta &ws_meta,
            (long long)wsrep_thd_trx_seqno(m_thd));
   WSREP_DEBUG("%s", m_thd->wsrep_info);
   thd_proc_info(thd, m_thd->wsrep_info);
+  thd->set_command(COM_SLEEP);
```
venkatesh-prasad-v, line 487: Can you please check if the same fix is required for NBO as well?

kamil-holubicki: NBO was not affected in the way reported in the PXC-4225, but I improved thread state reporting for it.

sql/wsrep_high_priority_service.cc
```diff
@@ -860,6 +890,28 @@ int Wsrep_applier_service::apply_nbo_begin(const wsrep::ws_meta &ws_meta,
...
+  DBUG_EXECUTE_IF("wsrep_signal_applier_thread", {
+    const char act[] =
+        "now "
+        "SIGNAL apply_nbo_begin_entered_nbo_mode.reached "
+        "WAIT_FOR apply_nbo_begin_entered_nbo_mode.continue";
+    assert(!debug_sync_set_action(m_thd, STRING_WITH_LEN(act)));
+  };);
+
+  /* WSREP applier thread created NBO background worker
+     and is read for processing of next writeset. */
```
venkatesh-prasad-v:
```suggestion
     and is ready for processing next writesets. */
```

sql/wsrep_high_priority_service.cc
```diff
@@ -810,13 +830,23 @@ int Wsrep_applier_service::apply_nbo_begin(const wsrep::ws_meta &ws_meta,
 
     ret = trans_commit(thd);
 
+    /* This tread is going to be removed soon and will be not visible in
```
venkatesh-prasad-v:
```suggestion
    /* This thread is going to be removed soon and will be not visible in
```

venkatesh-prasad-v (changes requested):

sql/wsrep_high_priority_service.cc
kamil-holubicki: fixed

venkatesh-prasad-v (approved):

PR #1809, opened 2023-06-22
PXC-4221: Merge PS-8.0.33 (MTR tests fixed)

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4221

Fixed MTR testcases
    ***
    Reverted our implementation of handling expected server abort on mysqld
    restart in MTR testcases
    
    in favor of upstream's implementation by
    
    Bug#35005463 MTR: IMPLEMENT A WAY TO HANDLE SERVER ABORT ON MYSQLD RESTART
    (commit 3437378a6e)
    
    While our implementation used "try:" for the expected server aborts,
    upstream implementation used "restart_abort:". As per this change, all
    the dependent testcases were modified to use "restart_abort:".
    
    ***
    galera_performance_schema test failed when used with the
    libgalera_smm.so library built using cmake command. This is because of
    the uninitialized WITH_PERFSCHEMA_STORAGE_ENGINE variable resulted in
    galera to be built without perfschema instrumentation.
    
    In order to fix this, galera build order has been modified.
    
    ***
    Some tests failing due to master-source changes have been re-recorded.
    
    ***
    SST Joiner thread code has been modified to avoid calling
    unireg_abort(), as it was observed that it could result in both mysqld
    main thread and sst joiner thread simultaneously calling mysqld_exit()
    resulting in double free of some memory structures.
    (commit 06bfad84918 has been partially reverted.

### DIFF
--- a/sql/wsrep_sst.cc
+++ b/sql/wsrep_sst.cc
@@ -672,8 +672,7 @@ static void *sst_joiner_thread(void *a) {
     thd->variables.transaction_isolation = ISO_READ_COMMITTED;
 
     if (wsrep_sst_complete(thd, -err)) {
-      WSREP_ERROR("Failure while signalling the completion of SST. Aborting.");
-      unireg_abort(1);
+      WSREP_WARN("Failure while signalling the completion of SST.");
     }
 
     WSREP_SYSTEM("SST completed");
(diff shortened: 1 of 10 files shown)


sql/wsrep_sst.cc
```diff
@@ -671,10 +671,7 @@ static void *sst_joiner_thread(void *a) {
     /* Read committed isolation to avoid gap locking */
     thd->variables.transaction_isolation = ISO_READ_COMMITTED;
 
-    if (wsrep_sst_complete(thd, -err)) {
-      WSREP_ERROR("Failure while signalling the completion of SST. Aborting.");
-      unireg_abort(1);
-    }
+    wsrep_sst_complete(thd, -err);
```
kamil-holubicki: Maybe leaving error/warning message here is not a bad idea here?

kamil-holubicki (approved):

PR #1811, opened 2023-06-23
PXC-4233: Cluster state interruption during NBO may lead to permanent cluster lock

kamil-holubicki:
https://jira.percona.com/browse/PXC-4233

Problem:
DDL transaction (ALTER TABLE) executed in NBO mode can cause BF-BF
conflict with other DML transaction (INSERT) which tries to modify data
in the same table.
This is a known limitation that both cannot be executed at the same time
but the problem is that BF-BF causes all cluster nodes abort and cluster
outage.

Cause:
When NBO transaction is replicated two things happen:
1. On local node TOI transaction indicating start of NBO ends, releasing
Commit and Apply Monitors in Galera. This is end of NBO phase 1. Then
the DDL goes as normal transaction. At the end we wait for NBO to
finish (NBO phase 2 begin). At this moment we wait for all cluster nodes
to confirm completion of NBO execution. We still hold all MDL locks
related to the transaction
2. On remote node, applier thread just spawns NBO thread which is
responsible for applying all DDL changes. Then applier thread finishes
processing TOI transaction related to local node's NBO phase 1.
At this moment no MDL locks related to NBO are acquired.
Commit and Apply Order Monitors are released as well.

At the moment, when on remote node NBO applier didn't acuire any MDL
locks yet, but the NBO-triggering TOI finished, another DML transaction
modifying the same table is free to go (no MDL conflicts).
Certification flow was implemented in the way that it considered only
other TOI and NBO transactions, ignoring normal and streaming
replication transactions. So the DML transaction was replicated.

Back to local node. It waits for NBO confirmation from remote node.
Replicated DML transaction arrives and is being applied. Applier thread
is attempting to lock MDL lock, which is already held by local NBO
transaction which ends up with BF-BF abort.

Solution:
Modify certification flow in the way that it checks for conflicts
against ongoing NBO transactions also for regular and streaming
replication transactions. If the transaction conflicts with ongoing NBO
transaction, it fails with deadlock error (the original behavior for
TOI)

### DIFF
--- a/mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
+++ b/mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
@@ -1,0 +1,56 @@
+--source include/have_debug_sync.inc
+--source include/have_debug.inc
+--source include/galera_cluster.inc
+
+--connection node_1
+CREATE TABLE t1 (a INT PRIMARY KEY, b INT);
+INSERT INTO t1 VALUES (0, 0), (1, 0);
+CREATE TABLE aux (a INT PRIMARY KEY);
+
+# Setup node_2 to stop NBO applier thread just before apply_events()
+--connection node_2
+SET GLOBAL DEBUG="+d,nbo_stop_before_apply_events";
+
+--connection node_1
+SET SESSION wsrep_osu_method = NBO;
+--send CREATE INDEX b_idx ON t1(b)
+
+--connection node_2
+SET DEBUG_SYNC = 'now WAIT_FOR nbo_before_apply_events_reached';
+
+# make sure NBO phase 1 finished
+INSERT INTO aux VALUES (0);
+
+# There is nothing which blocks the following transaction:
+# 1. Apply and Commit Monitors are released (NBO phase 1 finished)
+# 2. No MDL locks acquired by NBO applier thread yet
+# It should be be rejected at the certification stage. If not, it would replicate and cause BF-BF conflict on node 1
+--error ER_LOCK_DEADLOCK
+INSERT INTO t1 VALUES (2, 2);
+
+# Try with TOI
+CALL mtr.add_suppression(".*TO isolation failed.*");
+--error ER_LOCK_DEADLOCK
+ALTER TABLE t1 ADD COLUMN (c INT);
+
+# Try with SR
+SET SESSION wsrep_trx_fragment_size=1;
+SET SESSION wsrep_trx_fragment_unit="STATEMENTS";
+BEGIN;
+--error ER_LOCK_DEADLOCK
+INSERT INTO t1 VALUES (2, 2);
+ROLLBACK;
+SET SESSION wsrep_trx_fragment_size=default;
+SET SESSION wsrep_trx_fragment_unit=default;
+
+# Let NBO applier thread to continue
+SET DEBUG_SYNC = 'now SIGNAL nbo_before_apply_events_continue';
+
+--connection node_1
+--reap
+SET SESSION wsrep_osu_method=TOI;
+DROP TABLE t1;
+DROP TABLE aux;
+
+--connection node_2
+SET GLOBAL DEBUG="-d,nbo_stop_before_apply_events";
--- a/sql/service_wsrep.cc
+++ b/sql/service_wsrep.cc
@@ -112,7 +112,7 @@ extern "C" bool wsrep_thd_is_BF(const THD *thd, bool sync) {
     if (sync)
       mysql_mutex_lock(const_cast<mysql_mutex_t *>(&thd->LOCK_wsrep_thd));
     status = (wsrep_thd_is_applying(thd) || wsrep_thd_is_toi(thd) ||
-              wsrep_thd_is_in_rsu(thd));
+              wsrep_thd_is_in_rsu(thd)) || wsrep_thd_is_in_nbo(thd);
     if (sync)
       mysql_mutex_unlock(const_cast<mysql_mutex_t *>(&thd->LOCK_wsrep_thd));
   }
--- a/sql/wsrep_high_priority_service.cc
+++ b/sql/wsrep_high_priority_service.cc
@@ -30,6 +30,7 @@
 #include "transaction.h"
 
 #include <condition_variable>
+#include "my_dbug.h"
 #include "sql_base.h"  // close_temporary_table()
 
 extern handlerton *binlog_hton;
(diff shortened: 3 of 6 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1823/

sql/wsrep_high_priority_service.cc
```diff
@@ -809,6 +810,13 @@ int Wsrep_applier_service::apply_nbo_begin(const wsrep::ws_meta &ws_meta,
     Avoid over-writting of this XID by MySQL XID */
     thd->get_transaction()->xid_state()->get_xid()->set_keep_wsrep_xid(true);
 
+    DBUG_EXECUTE_IF("nbo_stop_before_apply_events", {
+      const char act[] =
+          "now signal nbo_before_apply_events_reached wait_for "
+          "nbo_before_apply_events_continue";
+      assert(!debug_sync_set_action(current_thd, STRING_WITH_LEN(act)));
```
venkatesh-prasad-v: Can it be done using just the `thd`?
```suggestion
      assert(!debug_sync_set_action(thd, STRING_WITH_LEN(act)));
```

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
```diff
@@ -0,0 +1,41 @@
...
+
+--connection node_2
+SET DEBUG_SYNC = 'now WAIT_FOR nbo_before_apply_events_reached';
+
+# make sure NBO phase 1 finished
+INSERT INTO aux VALUES (0);
+
+# There is nothing which blocks the following transaction:
+# 1. Apply and Commit Monitors are released (NBO phase 1 finished)
+# 2. No MDL locks acuquired by NBO applier thread yet
```
venkatesh-prasad-v:
```suggestion
# 2. No MDL locks acquired by NBO applier thread yet
```

sql/wsrep_binlog.cc
```diff
@@ -430,6 +430,8 @@ void wsrep_register_for_group_commit(THD *thd) {
 }
 
 bool wsrep_implicit_transaction(THD *thd) {
+  if (!thd) return false;
```
venkatesh-prasad-v: I think we already have this check in the `release-8.0.33` branch. Please revert this check and change the target branch of this PR to `release-8.0.33`.

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
```diff
@@ -0,0 +1,41 @@
+--source include/galera_cluster.inc
+--source include/have_debug_sync.inc
+--source include/have_debug.inc
```
venkatesh-prasad-v: It is better to have the restrictive inc files at the beginning of the test cases.
```suggestion
--source include/have_debug_sync.inc
--source include/have_debug.inc
--source include/galera_cluster.inc
```

sql/wsrep_high_priority_service.cc
kamil-holubicki: done

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
kamil-holubicki: done

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
kamil-holubicki: done

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1896/

sql/wsrep_mysqld.cc
```diff
@@ -3013,8 +3013,29 @@ bool wsrep_handle_mdl_conflict(const MDL_context *requestor_ctx,
     ticket->wsrep_report(wsrep_debug);
 
     mysql_mutex_lock(&granted_thd->LOCK_wsrep_thd);
-    if (wsrep_thd_is_toi(granted_thd) || wsrep_thd_is_in_nbo(granted_thd) ||
-        wsrep_thd_is_applying(granted_thd)) {
+    if (wsrep_thd_is_in_nbo(granted_thd) &&
+        wsrep_thd_is_applying(request_thd)) {
+      /* Here we've got the situation when NBO is in progress and applier tires
```
venkatesh-prasad-v: Probably you meant `tries`
```suggestion
      /* Here we've got the situation when NBO is in progress and applier tries
```

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
```diff
@@ -0,0 +1,41 @@
...
+
+# make sure NBO phase 1 finished
+INSERT INTO aux VALUES (0);
+
+# There is nothing which blocks the following transaction:
+# 1. Apply and Commit Monitors are released (NBO phase 1 finished)
+# 2. No MDL locks acquired by NBO applier thread yet
+# It should be be rejected at the certification stage. If not, it would replicate and cause BF-BF conflict on node 1
+--error ER_LOCK_DEADLOCK
+INSERT INTO t1 VALUES (2, 2);
```
venkatesh-prasad-v, line 29: Can you please extend the test to verify that we also take other TOI,NBO, SR into considertation during certification?

sql/wsrep_mysqld.cc
```diff
@@ -3013,8 +3013,29 @@ bool wsrep_handle_mdl_conflict(const MDL_context *requestor_ctx,
...
+         but not yet released MDL locks. Ideally, Galera should clean NBO
+         certification index during finishing TOI related to 2nd phase of NBO,
+         which happens after releasing all MDL locks, but it is as it is.
+         Instead of rewriting that part in Galera, we will block applier here
+         for the time of unblocking MDL locks by NBO thread. */
+      WSREP_MDL_LOG(
+          INFO,
+          "MDL conflict, NBO vs applier. Waiting for NBO to release MDL locks",
+          schema, schema_len, request_thd, granted_thd);
+      mysql_mutex_unlock(&granted_thd->LOCK_wsrep_thd);
```
venkatesh-prasad-v, line 3034: Not related to the bugfix, but as an improvement can we have a RAII object for the automatic release of `granted_thd->LOCK_wsrep_thd`? With this, we can avoid multiple `unlock()` calls in this function.

May be using `Mutex_lock` from `mutex_lock.h` will be useful here.

sql/service_wsrep.cc
```diff
@@ -112,7 +112,8 @@ extern "C" bool wsrep_thd_is_BF(const THD *thd, bool sync) {
     if (sync)
       mysql_mutex_lock(const_cast<mysql_mutex_t *>(&thd->LOCK_wsrep_thd));
     status = (wsrep_thd_is_applying(thd) || wsrep_thd_is_toi(thd) ||
-              wsrep_thd_is_in_rsu(thd));
+              wsrep_thd_is_in_rsu(thd)) ||
+             wsrep_thd_is_in_nbo(thd);
```
venkatesh-prasad-v: Why not
```suggestion
              wsrep_thd_is_in_rsu(thd) || wsrep_thd_is_in_nbo(thd));
```

sql/wsrep_mysqld.cc
kamil-holubicki: yup. done

sql/wsrep_mysqld.cc
kamil-holubicki: I'm not sure RAII would help here. Please note that it would be a problem with proper scoping. Sometimes we unlock and do something more, sometimes not...

mysql-test/suite/galera_nbo/t/galera_nbo_alter_bf3.test
kamil-holubicki: done

venkatesh-prasad-v (approved):
Thank you for fixing this issue. Great work!  Approved.

PR #1820, opened 2023-07-18
PXC-4268: ALTER DEFINER VIEW makes PXC node Disconnected and Inconsistent

kamil-holubicki:
https://jira.percona.com/browse/PXC-4268

Problem:
When ALTER DEFINER VIEW is executed by the user with insufficient privileges the query fails and the node self-leaves the cluster.

Cause:
This is yet another case when replicated TOI is allowed to proceed on the remote node, because it is executed in context of wsrep applier thread (root user), while it fails on local node. This is caused because TOI is replicated before checking privileges locally.

Solution:
Postponed TOI replication up to the point when privileges of the user executing query are validated.


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/1909/

venkatesh-prasad-v (approved):

PR #1823, opened 2023-07-24
PXC-4258: Failed to add foreign key results in Inconsistency

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4258

Problem
-------
ALTER TABLE on a table involving a FK relationship can sometimes result in cluster inconsistency due to different error messages.

Different messages are possible when the error message includes the name of the intermediate temporary table that gets created during the execution of the ALTER TABLE.

  Example:

  node-1: foreign key constraint fails (`test`.`#sql-25528d_2`, CONSTRAINT `child_ibfk_1`
  node-2: foreign key constraint fails (`test`.`#sql-22ed07_9`, CONSTRAINT `child_ibfk_1`

Solution
--------
Avoid printing the temporary table name and instead print "temp_table" to the error message for TOI and NBO operations.

Testing Done
----
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/1916/console

Failing tests
| Sl.No | Test | Remarks |
|--------|--------|--------|
| 1 | main.derived_condition_pushdown | Fixed in 8.0.33 |
| 2 | galera.galera_fc_auto_evict | Fixed in 8.0.33 |
| 3 | galera.galera_performance_schema | Fixed in 8.0.33 |
| 4 | galera.pxc_information_schema | Fixed in 8.0.33 |
| 5 | galera_sr.GCF-1043A | Fixed in 8.0.33 |
| 6 | galera_sr.gakera_sr_wsrep_error_handling | Test removed in 8.0.33 |
| 7 | galera_encryption.gcache_mk_rotation_enc_off | Sporadic |
| 7 | galera.galera_gcache_recover_manytrx | Sporadic |

### DIFF
--- a/mysql-test/include/mtr_warnings.sql
+++ b/mysql-test/include/mtr_warnings.sql
@@ -426,6 +426,7 @@ INSERT INTO global_suppressions VALUES
  ("Toggling wsrep_on to OFF will affect sql_log_bin"),
  ("Toggling wsrep_on to ON will affect sql_log_bin"),
  ("InnoDB High Priority being used"),
+ ("Query apply failed"),
 
  /* MySQL supression needed by Galera */
  ("Slave I/O.*: Get master clock failed with error:.*"),
--- a/mysql-test/suite/galera/r/pxc_fk_inconsistency_voting.result
+++ b/mysql-test/suite/galera/r/pxc_fk_inconsistency_voting.result
@@ -1,0 +1,14 @@
+CREATE TABLE parent (i INT PRIMARY KEY) ENGINE = InnoDB;
+CREATE TABLE child (j INT PRIMARY KEY) ENGINE = InnoDB;
+INSERT INTO child VALUES (10);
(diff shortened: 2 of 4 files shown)


mysql-test/r/venki_test.result
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE t2(c2 INT PRIMARY KEY);
```
kamil-holubicki: I think we don't need this file :)

kamil-holubicki (changes requested):

mysql-test/r/venki_test.result
venkatesh-prasad-v: Yeah. It must be a side effect of `git add mysql-test`. I will revert this change.

kamil-holubicki (approved):

PR #1826, opened 2023-07-25
PXC-4278: Renaming table crashes with NBO

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4278

Problem: ALTER TABLE RENAME command asserts with

  In wsrep::client_state::begin_nbo_phase_two()
  Assertion 'toi_mode_ == m_undefined' failed

Solution: Fixed the missing call to end_nbo_phase_one() during the ALTER TABLE RENAME command.

### DIFF
--- a/sql/sql_table.cc
+++ b/sql/sql_table.cc
@@ -16012,6 +16012,14 @@ static bool simple_rename_or_index_change(
     // It's now safe to take the table level lock.
     if (lock_tables(thd, table_list, alter_ctx->tables_opened, 0)) return true;
 
+#ifdef WITH_WSREP
+    /*
+      Locks are acquired for ALTER TABLE .. ENABLE | DISABLE KEYS
+      It is time to call nbo_phase_one_end().
+    */
+    WSREP_NBO_1ST_PHASE_END;
+#endif /* WITH_WSREP */
+
     if (keys_onoff == Alter_info::ENABLE) {
       DEBUG_SYNC(thd, "alter_table_enable_indexes");
       DBUG_EXECUTE_IF("sleep_alter_enable_indexes", my_sleep(6000000););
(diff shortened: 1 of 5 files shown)


kamil-holubicki (commented):
Could we add MTR test for this?

sql/sql_table.cc
```diff
@@ -17310,8 +17310,15 @@ bool mysql_alter_table(THD *thd, const char *new_db, const char *new_name,
                "LOCK=EXCLUSIVE");
       return true;
     }
+#ifdef WITH_WSREP
+    bool res = simple_rename_or_index_change(
+        thd, *new_schema, table_list, alter_info->keys_onoff, &alter_ctx);
+    WSREP_NBO_1ST_PHASE_END;
```
kamil-holubicki: In my opinion we should finish 1st phase just before calling mysql_rename_table here
https://github.com/percona/percona-xtradb-cluster/blob/065e70155dc60f6811e68e2a16ec394e04908db7/sql/sql_table.cc#L16137

In the way like it is implemented right now, it is in fact regular TOI

sql/sql_table.cc
```diff
@@ -16012,6 +16012,14 @@ static bool simple_rename_or_index_change(
     // It's now safe to take the table level lock.
     if (lock_tables(thd, table_list, alter_ctx->tables_opened, 0)) return true;
 
+#ifdef WITH_WSREP
+    /*
+      Locks are acquired for ALTER TABLE .. ENABLE | DISABLE KEYS
+      It is time to call nbo_phase_one_end().
+    */
+    WSREP_NBO_1ST_PHASE_END;
```
kamil-holubicki, line 16020: Won't we hit WSREP_NBO_1ST_PHASE_END twice?

`alter table t2 enable keys, rename to t1;`

venkatesh-prasad-v: Hi! Yes. It is possible that it calls `WSREP_NBO_1ST_PHASE_END` twice for query like `alter table t1 rename to t2, enable keys;`

But the first call to `nbo_end_phase_one()` changes the toi_mode to `m_undefined`.

So the second call to `WSREP_NBO_1ST_PHASE_END` just becomes a no-op, as the check if `(!wrep_thd_is_local_nbo(thd)) return;` in `wsrep_NBO_end_phase_one()`  performs an early return and we are safe.

kamil-holubicki (approved):

PR #1829, opened 2023-08-01
PXC-4284: GRANT without existing user leads to disconnected PXC node

kamil-holubicki:
https://jira.percona.com/browse/PXC-4284

Problem:
When GRANT statement related to nonexisting user was executed on server started with --skip-name-resolve option, the node was kicked-off from the cluster by inconsistency voting algorithm.

Cause:
Inconsistency voting uses hashed error messages for voting. In case of --skip-name-resolve additional warning message is added however, it was skipped for applier thread.

Solution:
1. Do not skip --skip-name-resolve part of message for of applier thread
2. Improved MTR scripts. Now it is possible to use socket connections.

### DIFF
--- a/mysql-test/include/galera_connect.inc
+++ b/mysql-test/include/galera_connect.inc
@@ -17,6 +17,9 @@
 # $galera_debug
 #   Print debug information.
 #
+# $galera_use_socket_client_connection
+#   Use socket connection instead of TCP/IP
+#
 
 if (!$galera_connection_name)
 {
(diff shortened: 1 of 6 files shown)


mysql-test/include/galera_connect.inc
```diff
@@ -28,18 +28,37 @@ if (!$galera_server_number)
   --die ERROR IN TEST: $galera_server_number must be set before sourcing include/galera_connect.inc
 }
 
---let $_galera_port= \$NODE_MYPORT_$galera_server_number
-if (!$_galera_port)
+
+--let $_galera_port=
+--let $_host=
+if (!$galera_use_socket_client_connection)
```
venkatesh-prasad-v, line 37: Can you add description about `galera_use_socket_client_connection` in the header of the `galera_cluster.inc` so that others will be aware of what to set to have connections use sockets?

kamil-holubicki: Added to `galera_connect.inc`

venkatesh-prasad-v (approved):

PR #1830, opened 2023-08-01
PXC-4219: Deadlock with concurrent START SLAVE and backup using PXB

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4219

Problem
-------
Server could enter into a deadlock when START SLAVE is issued immediately when PXB process starts to take a backup of the server.

Deadlock Summary
----------------
This is a 3 way deadlock

1. SQL thread holding rli->run_lock and is waiting for s_synced inside wait_until_state(). This will return only on successful SST.

   But SST is stuck at waiting for results from log_status query.

2. START SLAVE thread waiting for rli->run_lock(lock_cond_sql) in start_slave_thread() which is held by SQL thread.

3. A query on performance_schema.log_status which triggered by SST is waiting for channel_map_lock->wrlock() but it is held by START SLAVE thread.

i.e,
1. SQL thread

        holds: rli->run_lock acquired at line no 7128 in rpl_replica.cc

        waits for: SST/PXB

2. START SLAVE: holds: 1. mi->run_lock, this was reacquired after starting IO thread, usually released in unlock_slave_threads() in the end of start_slave(). 2. channel_map_lock held in start_slave_cmd() line no 747 in rpl_replica.cc

        waits for: rli->run_lock, held by SQL thread

3. SST/PXB:

        holds: nothing

        waits for: channel_map_lock, held by START SLAVE

Solution
--------
The server now performs galera synchronization (i.e, waits for the server to reach 'Synced' state before applying) after worker threads are started and before the events are read from the relay logs, to ensure that other threads get the required locks necessary for SST.

Testing Done
----
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/145/console

### DIFF
--- a/mysql-test/suite/galera/t/pxc_start_replica_during_sst_donate.test
+++ b/mysql-test/suite/galera/t/pxc_start_replica_during_sst_donate.test
@@ -1,0 +1,112 @@
+# ==== Purpose ====
+#
+# This test verifies that starting async replication while proccessing SST
+# request are handled properly by the server.
+#
+# ==== Implementation ====
+#
+# 1. Create a two node cluster.
+# 2. Shutdown node2 and remove the grastate.dat file.
+# 3. Execute some transactions on node1 so that node2 forces an SST when it
+#    joins next time.
+# 4. Add a debug point on node1 to halt just before donating state transfer.
+# 5. Start node2 in background and wait till node1 stops before donating SST
+# 6. Configure replication parameters and execute START REPLICA.
+# 7. Shutdown node2 started as a background process and restart
+# 8. Cleanup
+#
+# ==== References ====
+#
+# PXC-4219: Deadlock with concurrent START SLAVE and backup using PXB
+#
+
+
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+
+--echo #
+--echo # 1. Create a two node cluster.
+--source include/galera_cluster.inc
+
+--echo #
+--echo # 2. Shutdown node2 and remove the grastate.dat file.
+--connection node_2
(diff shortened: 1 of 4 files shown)


mysql-test/suite/galera/t/pxc_start_replica_during_sst_donate.test
```diff
@@ -0,0 +1,112 @@
...
+--echo #
+--echo # 3. Execute some transactions on node1 so that node2 forces an SST when it
+--echo #    joins next time.
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY) ENGINE=InnoDB;
+INSERT INTO t1 VALUES(1);
+INSERT INTO t1 VALUES(2);
+
+--echo #
+--echo # 4. Add a debug point on node1 to halt just before donating state transfer.
+--let $debug_point= halt_before_sst_donate
```
kamil-holubicki, line 53: The whole file: Inconsistent spaces around `=`

kamil-holubicki (commented):
In commit message:
`... required logs necessary for SST.`
Shouldn't it be locks instead of logs?

mysql-test/suite/galera/t/pxc_start_replica_during_sst_donate.test
venkatesh-prasad-v: Done. Used `= ` for `--let` mtr assignments and `  =  ` for SQL statements. And commit message was also updated.

kamil-holubicki (approved):

PR #1832, opened 2023-08-07
PXC-4238: UDF execution adds Errant GTID

venkatesh-prasad-v:
Problem
-------
Running asynchronous connection failover UDFs like asynchronous_connection_failover_add_source() or
asynchronous_connection_failover_delete_source(), added an errant GTID in the binary log that prevented a failover.

In debug builds it resulted in an assertion failure.

Analysis
--------
During the execution of the async connection failover UDFs, the server thread tries to generate writeset info from InnoDB about the newly added row to the mysql.replication_asynchronous_connection_failover table. Since the thread is marked as a wsrep thread internally, it expects the wsrep transaction state to be active.

Solution
--------
All such implicit writes to replication system tables have been marked as non-wsrep transactions so that they dont generate wsrep writesets from InnoDB.

Testing Done
-----
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/146/


mysql-test/suite/galera/t/pxc_async_conn_failover.test
```diff
@@ -0,0 +1,369 @@
...
+--source include/big_test.inc
+--source include/have_wsrep_provider.inc
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+
+--echo #
+--echo # 1. Create a chain replication between server 1, 2, 3 and 4.
+--echo #.
+--let $rpl_skip_reset_master_and_slave= 1
+--let $rpl_server_count= 4
```
kamil-holubicki, line 62: The whole file: Inconsistent spaces around `=`

kamil-holubicki (approved):

mysql-test/suite/galera/t/pxc_async_conn_failover.test
venkatesh-prasad-v: Done. Used `= ` for --let mtr assignments and `  =  ` for SQL statements.

PR #1838, opened 2023-08-31
PXC-4280: Add support for pxb 8.0.34 (ZSTD compression by default)

kamil-holubicki:
https://jira.percona.com/browse/PXC-4280

PXB 8.0.34 removed qpress compression and added ZSTD as the default way of compression.

SST script already supports qpress decompression. It is implemented externally to PXB (probably historical reason).
On the other hand, PXB is is able to automatically detect decompression method if called with --decompress option, but next versions of PXB will remove quicklz support.
Having said above, to be compatible with old donors, we need to leave the existing qpress decompression implementation.
New methods of compression will be handled by additional PXB call for automatic decompression. This step is executed always, because in case of not compressed backup, PXB just does nothing.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -1652,6 +1652,11 @@ function initialize_pxb_commands()
     xb_version=${xb_version# }
     wsrep_log_debug "pxb-version:$xb_version"
 
+    local decompress_threads=$(grep -c ^processor /proc/cpuinfo)
+    if [[ $decompress_threads -gt 1 ]]; then
+        decompress_threads=$((decompress_threads/2))
+    fi
+
     #
     # Use syslogger (for logging) if configured else continue
     # to use file-based logging.
@@ -1705,6 +1710,9 @@ function initialize_pxb_commands()
                 \$encrypt_backup_options --stream=\$sfmt \
                 --xtrabackup-plugin-dir="$pxb_plugin_dir" \
                 --target-dir=\$itmpdir 2> >(logger -p daemon.err -t ${ssystag}innobackupex-backup)"
+            INNODECOMPRESS="${pxb_bin_path} $disver --parallel=${decompress_threads} --decompress --remove-original \
+                --target-dir=\${DATA} 2>&1 | logger -p daemon.err -t ${ssystag}innobackupex-decompress "
+
         fi
     else
         # prepare doesn't look at my.cnf instead it has its own generated backup-my.cnf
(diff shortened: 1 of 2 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2031

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1705,6 +1710,9 @@ function initialize_pxb_commands()
                 \$encrypt_backup_options --stream=\$sfmt \
                 --xtrabackup-plugin-dir="$pxb_plugin_dir" \
                 --target-dir=\$itmpdir 2> >(logger -p daemon.err -t ${ssystag}innobackupex-backup)"
+            INNODECOMPRESS="${pxb_bin_path} $disver --parallel=${decompress_threads} --decompress --remove-original \
```
satya-bodapati, line 1713: if decompress fails, do we retry the operation on the same directory?

kamil-holubicki: No, if decompress fails, SST script fails (line 2552) and the node fails to join.
Then it is restarted, SST starts again, the old directory is removed, then downloaded, and so on

satya-bodapati: Thanks. OK to add --remove-original :)

satya-bodapati: do you have any PXC test with the logs? showing SST joiner able to decompress zstd backups from donor

altmannmarcelo (commented):
Changes looks good, I just want to understand how --decompress worked in the past and why we need extra command now. xtrabackup --decompress should be identical for .qp/.lz4/.zst

kamil-holubicki:
@altmannmarcelo If you look a few lines above, qp decompression is handled by SST script directly (no pxb --decompress call at all). Why? I have no idea. Historical reason?
Before my fix, there was no --decompress, only this SST-like qp decompression. So on the donor side, only default (qpress) compression could have been done.
As I said in the commit comment, I decided to leave qpress decompression as it is, because it is planned to be completely removed from PXB (--decompress will not be able to handle qpress)

altmannmarcelo:
@kamil-holubicki got it, it was done manually.

With regards to --decompress - FYI, we are not removing it on 8.0, only 8.1+.

altmannmarcelo (approved):
LGTM

scripts/wsrep_sst_xtrabackup-v2.sh
kamil-holubicki: Good point. I have enabled compression in galera_sst_xtrabackup-v2 MTR test.

venkatesh-prasad-v (approved):

PR #1840, opened 2023-09-27
PXC-4302: GRANT statement may be replicated in a wrong way if partial…

kamil-holubicki:
…_revokes=1

https://jira.percona.com/browse/PXC-4302

Problem:
If partial_revokes is enabled and the user executes on node_1 GRANT statement without explicit AS clause, the binlog on node_2 contains wrong event having "AS 'skip-grants user'@'skip-grants host'" which may break async replication if node_2 is a source node for such a chain.

Cause:
If partial_revokes is enabled, AS clause is not explicitly specified and the GRANT is on global (*.*) level, node_1 rewrites the query just before writing the event into the binlog, adding AS clause with the current user. This causes async replica node to explicitly see AS clause and not doing any rewrite (replica thread works in root user context)

However, we replicate TOI with the query being "as is" before local execution. It triggers the rewrite path on replica node, causing event with wrong AS clause being stored in its binlog.

Solution:
Rewrite query on source node for TOI replication in the same way as it is rewritten before storing into the binlog.

### DIFF
--- a/mysql-test/suite/galera/t/galera_account_management3.test
+++ b/mysql-test/suite/galera/t/galera_account_management3.test
@@ -1,0 +1,66 @@
+#
+# Test the account management statements - GRANT, REVOKE with partial_revokes enabled
+#
+
+
+--source include/galera_cluster.inc
+--source include/have_log_bin.inc
+--source include/count_sessions.inc
+
+#
+# CREATE USER
+#
+--connection node_1
+CREATE ROLE 'reader', 'writer';
+GRANT SELECT ON *.* TO 'reader';
+GRANT INSERT, UPDATE, DELETE ON *.* TO 'writer';
+
+CREATE USER 'user1'@'%' IDENTIFIED WITH mysql_native_password BY 'passwd';
+CREATE USER 'user2'@'%' IDENTIFIED WITH mysql_native_password BY 'passwd';
+
+GRANT ALL PRIVILEGES ON *.* TO 'user1'@'%' WITH GRANT OPTION ;
+GRANT 'reader', 'writer' TO 'user2'@'%';
+GRANT USAGE ON *.* TO 'user2'@'%' WITH GRANT OPTION;
+
+--connect node_1a, 127.0.0.1, user1, passwd, test, $NODE_MYPORT_1
+--connection node_1a
+GRANT ALL PRIVILEGES ON *.* TO 'user2'@'%' WITH GRANT OPTION ;
+--disconnect node_1a
+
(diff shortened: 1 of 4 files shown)


mysql-test/suite/galera/t/galera_account_management3.test
```diff
@@ -0,0 +1,62 @@
...
+--let $binlog_limit = 1
+--source include/show_binlog_events.inc
+
+--disconnect node_1b
+
+--connection node_2
+--let $binlog_file = mysqld-bin.000004
+--let $binlog_start = 2298
+--let $binlog_limit = 1
+--source include/show_binlog_events.inc
```
venkatesh-prasad-v: Can we instead make use of the `assert_binlog_events.inc` to verify that the rewritten query is logged in the binary log?
```
--let $event_sequence= Anonymous_Gtid # Query/.*GRANT ALL.*TO 'user2'.* WITH GRANT OPTION AS 'user1'.* # Anonymous_Gtid # Query/.*GRANT ALL.*TO 'user1'.* WITH GRANT OPTION AS 'user2'.*                                        
--source include/assert_binlog_events.inc 
```

sql/wsrep_mysqld.cc
```diff
@@ -2228,6 +2305,9 @@ static int wsrep_TOI_event_buf(THD *thd, uchar **buf, size_t *buf_len) {
     case SQLCOM_DROP_TABLE:
       err = wsrep_drop_table_query(thd, buf, buf_len);
       break;
+    case SQLCOM_GRANT:
```
venkatesh-prasad-v, line 2308: Does this bug affect REVOKE queries as well?

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2069/

sql/wsrep_mysqld.cc
kamil-holubicki: `mysql_grant` rewrites query only for GRANT. 
REVOKE does not have AS clause, nothing about implicit curren_user, so not affected.

mysql-test/suite/galera/t/galera_account_management3.test
kamil-holubicki: done

venkatesh-prasad-v (approved):
LGTM

kamil-holubicki:
Galera only: https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2129/

PR #1843, opened 2023-09-28
PXC-4255: Running ALTER USER/ SET PASSWORD and FLUSH PRIVILEGES stall…

kamil-holubicki:
… the PXC cluster

https://jira.percona.com/browse/PXC-4255

Problem:
During execution of SET PASSWORD / ALTER USER  on node_1 and simultanous execution of FLUSH PRIVILEGES on node_2 the whole cluster stuck in deadlock.

Cause:
Commits 487328e7 (PXC-3872) and a40b33f8 (PXC-3999) postponed starting TOI after open_grant_tables() call. In such a case some MDL locks are already acquired when TOI is being started.
Node_1:
1. Acquire MDLs
2. Replicate TOI
3. Acquire CommitOrder monitor in Galera

If just before 2. node_2 executes FLUSH PRIVILEGES (another TOI), it will be ordered before node_1's TOI, so node_1 will wait in 3 for replicated TOI to finish. If replicated TOI conflicts with local TOI it will try to abort local TOI, however Galera does not have a concept of aborting TOI, so node_1 still waits for CommitOrder monitor to enter.

Solution:
Avoid acquiring MDL locks before starting TOI. This is in contradiction to what we fixed in the above-mentioned commits, because we need to acquire ACL cache lock to validate query before TOI. On the other hand even original code (before above-mentioned fixes suffered the same problem, because MDL locks were acquired before TOI inside open_grant_tables() when
acl_tables_setup_for_write_and_acquire_mdl() was called.

1. Revoke fixes from 487328e7 and a40b33f8.
2. Move query validation before starting TOI, release all locks before TOI. This is kind of trade-off, because in the time without lock things can potentially change, but I don't know if we could do any better.
3. Start TOI before acl_tables_setup_for_write_and_acquire_mdl().
4. Revoke fix from faad2e23 as it is not needed anymore.

### DIFF
--- a/sql/auth/sql_user_table.cc
+++ b/sql/auth/sql_user_table.cc
@@ -1971,7 +1971,14 @@ static bool acl_tables_setup_for_write_and_acquire_mdl(THD *thd,
          this function.
 */
 
-int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables) {
+#ifdef WITH_WSREP
+int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables,
+                      const char *db [[maybe_unused]],
+                      const char *table [[maybe_unused]]) {
+#else
+int open_grant_tables(THD *thd, TABLE_LIST *tables,
+                      bool *transactional_tables) {
+#endif /* WITH_WSREP */
   DBUG_TRACE;
 
   if (!initialized) {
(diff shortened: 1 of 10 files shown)


venkatesh-prasad-v (approved):

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2126/

kamil-holubicki:
Fixed MTR failures:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2128/

sql/auth/sql_user_table.cc
```diff
@@ -2060,14 +2060,17 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
 #ifdef WITH_WSREP
 bool wsrep_check_system_user_privilege(THD *thd,
                                        const List<LEX_USER> &user_list) {
-  Acl_cache_lock_guard acl_cache_lock(thd, Acl_cache_lock_mode::READ_MODE);
 
-  if (!acl_cache_lock.lock()) {
-    return true;
-  }
+  if (WSREP(thd) && !thd->wsrep_applier) {
+    Acl_cache_lock_guard acl_cache_lock(thd, Acl_cache_lock_mode::WRITE_MODE);
```
venkatesh-prasad-v: Taking the lock in write mode may not be necessary here as read mode will just do the job.

sql/auth/sql_user_table.cc
```diff
@@ -1971,7 +1971,14 @@ static bool acl_tables_setup_for_write_and_acquire_mdl(THD *thd,
          this function.
 */
 
-int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables) {
+#ifdef WITH_WSREP
+int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables,
+                      const char *db __attribute_maybe_unused__,
+                      const char *table __attribute_maybe_unused__) {
```
venkatesh-prasad-v: Use the C++17 attribute `[[maybe_unused]]` instead of `__attribute_maybe_unused__`
```suggestion
                      const char *db [[maybe_unused]],
                      const char *table [[maybe_unused]]) {
```

PR #1846, opened 2023-10-10
PXC-4312: DROP EVENT IF EXISTS generates local GTID event

kamil-holubicki:
https://jira.percona.com/browse/PXC-4312

Problem:
When DROP EVENT IF EXISTS is executed for non existing event, the event is binlogged with the GTID containing UUID of local server instead of global, cluster-wide UUID.

Cause:
As per MySql documentation, DROP DATABASE IF EXISTS, DROP TABLE IF EXISTS, and DROP VIEW IF EXISTS have to be replicated always. Code analysis lead to the conclusion that DROP EVENT IF EXISTS is in this group as well.
We did TOI for all besides DROP EVENT IF EXISTS. If the query is not replicated, wsrep logic generates GTID using local server UUID instead of cluster-wide UUID.

Solution:
Replicate DROP EVENT IF EXISTS always. This is enough for GTID to be properly generated.

### DIFF
--- a/mysql-test/suite/galera/t/galera_drop_if_exists.test
+++ b/mysql-test/suite/galera/t/galera_drop_if_exists.test
@@ -1,0 +1,30 @@
+# Check that the following requirement is fulfiled
+# The DROP DATABASE IF EXISTS, DROP TABLE IF EXISTS, DROP VIEW IF EXISTS and DROP EVENT IF EXISTS statements are always replicated,
+# even if the database, table, or view to be dropped does not exist on the source.
+# This is to ensure that the object to be dropped no longer exists on either the source or the replica,
+# once the replica has caught up with the source.
+#
+# Additionally check that such statements generate cluster gtid instead of local gtid.
+
+--source include/galera_cluster.inc
+--source include/gtid_utils.inc
+
+--connection node_1
+--source include/gtid_step_reset.inc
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2136/

mysql-test/suite/galera/t/galera_drop_if_exists.test
```diff
@@ -0,0 +1,30 @@
...
+--connection node_1
+--source include/gtid_step_reset.inc
+DROP DATABASE IF EXISTS db1;
+DROP TABLE IF EXISTS t1;
+DROP VIEW IF EXISTS v1;
+DROP EVENT IF EXISTS ev1;
+
+# check that GTID is advanced by 4
+--let $gtid_step_count = 4
+--let $gtid_step_uuid = `SELECT LEFT(@@global.gtid_executed, 36);`
```
venkatesh-prasad-v: With this we don't ensure that the events are binlogged with cluster uuid. Can we instead try this way?
```suggestion
--let $gtid_step_uuid = `SELECT Variable_value FROM performance_schema.global_status WHERE Variable_name='wsrep_cluster_state_uuid'`
```

venkatesh-prasad-v (approved):
LGTM

PR #1848, opened 2023-10-16
PXC-4315: No BF-abort but 'MDL conflict ... solved by abort' printed

kamil-holubicki:
https://jira.percona.com/browse/PXC-4315

Problem:
Logs like 'MDL conflict ... solved by abort' are printed, but no transaction was actually aborted.

Cause:
The log is printed from MDL_lock::can_grant_lock() if wsrep_handle_mdl_conflict() returns 'false'.
wsrep_handle_mdl_conflict() always returns 'false'.

Solution:
Fix wsrep_handle_mdl_conflict() to return 'true' if the transactions was really aborted, 'false' otherwise and print the log message accordingly

### DIFF
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -2562,17 +2562,16 @@ bool MDL_lock::can_grant_lock(enum_mdl_type type_arg,
                   wsrep_thd_thread_id(requestor_ctx->wsrep_get_thd()),
                   wsrep_thd_query(requestor_ctx->wsrep_get_thd()));
               can_grant = true;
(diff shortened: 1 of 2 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2153/

kamil-holubicki:
I've changed base branch. Let's merge it after 8.0.34 release branch

venkatesh-prasad-v (approved):

PR #1854, opened 2023-11-08
PXC-4318: long semaphore wait crash due to ha_commit_low does not com…

kamil-holubicki:
…mit an empty transaction

https://jira.percona.com/browse/PXC-4318

Problem:
When PXC node acts as the async replica to some master and in parallel
the row is updated on PXC node and via replication the PXC node hangs.

Cause:
If the row is updated in parallel by local session on PXC node and via
replication in the way that both updates result in the same row, and
local transaction updated the row as the first one, the replicated
transaction is not logged into the binlog.
This is because when replicated transaction commits in InooDB,
SE returns HA_ERR_RECORD_IS_THE_SAME and binlogging is skipped
(logic inside handler::ha_update_row()). So the transaction becomes
an empty transaction. In this case slave protocol directly commits
gtid through gtid_end_transaction that invokes ordered_commit causing
handler as part of storage engine commit.

Commit 730b82ac tried to solve the above situation by removing
the thread from wsrep group commit queue in wsrep_after_commit() if it
was not removed by committing to SE.
The problem of this solution is when the thread (t1) registers in
wsrep_group_commit_queue, another thread (t2) may register after it
(t1 does not have to be a leader). The queue is processed by the leader
and each thread is supposed to remove itself from
wsrep_group_commit_queue during SE commit. If t1 does not do SE commit
it remains at the front of the queue, and we block during t2 processing
in wsrep_wait_for_turn_in_group_commit().
So unregistering t1 from the queue in wsrep_after_commit() (after the
whole commit queue is done) is too late.

Solution:
If the transaction does not cause commit in SE which would cause
removal of it from wsrep_group_commit_queue, remove it explicitly.
This is a similar approach as implemented in 730b82ac but moved directly
to after processing of the thread during COMMIT stage rather than having
it after the whole stage is finished.

### DIFF
--- a/sql/handler.cc
+++ b/sql/handler.cc
@@ -2071,6 +2071,29 @@ int ha_commit_low(THD *thd, bool all, bool run_after_commit) {
         Commit_order_manager::wait_and_finish(thd, error);
       }
     }
+  } else {
+      /* This function is common for group commit flow and the flow that skips
+      group commit. We register the thread in wsrep_group_commit_queue always.
+      For group commit it is done during binlog flush stage, for 1-phase commit
+      it is done in wsrep_before_commit().
+      Ideally, for one-phase (with binlog=off) or two-phase (with binlog=on)
+      this step would be executed when transaction commits in InnoDB.
+      If galera node is acting as async slave and replicated action from async
+      master result in empty changes on slave (slave directly applied the said
+      changes and has skipped error through skip-slave-error configuration) it
+      can result in said situation. In this case slave protocol directly commits
+      gtid through gtid_end_transaction that invokes ordered_commit causing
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2188/
Permanent failures (fail also on 8.0 branch):
```
innodb.ddl_crash_alter_partition
galera.mysql-wsrep#216
```

All other failures are not permanent:
```
innodb.ddl_kill
main.percona_heap_bug_ps296
main.acl_mdl_scalability
```

venkatesh-prasad-v (approved):
Great work, approved! Please update the commit message as discussed on Slack before merging the PR.

PR #1855, opened 2023-11-20
PXC-4298: Make Inconsistency Voting to consider only error codes for voting

venkatesh-prasad-v:
https://jira.percona.com/browse/PXC-4298

Added a testcase


kamil-holubicki (approved):
LGTM. Please remember about reverting `.gitmodules` file after merging galera and moving galera repo pointer.

PR #1856, opened 2023-12-04
PXC-4336: PXC node eviction due to CHECK CONSTRAINT

kamil-holubicki:
https://jira.percona.com/browse/PXC-4336

Problem:
When constraint, which fails, is set on table, inconsistency voting is kicked off and the node is evicted from the cluster.

Cause:
ALTER TABLE ... ADD CONSTRAINT ... is replicated as TOI before the actual execution on the source node. Than it fails to execute on source node and succeeds on replicated node. This activates inconsistency voting protocol and the source node gets evicted from the cluster.

When the query is executed, before the actual ALTER execution, check constraints specifications are created
(prepare_check_constraints_for_alter()).
On source node they are created, then validated, validatio fails, and the node finishes with error.
On replicated node constraints specifications creation is skipped, so constraints are not validated at all. It is because of the logic of is_slave_with_master_without_check_constraints_support() function, which forces constraints check skipping in case of source node is not supporting such checks. The decision is based on the version of the source node which is carried in Gtid event. Galera does not replicate Gtid events, so this information is missing and the logic assumes that the source node does not support constraints check.

Solution:
Allow constraints check for Galera replication.

### DIFF
--- a/mysql-test/suite/galera/t/galera_alter_add_constraint.test
+++ b/mysql-test/suite/galera/t/galera_alter_add_constraint.test
@@ -1,0 +1,38 @@
+
+# Test that contraint check is done on source and replicated node
+# and produces the same error, which leaves the cluster in consistent
+# state after consistency voting.
+
+--source include/galera_cluster.inc
+
+--connection node_2
+CALL mtr.add_suppression("Replica SQL: Error .Check constraint .CHK_i. is violated");
+CALL mtr.add_suppression("Replica SQL: Error 'Duplicate check constraint name .CHK_i.* on query");
+CALL mtr.add_suppression("Replica SQL: Error 'Constraint .CHK_i2. does not exist");
+
+--connection node_1
+
+CREATE TABLE t (id INT PRIMARY KEY, i INT NOT NULL);
+INSERT INTO t VALUES(1, 100);
+
+# TOI will replicate before local execution.
+# Then it will fail locally, as the constraint check fails.
+# It is expected that the replica will fail as well with the same error.
+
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/2224/

mysql-test/suite/galera/t/galera_alter_add_constraint.test
```diff
@@ -0,0 +1,23 @@
...
+
+CREATE TABLE t (id INT PRIMARY KEY, i INT NOT NULL);
+
+INSERT INTO t VALUES(1, 100);
+
+# This will replicate before local execution.
+# Then it will fail locally, as the constraint check fails.
+# It is expected that the replica will fail as well with the same error.
+--error ER_CHECK_CONSTRAINT_VIOLATED
+ALTER TABLE t ADD CONSTRAINT CHK_i CHECK (i <=75);
```
venkatesh-prasad-v, line 24: Can you add the test scenario covered in https://jira.percona.com/browse/PXC-4337 and https://jira.percona.com/browse/PXC-4338 as well?

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/galera_alter_add_constraint.test
kamil-holubicki: done

venkatesh-prasad-v (approved):
lgtm

PR #1861, opened 2023-12-14
(8.0) PXC-4034: Using sql_log_bin=0 breaks GTID consistency 

kamil-holubicki:
https://jira.percona.com/browse/PXC-4034

Problem:
If binlogging is enabled globally, but it is disabled in the session
scope, transactions executed within this session advance 'gtid_executed'
As transactions are not binlogged, after the node restart, 'gtid_purged'
is populated with gtids related to these transactions.
When the node acts as the source for async replica, it causes replica
to fail, because source is not able to serve not logged, but
gtid-related transactions.

Cause:
PXC, even if binlog is disabled, uses binlog infrastructure to cache
binlog events for the purpose of creation of Galera writesets. If binlog
is disabled, logging to the file is skipped.
GTID is generated during group commit binlog flush stage.
PXC flow goes through this stage if binlog is enabled generating GTIDs
for every transaction, even if later the transaction is not binlogged.

Solution:
If binlogging is disabled for the particular thread, force the execution
path as if gtid_mode was equal OFF.

### DIFF
--- a/sql/rpl_gtid_state.cc
+++ b/sql/rpl_gtid_state.cc
@@ -543,8 +543,24 @@ enum_return_status Gtid_state::generate_automatic_gtid(THD *thd,
     /* The caller must lock the sid_lock when locked_sidno is passed */
     sid_lock->assert_some_lock();
 
+#ifdef WITH_WSREP
+    bool skip_gtid = false;
+    /* If binlog is disabled permanently we don't get here.
+    If it is disabled only in a particular session, we need to
+    do PXC replication, however we will not binlog it.
+    So we need to skip GTID generation for this transaction.
+    Just pretend for a while that gtid_mode=OFF */
+    if (!(thd->variables.option_bits & OPTION_BIN_LOG) &&
+        !(thd->variables.option_bits & OPTION_BIN_LOG_INTERNAL_OFF))
+        {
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/2233/

sql/rpl_gtid_state.cc
```diff
@@ -1011,3 +1034,53 @@ enum_return_status Gtid_state::ensure_commit_group_sidnos(rpl_sidno sidno) {
...
+  DBUG_TRACE;
+  sid_lock->assert_some_lock();
+
+/* We are allowed to get here only in two cases:
+1. If this is WSREP-enabled thread
+    and binlog is disabled
+    and binlog is not disabled internally
+    This means that this is the session with sql_log_bin=OFF.
+2. if gtid_mode != ON (the original condition) */
+assert( (WSREP(thd) && !(thd->variables.option_bits & OPTION_BIN_LOG) &&
```
venkatesh-prasad-v: Please run clang-format once before merging the fix.

venkatesh-prasad-v (approved):

PR #1867, opened 2024-01-03
PXC-4341: Executing prepared statement can abort node after FLUSH TABLES

venkatesh-prasad-v:
https://perconadev.atlassian.net/browse/PXC-4341

Problem
-------
Transient errors during the execution of the DDL in a prepared statement mode can trigger inconsistency voting resulting in node eviction from the cluster.

Analysis & Fix
--------------
It can sometimes happen that the DDL executed as a prepared statement may fail with transient errors like ER_NEED_REPREPARE if the table metadata was changed between PREPARE and EXECUTE (due to FLUSH TABLES). In such cases, the statement is re-prepared and re-executed from Prepared_statement::execute_loop().

However, before the statement is reprepared, PXC's TOI_end() hook captures this transient error (i.e, ER_NEED_REPREPARE) and sends it to galera thereby triggering the inconsistenct voting and kicking this node out of the cluster.

So, in such cases, we need to delay sending the final error to galera until all reprepare attempts are over. This is done by attaching a temporary Diagnostics_area to THD and resetting it later if all the required conditions are met.

NOTE
----

The below mentioned commands reprepare the statement if there is any
metadata change (marked with CF_REEXECUTION_FRAGILE)
    
        SQLCOM_CREATE_TABLE
        SQLCOM_LOAD
        SQLCOM_CREATE_VIEW
        SQLCOM_UPDATE
        SQLCOM_UPDATE_MULTI
        SQLCOM_INSERT
        SQLCOM_INSERT_SELECT
        SQLCOM_DELETE
        SQLCOM_DELETE_MULTI
        SQLCOM_REPLACE
        SQLCOM_REPLACE_SELECT
        SQLCOM_SELECT
        SQLCOM_DO
        SQLCOM_CALL
    
Among the above commands, only SQLCOM_CREATE_TABLE and
SQLCOM_CREATE_VIEW are DDLs and thus was affected by this bug.

However in case of CREATE VIEW, table version check is performed before
the query is replicated as TOI, so CREATE TABLE was the only query that
was affected by this bug. So, a special handling has been done only for
the CREATE_TABLE query in this patch.

Testing Done
----
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/180/console

### DIFF
--- a/sql/sql_cmd_ddl_table.cc
+++ b/sql/sql_cmd_ddl_table.cc
@@ -516,17 +516,26 @@ bool Sql_cmd_create_table::execute(THD *thd) {
           mysql_mutex_lock(&LOCK_wsrep_alter_tablespace);
         }
 
-        /* Note we are explictly opening the macro as we need to perform
+        bool prepared_stmt_execution = thd->get_reprepare_observer() != nullptr;
+        bool should_start_toi = !prepared_stmt_execution ||
+                                (prepared_stmt_execution &&
+                                 !wsrep_thd_is_in_to_isolation(thd, false));
+        /* Note we are explicitly opening the macro as we need to perform
         cleanup action on TOI failure. */
-        if (WSREP(thd) && wsrep_to_isolation_begin(thd, create_table->db,
-                                                   create_table->table_name,
-                                                   NULL, NULL, &alter_info)) {
+        if (WSREP(thd) && should_start_toi &&
+            wsrep_to_isolation_begin(thd, create_table->db,
+                                     create_table->table_name, NULL, NULL,
+                                     &alter_info)) {
           if (!thd->lex->is_ignore() && thd->is_strict_mode())
             thd->pop_internal_handler();
           if (create_info.tablespace) {
             mysql_mutex_unlock(&LOCK_wsrep_alter_tablespace);
           }
           return true;
+        }
+
+        if (should_start_toi) {
+          thd->wsrep_prepared_statement_TOI_started = true;
         }
 
         if (create_info.tablespace) {
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -6093,10 +6093,60 @@ finish:
   }
 
 #ifdef WITH_WSREP
+  /*
+    It can sometimes happen that the DDL executed as a prepared statement may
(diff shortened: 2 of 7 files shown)


kamil-holubicki (approved):
lgtm

sql/sql_cmd_ddl_table.cc
```diff
@@ -516,11 +516,16 @@ bool Sql_cmd_create_table::execute(THD *thd) {
...
-        cleanup action on TOI failure. */
-        if (WSREP(thd) && wsrep_to_isolation_begin(thd, create_table->db,
-                                                   create_table->table_name,
-                                                   NULL, NULL, &alter_info)) {
+        bool prepared_stmt_execution = thd->get_reprepare_observer() != nullptr;
+        bool should_start_toi = !prepared_stmt_execution ||
+                                (prepared_stmt_execution &&
+                                 !wsrep_thd_is_in_to_isolation(thd, false));
+        /* Note we are explicitly opening the macro as we need to perform
+        cleanup action on TOI failure. */ // KH: here TOI begin
```
kamil-holubicki: Please remove 'KH:' comment

sql/sql_parse.cc
```diff
@@ -6093,10 +6093,60 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
+      } else {
+        // This means all retries are exhausted. We should call TOI_end now.
+        finish_toi = true;
+      }
+    }
+  }
   thd->wsrep_consistency_check = NO_CONSISTENCY_CHECK;
-  WSREP_TO_ISOLATION_END;
+  if (finish_toi) {
+    WSREP_TO_ISOLATION_END;  // KH: here toi end
```
kamil-holubicki: Please remove 'KH:' comment

kamil-holubicki (changes requested):

sql/sql_cmd_ddl_table.cc
venkatesh-prasad-v: Done

sql/sql_parse.cc
venkatesh-prasad-v: Done

kamil-holubicki (approved):

PR #1873, opened 2024-01-19
PXC-4362: PXC node evicted when create function by

kamil-holubicki:
user don't have super privilege and binary loggin is enabled

https://perconadev.atlassian.net/browse/PXC-4362

Problem:
If log_bin_trust_function_creators=OFF (default), CREATE FUNCTION statement fails locally with error "You do not have the SUPER privilege and binary logging is enabled (you *might* want to use the less safe log_bin_trust_function_creators variable)", but doesn't fail on the remote node which leads to node being evicted from the cluster.

Cause:
CREATE FUNCTION is replicated as TOI, before doing checks on local node. As it is executed from applier thread context on the remote node, it succeeds remotely, but fails locally. This kicks-in inconsistency voting protocol, which causes node to be evicted from the cluster.

Solution:
Do privileges checks before replicating TOI.

### DIFF
--- a/mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
+++ b/mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
@@ -1470,6 +1470,67 @@ REVOKE CREATE ROUTINE,EXECUTE,ALTER ROUTINE ON testdb.* FROM 'testme'@'%';
 SHOW GRANTS FOR 'testme';
 FLUSH PRIVILEGES;
 
+--echo #
+--echo # SQLCOM_CREATE_SPFUNCTION (with log_bin_trust_function_creators=OFF) access test
+--echo #
+#
+# All tests are executed with log_bin_trust_function_creators=ON
+# (set in default_mysqld.cnf)
+# however OFF is the default value. Test CREATE FUNCTION with default
+# log_bin_trust_function_creators=OFF
+#
+--inc $test_id
+
+--echo # connection node_1
+--connection node_1
+SET @old_log_bin_trust_function_creators= @@global.log_bin_trust_function_creators;
+SET global log_bin_trust_function_creators=OFF;
+
+GRANT ALL PRIVILEGES ON testdb.* TO 'testme'@'%';
+--replace_regex /\*[0-9A-Z]+/<SECRET>/
+SHOW GRANTS FOR 'testme';
+FLUSH PRIVILEGES;
+
+--echo # connection node_testme
+--disconnect node_testme
+--connect node_testme, 127.0.0.1, testme, secret,, $NODE_MYPORT_1
+--connection node_testme
+use testdb;
+
+--error ER_BINLOG_CREATE_ROUTINE_NEED_SUPER
+CREATE FUNCTION hello (s CHAR(20)) RETURNS CHAR(50) DETERMINISTIC RETURN CONCAT('Hello, ',s,'!');
+
+--eval INSERT INTO counter(count) VALUES($test_id);
+
+--echo # connection node_1
+--connection node_1
+--let $wait_condition = SELECT COUNT(*) = $test_id FROM counter;
+--source include/wait_condition.inc
+
+--let $assert_text = The hello function should not exist on node 1
+--let $assert_cond = COUNT(*) = 0 FROM information_schema.routines where ROUTINE_NAME="hello"
+--source include/assert.inc
+
+--echo # connection node_2
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = $test_id FROM counter;
+--source include/wait_condition.inc
+
+--let $assert_text = The hello function should not exist on node 2
+--let $assert_cond = COUNT(*) = 0 FROM information_schema.routines where ROUTINE_NAME="hello"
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2320/

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
```diff
@@ -1470,6 +1470,67 @@ REVOKE CREATE ROUTINE,EXECUTE,ALTER ROUTINE ON testdb.* FROM 'testme'@'%';
...
+# (set in default_mysqld.cnf)
+# however OFF is the default value. Test CREATE FUNCTION with default
+# log_bin_trust_function_creators=OFF
+#
+--inc $test_id
+
+--echo # connection node_1
+--connection node_1
+set @old_log_bin_trust_function_creators= @@global.log_bin_trust_function_creators;
+set global log_bin_trust_function_creators=OFF;
```
venkatesh-prasad-v: Use uppercase for SQL commands.

sql/sp.cc
```diff
@@ -802,6 +802,18 @@ bool sp_create_routine(THD *thd, sp_head *sp, const LEX_USER *definer,
...
   assert(!sp_already_exists);
 
+#ifdef WITH_WSREP
+  if (create_routine_precheck(thd, sp)) {
+    /* If this happens, an error should have been reported. */
+    return true;
+  }
+
+  if (WSREP(thd) &&
+      wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, nullptr, nullptr)) {
```
venkatesh-prasad-v, line 812: Can we pass the actual database name stored in `sp->m_db.str` instead of WSREP_MYSQL_DB?

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
```diff
@@ -1470,6 +1470,67 @@ REVOKE CREATE ROUTINE,EXECUTE,ALTER ROUTINE ON testdb.* FROM 'testme'@'%';
...
+GRANT ALL PRIVILEGES ON testdb.* TO 'testme'@'%';
+--replace_regex /\*[0-9A-Z]+/<SECRET>/
+SHOW GRANTS FOR 'testme';
+FLUSH PRIVILEGES;
+
+--echo # connection node_testme
+--disconnect node_testme
+--connect node_testme, 127.0.0.1, testme, secret,, $NODE_MYPORT_1
+--connection node_testme
+use testdb;
```
venkatesh-prasad-v, line 1498: No need to use  `use testdb;` command as connection can directly be made to `testdb` by passing the db name to `--connect`
```suggestion
--connect node_testme, 127.0.0.1, testme, secret,testdb, $NODE_MYPORT_1
--connection node_testme
```

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
```diff
@@ -1470,6 +1470,67 @@ REVOKE CREATE ROUTINE,EXECUTE,ALTER ROUTINE ON testdb.* FROM 'testme'@'%';
...
+--let $wait_condition = SELECT COUNT(*) = $test_id FROM counter;
+--source include/wait_condition.inc
+
+--let $assert_text = The hello function should not exist on node 2
+--let $assert_cond = COUNT(*) = 0 FROM information_schema.routines where ROUTINE_NAME="hello"
+--source include/assert.inc
+
+--echo # connection node_1
+--connection node_1
+# revoke what was cranted for the current check
```
venkatesh-prasad-v:
```suggestion
# revoke what was granted for the current check
```

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
kamil-holubicki: done

sql/sp.cc
kamil-holubicki: I'm not sure. This TOI begin was just moved from the caller to this location.

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
kamil-holubicki: This seems to be a convention of this file, so let's follow it.

mysql-test/suite/galera/t/galera_wsrep_ddl_access_checking2.test
kamil-holubicki: done

sql/sp.cc
venkatesh-prasad-v: Currently, at many places we are calling `wsrep_to_isolation_begin()` with `WSREP_MYSQL_DB` as the db name for the current TOI operation. As database name is also included in the writeset, there is a chance that two concurrent SPs with same name but in different database may cause certification failure. So, I think it is better we pass the actual DB name while calling TOI begin.

venkatesh-prasad-v (approved):
LGTM

PR #1876, opened 2024-02-09
PXC-4348: Cluster state interruption with MDL BF-BF conflict and exec…

kamil-holubicki:
…-mode:toi

https://perconadev.atlassian.net/browse/PXC-4348

Problem:
The Joiner node crashes with BF-BF abort during IST.

Cause:
Let's consider the following case in 2-nodes cluster: 0. Two wsrep_applier threads (A1, A2)
1. The schema consists of two tables: t1 - parent, t2 - child
2. Two sessions on node_1 execute workloads which conflict and cause certification failures, like:
session_1: OPTIMZE TABLE t1
session_2: DELETE FROM t2 WHERE ...; INSERT INTO t2 VALUES ...;
3. Queries are dependant (writeset's depends_seqno set to previous transaction), because of parent-child tables relation. However, because of certification conflicts, so-called dummy writesets are logged.
4. Dummy writeset does not depend on other writeset (depends_seqno == -1), but other writeset can depend on dummy writeset.
5. When DELETE FROM ... is executed we call wsrep_row_upd_check_foreign_constraints(). If foreign->referenced_table is nullptr we call dd_table_open_on_name() which creates explicit lock on t1 table. Later we call dict_table_close(), which closes the table, but does not release the lock. dd_table_close() should be called instead. This is the root cause of the problem.
6. Let's say during IST we've got the following chain of dependencies: T1 (DML) -> T2 (DDL, depends on T1) -> T3 (dummy, no dependency)
7. T1 is executed by A1. T2 waits. (A2 is bussy with other WS - doesn't matter)
8. A1 finishes with T1. Explicit lock on t1 is owned by A1.
9. T2 is executed by A2, T3 is executed by A1 - they have no dependency, so execution is in parallel. T2 brute forces T3 as they conflict on having t1 MDL.

Solution:
1. Release explicit MDL lock by calling dd_table_close()
2. Added assertions in server_state::on_apply() before and after writeset apply to catch situations when we start or end processing with MDL locks acquired.
3. In places when we check pxc_strict_mode, handled the situation when MDL lock can't be acquired. In such a case it shouldn't be released (sql_alter.cc, sql_truncate.cc and sql_tablespace.cc)
4. Removed not neded unlock of MDL lock in sql_truncate.cc. It is handled by wsrep_scope_guard mdl_lock.

### DIFF
--- a/sql/mdl.cc
+++ b/sql/mdl.cc
@@ -4479,6 +4479,21 @@ void MDL_context::release_locks_stored_before(enum_mdl_duration duration,
 void MDL_context::release_explicit_locks() {
   release_locks_stored_before(MDL_EXPLICIT, NULL);
 }
+
+void MDL_context::dump_locks() {
+  MDL_ticket *ticket;
+  WSREP_INFO("MDL_context::dump_locks() >>>\n");
+  for (int i = 0; i < MDL_DURATION_END; i++) {
+    enum_mdl_duration duration = static_cast<enum_mdl_duration>(i);
+
+    MDL_ticket_store::List_iterator it = m_ticket_store.list_iterator(duration);
+    while ((ticket = it++)) {
+      WSREP_INFO("duration: %d, name: %s", i, ticket->get_lock()->key.name());
+    }
+  }
+  WSREP_INFO("MDL_context::dump_locks() <<<\n");
+}
+
 #endif /* WITH_WSREP */
 
 /**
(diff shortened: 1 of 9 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2362/

venkatesh-prasad-v (approved):
LGTM

PR #1880, opened 2024-03-01
PXC-4367: InnoDB semaphore wait timeout failure after upgrade from 8.…

kamil-holubicki:
…0.34 to 8.0.35

https://perconadev.atlassian.net/browse/PXC-4367

Problem:
Under the heavy load, it happens sometimes that the server is stuck waiting for the InnoDB semaphore. After the 600secs timeout the server is intentionally stopped.

Cause:
This bug is the regression caused by commit 3dbdd49b related to PXC-4318 3dbdd49b moved unregistration of thread from wsrep_group_commit_queue from wsrep_after_commit() directly to ha_commit_low() to the execution branch reached when no SE is involved in the commit. However, this fix didn't consider a scenario when SE is involved, but actual innobase_commit() fails (because of different reasons, e.g. transaction abort or others) and the flow doesn't reach trx_sys_update_wsrep_checkpoint() where unregistration from wsrep_group_commit_queue should happen. In such a case the thread is left in front of wsrep_group_commit_queue blocking every other thread from processing (the other thread waits to be the 1st one to continue). Before the above-mentioned fix unregistration happened in wsrep_after_commit().

In other words, the scenario is:
1. thd_1 commits transaction. It calls wsrep_ordered_commit() to register itself in wsrep_group_commit_queue
2. thd_1 calls ha_commit_low , but is aborted -> no unregistration from wsrep_group_commit_queue
3. thd_1 calls wsrep_after_commit. Before 3dbdd49b there was unregistration from wsrep_group_commit_queue if not already unregistered. It was removed. So thd_1 stays on wsrep_group_commit_queue
4. thd_2 commits transaction, it calls wsrep_ordered_commit() to register itself in wsrep_group_commit_queue
5. thd_2 calls ha_commit_low and gets to wsrep_wait_for_turn_in_group_commit
6. thd_2 waits for being the 1st in the queue forever

Reverting that part of the previous fix is not enough, because it will still not work for the following scenario:
1. The following commit queue is being processed: thd_1, thd_2
2. thd_1 becomes the leader of the commit stage of group commit
3. thd_1 calls ha_commit_low, but it fails -> no unregistration from wsrep_group_commit_queue
4. thd_1 processes thd_2 commit
5. it gets to wsrep_wait_for_turn_in_group_commit and waits for thd_2 to be the 1st in the queue

Solution:
A thread has to be removed from wsrep_group_commit_queue after it is committed. No matter if the actual commit succeeded or failed. Call unregistration from wsrep_group_commit_queue unconditionally at the end of ha_commit_low()

### DIFF
--- a/mysql-test/suite/galera/t/pxc_error_during_group_commit.test
+++ b/mysql-test/suite/galera/t/pxc_error_during_group_commit.test
@@ -1,0 +1,36 @@
+# Test the following problematic case is handled correctly:
+# 1. group commit queue consist of one thread thd_1
+# 2. thd_1 is being processed, it registers in wsrep_group_commit_queue
+# 3. InnoDB commit fails (it may happen because of transaction abort or other reason)
+# 4. As the result thd_1 fails to unregister from wsrep_group_commit_queue
+# 5. thd_2 is being processed as another commit queue
+# 6. thd_2 commits in InnoDB and tries to unregister from wsrep_group_commit_queue
+# 7. thd_2 waits, because thd_1 is still the 1st item in the wsrep_group_commit_queue
+
+--source include/have_debug.inc
+--source include/have_debug_sync.inc
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+CALL mtr.add_suppression("Transaction not registered for MySQL 2PC");
+
+CREATE TABLE t1 (a INT PRIMARY KEY);
+
+# Force InnoDB commit failure, which will cause skip of unregistration from wsrep_group_commit_queue
+SET DEBUG = '+d,innobase_commit_low_fail';
+INSERT INTO t1 VALUES (0);
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+--connection node_1a
+
+# node_1 connection is the 1st one in wsrep_group_commit_queue, so node_1a will assert in debug build
+# For release build it would cause '[InnoDB] A long semaphore wait', but unfortunately debug_sync
+# is not available, so no way to simulate. 
+INSERT INTO t1 VALUES (1);
+--assert(`SELECT COUNT(*) = 2 FROM t1`)
+
+# cleanup
+--disconnect node_1a
+--connection node_1
+DROP TABLE t1;
+--source include/wait_until_count_sessions.inc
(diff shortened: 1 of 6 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2386/

sql/wsrep_trans_observer.h
```diff
@@ -378,6 +378,12 @@ static inline int wsrep_after_commit(THD *thd, bool all) {
 
   /* It has to already unregistered from wsrep_group_commit_queue either
   during SE commit or directly in ha_commit_low if SE commit was not done. */
+  if (thd->wsrep_enforce_group_commit) {
```
venkatesh-prasad-v: I think this commit is not required now.

The original comment was this.
```
    /* Ideally, for one-phase (with binlog=off) or two-phase (with binlog=on)
    this step would be executed when transaction commits in InnoDB.
    If galera node is acting as async slave and replicated action from async
    master result in empty changes on slave (slave directly applied the said
    changes and has skipped error through skip-slave-error configuration) it
    can result in said situation. In this case slave protocol directly commits
    gtid through gtid_end_transaction that invokes ordered_commit causing
    handler as part of storage engine commit. Handle unregistration here. */
```
Can you restore this comment if it is still valid now?

kamil-holubicki: The original comment is not relevant in this place. It was moved to `ha_commit_low`
I updated the current comment.

mysql-test/suite/galera/t/pxc_error_during_group_commit.test
```diff
@@ -0,0 +1,35 @@
+# Test the following problematic case is handled correctly:
+# 1. group commit queue consist of one thread thd_1
+# 2. thd_1 is being processed, it registers in wsrep_group_commit_queue
+# 3. InnoDB commit fails (it may happen because of transaction abort or other reason)
+# 4. As the result thd_1 fails to unregister from wsrep_group_commit_queue
+# 5. thd_2 is being processed as another commit queue
+# 6. thd_2 commits in InnoDB and tries to unregister from wsrep_group_commit_queue
+# 7. thd_2 waits, because thd_1 is still the 1st item in the wsrep_group_commit_queue
+
+--source include/have_debug_sync.inc
```
venkatesh-prasad-v, line 11: Also include `include/have_debug.inc` to skip this test on release builds.

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/pxc_error_during_group_commit.test
kamil-holubicki: Done

venkatesh-prasad-v (approved):
LGTM

PR #1882, opened 2024-03-19
PXC-4340: Crash in row_sel_convert_mysql_key_to_innobase

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4340

Problem:
Server crashes during execution of the complicated query with 9 CTEs.

Cause:
For CTE queries, the optimizer can request the creation of a temp table for a particular CTE. If such a temp result is used in table joins, indexes are added to such a table, based on join columns. All possible indexes are added at the 1st stage and then the optimizer decides which of them will be used, removing unused ones. If the CTE temp result is used in joins in other CTEs of the same query, the process of finding indexes is repeated several times.

Indexes are created in TABLE::add_tmp_key() as <auto_keyARRAY_INDEX>, where ARRAY_INDEX is the index in TABLE_SHARE::key_names array. Let's say <auto_key0>, <auto_key1>, <auto_key2> are created. Then the query optimizer decides that only <auto_key2> will be used, so <auto_key0> and <auto_key1> are removed. <auto_key2> is shifted to the 1st unused position (TABLE::move_tmp_key()),
so TABLE_SHARE::key_names contains <auto_key2> on position 0.

Then the query optimizer makes a plan for another join, repeating the above steps. It adds new keys to TABLE_SHARE::key_names and it results with <auto_key2>, <auto_key1>, <auto_key2>. So we've got two <auto_key2> keys, having different definitions.

Then the temp table is created in the InnoDB world.

Then the query is executed. We get to ha_innobase::open() which calls dict_table_get_index_on_name() requesting <auto_key2> (the 2nd one). But the function returns the 1st index.
innobase_match_index_columns() is called to check the consistency between MySql and InnoDB index definition and here we get the message:

[ERROR] [MY-010880] [InnoDB] Found index <auto_key2> whose column info does not match that of MySQL. [ERROR] [MY-010882] [InnoDB] Build InnoDB index translation table for Table /var/lib/mysql/tmp/#sql3dad_13_4 failed

Then we get to ha_innobase::index_read() where
row_sel_convert_mysql_key_to_innobase() is called. The function doesn't do many checks, just converts the key. As the definition of MySql key is longer (e.g. 8 + 8 bytes) than InnoDB's (e.g. 8 + 6), we try to read too much, causing a segmentation fault.

The problem is in the way of naming auto keys by the optimizer. The current algorithm does not guarantee the uniqueness of key names, causing name clashes in InnoDB world.

Solution:
Introduce a variable that will serve the unique number for auto_key names causing index names to be unique.


sql/table.cc
```diff
@@ -6069,7 +6069,7 @@ bool TABLE::add_tmp_key(Field_map *key_parts, bool invisible,
     s->max_key_length = std::max(s->max_key_length, key_len);
     s->key_parts += key_part_count;
     assert(s->keys < s->max_tmp_keys);
-    sprintf(s->key_names[s->keys].name, "<auto_key%d>", s->keys);
+    sprintf(s->key_names[s->keys].name, "<auto_key%d>", s->auto_key_name_server++);
```
satya-bodapati: better to use atomic add here?

sql/table.h
```diff
@@ -842,6 +842,7 @@ struct TABLE_SHARE {
   uint fields{0};            /* Number of fields */
   uint rec_buff_length{0};   /* Size of table->record[] buffer */
   uint keys{0};              /* Number of keys defined for the table*/
+  uint auto_key_name_server{0}; /* Serves the unique number for <auto_keyN>*/
```
satya-bodapati: or use std::atomic_int64_t/32_t

sql/table.cc
kamil-holubicki: I don't know if it is better. It is a single thread. Previously `keys` was used and it is not atomic as well.

sql/table.h
kamil-holubicki: see above. keys was used before. Single thread here :)

sql/table.cc
satya-bodapati: can  TABLE* created be created from TABLE_SHARE concurrently? ie from two connections?
Imagine two such big CTE queries from two connections? If answer is yes, I would recommend atomic

kamil-holubicki: I think we are safe here.
`add_tmp_keys()` is called only for temp tables.
As per description: `Add one key to a materialized derived table.`

The caller, `Table_ref::generate_keys()`, has assert for this at its beginning.

satya-bodapati: Ok, if it is for temp table, then other connections/threads cannot use the same TABLE_SHARE.
I would recommend a name that indicates this variable is meant for temp-table. Something like temp_table_key_id?

satya-bodapati: I also wonder if it is wise to add a debug assertion that validates duplicate keynames are disallowed? (they are disallowed by design for regular tables but temp tables got away with it).

My intention is that server layer catch/report this problem, not SEs.

kamil-holubicki: Changed the name of the variable.
As for assert, after the fix, I think it is impossible by design to have two keys in the temp table with the same name.

satya-bodapati: Thanks for the fixing the name. Optional: better to remove the server word suffix

right, I always feel the same. But there is always a new branch code that can populate TABLE* and break the assertion. But it all depends on where we do the check.

I will not insist on this. if there is a good place to check, we may do it.
@dlenev may be has some ideas/thoughts on the assertion? :thinking:

kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2486/

Long test (timeout on Jenkins):
main.mysql_upgrade_grant

Fail without fix as well:
fix galera.mysql-wsrep#216
galera-x.crud_modify_view

Fixed after the Jenkins run:
main.derived
main.derived_condition_pushdown
main.view

satya-bodapati (approved):
LGTM! 
[optional suggestion] to implement serverside duplicate name check.

percona-ysorokin (approved):
LGTM

kamil-holubicki:
Hi @satya-bodapati 
I think we already have such a check in InnoDB
https://github.com/percona/percona-xtradb-cluster/blob/f95c91e01ef3d1cb371370b74e974fae88d68e0b/storage/innobase/dict/dict0dd.cc#L2770
`dd_find_index` is called when the table is created. So for InnoDB we are good.

To handle all SEs, I think we could add similar check in `create_tmp_table_with_fallback()` on the server side. 

But anyway, as we see the already existing code was never hit in debug, so I just doubt if having such assertion brings any value. Having this detected and reporting error to the caller seems more reasonable, but it is a bigger thing... 
What is your opinion?

PR #1886, opened 2024-03-27
PXC-4385: MDL conflict leads to locked nodes

kamil-holubicki:
https://perconadev.atlassian.net/issues/PXC-4385

Problem:
Execution of queries 'ALTER USER ... ACCOUNT LOCK' and 'SET PASSWORD FOR ... = ...' in parallel on different nodes may lead to the cluster stuck.

Cause:
TOI replication is done in open_grant_tables() after checking replication filters. Validation of replication filters requires call to acl_tables_setup_for_write_and_acquire_mdl() which acquires MDL locks. Then local TOI is replicated.
If at the time when local TOI is about to replicate on Galera level, another, replicated TOI is being processed and calls open_grant_tables() as well, we end up with the situation when replicated TOI should B-F abort local TOI. But local TOI waits for its turn in Galera, so not able to unlock acquired MDL locks. Replicated TOI waits on MDL locks held by local TOI which leads to deadlock.

Solution:
After examining replication filters, release all MDL locks and replicate TOI. Then reacquire needed locks and continue original flow. This is the best we can do.

### DIFF
--- a/sql/auth/sql_user_table.cc
+++ b/sql/auth/sql_user_table.cc
@@ -1990,6 +1990,65 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
 
   DEBUG_SYNC(thd, "wl14084_acl_ddl_before_mdl_acquisition");
 
+#ifdef WITH_WSREP
+  /* We have to check replication filters before TOI.
+     To do the check, we need to acquire MDL locks on ACL tables.
+     The check is done only for replica threads if filters are enabled.
+     Here we decide if async replicated transaction should be  PXC-replicated.
+     If so, we will start TOI, but can't do it while holding MDL lock, because
+     if we call TOI begin here and get to the point where Galera is just about
+     to replicate, and at the same time another, replicated TOI arrives, we will
+     end up waiting in Galera for our turn.
+     If that replicated TOI relates to ACL as well (e.g. SET PASSWORD) and calls
+     acl_tables_setup_for_write_and_acquire_mdl(), it will try to abort us,
+     but we are waiting in Galera, so not able to release locks.
+     The result is a deadlock.
+
+     Let's close ACL tables here and and lock them again after TOI, hoping
+     that replication filters won't change in the meantime. */
+  if (thd->slave_thread && thd->rli_slave->rpl_filter->is_on()) {
+    if (acl_tables_setup_for_write_and_acquire_mdl(thd, tables)) return -1;
+    /*
+      The tables must be marked "updating" so that tables_ok() takes them into
+      account in tests.
+    */
+    for (auto i = 0; i < ACL_TABLES::LAST_ENTRY; i++) tables[i].updating = true;
+
+    if (!(thd->sp_runtime_ctx ||
+          thd->rli_slave->rpl_filter->tables_ok(nullptr, tables))) {
+      thd->mdl_context.release_transactional_locks();
+      return 1;
+    }
+
+    for (auto i = 0; i < ACL_TABLES::LAST_ENTRY; i++)
+      tables[i].updating = false;
+
+    thd->mdl_context.release_transactional_locks();
+  }
+
+  /* CREATE/DROP function/procedure implicitly grant priviliges.
+  Check for detail comment in respected switch handler in sql_parse.cc
+  Since the original statement is already replicated using TOI
+  sub-action of grant/revoke privilges doesn't need to get replicated. */
+  bool skip_toi = (thd->lex->sql_command == SQLCOM_CREATE_SPFUNCTION ||
+                   thd->lex->sql_command == SQLCOM_CREATE_PROCEDURE ||
+                   thd->lex->sql_command == SQLCOM_DROP_FUNCTION ||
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
All tests:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2493/

New test only:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2494/

sql/auth/sql_user_table.cc
```diff
@@ -2026,15 +2038,39 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
...
     replicating commands that won't be applied locally (due to a filter).
   */
-  /* Doing TOI here is not the ideal solution, as we are holding
-  MDL locks already and the current thread can still be BF-aborted.
-  But we have to check replication filters before TOI and for this
-  we need MDL locks. */
   if (WSREP(thd) && !skip_toi &&
       wsrep_to_isolation_begin(thd, db, table, nullptr)) {
     WSREP_ERROR("Fail to replicate: %s", thd->query().str);
     return -1;
```
venkatesh-prasad-v, line 2044: We need to call `thd->mdl_context.release_transactional_locks();` here as well.
```suggestion
    WSREP_ERROR("Fail to replicate: %s", thd->query().str);
    thd->mdl_context.release_transactional_locks();
    return -1;
```

sql/auth/sql_user_table.cc
```diff
@@ -2011,9 +2022,10 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
 
     for (auto i = 0; i < ACL_TABLES::LAST_ENTRY; i++)
       tables[i].updating = false;
+
+    thd->mdl_context.release_transactional_locks();
```
venkatesh-prasad-v, line 2026: Can we have this logic of temporary release of lock and re-acquire to apply only for WSREP enabled threads?
```suggestion
    if (WSREP(thd)) thd->mdl_context.release_transactional_locks();
```

sql/auth/sql_user_table.cc
```diff
@@ -2026,15 +2038,39 @@ int open_grant_tables(THD *thd, TABLE_LIST *tables,
...
   if (WSREP(thd) && !skip_toi &&
       wsrep_to_isolation_begin(thd, db, table, nullptr)) {
     WSREP_ERROR("Fail to replicate: %s", thd->query().str);
     return -1;
   }
+
+  /* For replica thread we checked replication filters before TOI and then
+     released MDL locks.
+     Original flow continues with tables locked, so lock them again. */
+  if (acl_tables_setup_for_write_and_acquire_mdl(thd, tables)) return -1;
```
venkatesh-prasad-v, line 2050: Can we have this logic of temporary release of lock and re-acquire to apply only for WSREP enabled threads?
```suggestion
  if (WSREP(thd) && acl_tables_setup_for_write_and_acquire_mdl(thd, tables)) return -1;
```

venkatesh-prasad-v (changes requested):

sql/auth/sql_user_table.cc
kamil-holubicki: Please imagine the flow for non-slave thread. 
This part is executed only for slave thread.

sql/auth/sql_user_table.cc
kamil-holubicki: Please imagine the flow for non-slave thread. 
We have to lock always. For non-slave thread there will be no lock-check-unlock-toi-lock. There will be toi-lock only

sql/auth/sql_user_table.cc
kamil-holubicki: Well, we are not holding locks during TOI replication so no unlocking if it fails.

venkatesh-prasad-v: Thanks for clarifying. Got it! 👍🏼

sql/auth/sql_user_table.cc
venkatesh-prasad-v: I meant if we can skip this temporary release of locks if it is not a WSREP enabled thread, but yet an applier thread i(n other words async replica threads).

venkatesh-prasad-v: Currently it looks that **lock & check-unlock & reacquire** happens even for non-wsrep async replica threads
``` c++
if (slave_thread) {
  if (acl_tables_setup_for_write_and_acquire_mdl(thd, tables)) return -1;
  ..
  thd->mdl_context.release_transactional_locks();
}

if (WSREP(thd) && wsrep_to_isolation_begin()) {..}
...

if (acl_tables_setup_for_write_and_acquire_mdl(thd, tables)) return -1;
```

kamil-holubicki:
Rerun of failed tests
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/2495/

venkatesh-prasad-v (approved):
LGTM

PR #1888, opened 2024-03-29
PXC-4397: ALTER DATABASE and DROP TRIGGER hangs the instance

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4397

Problem:
When ALTER DATABASE and DROP TRIGGER are executed in parallel on the same node, the node hangs.

Cause:
connection_1: DROP TRIGGER acquires database level MDL lock and trigger-related MDL locks before replicating TOI. connection_2: ALTER DATABASE acquires database level MDL lock after replicating TOI.
If we are about to replicate DROP TRIGGER to the Galera channel and at the same time ALTER DATABASE is replicated,
connection_1 waits for connection_2 to finish. connection_2 tries to acquire MDL lock and should BF-abort connection_1, but as it waits in Galera for its turn, it is not possible to release MDL locks. We end up in deadlock state.

I found that CREATE TRIGGER is affected in the same way.

Solution:
After initial checks in CREATE/DROP TRIGGER, release all MDL locks, replicate TOI and re-acquire MDL locks.
This is the best we can do.

### DIFF
--- a/sql/sql_trigger.cc
+++ b/sql/sql_trigger.cc
@@ -444,9 +444,17 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
   }
 
 #ifdef WITH_WSREP
-  if (WSREP(thd) && wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL,
-                                             m_trigger_table, NULL)) {
-    return true;
+  if (WSREP(thd)) {
+    schema_mdl_locker.unlock();
+    assert(!thd->mdl_context.has_locks());
+    if (wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL, m_trigger_table,
+                                 NULL)) {
+      return true;
+    }
+    if (thd->locked_tables_mode &&
+        schema_mdl_locker.ensure_locked(m_trigger_table->db)) {
+      return true;
+    }
   }
 #endif /* WITH_WSREP */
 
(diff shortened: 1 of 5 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/2520/

sql/sql_trigger.cc
```diff
@@ -444,9 +444,16 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
...
+  if (WSREP(thd)) {
+    schema_mdl_locker.unlock();
+    assert(!thd->mdl_context.has_locks());
+    if (wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL, m_trigger_table,
+                                 NULL)) {
+      return true;
+    }
+    if (thd->locked_tables_mode) {
+      schema_mdl_locker.ensure_locked(m_trigger_table->db);
+    }
```
venkatesh-prasad-v: `ensure_locked()` returns bool and any error occured during the re-acquisition also must be handled. Can we have it like this?
```C++
   if (thd->locked_tables_mode &&                                                                                                                                                                               
       schema_mdl_locker.ensure_locked(m_trigger_table->db))                                                                                                                                                    
     return true;
```

kamil-holubicki: done

VarunNagaraju (approved):
LGTM

venkatesh-prasad-v (approved):
LGTM

satya-bodapati:
@kamil-holubicki @venkatesh-prasad-v If you release schema MDL lock, there is no protection from DROP DATABASE, isn't it? What happens, if DROP DATABASE is executed after u release the schema MDL lock in trigger code?

PR #1889, opened 2024-03-29
(8.0) PXC-4403: Wrong field metadata cause assertion in protocol_classic.cc::store_string()

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4403

Problem:
When using debug build, sometimes we hit the following assertion:

mysqld: protocol_classic.cc:3423:
virtual bool Protocol_classic::store_string(const char*, size_t, const CHARSET_INFO*): Assertion `send_metadata || field_types == nullptr || field_types[field_pos] == MYSQL_TYPE_DECIMAL || field_types[field_pos] == MYSQL_TYPE_BIT ||
field_types[field_pos] == MYSQL_TYPE_NEWDECIMAL || field_types[field_pos] == MYSQL_TYPE_NEWDATE ||
field_types[field_pos] == MYSQL_TYPE_JSON ||
(field_types[field_pos] >= MYSQL_TYPE_ENUM && field_types[field_pos] <= MYSQL_TYPE_GEOMETRY)' failed.

Cause:
Investigation shows that field_types[field_pos] has an unknown value. There are two reasons for this:
1. field_types is allocated in Protocol_classic::start_result_metadata() from threads's mem_root. This memory is disposed after the command in Thd::cleanup_after_query(). If we are retrying the query, we are not sending metadata again, so field_types is a dangling pointer.

2. Even after fixing the above issue, the assertion is still hit. This time the investigation shows, that field_types[field_pos] contains not allowed type (e.g. MYSQL_TYPE_LONGLONG). This is because we are not resending metadata if we retry the query. But it may happen that the query is aborted before sending metadata. In such a case metadata of the previous query will be used.

Solution:
1. Allocate field_types from heap.
2. Do not re-send metadata only if it was already sent.

### DIFF
--- a/sql/protocol_classic.cc
+++ b/sql/protocol_classic.cc
@@ -3109,8 +3109,27 @@ bool Protocol_classic::start_result_metadata(uint num_cols_arg, uint flags,
     ::storeXXX() method's asserts failures.
   */
   if (m_thd->variables.resultset_metadata == RESULTSET_METADATA_FULL)
+#ifdef WITH_WSREP
+  {
+    /* field_types will be disposed from
+    dispatch_sql_command() -> Thd::cleanup_after_query() if it is allocated by
+    m_thd->alloc().
+    If we are retrying the query, we are not resending metadata,
+    so fields_types will be a dangling pointer causing problem in store_string()
+    and others referring to it. */
+    enum_field_types *mem =
+        new enum_field_types[sizeof(field_types) * num_cols];
+    field_types_ptr =
+        std::shared_ptr<enum_field_types>(mem, [this](enum_field_types *p) {
+          delete[] p;
+          field_types = nullptr;
+        });
+    field_types = field_types_ptr.get();
+  }
+#else
     field_types =
         (enum_field_types *)m_thd->alloc(sizeof(field_types) * num_cols);
+#endif /* WITH_WSREP */
   else
     field_types = nullptr;
   count = 0;
--- a/sql/query_result.cc
+++ b/sql/query_result.cc
@@ -73,8 +73,8 @@ uint Query_result::field_count(const mem_root_deque<Item *> &fields) const {
 bool Query_result_send::send_result_set_metadata(
     THD *thd, const mem_root_deque<Item *> &list, uint flags) {
 #ifdef WITH_WSREP
-  if (WSREP(thd) && thd->wsrep_retry_query) {
-    /* Metadata is already send during first try that failed so avoid
+  if (WSREP(thd) && thd->wsrep_retry_query && is_result_set_started) {
+    /* If metadata is already sent during first try that failed so avoid
     resending it. Just plan to send the result. */
     WSREP_DEBUG("Skip resending metadata if query is being re-tried");
     return false;
--- a/sql/protocol_classic.h
+++ b/sql/protocol_classic.h
@@ -36,6 +36,10 @@
 #include "sql/protocol.h"  // Protocol
 #include "violite.h"
 
+#ifdef WITH_WSREP
+#include <memory>
+#endif
+
 class Item_param;
 class Send_field;
 class String;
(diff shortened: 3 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/job/pxc-8.0-pipeline-parallel-mtr/2523

sql/protocol_classic.cc
```diff
@@ -3108,10 +3108,27 @@ bool Protocol_classic::start_result_metadata(uint num_cols_arg, uint flags,
...
+#ifdef WITH_WSREP
+    /* field_types will be disposed from
+    dispatch_sql_command() -> Thd::cleanup_after_query() if it is allocated by
+    m_thd->alloc().
+    If we are retrying the query, we are not resending metadata,
+    so fields_types will be a dangling pointer causing problem in store_string()
+    and others referring to it. */
+    enum_field_types *mem =
+        new enum_field_types[sizeof(field_types) * num_cols];
+    field_types_ptr =
```
venkatesh-prasad-v, line 3122: Please check if this will work in release builds. As i see, `field_types_ptr` is declared inside `#ifndef NDEBUG`  which is for debug builds IIRC.

sql/query_result.cc
```diff
@@ -73,8 +73,8 @@ uint Query_result::field_count(const mem_root_deque<Item *> &fields) const {
 bool Query_result_send::send_result_set_metadata(
     THD *thd, const mem_root_deque<Item *> &list, uint flags) {
 #ifdef WITH_WSREP
-  if (WSREP(thd) && thd->wsrep_retry_query) {
-    /* Metadata is already send during first try that failed so avoid
+  if (WSREP(thd) && thd->wsrep_retry_query && is_result_set_started) {
+    /* If metadata is already send during first try that failed so avoid
```
venkatesh-prasad-v:
```suggestion
    /* If metadata is already sent during first try that failed so avoid
```

sql/protocol_classic.cc
```diff
@@ -3108,10 +3108,27 @@ bool Protocol_classic::start_result_metadata(uint num_cols_arg, uint flags,
     Set it to NULL if we skip resultset metadata to avoid
     ::storeXXX() method's asserts failures.
   */
-  if (m_thd->variables.resultset_metadata == RESULTSET_METADATA_FULL)
+  if (m_thd->variables.resultset_metadata == RESULTSET_METADATA_FULL) {
+#ifdef WITH_WSREP
```
venkatesh-prasad-v: Can we have it this way?
```suggestion
  if (m_thd->variables.resultset_metadata == RESULTSET_METADATA_FULL)
#ifdef WITH_WSREP
  {
```

sql/protocol_classic.cc
```diff
@@ -3108,10 +3108,27 @@ bool Protocol_classic::start_result_metadata(uint num_cols_arg, uint flags,
...
+          delete[] p;
+          field_types = nullptr;
+        });
+    field_types = field_types_ptr.get();
+#else
     field_types =
         (enum_field_types *)m_thd->alloc(sizeof(field_types) * num_cols);
-  else
+#endif
+  } else
```
venkatesh-prasad-v: Can we have it this way?
```suggestion
}
#endif
  else
```

sql/protocol_classic.cc
kamil-holubicki: Yes, It is for debug build only. `field_types` is for debug-only as well.
Here we are inside `#ifndef NDEBUG` block.

sql/query_result.cc
kamil-holubicki: fixed

sql/protocol_classic.cc
kamil-holubicki: sure, original code does not had curly brackets. Fixed

sql/protocol_classic.cc
kamil-holubicki: fixed

venkatesh-prasad-v (approved):
LGTM

VarunNagaraju (approved):
LGTM

PR #1893, opened 2024-04-01
PXC-4341: Executing prepared statement can abort node after FLUSH TABLES (trunk)

venkatesh-prasad-v:
Merge remote-tracking branch 'venki/PXC-4341-8.0' into PXC-4341-trunk
    
    ***
    Problem
    -------
    Transient errors during the execution of the DDL in a prepared statement
    mode can trigger inconsistency voting resulting in node eviction from
    the cluster.
    
    Analysis & Fix
    --------------
    It can sometimes happen that the DDL executed as a prepared statement
    may fail with transient errors like ER_NEED_REPREPARE if the table
    metadata was changed between PREPARE and EXECUTE (due to FLUSH TABLES).
    In such cases, the statement is re-prepared and re-executed from
    Prepared_statement::execute_loop().
    
    However, before the statement is reprepared, PXC's TOI_end() hook captures
    this transient error (i.e, ER_NEED_REPREPARE) and sends it to galera thereby
    triggering the inconsistenct voting and kicking this node out of the
    cluster.
    
    So, in such cases, we need to delay sending the final error to galera until
    all reprepare attempts are over. This is done by attaching a temporary
    Diagnostics_area to THD and resetting it later if all the required conditions
    are met.
    
    NOTE
    ----
    
    The below mentioned commands reprepare the statement if there is any
    metadata change (marked with CF_REEXECUTION_FRAGILE)
    
        SQLCOM_CREATE_TABLE
        SQLCOM_LOAD
        SQLCOM_CREATE_VIEW
        SQLCOM_UPDATE
        SQLCOM_UPDATE_MULTI
        SQLCOM_INSERT
        SQLCOM_INSERT_SELECT
        SQLCOM_DELETE
        SQLCOM_DELETE_MULTI
        SQLCOM_REPLACE
        SQLCOM_REPLACE_SELECT
        SQLCOM_SELECT
        SQLCOM_DO
        SQLCOM_CALL
    
    Among the above commands, only SQLCOM_CREATE_TABLE and
    SQLCOM_CREATE_VIEW are DDLs and thus was affected by this bug.
    
    However in case of CREATE VIEW, table version check is performed before
    the query is replicated as TOI, so CREATE TABLE was the only query that
    was affected by this bug. So, a special handling has been done only for
    the CREATE_TABLE query in this patch.

### DIFF
--- a/mysql-test/suite/galera_3nodes/r/pxc_prepared_statement_toi_failure.result
+++ b/mysql-test/suite/galera_3nodes/r/pxc_prepared_statement_toi_failure.result
@@ -1,0 +1,28 @@
+PREPARE stmt1 FROM "CREATE TABLE IF NOT EXISTS pxc4341 (id INT PRIMARY KEY AUTO_INCREMENT, a INT)";
+EXECUTE stmt1;
+include/assert.inc [Table pxc4341 was created on node1.]
+FLUSH TABLES;
+EXECUTE stmt1;
+Warnings:
+Note	1050	Table 'pxc4341' already exists
+SHOW TABLES;
+Tables_in_test
+pxc4341
+INSERT INTO pxc4341 VALUES(1,1),(2,2),(3,3);
+include/assert.inc [Table pxc4341 was created on node2.]
+include/assert.inc [Node 2 is up, running and replicating.]
+include/assert.inc [Table pxc4341 was created on node3.]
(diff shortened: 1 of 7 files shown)


VarunNagaraju (approved):
LGTM

kamil-holubicki (approved):

PR #1894, opened 2024-04-01
PXC-4365: PXC nodes leave cluster when row size is too large and has more than 3 NVARCHAR columns (trunk)

venkatesh-prasad-v:
https://perconadev.atlassian.net/browse/PXC-4365

Problem
-------
Server evicts from the cluster if the query failed with a long error message > 1024 characters.

Analysis
--------
While creating a table, if the table structure has any NVARCHAR columns, the server will generate a warning saying

    NATIONAL/NCHAR/NVARCHAR implies the character set UTF8MB3, which will
    be replaced by UTF8MB4 in a future release. Please consider using
    CHAR(x) CHARACTER SET UTF8MB4 in order to be unambiguous.

If row length of the new table exceeds the max row length supported by the storage engine, then the query will fail with error

    Row size too large. The maximum row size for the used table type, not
    counting BLOBs, is 65535. This includes storage overhead, check the
    manual. You have to change some columns to TEXT or BLOBs

When we try to create a table having nvarchar columns and if the total row length exceeds the value supported by storage enging, then the query fails with both warnings and error.

When such a query fails in a clustered environment, warnings and errors generated by the query are concatenated into a single string and is sent to the cluster to peform inconsistency voting.

However, if this concatenated error message has more than 991 characters, then it undergoes truncation to 991 characters which caused the voting mechanism to calculate different votes, eventually resulting in node eviction from the cluster.

Solution
--------
The size of the buffer (used while sending the vote message to the cluster) has been increased to 2070 characters (2048 for error message + 32 for coded message) to accomodate even the longest error message sent by server.

(cherry picked from commit d92abd5161b19c32210052442ecbb3b64db8ced2)

### DIFF
--- a/mysql-test/suite/galera/r/pxc_long_error_message.result
+++ b/mysql-test/suite/galera/r/pxc_long_error_message.result
@@ -1,0 +1,45 @@
+CREATE TABLE TestTable (
+ID INT NOT NULL,
+A1 VARCHAR(512),
+A2 VARCHAR(512),
+A3 VARCHAR(512),
+A4 VARCHAR(512),
+A5 VARCHAR(512),
+A6 VARCHAR(512),
+A7 VARCHAR(512),
+A8 VARCHAR(512),
+A9 VARCHAR(512),
+A10 VARCHAR(512),
+A11 VARCHAR(512),
+A12 VARCHAR(512),
+A13 VARCHAR(512),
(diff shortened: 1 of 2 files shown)


VarunNagaraju (approved):
LGTM

kamil-holubicki (approved):

PR #1896, opened 2024-04-03
(trunk) PXC-4403: Wrong field metadata cause assertion in protocol_classic.cc::store_string()

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4403

Problem:
When using debug build, sometimes we hit the following assertion:

mysqld: protocol_classic.cc:3423:
virtual bool Protocol_classic::store_string(const char*, size_t,
const CHARSET_INFO*): Assertion `send_metadata || field_types == nullptr
|| field_types[field_pos] == MYSQL_TYPE_DECIMAL ||
field_types[field_pos] == MYSQL_TYPE_BIT ||
field_types[field_pos] == MYSQL_TYPE_NEWDECIMAL ||
field_types[field_pos] == MYSQL_TYPE_NEWDATE ||
field_types[field_pos] == MYSQL_TYPE_JSON ||
(field_types[field_pos] >= MYSQL_TYPE_ENUM && field_types[field_pos]
<= MYSQL_TYPE_GEOMETRY)' failed.

Cause:
Investigation shows that field_types[field_pos] has unknown value.
There are two reasons for this:
1. field_types is allocated in Protocol_classic::start_result_metadata()
from threads's mem_root. This memory is disposed after the command in
Thd::cleanup_after_query(). If we are retrying the query, we are not
sending metadata again, so field_types is a dangling pointer.

2. Even after fixing the above issue, the assertion is still hit.
This time the investigation shows, that field_types[field_pos] contains
not allowed type (e.g. MYSQL_TYPE_LONGLONG). This is because we are not
resending metadata if we retry the query. But it may happen that the
query is aborted before sending metadata. In such a case metadata of
the previous query will be used.

Solution:
1. Allocate field_types from heap.
2. Do not re-send metadata only if it was already sent.

### DIFF
--- a/sql/protocol_classic.cc
+++ b/sql/protocol_classic.cc
@@ -3109,8 +3109,27 @@ bool Protocol_classic::start_result_metadata(uint num_cols_arg, uint flags,
     ::storeXXX() method's asserts failures.
   */
   if (m_thd->variables.resultset_metadata == RESULTSET_METADATA_FULL)
+#ifdef WITH_WSREP
+  {
+    /* field_types will be disposed from
+    dispatch_sql_command() -> Thd::cleanup_after_query() if it is allocated by
+    m_thd->alloc().
+    If we are retrying the query, we are not resending metadata,
+    so fields_types will be a dangling pointer causing problem in store_string()
+    and others referring to it. */
+    enum_field_types *mem =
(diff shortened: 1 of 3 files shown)


VarunNagaraju (approved):
LGTM

venkatesh-prasad-v (approved):
LGTM

PR #1927, opened 2024-06-12
PXC-4173: PXC node stalls with parallel replication workers executing DDLs via async node

venkatesh-prasad-v:

    Problem
    =======
    PXC replica could enter into a deadlock with multi threaded replication when
    --replica-preserve-commit-order is enabled.
    
    The deadlock is more likely to occur on a replica with
    --replica-preserve-commit-order enabled when the source server's workload
    consists of concurrent executions of DMLs and DDLs. In such a scenario, the
    replica shall try to commit them in the same order as that of the source which
    will conflict with galera's ordering resulting in a deadlock.
    
    MySQL ordering: Commits are serialized as per their commit order on source server.
    Galera ordering: Commits are serialized in a FIFO manner.
    
    Scenario
    ========
    Let's assume we have the following two transactions in the relay log with same
    last_committed.
    
    T1: DML, seqno=10
    T2: DDL, seqno=11
    
    +-----------+--------------+-------------------------------------------------------+
    | Timestamp |    Thread    |                        Activity                       |
    +-----------+--------------+-------------------------------------------------------+
    |     t0    | Co-ordinator | Assigns T1 to Applier-1                               |
    +-----------+--------------+-------------------------------------------------------+
    |     t1    | Co-ordinator | Assigns T2 to Applier-2                               |
    +-----------+--------------+-------------------------------------------------------+
    |     t2    |   Applier-1  | DML starts with seqno = 10                            |
    +-----------+--------------+-------------------------------------------------------+
    |     t3    |   Applier-2  | DDL starts with seqno = 11                            |
    +-----------+--------------+-------------------------------------------------------+
    |     t4    |   Applier-2  | DDL calls TOI_begin(), acquires commit monitor        |
    +-----------+--------------+-------------------------------------------------------+
    |     t5    |   Applier-2  | DDL executes                                          |
    +-----------+--------------+-------------------------------------------------------+
    |     t5    |   Applier-1  | DML executes                                          |
    +-----------+--------------+-------------------------------------------------------+
    |     t6    |   Applier-2  | DDL reaches commit_order_manager, finds that          |
    |           |              | it needs to wait until Applier-1 is committed         |
    +-----------+--------------+-------------------------------------------------------+
    |     t7    |   Applier-1  | DML reaches commit, calls commit_order_enter_local()  |
    |           |              | and will wait for Applier-2 to release commit monitor |
    +-----------+--------------+-------------------------------------------------------+
    
    In the end it will result in
    Applier-1: DML executing Xid_log_event::commit waiting for DDL to release commit order
    Applier-2: DDL executing ALTER TABLE waiting for applier-1 to commit first
    
    Solution
    ========
    This commit introduces "Async Monitor" at server layer to ensure that the
    ordering in galera is done in the same order in the relay log.
    
    Async Monitor is similar to Commit Monitor used in galera, but differs in how
    seqnois are tracked within the monitor.
    
    The Async Monitor has the following methods.
    1. schedule() - Schedules the transactions in as per their order in the
                    the relay log.
    2. enter() - Ensures that threads are serialized as per their order.
    3. leave() - Ensures that threads are removed from the monitor in the serialized order.
    4. skip()  - Provides some flexibilty to the monitor to skip transactions. This
                 will be required for handling skipped and filtered transactions.
    
    Lifetime of the Async Monitor:
    Async Monitor is created on START REPLICA and is destroyed on STOP REPLICA, or
    whenever applier thread exits.


sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,38 @@
...
+  // Wait for its turn before entering
+  m_cond.wait(lock, [this, seqno]() { return seqno == m_last_left + 1; });
+  m_last_entered = (seqno > m_last_entered) ? seqno : m_last_entered;
+  fprintf(stderr, "Entered the monitor with seqno: %llu\n", seqno);
+}
+
+void Wsrep_async_monitor::leave(seqno_t seqno) {
+  // Wait for its turn before leaving
+  std::unique_lock<std::mutex> lock(m_mutex);
+  m_cond.wait(lock, [this, seqno]() { return seqno == m_last_left + 1; });
```
kamil-holubicki: Is this cond.wait needed? I would add assert instead of this. The only valid situation is when the one who leaves monitor is 1 more than the previous one.

sql/wsrep_mysqld.cc
```diff
@@ -2870,6 +2870,57 @@ static void wsrep_RSU_end(THD *thd) {
   thd->disable_binlog_guard.reset();
 }
 
+bool thd_enter_apply_monitor(THD* thd) {
```
kamil-holubicki: return value is not evaluated by the caller

sql/wsrep_mysqld.cc
```diff
@@ -2870,6 +2870,57 @@ static void wsrep_RSU_end(THD *thd) {
...
+
+  if (thd->killed != THD::NOT_KILLED) {
```
kamil-holubicki: shouldn't it be under if (thd->slave_thread)?

sql/wsrep_mysqld.cc
```diff
@@ -2870,6 +2870,57 @@ static void wsrep_RSU_end(THD *thd) {
...
+    return true;
+  }
+  if (thd->slave_thread) {
+    Slave_worker * sw = dynamic_cast<Slave_worker*>(thd->rli_slave);
```
kamil-holubicki: Can we use static_cast here? If not, we should check for sw not being nullptr

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,38 @@
...
+#include "sql/wsrep_async_monitor.h"
+#include <algorithm>
+#include <cassert>
+
+void Wsrep_async_monitor::enter(seqno_t seqno) {
+  std::unique_lock<std::mutex> lock(m_mutex);
+
+  // Wait for its turn before entering
+  m_cond.wait(lock, [this, seqno]() { return seqno == m_last_left + 1; });
+  m_last_entered = (seqno > m_last_entered) ? seqno : m_last_entered;
```
kamil-holubicki: I don't understand this. Is it possible that seqno <= m_last_entered? I think the only possibility enforced by above condition.wait is that seqno grows monolithically.

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,38 @@
...
+  fprintf(stderr, "Entered the monitor with seqno: %llu\n", seqno);
```
kamil-holubicki: Remember to remove fprintfs in the final version

sql/wsrep_mysqld.cc
venkatesh-prasad-v: `Slave_worker` is a child class of `Relay_log_info` which is the type of `rli_slave`. So `dynamic_cast` is the better option here.

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Hi! In an ideal case, it is not possible to have seqno < m_last_entered. I will change it just to `m_last_entered = seqno`

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Sure. They are only added for QA. Once we get a green flag from the QA. I will remove these prints.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Agree. Moved it to under `if (thd->slave_thread)`.

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Done

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Changed the return type to `void` as `thd->killed` is the only condition that performs early return and it will be handled by the caller functions.

sql/log_event.cc
```diff
@@ -2704,6 +2704,14 @@ Slave_worker *Log_event::get_slave_worker(Relay_log_info *rli) {
 
           Gtid_log_event *gtid_log_ev = static_cast<Gtid_log_event *>(this);
           rli->started_processing(gtid_log_ev);
+#ifdef WITH_WSREP
+          Wsrep_async_monitor *wsrep_async_monitor {rli->get_wsrep_async_monitor()};
+          if (wsrep_async_monitor) {
+            auto seqno = gtid_log_ev->sequence_number;
+            wsrep_async_monitor->schedule(seqno);
+            fprintf(stderr, "Scheduled the seqno: %llu in async monitor\n", seqno);
```
kamil-holubicki: fprintf
multiple occurences

sql/rpl_gtid_execution.cc
```diff
@@ -37,6 +37,9 @@
 #include "sql/mysqld.h"  // connection_events_loop_aborted
 #include "sql/rpl_gtid.h"
 #include "sql/rpl_rli.h"    // Relay_log_info
+#ifdef WITH_WSREP
```
kamil-holubicki: Can we keep WSREP-dependant includes after the standard ones, like we do in other cases?

sql/rpl_rli.h
```diff
@@ -63,6 +63,9 @@
 #include "sql/sql_class.h"    // THD
 #include "sql/system_variables.h"
 #include "sql/table.h"
+#ifdef WITH_WSREP
+#include "sql/wsrep_async_monitor.h"
```
kamil-holubicki: It looks like forward type declaration is enough, and the header can be included in *.cc file

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,94 @@
...
+    return !scheduled_seqnos.empty() && scheduled_seqnos.front() == seqno;
+  });
+}
+
+// Method to be called after DDL/DML processing is complete
+void Wsrep_async_monitor::leave(seqno_t seqno) {
+  std::unique_lock<std::mutex> lock(m_mutex);
+
+  // Check if the sequence number matches the front of the queue.
+  // In a correctly functioning monitor this should not happen
```
kamil-holubicki: In a correctly functioning monitor this should not happen ->
In a correctly functioning monitor this should always be true

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,94 @@
...
+void Wsrep_async_monitor::skip(unsigned long seqno) {
+  std::unique_lock<std::mutex> lock(m_mutex);
+
+  // Check if the seqno is already marked as skipped
+  if (skipped_seqnos.count(seqno) > 0) {
+    return;  // Already skipped, so do nothing
+  }
+
+  // Mark the seqno as skipped
+  skipped_seqnos.insert(seqno);
```
kamil-holubicki, line 114: I can't see any place where seqnos are removed from `skipped_seqnos`

sql/wsrep_mysqld.cc
```diff
@@ -2858,6 +2858,61 @@ static void wsrep_RSU_end(THD *thd) {
...
+    // If the thread is already killed, leave it to the called to handle it.
+    if (thd->killed != THD::NOT_KILLED) {
+      return;
+    }
+    Slave_worker *sw = dynamic_cast<Slave_worker*>(thd->rli_slave);
+    Wsrep_async_monitor *wsrep_async_monitor {sw->get_wsrep_async_monitor()};
+    if (wsrep_async_monitor) {
+      auto seqno = sw->sequence_number();
+      assert(seqno > 0);
+      // TODO: If requied, we must set current_mutex and current_cond here
```
kamil-holubicki, line 2879: TODO

sql/wsrep_mysqld.cc
```diff
@@ -2858,6 +2858,61 @@ static void wsrep_RSU_end(THD *thd) {
...
+      // i.e, thd->enter_cond();
+      wsrep_async_monitor->enter(seqno);
+    } else {
+      // Error out and exit if we have more than one parallel worker and async monitor
+      // is not set yet.
+      if (opt_replica_preserve_commit_order && opt_mts_replica_parallel_workers > 1) {
+        WSREP_ERROR("Async replica thread %u is not monitored by wsrep async monitor",
+                     thd->thread_id());
+        exit(1);
```
kamil-holubicki: exit?
maybe unireg_abort(), or assert and handle it somehow?

sql/wsrep_mysqld.cc
```diff
@@ -2858,6 +2858,61 @@ static void wsrep_RSU_end(THD *thd) {
...
+        else
+          fprintf(stderr, "DML: thd_leave_apply_monitor: seqno=%llu\n", seqno);
+      }
+    } else {
+      // Error out and exit if we have more than one parallel worker and async monitor
+      // is not set yet.
+      if (opt_replica_preserve_commit_order && opt_mts_replica_parallel_workers > 1) {
+        WSREP_ERROR("Async replica thread %u is not monitored by wsrep async monitor",
+                     thd->thread_id());
+        exit(1);
```
kamil-holubicki: exit?

sql/log_event.cc
```diff
@@ -2704,6 +2704,14 @@ Slave_worker *Log_event::get_slave_worker(Relay_log_info *rli) {
 
           Gtid_log_event *gtid_log_ev = static_cast<Gtid_log_event *>(this);
           rli->started_processing(gtid_log_ev);
+#ifdef WITH_WSREP
+          Wsrep_async_monitor *wsrep_async_monitor {rli->get_wsrep_async_monitor()};
+          if (wsrep_async_monitor) {
+            auto seqno = gtid_log_ev->sequence_number;
+            wsrep_async_monitor->schedule(seqno);
```
kamil-holubicki, line 2712: Is my understandig correct, that in this point transactions appear in the order they were committed on the source node?
We create the queue of ordered transactions, and will use AsyncMonitor to enforce this order of commit in Galera, right?

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,94 @@
...
+
+#ifdef WITH_WSREP
+#include "sql/wsrep_async_monitor.h"
+#include <algorithm>
+#include <cassert>
+
+// Method for main thread to add scheduled seqnos
+void Wsrep_async_monitor::schedule(seqno_t seqno) {
+  std::unique_lock<std::mutex> lock(m_mutex);
+  scheduled_seqnos.push(seqno);
```
kamil-holubicki, line 25: seqons should be growing monolithically, right?
Can we add an assert like seqno = prev+1?
If not monolithically growing (why?), at least let's add an assert seqno > prev.
A good code comment would be appreciated

sql/wsrep_async_monitor.cc
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Method for both DDL and DML to enter the monitor
+void Wsrep_async_monitor::enter(seqno_t seqno) {
+  std::unique_lock<std::mutex> lock(m_mutex);
+
+  // Wait until this transaction is at the head of the scheduled queue
+  m_cond.wait(lock, [this, seqno] {
+    // Remove skipped transactions
```
kamil-holubicki: When this can happen? Could you add a code comment? 
I suspect this is the case when A, B, C have been scheduled, then B skipped, but skip() removes only from the front of scheduled_seqno queue

kamil-holubicki (changes requested):

kamil-holubicki:
The commit message says about ordering transactions basing on their seqnos i not way this = prev + 1. But the code does it in a different way (I can understand that logically it is still the same). It would be good to add more details about how the current implementation actually works.

sql/log_event.cc
venkatesh-prasad-v: I've added it for debugging purpose. I will remove all of them in the final version.

sql/rpl_gtid_execution.cc
venkatesh-prasad-v: Done

sql/rpl_rli.h
venkatesh-prasad-v: Done

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Done.

sql/log_event.cc
venkatesh-prasad-v: Yes. `schedule()`  is for updating the `scheduled_seqnos`  queue in the montior. This order defines the order in which the transactions will be entering the monitor.

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: In general, sequence numbers are generated monotonically. However that's only true within one binary log.  When the binlog gets rotated, the sequence number is reset and the first transaction in the newly created binlog will again start from seq no `1`. So, there is no need to add the assert `seqno = prev+1`.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Good catch. Fixed.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: Fixed

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Fixed it.

sql/wsrep_mysqld.cc
venkatesh-prasad-v: I would prefer having it as-is, as adding a new PFS state and using it with `enter_cond()` would make the thread's status visible in the SHOW PROCESSLIST. For example: we could make the thread to display `Waiting for async monitor` when it is waiting for its turn.

sql/wsrep_async_monitor.cc
venkatesh-prasad-v: Yes. Imagine a scenario where scheduled seqnos is 1,2,3 and threads enter in an out of order manner.

- 3 enters the monitor first, since it is not in the front of the queue, it goes to cond_wait
- 2 enters the monitor, adds 2 to `skipped_seqnos` 
- 1 enters the monitor, since it is in the front, it acquires the monitor, does its job and leaves the monitor by removing itself from the `scheduled_seqnos` and signals 3.
- 3 wakes up, but it will see that 2 in the front of the queue. To proceed, it needs to remove all the skipped seqnos in the `scheduled_seqnos` queue.

venkatesh-prasad-v:
@kamil-holubicki Description is updated

sql/wsrep_mysqld.cc
kamil-holubicki: I'm just referring to leaving a TODO comment. Either implement it or remove the comment

sql/wsrep_async_monitor.cc
kamil-holubicki: Could you add a comment to the code explaining the purpose of this part?

venkatesh-prasad-v: Done

venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-param-parallel-mtr/248/consoleFull

kamil-holubicki (approved):
LGTM

PR #1931, opened 2024-06-14
PXC-4428: Garbd terminates SST script's children after SST is complete

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4428

MTR tests


venkatesh-prasad-v (approved):
Approved!

PR #1950, opened 2024-09-12
PXC-4433: Create file with current MySQL state

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4433

server.state file containing current server state is being created in data directory if the server is started with
create_server_state_file=ON option.
create_server_state_file default value is OFF.


kamil-holubicki:
Not needed. The same can be achieved via already existing socket.

PR #1963, opened 2024-10-09
PXC-4500: When innodb_thread_concurrency is set, the cluster can get stuck during SST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4500

Problem 1:
When innodb_thread_concurrency is set, and the node has a heavy user workload, it can become stuck when it receives a CC event from Galera.

Cause:
Let's say innodb_thread_concurrency = 2.
1. We have two user threads that have been granted access to InnoDB. They will release the InnoDB access lock after Galera certification, but for certification, they need to acquire LocalMonitor.
2. At the same time CC happens. The Applier thread acquires LocalMonitor and notifies the application about the new view. The application tries to store the view in wsrep schema. Before doing that it turns off wsrep_on thread's variable (the thread is not marked as wsrep-enabled thread). Then, it tries to enter InnoDB. As we already have two user threads in InnoDB and our thread is not wsrep thread, we have to wait.

The above results in a deadlock:
1. The user thread is holding InnoDB lock, waiting for LocalMonitor
2. The Applier thread is holding LocalMonitor, waiting for InnoDB lock

Solution:
The wrong condition was used to detect the wsrep applier thread in innobase_srv_conc_enter_innodb(). The applier thread should always be granted access. Fixed.

Problem 2:
Even after fixing Problem 1, the cluster was stuck during SST.

Cause:
The SST thread creates an SST user. For this, it needs to enter InnoDB. We end up in a similar situation as in Problem 1. The applier thread holds LocalMonitor and waits for SST to finish. The SST thread waits for InnoDB. User threads hold the InnoDB lock and wait for LocalMonitor.

Solution:
Allow SST thread to enter InnoDB always.

### DIFF
--- a/mysql-test/suite/galera/t/galera_innodb_thread_concurrency.test
+++ b/mysql-test/suite/galera/t/galera_innodb_thread_concurrency.test
@@ -1,0 +1,65 @@
+# Thest that Galera events processing is not blocked by user threads
+# that have been granted InnoDB access when innodb_thread_concurrency
+# is limited
+
+--source include/have_debug.inc
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+--connect node_1a, 127.0.0.1, root, , test, $NODE_MYPORT_1
+--connect node_1b, 127.0.0.1, root, , test, $NODE_MYPORT_1
+
+# Stop node 2 and force SST
+--connection node_2
+--source include/shutdown_mysqld.inc
+
+# Remove the grastate.dat file to force an SST
+--remove_file $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+
+--connection node_1
+CREATE TABLE t1 (a INT PRIMARY KEY);
+SET debug_sync='ib_after_row_insert SIGNAL insert1_done WAIT_FOR continue';
+--send INSERT INTO t1 VALUES (0)
+
+--connection node_1a
+SET debug_sync='ib_after_row_insert SIGNAL insert2_done WAIT_FOR continue';
+--send INSERT INTO t1 VALUES (1)
+
+--connection node_1b
+SET debug_sync='now WAIT_FOR insert1_done';
+SET debug_sync='now WAIT_FOR insert2_done';
+
+# Now we have 2 threads granted access to InnoDB
+# No more threads allowed (innodb_thread_concurrency=2)
+# But wsrep applier and SST donor thread should be the exemption
+
+# Start node_2
+--connection node_2
+--source include/start_mysqld.inc
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size';
+--source include/wait_condition.inc
+
+# Now let paused inserts to continue
+--connection node_1b
+SET debug_sync='now SIGNAL continue';
+SET debug_sync='now SIGNAL continue';
+
+--connection node_1
+--reap
+--connection node_1a
+--reap
+
+# Replication should still work
+INSERT INTO t1 VALUES (2);
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 3 FROM t1
+--source include/wait_condition.inc
+
+# cleanup
+--connection node_1
+DROP TABLE t1;
+
+--disconnect node_1a
+--disconnect node_1b
+--source include/wait_until_count_sessions.inc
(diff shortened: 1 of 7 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/2895/

mysql-test/suite/galera/t/galera_innodb_thread_concurrency.test
```diff
@@ -0,0 +1,65 @@
...
+SET debug_sync='ib_after_row_insert SIGNAL insert2_done WAIT_FOR continue';
+--send INSERT INTO t1 VALUES (1)
+
+--connection node_1b
+SET debug_sync='now WAIT_FOR insert1_done';
+SET debug_sync='now WAIT_FOR insert2_done';
+
+# Now we have 2 threads granted access to InnoDB
+# No more threads allowed (innodb_thread_concurrency=2)
+# But wsrep applier and SST donor thread should be the excemption
```
venkatesh-prasad-v:
```suggestion
# But wsrep applier and SST donor thread should be the exemption
```

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2068,8 +2068,14 @@ static inline dberr_t innobase_srv_conc_enter_innodb(row_prebuilt_t *prebuilt) {
...
-  if (wsrep_on(prebuilt->trx->mysql_thd) &&
-      wsrep_thd_is_BF(prebuilt->trx->mysql_thd, false))
+  // Actions executed from the sst_donor thread should not be blocked as well.
+  // sst_donor thread runs with Galera LocalMonitor acquired.
+  // If we have user threads entered innodb and about to commit, they will try
+  // to acquire LocalMonitor as well, but will block. So without letting
+  // sst_donor thread to move on here, we would end up in deadlock.
+  THD *thd = prebuilt->trx->mysql_thd;
+  if (thd && WSREP_ON &&
+      (wsrep_thd_is_applying(thd)|| thd->wsrep_applier || thd->wsrep_sst_donor))
```
venkatesh-prasad-v:
```suggestion
      (wsrep_thd_is_applying(thd) || thd->wsrep_applier || thd->wsrep_sst_donor))
```

venkatesh-prasad-v: The earlier code made a call to `wsrep_thd_is_BF()` which used the following condition
```
114     status = (wsrep_thd_is_applying(thd) || wsrep_thd_is_toi(thd) ||                                
115     |   |   | wsrep_thd_is_in_rsu(thd)) || wsrep_thd_is_in_nbo(thd);     
```

However, in the current logic we only have 
```
(wsrep_thd_is_applying(thd)|| thd->wsrep_applier || thd->wsrep_sst_donor)
```
that means we no longer check for TOI, RSU and NBO.

storage/innobase/handler/ha_innodb.cc
```diff
@@ -2108,8 +2114,9 @@ static inline void innobase_srv_conc_exit_innodb(row_prebuilt_t *prebuilt) {
   // innodb_thread_concurreny limit how many thread can work in innodb
   // at any given time. This limit is not applicable to wsrep-applier
   // threads given they are high priority threads.
-  if (wsrep_on(prebuilt->trx->mysql_thd) &&
-      wsrep_thd_is_BF(prebuilt->trx->mysql_thd, false))
+  THD *thd = prebuilt->trx->mysql_thd;
+  if (thd && WSREP_ON &&
+      (wsrep_thd_is_applying(thd)|| thd->wsrep_applier || thd->wsrep_sst_donor))
```
venkatesh-prasad-v: Same as above

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/galera_innodb_thread_concurrency.test
kamil-holubicki: fixed

storage/innobase/handler/ha_innodb.cc
kamil-holubicki: Good catch. Fixed.

storage/innobase/handler/ha_innodb.cc
kamil-holubicki: Fixed

venkatesh-prasad-v (approved):

PR #1964, opened 2024-10-16
(trunk) PXC-4500: When innodb_thread_concurrency is set, the cluster can get stuck during SST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4500

Problem 1:
When innodb_thread_concurrency is set and the node has a heavy user
workload, it can become stuck when it receives a CC event from Galera.

Cause:
Let's say innodb_thread_concurrency = 2.
1. We have two user threads that have been granted access to InnoDB.
They will release the InnoDB access lock after Galera certification,
but for certification, they need to acquire LocalMonitor.
2. At the same time CC happens. The Applier thread acquires LocalMonitor
and notifies the application about the new view. The application tries
to store the view in wsrep schema. Before doing that it turns off
wsrep_on thread's variable (the thread is not marked as wsrep-enabled
thread). Then, it tries to enter InnoDB. As we already have two user
threads in InnoDB and our thread is not wsrep thread, we have to wait.

The above results in a deadlock:
1. The user thread is holding InnoDB lock, waiting for LocalMonitor
2. The Applier thread is holding LocalMonitor, waiting for InnoDB lock

Solution:
The wrong condition was used to detect the wsrep applier thread in
innobase_srv_conc_enter_innodb(). The applier thread should always be
granted access. Fixed.

Problem 2:
Even after fixing Problem 1, the cluster was stuck during SST.

Cause:
The SST thread creates an SST user. For this, it needs to enter InnoDB.
We end up in a similar situation as in Problem 1. The applier thread
holds LocalMonitor and waits for SST to finish. The SST thread waits for
InnoDB. User threads hold the InnoDB lock and wait for LocalMonitor.

Solution:
Allow SST thread to enter InnoDB always.

### DIFF
--- a/mysql-test/suite/galera/r/galera_innodb_thread_concurrency.result
+++ b/mysql-test/suite/galera/r/galera_innodb_thread_concurrency.result
@@ -1,0 +1,12 @@
+CREATE TABLE t1 (a INT PRIMARY KEY);
+SET debug_sync='ib_after_row_insert SIGNAL insert1_done WAIT_FOR continue';
+INSERT INTO t1 VALUES (0);
+SET debug_sync='ib_after_row_insert SIGNAL insert2_done WAIT_FOR continue';
+INSERT INTO t1 VALUES (1);
+SET debug_sync='now WAIT_FOR insert1_done';
+SET debug_sync='now WAIT_FOR insert2_done';
+# restart
+SET debug_sync='now SIGNAL continue';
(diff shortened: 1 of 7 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/232/

venkatesh-prasad-v (approved):

PR #1967, opened 2024-10-25
PXC-3113: Custom implementation of row hash calculation

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-3113

When pxc_strict_mode is ENFORCING/MASTER, PXC prohibits writes to tables without primary keys. This is the default setting. When pxc_strict_mode is PERMISSIVE/DISABLED, writes to such tables without PK is allowed. In such a case the row is identifed by MD5 checksum calculated from all columns values. OpenSSL library is used for MD5 calculation.

On platform where FIPS mode is enabled, OpenSSL does not provide the functionality of MD5 calculation. PXC could switch to use different algorithm (like SHA), but that would mean that different version nodes could not cooperate in the same cluster. MD5 is not used for cryptography, only for hashing purpose (any algorithm would work as long as it is the same on all nodes), so using it for this purpose is FIPS compliant.

The solution is to fallback to custom implementation of MD5 hashing for platforms where FIPS mode is enabled.

Unit tests introduced.
No new MTR tests introduced. We have several ones (galera_nopk_*.test) that test usage of PXC without primary keys. When executed on FIPS platform, they should be still OK.

### DIFF
--- a/mysys/my_md5.cc
+++ b/mysys/my_md5.cc
@@ -41,6 +41,10 @@
 #include <openssl/md5.h>
 #endif /* OPENSSL_VERSION_NUMBER >= 0x30000000L */
 
+#ifdef WITH_WSREP
+#include "md5.h"
+#endif
+
 static void my_md5_hash(unsigned char *digest, unsigned const char *buf,
                         size_t len) {
 #if OPENSSL_VERSION_NUMBER >= 0x30000000L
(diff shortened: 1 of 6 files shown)


lukin-oleksiy (approved):
lgtm

mysys/my_md5.cc
```diff
@@ -95,38 +99,86 @@ then a md5-hash (16 bytes) string is generated using the complete record.
...
+
+void wsrep_init_fips_mode() {
+  wsrep_md5_init_fn = &wsrep_md5_init_custom;
+  wsrep_md5_update_fn = &wsrep_md5_update_custom;
+  wsrep_compute_md5_hash_fn = &wsrep_compute_md5_hash_custom;
+}
+
+/* The following function should be used only by unit tests.
+   It switches the backend implementation back to OpenSSL */
+void wsrep_disable_fips_mode() {
```
venkatesh-prasad-v, line 178: If it not too much to ask, can we have the function names as either `wsrep_init_fips_mode() & wsrep_deinit_fips_mode()` or `wsrep_enable_fips_mode() & wsrep_disable_fips_mode()`?

venkatesh-prasad-v (approved):

PR #1971, opened 2024-10-30
PXC-4546: databases_size value in the json file created by the percona_telemetry is zero

kamil-holubicki:
PXC-4545: percona_telemetry is not able to fetch the databases_count with mysql.session user

https://perconadev.atlassian.net/browse/PXC-4546
https://perconadev.atlassian.net/browse/PXC-4545

Created a dedicated user with minimum required privileges for the Percona Telemetry Component.


percona-ysorokin (approved):
LGTM

PR #1983, opened 2024-12-05
PXC-4544: Wrong Error message for 'RESET BINARY LOGS AND GTIDS' on cluster node.

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4544

Updated error log message to not use offensive terminology.


sql/rpl_source.cc
```diff
@@ -1222,11 +1222,13 @@ bool reset_binary_logs_and_gtids(THD *thd, bool unlock_global_read_lock) {
     */
 #ifdef WITH_WSREP
     if (WSREP(thd) && global_gtid_mode.get() > Gtid_mode::OFF) {
-      /* RESET MASTER will initialize GTID sequence, and that would happen
-         locally in this node only, so better reject it
+      /* RESET BINARY LOGS AND GTIDS will initialize GTID sequence, and that
```
satya-bodapati, line 1225: this looks good and simple but I wonder why we disallow? pxc wsrep replication doesn't rely on binlog and gtids ? :thinking:

kamil-holubicki: All nodes have to generate the same sequence of GTIDs. If we reset only one node, they will become inconsistent.

satya-bodapati (approved):
LGTM!

PR #1992, opened 2025-01-06
[8.4] PXC-4173: PXC node stalls with parallel replication workers executing DDLs via async node

venkatesh-prasad-v:

    Problem
    =======
    PXC replica could enter into a deadlock with multi threaded replication when
    --replica-preserve-commit-order is enabled.
    
    The deadlock is more likely to occur on a replica with
    --replica-preserve-commit-order enabled when the source server's workload
    consists of concurrent executions of DMLs and DDLs. In such a scenario, the
    replica shall try to commit them in the same order as that of the source which
    will conflict with galera's ordering resulting in a deadlock.
    
    MySQL ordering: Commits are serialized as per their commit order on source server.
    Galera ordering: Commits are serialized in a FIFO manner.
    
    Scenario
    ========
    Let's assume we have the following two transactions in the relay log with same
    last_committed.
    
    T1: DML, seqno=10
    T2: DDL, seqno=11
    
    +-----------+--------------+-------------------------------------------------------+
    | Timestamp |    Thread    |                        Activity                       |
    +-----------+--------------+-------------------------------------------------------+
    |     t0    | Co-ordinator | Assigns T1 to Applier-1                               |
    +-----------+--------------+-------------------------------------------------------+
    |     t1    | Co-ordinator | Assigns T2 to Applier-2                               |
    +-----------+--------------+-------------------------------------------------------+
    |     t2    |   Applier-1  | DML starts with seqno = 10                            |
    +-----------+--------------+-------------------------------------------------------+
    |     t3    |   Applier-2  | DDL starts with seqno = 11                            |
    +-----------+--------------+-------------------------------------------------------+
    |     t4    |   Applier-2  | DDL calls TOI_begin(), acquires commit monitor        |
    +-----------+--------------+-------------------------------------------------------+
    |     t5    |   Applier-2  | DDL executes                                          |
    +-----------+--------------+-------------------------------------------------------+
    |     t5    |   Applier-1  | DML executes                                          |
    +-----------+--------------+-------------------------------------------------------+
    |     t6    |   Applier-2  | DDL reaches commit_order_manager, finds that          |
    |           |              | it needs to wait until Applier-1 is committed         |
    +-----------+--------------+-------------------------------------------------------+
    |     t7    |   Applier-1  | DML reaches commit, calls commit_order_enter_local()  |
    |           |              | and will wait for Applier-2 to release commit monitor |
    +-----------+--------------+-------------------------------------------------------+
    
    In the end it will result in
    Applier-1: DML executing Xid_log_event::commit waiting for DDL to release commit order
    Applier-2: DDL executing ALTER TABLE waiting for applier-1 to commit first
    
    Solution
    ========
    This commit introduces "Async Monitor" at server layer to ensure that the
    ordering in galera is done in the same order in the relay log.
    
    Async Monitor is similar to Commit Monitor used in galera, but differs in how
    seqnois are tracked within the monitor.
    
    The Async Monitor has the following methods.
    1. schedule() - Schedules the transactions in as per their order in the
                    the relay log.
    2. enter() - Ensures that threads are serialized as per their order.
    3. leave() - Ensures that threads are removed from the monitor in the serialized order.
    4. skip()  - Provides some flexibilty to the monitor to skip transactions. This
                 will be required for handling skipped and filtered transactions.
    
    Lifetime of the Async Monitor:
    Async Monitor is created on START REPLICA and is destroyed on STOP REPLICA, or
    whenever applier thread exits.


kamil-holubicki (approved):
LGTM

venkatesh-prasad-v:
Jenkins: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-param-parallel-mtr/38/console

venkatesh-prasad-v:
Jenkins with test failure fixes: https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-param-parallel-mtr/39/console

kamil-holubicki (approved):

PR #1999, opened 2025-01-21
removing new line

Tusamarco:
fixing clone result report


Tusamarco:
Fix some MTR then we have many of the error coming from test in main. like [.main.main.component-upgrade](https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3054/testReport/junit/ubuntu-noble.Debug.main/main/component_upgrade/) 
These Privileges difference comes from the file <PXC root>/scripts/mysql_system_users.sql.
I tried commenting the changes done there and use previous Grants. Test complete ok, then I put the new grant back and it fails.
To note that the error comes from the check test run at the start and at the end identifying that the privileges at start for user `mysql.pxc.sst.role` are different.

PR #2009, opened 2025-01-28
PXC-4504: DROP PROCEDURE/FUNCTION IF EXISTS generates local GTID event

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4504

Problem:
When DROP PROCEDURE/FUNCTION IF EXISTS is executed for non existing object, the event is binlogged with the GTID containing UUID of local server instead of global, cluster-wide UUID.

This case is similar to the one already fixed in commit cc8ec98a (PXC-4312)

Cause:
DROP PROCEDURE/FUNCTION IF EXISTS is not replicated if the object does not exist, however is binlogged. If the query is not replicated, wsrep logic generates GTID using local server UUID instead of cluster-wide UUID.

Solution:
Replicate DROP PROCEDURE/FUNCTION IF EXISTS always. This is enough for GTID to be properly generated.

### DIFF
--- a/sql/sp.cc
+++ b/sql/sp.cc
@@ -968,7 +968,22 @@ enum_sp_return_code sp_drop_routine(THD *thd, enum_sp_type type,
         name->m_db.str, name->m_name.str, &routine);
   if (error) return SP_INTERNAL_ERROR;
 
+#ifdef WITH_WSREP
+  if (routine == nullptr) {
+    if (thd->lex->drop_if_exists) {
+      /* If 'IF EXISTS' clause is present, we replicate always.
+         In such a case binlogging part is done on the caller level. */
+      if (WSREP(thd) &&
+          wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL, NULL)) {
+        return SP_INTERNAL_ERROR;
+      }
+    }
+    return SP_DOES_NOT_EXISTS;
+  }
+#else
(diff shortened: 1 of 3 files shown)


sql/sp.cc
```diff
@@ -968,7 +968,22 @@ enum_sp_return_code sp_drop_routine(THD *thd, enum_sp_type type,
         name->m_db.str, name->m_name.str, &routine);
   if (error) return SP_INTERNAL_ERROR;
 
+#ifdef WITH_WSREP
+  if (routine == nullptr) {
+    if (thd->lex->drop_if_exists) {
+      /* If 'IF EXISTS' clause is present, we replicate always.
+         In such a case binlogging part is done on the caller level. */
+      if (WSREP(thd) &&
+          wsrep_to_isolation_begin(thd, WSREP_MYSQL_DB, NULL, NULL)) {
```
venkatesh-prasad-v, line 977: I know that we use WSREP_MYSQL_DB (mysql db) for all SPs during TOI. I wonder why don't we use the exact name of the database?
```suggestion
          wsrep_to_isolation_begin(thd, name->m_db.str, NULL, NULL)) {
```

venkatesh-prasad-v (approved):
Other than the comment related to sending database name, I do not have any other comments. Approved!

sql/sp.cc
kamil-holubicki: As discussed on Slack, for now, let's leave it consistent across the current code

PR #2018, opened 2025-01-31
Pxc 4607

Tusamarco:
Fix for  PXC-4607 (synchronize clone plugin persistency)


Tusamarco:
@ParveezBaig @kamil-holubicki can you check it please ? this should fix the 4067

scripts/wsrep_sst_clone.sh
```diff
@@ -358,12 +358,14 @@ setup_clone_plugin()
             WSREP_OFF="" # joiner does not have replication enabled
         fi
         $MYSQL_ACLIENT -e "${WSREP_OFF}INSTALL PLUGIN CLONE SONAME 'mysql_clone.so';"
-        CLEANUP_CLONE_PLUGIN="yes"
+        if [ "$ROLE" == "donor" ]; then
```
kamil-holubicki, line 361: It is not consistent across the script how we do the comparison. Sometimes it is `==` and sometimes it it `=` (just 4 lines above). 
`=` is posix compatible and should be used with single brackets `[...]`
`==` is bash only compatible and should be used with double brackets `[[...]]`

Would it be possible to have it consistent?

scripts/wsrep_sst_clone.sh
```diff
@@ -392,7 +394,7 @@ setup_clone_plugin()
     [ "$SERVER_SSL_KEY" = "NULL" ] && SERVER_SSL_KEY=
     [ "$SERVER_SSL_CA" = "NULL" ] && SERVER_SSL_CA=
 
-    if [ "$CLONE_SSL_CERT" = "NULL" -o "$CLONE_SSL_KEY" = "NULL" ]
+    if [ "$CLONE_SSL_CERT" == "NULL" -o "$CLONE_SSL_CERT" = "" ]
```
kamil-holubicki, line 397: Mixture of single `=` and double `==` in one `if`

kamil-holubicki (changes requested):

scripts/wsrep_sst_clone.sh
Tusamarco: will review all in a different branch/pr

PR #2020, opened 2025-02-02
Pxc 4609

Tusamarco:
fixing PXC-4609


sql/wsrep_sst.cc
```diff
@@ -886,7 +888,16 @@ std::string wsrep_sst_prepare() {
 
   const char *method_ptr(ret.data());
   const char *addr_ptr(ret.data() + strlen(method_ptr) + 1);
-  WSREP_INFO("Prepared SST request: %s|%s", method_ptr, addr_ptr);
+  /* we check for @ as separator in user:pw@address format and return only the address part
```
kamil-holubicki, line 891: Wouldn't it be better to do some regex matching and use `std::regex_replace`?
ChatGPT proposal:
```
#include <iostream>
#include <regex>
#include <string>

std::string maskPassword(const std::string& input) {
    std::regex pattern(R"((\b[^:@]+):([^:@]+)@)"); // Captures 'user:password@'
    return std::regex_replace(input, pattern, "$1:********@"); // Replace only password with '********'
}

int main() {
    std::string input = "clone|clone_sst:8ead8edcf31f0b3c113183462eaabb7aAa@127.0.0.1:23020";
    std::string masked = maskPassword(input);
    std::cout << "Masked Output: " << masked << std::endl;
    return 0;
}
```
The output is: 
`clone|clone_sst:********@127.0.0.1:23020`

It will be a generic solution that masks password in all strings like user:<password>@host:port

scripts/wsrep_sst_clone.sh
```diff
@@ -216,22 +216,22 @@ wsrep_sst_normalize_state_string()
 cleanup_donor()
 {
     wsrep_log_info "Cleanup DONOR."
-    wsrep_log_debug "Cleanup MySQL ADMIN_PSWD: $ADMIN_PSWD"
+#    wsrep_log_debug "Cleanup MySQL ADMIN_PSWD: $ADMIN_PSWD"
```
kamil-holubicki, line 219: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -520,9 +520,9 @@ then
 
     wsrep_log_debug "-> WSREP_SST_OPT_LPORT: $WSREP_SST_OPT_LPORT "
     wsrep_log_debug "-> MYSQL_ACLIENT: $MYSQL_ACLIENT "
-    wsrep_log_debug "-> MYSQL_ACLIENT ADMIN_PSWD: $ADMIN_PSWD"
+    #wsrep_log_debug "-> MYSQL_ACLIENT ADMIN_PSWD: $ADMIN_PSWD"
```
kamil-holubicki, line 523: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -520,9 +520,9 @@ then
...
     wsrep_log_debug "-> MYSQL_RCLIENT: $MYSQL_RCLIENT "
-    wsrep_log_debug "-> MYSQL_RUSER: $WSREP_SST_OPT_REMOTE_JOINER_USER | PWD  $WSREP_SST_OPT_REMOTE_JOINER_PSWD"
+    #wsrep_log_debug "-> MYSQL_RUSER: $WSREP_SST_OPT_REMOTE_JOINER_USER | PWD  $WSREP_SST_OPT_REMOTE_JOINER_PSWD"
```
kamil-holubicki, line 525: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -553,7 +553,7 @@ GRANT EXECUTE ON *.* TO "$WSREP_SST_OPT_REMOTE_JOINER_USER"@"%";
 EOF
         RC=0
         export MYSQL_PWD=$ADMIN_PSWD
-        wsrep_log_debug "-> connecting to donor: $CLONE_PREPARE_SQL $MYSQL_PWD $MYSQL_ACLIENT"
+#        wsrep_log_debug "-> connecting to donor: $CLONE_PREPARE_SQL $MYSQL_PWD $MYSQL_ACLIENT"
```
kamil-holubicki, line 556: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -607,12 +608,12 @@ EOF
 
         wsrep_log_debug "JOINER CLONE ACTION SQL: $CLONE_EXECUTE_SQL $MYSQL_PWD $MYSQL_RCLIENT"
         CLONE_EXECUTE=`cat $CLONE_EXECUTE_SQL` || :
-        wsrep_log_debug "-> $CLONE_EXECUTE"
+#        wsrep_log_debug "-> $CLONE_EXECUTE"
```
kamil-holubicki, line 611: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -996,7 +1000,7 @@ EOF
 
     wsrep_log_info "Launching clone recipient daemon"
     wsrep_log_info "-> using: $CLONE_ENV $CLONE_BINARY_SAFE $DEFAULT_OPTIONS "
-    wsrep_log_debug "-> Test connection as: -u$CLONE_USER -p$CLONE_PSWD -h $JOINER_CLONE_HOST -P $JOINER_CLONE_PORT"
+    wsrep_log_debug "-> Test connection as: -u$CLONE_USER -pCLONE_PSWD -h $JOINER_CLONE_HOST -P $JOINER_CLONE_PORT"
```
kamil-holubicki: instead of `-pCLONE_PSWD`, please use -p*******

scripts/wsrep_sst_clone.sh
```diff
@@ -1030,7 +1034,7 @@ EOF
     trap cleanup_joiner EXIT
 
     export MYSQL_PWD=$CLONE_PSWD
-    wsrep_log_debug "-> Exported MySQL password $MYSQL_PWD"
+#    wsrep_log_debug "-> Exported MySQL password $MYSQL_PWD"
```
kamil-holubicki, line 1037: remove

scripts/wsrep_sst_clone.sh
```diff
@@ -1075,19 +1079,24 @@ EOF
...
+    CLEANUP_CLONE_PLUGIN_SQL=""
+    if [ "${CLEANUP_CLONE_PLUGIN}" == "yes" ]; then
+            CLEANUP_CLONE_PLUGIN_SQL="UNINSTALL PLUGIN CLONE;"
+    fi
 cat << EOF > "$CLONE_SQL"
 SET SESSION sql_log_bin=OFF;
 DROP USER IF EXISTS $CLONE_USER;
 DROP USER IF EXISTS $CLONE_USER@'localhost';
-DROP USER IF EXISTS 'mysql.pxc.sst.user'@'localhost';
+DROP USER IF EXISTS 'mysql.pxc.sst.user'@'localhost'; $CLEANUP_CLONE_PLUGIN_SQL
```
kamil-holubicki, line 1099: whitespace at the end of line

kamil-holubicki (changes requested):

kamil-holubicki:
Thsi PR contains also changes from https://github.com/percona/percona-xtradb-cluster/pull/2018. Is it needed?

sql/wsrep_sst.cc
Tusamarco: Personally no, I prefer to remove it completely instead masking. 
I was thinking to that as well, but at the end I do not see it consistent with XB. If some user has any grep on the log for any use this will be incompatible and inconsistent.

scripts/wsrep_sst_clone.sh
Tusamarco: @kamil-holubicki (valid for this and all the following) I did not remove them on purpose.
One thing is to do not print them out (also if debug) another is to remove the.
In case of deep debug having there already the debug string can significantly speed-up the problem identification and resolution.
As comment in a bash there is no additional risk of exposure as if you expose the variable.

scripts/wsrep_sst_clone.sh
Tusamarco: resolved

scripts/wsrep_sst_clone.sh
Tusamarco: done

scripts/wsrep_sst_clone.sh
kamil-holubicki: Not a critical thing. I just try to avoid leaving 'handy things that could potentially be useful". If someone debugs and needs this info, he can add the needed debug print. 
You can leave it as it is, or remove. Up to you.

PR #2022, opened 2025-02-03
PXC-4620: Fix MTR tests that fail after changes needed for clone SST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4620

fixed:
galera.galera_sst_clone - aligned with galera_sst_xtrabackup-v2 test

Missing privileges granted / privileges cleanup:
main.grant_dynamic_flush_upgrade
main.roles-upgrade
main.mysql_upgrade_slave_master_info
main.plugin_auth
main.mysql_57_inplace_upgrade
main.mysql_80_inplace_upgrade
main.component-upgrade
main.grant_dynamic

re-recorded:
perfschema.privilege_table_io
funcs_1.is_table_privileges
funcs_1.is_schema_privileges
main.transactional_acl_tables
main.ps_sys_upgrade


scripts/mysql_system_users.sql
```diff
@@ -78,7 +78,10 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.internal.session'@localhost;
 
 -- Due to bugs with roles, we need to grant superuser access here
 GRANT ALL PRIVILEGES ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
-GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, SHUTDOWN,CLONE_ADMIN, SYSTEM_USER,GROUP_REPLICATION_STREAM  ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
+-- Needed by sst_xtrabackup
+GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
+-- Additionally needed by sst_clone
+GRANT SHUTDOWN,CLONE_ADMIN, SYSTEM_USER,GROUP_REPLICATION_STREAM  ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v: Why is `GROUP_REPLICATION_STREAM` privilege needed for clone?

scripts/mysql_system_users.sql
```diff
@@ -91,12 +94,23 @@ GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, SHU
...
+-- Common. WITH GRANT OPTION is needed by clone-sst script
+GRANT BACKUP_ADMIN, EXECUTE ON *.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
+GRANT SELECT ON performance_schema.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
+
+-- Additional grants needed by sst_clone
+-- Together with CREATE USER, grant CREATE/DROP ROLE, because during the server upgrade
+-- they will be set anyway (and it affects mtr_check)
+-- See mysql_system_tables_fix.sql:
+-- UPDATE user SET Create_role_priv= 'Y', Drop_role_priv= 'Y' WHERE Create_user_priv = 'Y';
+GRANT CLONE_ADMIN, GROUP_REPLICATION_STREAM, SYSTEM_USER, SHUTDOWN, CONNECTION_ADMIN, CREATE USER, CREATE ROLE, DROP ROLE, SYSTEM_VARIABLES_ADMIN ON *.* TO 'mysql.pxc.sst.role'@localhost;
```
venkatesh-prasad-v: Do we need `CREATE ROLE`, `DROP ROLE` privileges for the `mysql.pxc.sst.role`? IIUC, `mysql.pxc.sst.role` role will be granted to `mysql.pxc.sst.user` only during the SST and I doubt the user `mysql.pxc.sst.user` will create any new roles during the SST.

scripts/mysql_system_users.sql
```diff
@@ -91,12 +94,23 @@ GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, SHU
...
-GRANT CLONE_ADMIN, SYSTEM_VARIABLES_ADMIN, SUPER, SHUTDOWN, EXECUTE ON *.* to 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
-GRANT SELECT,UPDATE,INSERT ON performance_schema.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
-
+-- Needed by sst_xtrabackup-v2
+GRANT LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER ON *.* TO 'mysql.pxc.sst.role'@localhost;
 GRANT ALTER, CREATE, SELECT, INSERT ON PERCONA_SCHEMA.xtrabackup_history TO 'mysql.pxc.sst.role'@localhost;
+
+-- Common. WITH GRANT OPTION is needed by clone-sst script
+GRANT BACKUP_ADMIN, EXECUTE ON *.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
+GRANT SELECT ON performance_schema.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v, line 109: Do we need the GRANT command in line number 115, when we are doing the same here, but with `GRANT OPTION`?
```
114: -- For some reason this is also needed, although the docs say BACKUP_ADMIN is enough
115: GRANT SELECT ON performance_schema.* TO 'mysql.pxc.sst.role'@localhost;
```

scripts/mysql_system_users.sql
```diff
@@ -78,7 +78,10 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.internal.session'@localhost;
...
 -- Due to bugs with roles, we need to grant superuser access here
 GRANT ALL PRIVILEGES ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
-GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, SHUTDOWN,CLONE_ADMIN, SYSTEM_USER,GROUP_REPLICATION_STREAM  ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
+-- Needed by sst_xtrabackup
+GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
+-- Additionally needed by sst_clone
+GRANT SHUTDOWN,CLONE_ADMIN, SYSTEM_USER,GROUP_REPLICATION_STREAM  ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
 -- GRANT CREATE USER ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
 -- GRANT SUPER ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
 -- GRANT RELOAD ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v, line 91: Please remove these extra comments as they are not needed anymore.

scripts/mysql_system_users.sql
```diff
@@ -78,7 +78,10 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.internal.session'@localhost;
 
 -- Due to bugs with roles, we need to grant superuser access here
 GRANT ALL PRIVILEGES ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
-GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, SHUTDOWN,CLONE_ADMIN, SYSTEM_USER,GROUP_REPLICATION_STREAM  ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
+-- Needed by sst_xtrabackup
+GRANT BACKUP_ADMIN, LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v: If possible, please add a comment on why every listed privilege is required. You may refer to line numbers 69-73 for the already existing comments and extend it.

venkatesh-prasad-v (changes requested):

scripts/mysql_system_users.sql
kamil-holubicki: Good point. Fixed

scripts/mysql_system_users.sql
kamil-holubicki: As I explained on Slack:
```
There is SQL code that is executed during the upgrade (eg. --upgrade=FORCE). It  is mentioned int the comment
If you grant the user CREATE USER priviege, CREATE/DROP ROLE privs are not granted automatically. But if you start the server with --upgrade=FORCE (or do other upgrade path), they will be added.

We've got several mtr tests that do upgrade. MTR does mtr_check before and after the test
Without granting CREATE/DROP ROLE upfront, after --upgrade=FORCE the check ends up in the following state:

-localhost	mysql.pxc.sst.role	N	N	N	N	N	N	Y	Y	Y	NY	N	N	N	N	Y	N	Y	Y	N	Y	N	N	N	NY	N	N	N					0	0	0	0	caching_sha2_password		Y	NULL	Y	N	N	NULL	NULL	NULL	NULL
+localhost	mysql.pxc.sst.role	N	N	N	N	N	N	Y	Y	Y	NY	N	N	N	N	Y	N	Y	Y	N	Y	N	N	N	NY	N	N	N					0	0	0	0	caching_sha2_password		Y	NULL	Y	Y	Y	NULL	NULL	NULL	NULL

as you see it is about CREATE/DROP ROLE being added by upgrade.

It is unclear why upgrade adds these grants, but if you grant CREATE USER they are not set automatically. I would say there is some inconsistency in MySql logic
but answering your question: yes, for now, only in MTR we observe the effect of this inconsistency
```

scripts/mysql_system_users.sql
kamil-holubicki: Done

scripts/mysql_system_users.sql
```diff
@@ -78,7 +78,10 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.internal.session'@localhost;
 
 -- Due to bugs with roles, we need to grant superuser access here
 GRANT ALL PRIVILEGES ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v, line 82: Here, we are granting all privileges to `mysql.pxc.internal.session`, is it still required to grant individual privileges separately to `mysql.pxc.internal.session`?

kamil-holubicki: As we discussed, no, it is not needed. Some historical mess. Removing.

scripts/mysql_system_users.sql
kamil-holubicki: Actually I have left it with the comment. This is the way we should do, not GRANT ALL

scripts/mysql_system_users.sql
Tusamarco: Well as far as I know:
LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER -> are required by the XB process 
BACKUP_ADMIN, SHUTDOWN AND CLONE_ADMMIN are required by clone. 
However only BACKUP_ADMIN is needed by pxc.internal given the other are granted on client(receiver) during init phase so by root

scripts/mysql_system_users.sql
Tusamarco: TBH I do not know why is there, I have no memory of it, but was not in before so for some reason I had add it, but it is not used by clone. that is gr only

venkatesh-prasad-v: So, can we remove it?

Tusamarco: Yes but as you comment in the other comment ... 
WHY are we doing this if we grant all after??

scripts/mysql_system_users.sql
kamil-holubicki: Not relevant anymore. I've remove this grant.

scripts/mysql_system_users.sql
kamil-holubicki: Not needed. Removed.

Tusamarco (approved):
LGTM

scripts/mysql_system_users.sql
```diff
@@ -100,19 +101,18 @@ GRANT ALTER, CREATE, SELECT, INSERT ON PERCONA_SCHEMA.xtrabackup_history TO 'mys
 
 -- Common. WITH GRANT OPTION is needed by clone-sst script
 GRANT BACKUP_ADMIN, EXECUTE ON *.* TO 'mysql.pxc.sst.role'@localhost WITH GRANT OPTION;
+-- For some reason this is also needed, although the docs say BACKUP_ADMIN is enough
```
venkatesh-prasad-v: The reason why we need `SELECT` on `performance_schema.*` is because both the SST script and PXB need to query keyring status and log_status PFS tables respectively. 

Perhaps it is better we add a comment saying why we need it.

scripts/mysql_system_users.sql
```diff
@@ -78,11 +78,15 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.internal.session'@localhost;
...
--- GRANT RELOAD ON *.* TO 'mysql.pxc.internal.session'@localhost WITH GRANT OPTION;
 
+-- TODO: Investigate why commit 7e467c54 claims that there is a "bug with roles"
+-- and grants all privileges. It should be enough to use only the following grants
+-- as per the comment above.
+-- We need mysql.pxc.internal.session only for one purpose: create SST user
+-- and grant mysql.pxc.sst.role role to it.
+-- GRANT CREATE USER ON *.* TO 'mysql.pxc.sst.root'@localhost WITH GRANT OPTION;
+-- GRANT SUPER ON *.* TO 'mysql.pxc.sst.root'@localhost WITH GRANT OPTION;
+-- GRANT RELOAD ON *.* TO 'mysql.pxc.sst.root'@localhost WITH GRANT OPTION;
```
venkatesh-prasad-v: `mysql.pxc.sst.root` is not a valid user. I guess you meant `mysql.pxc.internal.session`

scripts/mysql_system_users.sql
kamil-holubicki: Yup, it is good to have a real explanation rather than "for some reason" :).
Added.

venkatesh-prasad-v (changes requested):

scripts/mysql_system_users.sql
kamil-holubicki: That was old stuff, but let's update the username

venkatesh-prasad-v (approved):
Great work on this 👏🏼

venkatesh-prasad-v (approved):
Please squash all commits before merging

PR #2027, opened 2025-02-06
PXC-4585: Clone SST on joiner waits indefinitely if the donor node is killed while SST.

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4585

Problem 1:
If the donor node is gone after SST process is started, the joiner waits indefinitely.

Problem 2:
If the joiner is killed during the SST process, the clone instance remains alive in background.

Solution:
Introduced sst progress monitor on Joiner side. After initiating clone transfer it monitors a few things:
1. If the clone instance is still alive (if clone transfer is finished)
2. If the parent process is still alive. Actually this mechanism is redundant, because if partent process dies, it should terminate its children.
3. If there is any data transfer progress. If it stalls for sst-idle-timeout (default 120 sec), SST process is terminated.

If we detect that SST has to be terminated (timeout, parent process died) the script exits and the cleanup is done in EXIT handler.

If the parent process is terminated, it should also terminate its children. It termination is handle in TERM handler.

The similar handling is for netcat process.

Additionally converted all double brackets [[ ... ]] into single brackets for consistency.

### DIFF
--- a/scripts/wsrep_sst_clone.sh
+++ b/scripts/wsrep_sst_clone.sh
@@ -94,11 +94,13 @@ set -o nounset -o errexit
 readonly EINVAL=22
 readonly EPIPE=32
 readonly ETIMEDOUT=110
-PARENT_PID=""
 CLEANUP_CLONE_PLUGIN=""
 CLEANUP_CLONE_SSL=""
 CLONE_USER=""
+PARENT_PID=""
 NC_PID=""
+CLONE_INSTANCE_PID=""
+
 # The following variable will store the position (GTID:POS) coming from the donor in the message exchange. 
 # It will be used ONLY in emergency as last measure if for any reason the position recovery (at the end) will fail 
 RP_PURGED_EMERGENCY=""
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_clone.sh
```diff
@@ -242,14 +244,19 @@ cleanup_donor()
...
-cleanup_joiner()
+sig_cleanup_joiner()
 {
-    wsrep_log_info "Joiner cleanup. SST daemon PID: $CLONE_REAL_PID"
+    if [ -n "$CLONE_INSTANCE_PID" ]; then
+        kill -SIGKILL $CLONE_INSTANCE_PID /dev/null 2>&1 ||:
+    fi
+
+    if [ -n "$NC_PID" ]; then
+        kill -SIGKILL $NC_PID /dev/null 2>&1 ||:
```
Tusamarco, line 254: I know I am redundant but can we have a log message (in debug) that we are killing things (pid)

scripts/wsrep_sst_clone.sh
```diff
@@ -458,6 +460,76 @@ setup_clone_plugin()
...
+        current_timeout=$((current_timeout + 1))
+        wsrep_log_debug "sst idle for ${current_timeout} sec"
+    fi
+
+    if [ ${current_timeout} -eq  ${timeout} ]; then
+      return 2
+    fi
+
+    # Report the current progress
+    current_progress=`$MYSQL_ACLIENT -NB -e "select format(((data/estimate)*100),2) 'completed%' from performance_schema.clone_progress where stage like 'FILE_COPY';" 2> /dev/null` || :
```
Tusamarco, line 520: mm not sure about this.
In very large datasets this % can be not good meaning that some changes may stay longer than 5 seconds.
like 20.45 can move to 20.46 in 10 seconds.
I would take the advantage of the fact we have many numbers after 0. and use them not formatting the %

scripts/wsrep_sst_clone.sh
```diff
@@ -458,6 +460,76 @@ setup_clone_plugin()
...
+    # check if clone instance process is still alive
+    kill -0 $clone_instance_pid 2> /dev/null
+    if [ $? -ne 0 ]; then
+      # Clone instance is done with its job
+      return 0;
+    fi
+
+    # check if there is any SST progress
+    command_output=$(eval "$progress_monitor_command" 2> /dev/null ||:)
+    if [ "${command_output}" != "${previous_command_output}" ]; then
```
Tusamarco, line 507: Myabe it is me but where do you assign value to `previous_command_output`?
at line 476 you initialize it but I do not see the assign

Tusamarco: ok

Tusamarco (approved):
lgtm

scripts/wsrep_sst_clone.sh
kamil-holubicki: Good catch. Fixed.

Tusamarco:
@kamil-holubicki can you merge this?

PR #2028, opened 2025-02-07
(8.4) PXC-4498: PXC node "stalls" and causes the cluster to stop responding

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4498
https://perconadev.atlassian.net/browse/PXC-4390

Removed wsrep_group_commit_queue.

The queue works only for variables used for debug assertions. The actual
write to sys_header is done without the protection of the queue.

As the queue is not enforcing proper order of xid storing in sys_header,
but was in the past and still is the source of deadlocks - removing.


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/377
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/378

All failures are instabilities or timeouts

venkatesh-prasad-v (approved):

PR #2031, opened 2025-02-11
(8.4) PXC-4573: SST Role lacks INNODB_REDO_LOG_ARCHIVE

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4573

Problem:
When SST xtrabackup is configured with redo-log-arch-dir option, pxb
complains about the lack of INNODB_REDO_LOG_ARCHIVE privilege.

Solution:
Missing privilege granted for mysql.pxc.sst.role.


venkatesh-prasad-v (approved):
LGTM

PR #2034, opened 2025-02-13
PXC-4622:Joining node with clone SST method fails with component_keyrng_file based encryption

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4622

Problem:
During SST we initialize a fresh directory for Joiner's clone instance. During this initialization we use Joiner's original my.cnf. If this file contains any encryption option enabled that expects keyring component to be loaded the initialization fails if the server is configured to use local manifest/keyring component config.

Cause:
This is because during the initialization the server tries to use keyring component, as it is not loaded, initialization fails. We can't place manifest/config in target directory, because initialization expects empty data directory.

Solution:
For initialization time disable all encryption-related options that can be reenabled after initialization.

There is only one option, innodb_sys_tablespace_encrypt, that requires to be enabled during the intialization if we need encrypted system tablespace. So this is incompatible with local manifest/config and we can't disable it temporariliy during the initialization. Detect such a case print error and terminate SST.

Additionally:
1. Moved get_json_value to wsrep_sst_common.sh
2. Introduced helper get_keyring_manifest_and_config_types function. Note that it is similar to already existing get_keyring_manifest_and_config. Consider refactoring.
3. Fixed hardcodded manifest name in wsrep_sst_xtrabackup-v2.sh (spotted during the investigation)

### DIFF
--- a/scripts/wsrep_sst_clone.sh
+++ b/scripts/wsrep_sst_clone.sh
@@ -100,8 +100,8 @@ CLONE_USER=""
 NC_PID=""
 CLONE_INSTANCE_PID=""
 
-# The following variable will store the position (GTID:POS) coming from the donor in the message exchange. 
-# It will be used ONLY in emergency as last measure if for any reason the position recovery (at the end) will fail 
+# The following variable will store the position (GTID:POS) coming from the donor in the message exchange.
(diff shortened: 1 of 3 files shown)


Tusamarco (approved):
Ok

PR #2035, opened 2025-02-14
Modify the SSL certification assignment and cleanup

Tusamarco:
Donor doesn't need any clone_ssl but it uses the ssl_% server section While joiner requires the clone_ssl_%


Tusamarco:
Still working on the other fixes

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
-    local CLIENT_SSL_CA=$(parse_cnf sst ssl_ca "")
-    local CLIENT_SSL_MODE=$(parse_cnf sst ssl_mode "")
+    # Client info
+    CLIENT_SSL_CERT=$(parse_cnf sst ssl_cert "")
+    CLIENT_SSL_KEY=$(parse_cnf sst ssl_key "")
+    CLIENT_SSL_CA=$(parse_cnf sst ssl_ca "")
+    CLIENT_SSL_MODE=$(parse_cnf sst ssl_mode "")
 
-    if [ -z "$CLIENT_SSL_CERT" -o "$CLIENT_SSL_CERT" == "" ]
+    if [ -z "$CLIENT_SSL_CERT" ] || [ "$CLIENT_SSL_CERT" == "" ]
```
kamil-holubicki: `if [ -z "$CLIENT_SSL_CERT" ]` is enough

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
-                CLEANUP_CLONE_SSL="yes"
+            # We actually modify the variables only if we are on the Joiner
+            if [ -n "$CLONE_SSL_CERT" ] && [ -n "$CLONE_SSL_KEY" ] && [ "$ROLE" == "joiner" ]
+            then
+                wsrep_log_info "Using SSL configuration from MySQL Server."
+                $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_cert='$CLONE_SSL_CERT'"
+                $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_key='$CLONE_SSL_KEY'"
+                $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_ca='$CLONE_SSL_CA'"
+                if [ "$CLEANUP_CLONE_PLUGIN" == "yes" ]; then
+                    CLEANUP_CLONE_SSL="yes"
```
kamil-holubicki: `CLEANUP_CLONE_SSL` is unused variable now

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
+        then
+            CLONE_SSL_CERT=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_cert")
+            CLONE_SSL_KEY=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_key")
+            CLONE_SSL_CA=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_ca")
+        fi
     fi
 
-    wsrep_log_debug "-> CLONE_SSL_CERT: $CLONE_SSL_CERT; CLONE_SSL_KEY: $CLONE_SSL_KEY; CLONE_SSL_CA: $CLONE_SSL_CA"
+    # Once identify and configured the CLONE plugin we need to check if the DONOR and the 
+    # Joiner handle encrypted connctions.
```
kamil-holubicki: s/connctions/connections/

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
+        if [ "$ROLE" == "joiner" ]
+        then
+            CLONE_SSL_CERT=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_cert")
+            CLONE_SSL_KEY=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_key")
+            CLONE_SSL_CA=$($MYSQL_ACLIENT -e "SELECT @@clone_ssl_ca")
+        fi
     fi
 
-    wsrep_log_debug "-> CLONE_SSL_CERT: $CLONE_SSL_CERT; CLONE_SSL_KEY: $CLONE_SSL_KEY; CLONE_SSL_CA: $CLONE_SSL_CA"
+    # Once identify and configured the CLONE plugin we need to check if the DONOR and the
```
kamil-holubicki: spaces at the end of lines

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
+    DATA=$($MYSQL_ACLIENT -e "SELECT @@datadir")
+    PXC_ENCRYPT_CLUSTER_TRAFFIC=$($MYSQL_ACLIENT -e "SELECT @@pxc_encrypt_cluster_traffic")
+    PXC_ENCRYPT_CLUSTER_TRAFFIC=$(normalize_boolean "$PXC_ENCRYPT_CLUSTER_TRAFFIC" "on")
+    CLIENT_SSL_VALID="no"
+    # We set yes as default for the server given the joiner
+    # will not need to perform these checks and we get them for granted
+    SERVER_SSL_VALID="yes"
+
+    # If pxc_encrypt_cluster_traffic is set ON then we go through the ssl checks
+    # otherwise we skip and do not set SSL
```
kamil-holubicki: spaces at the EOL

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
+    SERVER_SSL_VALID="yes"
+
+    # If pxc_encrypt_cluster_traffic is set ON then we go through the ssl checks
+    # otherwise we skip and do not set SSL    
+    wsrep_log_debug ":->(PRE) CLIENT_SSL_VALID=$CLIENT_SSL_VALID SERVER_SSL_VALID=$SERVER_SSL_VALID PXC_ENCRYPT_CLUSTER_TRAFFIC=$PXC_ENCRYPT_CLUSTER_TRAFFIC"
+    if [ "$PXC_ENCRYPT_CLUSTER_TRAFFIC" == "on" ]
     then
-        wsrep_log_info "CLONE SSL not configured. Checking general MySQL SSL settings."
+        # Client information must check on both side to allow connection from Donor to Joiner
+        wsrep_log_debug "-> CLONE_SSL_CERT: $CLONE_SSL_CERT; \
```
kamil-holubicki: space at the EOL. \ is preceding the space, so does not take effect

scripts/wsrep_sst_clone.sh
```diff
@@ -328,85 +320,157 @@ setup_clone_plugin()
...
-            wsrep_log_info "Using SSL configuration from MySQL Server."
-            $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_cert='$CLONE_SSL_CERT'"
-            $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_key='$CLONE_SSL_KEY'"
-            $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_ca='$CLONE_SSL_CA'"
-            if [ "$CLEANUP_CLONE_PLUGIN" == "yes" ]; then
-                CLEANUP_CLONE_SSL="yes"
+            # We actually modify the variables only if we are on the Joiner
+            if [ -n "$CLONE_SSL_CERT" ] && [ -n "$CLONE_SSL_KEY" ] && [ "$ROLE" == "joiner" ]
+            then
+                wsrep_log_info "Using SSL configuration from MySQL Server."
```
kamil-holubicki: Let's say donor has clone plugin installed and set server certificates. `pxc_encrypt_cluster_traffic=off`. This means:
1. cloning from donor will require clone client to use secure connection
2. galera communication channel is not encrypted
Am I right?

Havig the above logic on Joiner side, clone plugin will be installed, but `clone_ssl_*` will not be set.

In my opinion the logic on joiner node should be something like:
1. Check if clone plugin installed. If not -> install it
2. Check if clone_ssl_ certificates set. If set -> leave as it is
3. if not set -> use certificates from [sst] section, but if pxc_encrypt_cluster_traffic=on use certificates from [mysqld] section if available

scripts/wsrep_sst_clone.sh
```diff
@@ -328,98 +299,160 @@ setup_clone_plugin()
         if [ "$ROLE" == "donor" ]; then
```
kamil-holubicki: We've got exactly the same `if` a few lines above, so we can set `CLEANUP_CLONE_PLUGIN="yes"` there

scripts/wsrep_sst_clone.sh
```diff
@@ -328,98 +299,160 @@ setup_clone_plugin()
...
-    [ "$SERVER_SSL_CA" = "NULL" ] && SERVER_SSL_CA=
+    # Server info
+    SERVER_SSL_CERT=$($MYSQL_ACLIENT -e "SELECT @@ssl_cert")
+    SERVER_SSL_KEY=$($MYSQL_ACLIENT -e "SELECT @@ssl_key")
+    SERVER_SSL_CA=$($MYSQL_ACLIENT -e "SELECT @@ssl_ca")
+    DATA=$($MYSQL_ACLIENT -e "SELECT @@datadir")
+    CLIENT_SSL_VALID="no"
+    SERVER_SSL_VALID="no"
+
+    # Checking CLient certificates
```
kamil-holubicki: s/CLient/client/

scripts/wsrep_sst_clone.sh
```diff
@@ -328,98 +299,160 @@ setup_clone_plugin()
...
             $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_cert='$CLONE_SSL_CERT'"
             $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_key='$CLONE_SSL_KEY'"
             $MYSQL_ACLIENT -e "SET GLOBAL clone_ssl_ca='$CLONE_SSL_CA'"
-            if [ "$CLEANUP_CLONE_PLUGIN" == "yes" ]; then
-                CLEANUP_CLONE_SSL="yes"
-            fi
-        fi
+         fi
+        # We set that client certificates are valid
+        CLIENT_SSL_VALID="yes"
```
kamil-holubicki: As we spoke today, this whole part should happen only on joiner. It is about setting certificates on clone instance to be able to connect to donor instance. It has nothing to client SSL certificates validity. clone_ssl_cert plays any role only on Joiner side.

Client certificates are used ONLY on Donor side, when Donor connects to Joiner clone instance and triggers cloning. So it is not valid to say that CLIENT_SSL_VALID=true if CLONE_SSL_CERT is set. These are not related variables.

scripts/wsrep_sst_clone.sh
```diff
@@ -328,98 +299,160 @@ setup_clone_plugin()
...
-        if [ -n "$CLONE_SSL_CERT" -a -n "$CLONE_SSL_KEY" ]
-        then
-            wsrep_log_info "CLONE SSL already configured. Using it."
+        wsrep_log_info "CLONE SSL variables and SSL client are not correctly set: @@clone_ssl_cert='$CLONE_SSL_CERT', @@clone_ssl_key='$CLONE_SSL_KEY'"
+    fi
+
+    # Checking Server certificates
+    # Check that we have all the files
+    # If they have not been explicitly specified, check the datadir
+    if [ -z "$SERVER_SSL_CA" ]; then
```
kamil-holubicki: As agreed, let's remove these 3 ifs

scripts/wsrep_sst_clone.sh
```diff
@@ -328,98 +299,160 @@ setup_clone_plugin()
...
+        SERVER_SSL_VALID="yes"
+    else
+        SERVER_SSL_VALID="no"
     fi
 
+    # We reset the variable to pass ssl information and will fill it only if both SERVER and CLIENT 
     CLIENT_SSL_OPTIONS=""
-    if [ -n "$CLONE_SSL_CERT" -a -n "$CLONE_SSL_KEY" ]
+    wsrep_log_debug ":->(POST) CLIENT_SSL_VALID=$CLIENT_SSL_VALID SERVER_SSL_VALID=$SERVER_SSL_VALID"
+    if [ "$CLIENT_SSL_VALID" == "yes" ] && [ "$SERVER_SSL_VALID" == "yes" ]
```
kamil-holubicki: As we discussed, everything what happens here, affects only donor, so should happen only on donor (the result is setting 
1. `CLIENT_SSL_OPTIONS` - used during Donor -> clone instance connection
2. `REQUIRE_SSL` - creation of clone user on Donro server.
These are two independent things

scripts/wsrep_sst_clone.sh
Tusamarco: done

scripts/wsrep_sst_clone.sh
Tusamarco: done

scripts/wsrep_sst_clone.sh
Tusamarco: done

scripts/wsrep_sst_clone.sh
Tusamarco: done

kamil-holubicki (approved):

PR #2048, opened 2025-02-26
PXC-4593: [Dev] Code refresh for - PXC 8.0.41

kamil-holubicki:


VarunNagaraju:
Hi @kamil-holubicki, 
I see that the PR is bombarded with the comments from clag-tidy static analyzer, most originating from the third party libraries from extra directory. Could you please add a file .clang-tidy with contents similar to https://github.com/percona/percona-server/blob/8.4/include/boost_1_84_0/patches/boost/uuid/.clang-tidy in order to skip the analysis on the extra directory?

PR #2056, opened 2025-03-10
PXC-4542: Code refresh for - PXC 9.1.0 (Q4 2024)

kamil-holubicki:


mysql-test/suite/group_replication/r/gr_autorejoin_majority_loss_rejoin_failed_readonly.result
```diff
@@ -40,6 +40,8 @@ include/start_group_replication.inc
 [connection server1]
 SET GLOBAL group_replication_autorejoin_tries = 3;
 include/gr_provoke_majority_loss.inc
+# Kill and restart
+# Kill and restart
```
venkatesh-prasad-v: Why this change? Is it because upstream started printring "Kill and restart" to the logs?

mysql-test/suite/group_replication/r/gr_primary_key_conflict_xa.result
```diff
@@ -89,7 +89,6 @@ include/rpl/sync_to_replica.inc
 [connection server1]
 SET DEBUG_SYNC='now SIGNAL waiting';
 SET @@GLOBAL.DEBUG= @debug_save;
-SET DEBUG_SYNC='RESET';
```
venkatesh-prasad-v: Here and at all places where we reset debug_sync..RESET was added by commit f9d0ab1fd. We need to have this for stable gr tests on Jenkins.

scripts/mysql_system_tables_fix.sql
```diff
@@ -1383,7 +1383,6 @@ INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.internal.session', 'lo
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.internal.session', 'localhost', 'SESSION_VARIABLES_ADMIN', 'Y');
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.internal.session', 'localhost', 'SET_ANY_DEFINER', 'Y');
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.internal.session', 'localhost', 'ALLOW_NONEXISTENT_DEFINER', 'Y');
-INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.internal.session', 'localhost', 'SET_USER_ID', 'Y');
```
venkatesh-prasad-v: Why is that the privilege `SET_USER_ID` is revoked from `mysql.pxc.internal.session` user?

scripts/mysql_system_tables.sql
```diff
@@ -335,7 +335,7 @@ SET @cmd = "CREATE TABLE IF NOT EXISTS procs_priv
   Db char(64) binary DEFAULT '' NOT NULL,
   User char(32) binary DEFAULT '' NOT NULL,
   Routine_name char(64) COLLATE utf8mb3_general_ci DEFAULT '' NOT NULL,
-  Routine_type enum('FUNCTION','PROCEDURE','LIBRARY') NOT NULL,
```
venkatesh-prasad-v: May be we need to have the `LIBRARY` enum value here.

venkatesh-prasad-v (changes requested):

scripts/mysql_system_tables.sql
kamil-holubicki: It was my fault during conflict solving. LIBRARY was added in 9.2.
I removed this change from history.

scripts/mysql_system_tables_fix.sql
kamil-holubicki: SET_USER_ID has been removed in 9.1

mysql-test/suite/group_replication/r/gr_autorejoin_majority_loss_rejoin_failed_readonly.result
kamil-holubicki: `# Kill and restart` is printed by `include/kill_and_restart_mysqld.inc`. This is and old behavior however, for unknown reason some replication tests in PXC were not the same as in PS. I have aligned them to PS version and also aligned result. `gr_communication_protocol_and_different_member_states.test` is one of such tests.

mysql-test/suite/group_replication/r/gr_primary_key_conflict_xa.result
kamil-holubicki: To be honest, I didn't see any instabilities. That's why I decided to align these tests with the PS version. If they were unstable, they should be unstable in PS as well and should be fixed there.

venkatesh-prasad-v: Ok. Agreed. Let's re-apply the commit f9d0ab1 in PS if these tests start to fail again.

venkatesh-prasad-v (approved):

PR #2073, opened 2025-04-16
PXC-4512: 2 cluster nodes go in to NON-PRIMARY state MDL BF-BF conflict

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4512

Problem:
There are two tables: parent' and 'child'.
When RENAME TABLE 'child' and INSERT/UPDATE INTO 'parent' are executed
simultaneously, BF-BF abort is triggered on replica.

Cause:
RENAME TABLE 'child' is replicated as TOI. Execution of this DDL
acquires MDL lock on 'child' and 'parent' tables.
INSERT/UPDATE INTO 'parent' acquires MDL lock only on 'parent' table.
So these two transactions can not apply in parallel.
The problem was that DDL writeset didn't contain certification keys
related to 'parent' table.

1. Let' say transaction with global_seqno = N was the last commited one.
2. DDL was replicated and certified but not committed yet.
Its global_seqno = N+1 and depends_seqno=N. It adds 'child' to the
certification index.
3. Then DML was replicated. Its global_seqno = N+2 and
depends_seqno = N (before certification). Then, when we get to
the certification, we see that its key is 'parent', so it doesn't
conflict with DDL, so we don't have to set the dependency on DDL.
4. As the result, we have both DDL and DML depending on N, so they can
be applied in parallel
5. When it comes to the replicated node side, applier threads apply
both transactions in parallel. If DDL applier locks 'child' and 'parent'
and DML applier tries to acquire 'parent' which ends up with BF-BF abort

Solution:
Add parent table as the certification key of RENAME TABLE.

Note:
RENAME TABLE 'parent' locks also 'child' table. In general, it should
not be a problem, because INSERT/UPDATE 'child' adds 'parent' to the
certification keys, so certification of DML should detect the dependency
on DDL. On the flip side, adding 'child' as the certification key to
the DDL will not make any harm, so do it as well.


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3128/
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3129/

venkatesh-prasad-v (approved):

PR #2074, opened 2025-04-18
Suppressed warning that can be issued when GCache

kamil-holubicki:
encryption is enabled and but mlock fails.
The functionality still works. The warning is to inform the user about possible performance drop.
When Jenkins executes MTR tests, in some Docker environments the above can happen, affecting tests results.


venkatesh-prasad-v (approved):
LGTM

aybek (approved):

PR #2077, opened 2025-04-22
(trunk) PXC-4629 Creating temporary table as select puts the whole session in…

aybek:
… TOI/NBO not allowing to kill any query

https://perconadev.atlassian.net/browse/PXC-4629


mysql-test/suite/galera/t/galera_create_table_as_select.test
```diff
@@ -182,3 +182,29 @@ SHOW TABLES;
...
 DROP TABLE t1, t2;
+
+--connection node_1
+CREATE TABLE t2 (f1 INTEGER) ENGINE=InnoDB;
+INSERT INTO t2 VALUES (1),(2),(3),(4),(5);
+
+CREATE TEMPORARY TABLE t1 AS SELECT * FROM t2;
+
+# Add a sleep to delay the completion of the query
+--sleep 100
```
venkatesh-prasad-v: This is a wait caused by the MTR's `--sleep` command is not the `SELECT SLEEP(100)` which is done by the connection thread. So, the kill mechanism will fail as the query `SELECT ID FROM INFORMATION_SCHEMA.PROCESSLIST WHERE INFO LIKE '%SLEEP%' LIMIT 1` shall return empty value.

mysql-test/suite/galera/t/galera_create_table_as_select.test
```diff
@@ -182,3 +182,29 @@ SHOW TABLES;
...
+
+--connection node_1a
+let $thread_id = `SELECT ID FROM INFORMATION_SCHEMA.PROCESSLIST WHERE INFO LIKE '%SLEEP%' LIMIT 1`;
+--disable_query_log
+--error 1317
+--eval KILL $thread_id;
```
venkatesh-prasad-v: Having `--error 1317` means the KILL query failed with an error, which should not happen.

mysql-test/suite/galera/t/galera_create_table_as_select.test
```diff
@@ -182,3 +182,29 @@ SHOW TABLES;
...
+--enable_query_log
+
+--connection node_1
+let $wait_condition = SELECT COUNT(*) = 0 FROM INFORMATION_SCHEMA.PROCESSLIST WHERE ID = $thread_id;
+--source include/wait_condition.inc
+
+--connection node_1
+DROP TABLE IF EXISTS t1, t2;
+
+CALL mtr.add_suppression("Replica SQL: Error 'Unknown table 'test.t1'' on query");
```
venkatesh-prasad-v: Why is there a `Unknown table` error on node1?

venkatesh-prasad-v (changes requested):

mysql-test/suite/galera/t/galera_create_table_as_select.test
aybek: thanks, updated the test

mysql-test/suite/galera/t/galera_create_table_as_select.test
aybek: thanks, updated the test

mysql-test/suite/galera/t/galera_create_table_as_select.test
aybek: thanks, updated the test

kamil-holubicki:
Hi, I can't see a PR for 8.4.

PRs should follow GCA process.

venkatesh-prasad-v (approved):

kamil-holubicki (approved):

aybek:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/489/#showFailuresLink

PR #2081, opened 2025-04-25
(8.4) PXC-4629 Creating temporary table as select puts the whole session in…

aybek:
… TOI/NBO not allowing to kill any query

https://perconadev.atlassian.net/browse/PXC-4629


kamil-holubicki (approved):

aybek:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/488/#showFailuresLink

PR #2089, opened 2025-05-05
(8.4) PXC-4652: SIGSEV on rpl_gtid_owned > unordered_map - with high activity

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4652

Problem:
When there is a high write activity from multiple sessions, local node
or replicated node that have wsrep_applier_threads > 1 fails with
sigsegv.

Cause:
Commit 674df5a8 reworked the way sidno is locked for the time of gtid generation.
Before, the sidno lock was done after the detection that the transaction uses wsrep_sidno (in Gtid_state::generate_automatic_gtid()), so wsrep_sidno hash was locked properly.
After the rework, the locking is done in the caller of generate_automatic_gtid(), in
MYSQL_BIN_LOG::assign_automatic_gtids_to_flush_group(). As WSREP flow relied on the old logic that made the decision about lock after detecting sidno of transaction, wsrep_sidno was not locked at all causing concurrent access to sidno hash which caused memory corruption.

Solution:
Detect if any of the transactions in group uses wsrep_sidno and add it to locked_sidno_set for locking.

Addtionally, moved the logic of assigning wsrep_sidno to gtid out of generate_automatic_gtid() method.
This method allows the caller to specify sidno and gno, so wsrep detection can be done by the caller.

### DIFF
--- a/sql/binlog.cc
+++ b/sql/binlog.cc
@@ -1725,6 +1725,14 @@ bool MYSQL_BIN_LOG::assign_automatic_gtids_to_flush_group(THD *first_seen) {
   DBUG_TRACE;
   bool error = false;
   bool is_global_tsid_locked = false;
+#ifdef WITH_WSREP
+  /* It would work without this flag, as multiple calls to add_lock_for_sidno()
+   with the same sidno work, but take time (log(N)).
+   As we are in a busy area of group commit, let's detect that we already added
+   wsrep_sidno and skip following calls avoiding any overhead.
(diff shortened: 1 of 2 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/496

aybek (approved):

PR #2092, opened 2025-05-08
PXC-4657: MDL BF-BF conflict between two applier threads

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4657

Problem:
We have tables t1 and t2 and an update trigger on t1 that inserts into t2. If update on t1 doesn't change a row, no events related to update are created however, t2 insert event is created by the trigger. As the result a writeset corresponding to t1 update contains Table_map_log_event related to t1 and t2. On the applier side it causes locking of t1 and t2 tables, even if only t2 is modified. But as the update to t1 was skipped on the source node, a writeset doesn't contain t1-related certification keys.
If at the same time a DDL is executed on t1, Galera certification mechanism doesn't see the dependency between these two transactions and order them for parallel apply.
As both transactions acquire MDL locks on t1, on the applier side we end up with BF-BF abort.

Cause:
Writeset related to t1 update does not contain t1-related certification keys.

Solution:
In 'normal' case, when the row is updated, certification keys are added from InnoDB level during row update. The update is skipped, but we still need to add t1 to the certification keys in case there are triggers on t1.

Note: No GCA, because this commit modifies the test introduced after the latest GCA


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3154/

aybek (approved):

kamil-holubicki:
Merged via https://github.com/percona/percona-xtradb-cluster/pull/2094

PR #2093, opened 2025-05-12
(8.0) PXC-3576 PXC create a role mapping an upper case database with lower_…

aybek:
…case_table_names=1

Temporary workaround until the issue is fixed on PXB

### DIFF
--- a/scripts/mysql_system_tables_fix.sql
+++ b/scripts/mysql_system_tables_fix.sql
@@ -1415,8 +1415,30 @@ INSERT IGNORE INTO mysql.user VALUES ('localhost','mysql.pxc.sst.role','N','N','
 #  pxb-sst
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.sst.role', 'localhost', 'BACKUP_ADMIN', 'Y');
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.sst.role', 'localhost', 'INNODB_REDO_LOG_ARCHIVE', 'N');
-INSERT IGNORE INTO mysql.tables_priv VALUES ('localhost', 'PERCONA_SCHEMA', 'mysql.pxc.sst.role', 'xtrabackup_history', 'root\@localhost', CURRENT_TIMESTAMP, 'Alter,Select,Insert,Create', '');
-INSERT IGNORE INTO mysql.db VALUES ('localhost', 'PERCONA_SCHEMA', 'mysql.pxc.sst.role','N','N','N','N','Y','N','N','N','N','N','N','N','N','N','N','N','N','N','N');
+DROP PROCEDURE IF EXISTS conditional_grants;
+DELIMITER $$
+CREATE PROCEDURE conditional_grants()
+BEGIN
+  DECLARE lctn INT DEFAULT 0;
+
+  -- Get the value of lower_case_table_names
+  SET lctn = @@lower_case_table_names;
+
+  IF lctn = 1 THEN
+    -- Case-insensitive system: use lowercase schema/table
+    INSERT IGNORE INTO mysql.tables_priv VALUES ('localhost', 'percona_schema', 'mysql.pxc.sst.role', 'xtrabackup_history', 'root\@localhost', CURRENT_TIMESTAMP, 'Alter,Select,Insert,Create', '');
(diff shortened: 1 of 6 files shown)


aybek:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3169/

scripts/mysql_system_tables_fix.sql
```diff
@@ -1415,8 +1415,30 @@ INSERT IGNORE INTO mysql.user VALUES ('localhost','mysql.pxc.sst.role','N','N','
...
+CREATE PROCEDURE conditional_grants()
+BEGIN
+    DECLARE lctn INT DEFAULT 0;
+
+    -- Get the value of lower_case_table_names
+    SET lctn = @@lower_case_table_names;
+
+    IF lctn = 1 THEN
+        -- Case-insensitive system: use lowercase schema/table
+    INSERT IGNORE INTO mysql.tables_priv VALUES ('localhost', 'percona_schema', 'mysql.pxc.sst.role', 'xtrabackup_history', 'root\@localhost', CURRENT_TIMESTAMP, 'Alter,Select,Insert,Create', '');
```
kamil-holubicki, line 1429: I think there is something wrong with indentation here and in following conditional branches.

scripts/mysql_system_tables_fix.sql
```diff
@@ -1415,8 +1415,30 @@ INSERT IGNORE INTO mysql.user VALUES ('localhost','mysql.pxc.sst.role','N','N','
 #  pxb-sst
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.sst.role', 'localhost', 'BACKUP_ADMIN', 'Y');
 INSERT IGNORE INTO mysql.global_grants VALUES ('mysql.pxc.sst.role', 'localhost', 'INNODB_REDO_LOG_ARCHIVE', 'N');
-INSERT IGNORE INTO mysql.tables_priv VALUES ('localhost', 'PERCONA_SCHEMA', 'mysql.pxc.sst.role', 'xtrabackup_history', 'root\@localhost', CURRENT_TIMESTAMP, 'Alter,Select,Insert,Create', '');
-INSERT IGNORE INTO mysql.db VALUES ('localhost', 'PERCONA_SCHEMA', 'mysql.pxc.sst.role','N','N','N','N','Y','N','N','N','N','N','N','N','N','N','N','N','N','N','N');
+DELIMITER $$
+DROP PROCEDURE IF EXISTS conditional_grants $$
```
kamil-holubicki: Can we use it with standard delimiter and redefine delimiter after it (swap this and above lines)?

scripts/mysql_system_tables_fix.sql
```diff
@@ -1415,8 +1415,30 @@ INSERT IGNORE INTO mysql.user VALUES ('localhost','mysql.pxc.sst.role','N','N','
...
+CREATE PROCEDURE conditional_grants()
```
kamil-holubicki, line 1420: Not a strong opinion. To consider:
In the case of conditional execution of the query accros the file, we use the approach like in line [1081](https://github.com/percona/percona-xtradb-cluster/blob/9549ba275ea8f8fafe09d5521c1b724ff60faaa1/scripts/mysql_system_tables_fix.sql#L1081) which gathers the condition then IFs cmd basing on condition, then creates a prepared stmt, executes, drops.
I don't know if it would be more readable or not, that's why it is not a strong opinion :)

scripts/mysql_system_users.sql
```diff
@@ -99,7 +99,37 @@ REVOKE ALL PRIVILEGES, GRANT OPTION FROM 'mysql.pxc.sst.role'@localhost;
 
 -- Needed by sst_xtrabackup-v2
 GRANT LOCK TABLES, PROCESS, RELOAD, REPLICATION CLIENT, SUPER, INNODB_REDO_LOG_ARCHIVE ON *.* TO 'mysql.pxc.sst.role'@localhost;
-GRANT ALTER, CREATE, SELECT, INSERT ON PERCONA_SCHEMA.xtrabackup_history TO 'mysql.pxc.sst.role'@localhost;
+
+DELIMITER $$
```
kamil-holubicki, line 104: likewise
(all comments from the other file apply here as well)

kamil-holubicki (changes requested):
Would it be possible to create MTR testcase?

scripts/mysql_system_tables_fix.sql
aybek: Left as it is, this way it is more readable to me

kamil-holubicki (approved):

PR #2094, opened 2025-05-13
(8.0) PXC-4657: MDL BF-BF conflict between two applier threads

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4657

Fixes also:
PXC-4684: UPDATE query that matches two tables but modifies only one causes an MDL BF-BF conflict on other nodes.

https://perconadev.atlassian.net/browse/PXC-4684

PXC-4657:

Problem:
We have tables t1 and t2 and an update trigger on t1 that inserts into t2. If update on t1 doesn't change a row, no events related to update are created however, t2 insert event is created by the trigger. As the result a writeset corresponding to t1 update contains Table_map_log_event related to t1 and t2. On the applier side it causes locking of t1 and t2 tables, even if only t2 is modified. But as the update to t1 was skipped on the source node, a writeset doesn't contain t1-related certification keys.
If at the same time a DDL is executed on t1, Galera certification mechanism doesn't see the dependency between these two transactions and order them for parallel apply.
As both transactions acquire MDL locks on t1, on the applier side we end up with BF-BF abort.

Cause:
Writeset related to t1 update does not contain t1-related certification keys.

PXC-4684:

Problem:
For queries like
UPDATE t2 JOIN t1 USING (i) SET t2.d = t2.d+1, t1.d = t1.d; t1 is not updated, as MySql detects that t1 row didn't change. The result is a writeset having Table_map_log_event related to t1 and t2 but having only t2 certification keys.

Cause:
Transaction dependencies calculation is affected and BF-BF abort is seen on replicated node that have more than one applier thread.

Solution:
In 'normal' case, when the row is updated, certification keys are added from InnoDB level during row update. The update is skipped, but we still need to add t1 to the certification keys in case there are triggers on t1.

### DIFF
--- a/mysql-test/suite/galera/r/galera_bf_bf_2.result
+++ b/mysql-test/suite/galera/r/galera_bf_bf_2.result
@@ -1,10 +1,11 @@
+############################################################
+# Testcase:
+# RENAME TABLE <child_table> and concurrent INSERT INTO <parent_table>
+############################################################
 SET SESSION wsrep_sync_wait = 0;
 SET SESSION wsrep_sync_wait = 0;
 SET SESSION wsrep_sync_wait = 0;
-CREATE TABLE parent (id INT PRIMARY KEY);
-CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, FOREIGN KEY(parent_id) REFERENCES parent(id) ON DELETE CASCADE ON UPDATE CASCADE);
-INSERT INTO parent values (0);
-INSERT INTO child values (0, 0);
+CREATE TABLE parent (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, FOREIGN KEY(parent_id) REFERENCES parent(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent values (0); INSERT INTO child values (0, 0);;
 SET GLOBAL wsrep_provider_options = 'dbug=d,after_cert_and_catch';
 SET SESSION wsrep_sync_wait = 0;
 SET GLOBAL wsrep_applier_threads = 2;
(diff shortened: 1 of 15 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3170/

sql/wsrep_trans_observer.h
```diff
@@ -527,7 +527,9 @@ static inline void wsrep_commit_empty(THD *thd, bool all) {
     /* @todo CTAS with STATEMENT binlog format and empty result set
        seems to be committing empty. Figure out why and try to fix
        elsewhere. */
-    assert(!wsrep_has_changes(thd) || thd->wsrep_stmt_transaction_rolled_back ||
+    assert(!wsrep_has_changes(thd) ||
+           thd->wsrep_transaction_added_extra_cert_key ||
```
satya-bodapati, line 531: is this debug build only variable? used anywhere else except this place (assert())?

mysql-test/suite/galera/r/galera_bf_bf_2.result
```diff
@@ -52,7 +53,64 @@ SET SESSION wsrep_on = 1;
...
 SET GLOBAL wsrep_provider_options = 'signal=after_cert_and_catch';
 DROP TABLE child, parent_new;
+############################################################
+# Testcase:
+# t1 update trigger replication after noop update
+############################################################
+SET SESSION wsrep_sync_wait = 0;
+SET SESSION wsrep_sync_wait = 0;
+SET SESSION wsrep_sync_wait = 0;
+CREATE TABLE t1 (i INT NOT NULL AUTO_INCREMENT, d INT, PRIMARY KEY(i)); CREATE TABLE t2 LIKE t1; CREATE TRIGGER t1_on_update AFTER UPDATE ON t1 FOR EACH ROW INSERT INTO t2 (d) VALUES (NEW.d); INSERT INTO t1 VALUES (1, 1);;
```
satya-bodapati, line 66: can you also add UPDATE t2 scenario?
Also may be have trigger on update t2 to insert into t3? :) 

I guess there can be no limits to madness :joy:

storage/innobase/handler/ha_innodb.h
```diff
@@ -492,7 +492,7 @@ class ha_innobase : public handler {
 
 #ifdef WITH_WSREP
   int wsrep_append_keys(THD *thd, Wsrep_service_key_type key_type,
```
satya-bodapati, line 494: I am trying to understand whether wsrep_append_keys() should be in server or InnoDB.

it is required in InnoDB because of FKs, and also required in server because of triggers?

kamil-holubicki: Original idea is that writeset keys are added in two places:
1. During TOI start - there is a dedicated part
2. During DML execution - from InnoDB level. Every time the row is inserted/updated/deleted, from InnoDB level we add a key identifying the row.

The above idea is not enough, in case when on server level we detect that row will not be updated. In such a case we don't get to InnoDB world (we skip it), so certification key for the row that was considered to update is missing.
But we still have to add this cert key to the writeset for the reason described in commit message.

mysql-test/suite/galera/r/galera_bf_bf_2.result
kamil-holubicki: My proposal is to keep it simple. update t2 with t2 having trigger inserting into t3  is the same case. 
Enough for now :)

sql/wsrep_trans_observer.h
kamil-holubicki: Yes, only debug. I wrapped it wih NDEBUG

satya-bodapati (approved):

PR #2095, opened 2025-05-19
Bug fix for PXC-4687

Tusamarco:
PXC-4687: Summary
When executing SST with clone script and not using SSL and pxc_encrypt_cluster_traffic is disabled the script fails to connect to donor given a bug.

[PXC-4687](https://perconadev.atlassian.net/browse/PXC-4687)
Problem:
The script was still keeping the SSL as required and fails given the server settings are not enforcing SSL usage.

Cause:
During check of SSL settings one condition was missed

Solution:
Implemented a check on SSL server variable server_ssl_ca to not be equal NULL. When null MySQL is passing the string "NULL" as such the string is not empty.

[PXC-4687]: https://perconadev.atlassian.net/browse/PXC-4687?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/wsrep_sst_clone.sh
+++ b/scripts/wsrep_sst_clone.sh
@@ -323,7 +323,7 @@ setup_certificates_for_donor()
 
     # REQUIRE_SSL is used during clone user creation on Donor server
     REQUIRE_SSL=""
-    if [ -n "$server_ssl_ca" ] && [ -n "$server_ssl_cert" ] && [ -n "$server_ssl_key" ]; then
+    if [ -n "$server_ssl_ca" ] && [ -n "$server_ssl_cert" ] && [ -n "$server_ssl_key" ] && [ "$server_ssl_ca" != "NULL" ]; then
         REQUIRE_SSL="REQUIRE SSL"
     fi
     wsrep_log_debug "Clone SSL Server settings on $role from runtime:  CERT=$server_ssl_cert, KEY=$server_ssl_key, CA=$server_ssl_ca"
--- a/scripts/wsrep_sst_clone.sh
+++ b/scripts/wsrep_sst_clone.sh
@@ -323,7 +323,7 @@ setup_certificates_for_donor()
 
     # REQUIRE_SSL is used during clone user creation on Donor server
     REQUIRE_SSL=""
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_clone.sh
```diff
@@ -323,7 +323,7 @@ setup_certificates_for_donor()
 
     # REQUIRE_SSL is used during clone user creation on Donor server
     REQUIRE_SSL=""
-    if [ -n "$server_ssl_ca" ] && [ -n "$server_ssl_cert" ] && [ -n "$server_ssl_key" ]; then
+    if [ -n "$server_ssl_ca" ] && [ -n "$server_ssl_cert" ] && [ -n "$server_ssl_key" ] && [ ! "$server_ssl_ca" == "NULL" ]; then
```
kamil-holubicki: In Bash, for string equality, you should use = rather than == when inside [ ... ] (the classic test command). == is supported by [[ ... ]] (the extended test syntax), but not guaranteed with [ ... ] in all POSIX shells.

Please change to `[ "$server_ssl_ca" != "NULL" ]`

kamil-holubicki (commented):
Please change the commit message to follow the convention used in PXC:
```
PXC-XYZ: Summary

<link to PXC-XYZ jira ticket>

Problem:
<WHAT - describe problem here>

Cause:
<WHY - describe the root cause here>

Solution:
<HOW - describe how it was solved. Do not paste code. In this case it should be like: added server_ssl_ca validation against string "NULL">
```

kamil-holubicki:
Please also change a target branch. It should be 8.4

scripts/wsrep_sst_clone.sh
Tusamarco: @kamil-holubicki changes done but I cannot see the option to change the target, maybe I do not have the grants for that.

Tusamarco: Forget it I logout login again and I was able to see the dropdown now... @kamil-holubicki

PR #2096, opened 2025-05-20
PXC-4673: Code refresh for - PXC 8.0.42

aybek:


mysql-test/t/all_persisted_variables.test
```diff
@@ -43,9 +43,6 @@ call mtr.add_suppression("\\[Warning\\] .*MY-\\d+.* Changing innodb_extend_and_i
 call mtr.add_suppression("Failed to initialize TLS for channel: mysql_main");
 
 let $total_global_vars=`SELECT COUNT(*) FROM performance_schema.global_variables where variable_name NOT LIKE 'ndb_%' AND variable_name NOT LIKE 'debug_%'`;
-<<<<<<< HEAD
```
kamil-holubicki: Let's use the old version of the test:
[This](https://github.com/kamil-holubicki/percona-xtradb-cluster/commit/0d4521af859ee1bdedd56dffd7c5ff411d1a1cfe) is how I solved it in 8.4

sql/wsrep_client_service.cc
```diff
@@ -347,6 +347,10 @@ void Wsrep_client_service::debug_crash(const char *crash_point
   DBUG_EXECUTE_IF(crash_point, DBUG_SUICIDE(););
 }
 
+void Wsrep_client_service::notify_state_change() {
```
kamil-holubicki: This is not a conflict resolution. I think it would be more clear to add it as a separate commit.

kamil-holubicki (commented):
In overall looks fine. A few cosmetic changes.
Please enable diff3 conflictstyle. Conflicts look like they used conflictstyle=merge.
Here we don't have complicated conflicts, so I think we can leave it as is, but for the future, let's use diff3. 
Or you can merge it once again, there is not such much conflicts :). Up to you.

kamil-holubicki (approved):

PR #2104, opened 2025-05-29
(8.0) PXC-4664: Executing transactions with the same GTID on different nodes causes the node to exit

kamil-holubicki:
PXC-4688: Empty transaction stalls the replica thread

https://perconadev.atlassian.net/browse/PXC-4664
https://perconadev.atlassian.net/browse/PXC-4688

Both issues are the regression caused by fix for PXC-4173 (introduction
of wsrep_async_monitor).

Problem 1:
When node1 and node2 commit the transaction with the same gtid (after
setting 'gtid_next' to the same value), replicated node exits with
sigsegv.

Cause:
If the transaction with a given GTID was committed on local node, and
another, replicated transaction with the same GTID is being applied by
wsrep applier thread, we detect that this GTID was already committed and
attempt to skip the transaction
(gtid_pre_statement_checks() => skip_statement()).
In skip_statement() there is a code that marks a given transaction to be
skipped by wsrep_async_monitor. wsrep_async_monitor-related code is
valid only for async replica worker thread, not for wsrep applier.
We distinguish both by checking if thd is SYSTEM_THREAD_SLAVE_WORKER
(both galera and async worker) and 'wsrep_applier' flag. Check for
wsrep_applier was missing, so it was incorrectly detected, that we are
in the context of async worker thread. Then we did a dynamic_cast of
thd->rli_slave (of type Relay_log_info) to Slave_worker type.
Such a conversion gives a valid pointer only for async worker thread.
In case of wsrep applier, the object type is Relay_log_info, so
dynamic_cast results in nullptr. Then we dereference a pointer, and as
a consequence of being wsrep applier, we dereference nullptr.

Solution:
Added missing check for wsrep_applier, to access wsrep_async_monitor
path only from async worker thread.

For all similar places:
Added debug assertion.
Additionally, wrapped the block where we try to dereference a pointer
with a check against nullptr.

Note:
After fixing sigsegv, it is still possible to end up with a deadlock
(PXC-4665). It is an old issue.

Problem 2:
Committing an empty transaction on an async replica source causes the
PXC async replica node to stall.

Cause:
If an empty transaction is replicated from async source, PXC flow
commits empty transaction in Galera. This caused entering and releasing
wsrep_async_monitor. As this is an empty transaction, we need to commit
(and replicate) its GTID. It happens in
MYSQL_BIN_LOG::gtid_end_transaction(). Here we replicate and commit
an empty transaction with a given gtid. This path tried to acquire
wsrep_async_monitor again with the same sequence number, which caused
infinite wait.

Solution:
If we commit an empty transaction in Galera, but thd owns a GTID, it
means it will be commited later. In such a case postpone
entering/leaving releasing wsrep_async_monitor to the point when real
replication takes place.

Added MTR tests for both issues, however providing a reliable test for
PXC-4664 is not possible without fixing PXC-4665. PXC-4665 is an old
issue which causes a deadlock between wsrep applier and local session
(gtid lock vs galera commit_monitor).

Improved fix for PXC-4173

Fix for PXC-4173 introduced wsrep_async_monitor.

Problem:
The original fix disables group commit for transactions replicated from
async source.

Cause:
Galera orders transactions during replication (when every transaction
gets its global sequence number). Then transactions are commited in
that order.
TOI transactions are replicated before execution, DML transactions are
replicated during the commit. This was the source of the problems in
PXC-4173: transactions are dispatched to parallel async workers, so it
may happen that DDL that should commit after DML from async replication
perspective gets ordered in Galera before DML. This caused a deadlock
(Galera wants to commit in the other order than server's commit manager)

The idea behind introducing wsrep_async_monitor was  to enforce the same
galera ordering as the node receives from async source.

1. The original implementation acquired wsrep_async_monitor at TOI start
and kept it acquired for the whole time of its execution until TOI end.
2. The original implementation acquired wsrep_async_monitor before
group commit pipeline and released it after.

As the consequence transactions were allowed to get into group commit
pipeline one by one (only one transaction at a time), effectively
disabling group commit (there was no possibility to group transactions).

Solution:
It is enough to enforce the order of replicating transactions in Galera
world to be the same as for async replication. Once the transaction is
replicated in galera, it has the same commit order as for async
replication. So there is no point in keeping wsrep_async_monitor for
a long time (especially for commit time). Doing so will allow multiple
transactions to enter group commit pipeline and be commited together.

A simple benchmark test with 4 async workes shows that this approach
enables back the group commit and improves the performance 3-4 times.

Added MTR test for wsrep_async_monitor.

### DIFF
--- a/sql/log_event.cc
+++ b/sql/log_event.cc
@@ -11261,12 +11261,17 @@ static enum_tbl_map_status check_table_map(Relay_log_info const *rli,
       rli->info_thd->system_thread == SYSTEM_THREAD_SLAVE_WORKER &&
       !thd_is_wsrep_applier && res == FILTERED_OUT) {
     Slave_worker *sw =
-        static_cast<Slave_worker *>(const_cast<Relay_log_info *>(rli));
-    Wsrep_async_monitor *wsrep_async_monitor{sw->get_wsrep_async_monitor()};
-    if (wsrep_async_monitor) {
-      auto seqno = sw->sequence_number();
-      assert(seqno > 0);
-      wsrep_async_monitor->skip(seqno);
+        dynamic_cast<Slave_worker *>(const_cast<Relay_log_info *>(rli));
+    // It should never happen. If this is SYSTEM_THREAD_SLAVE_WORKER, but it
+    // is not wsrep applier, it has to be Slave_worker.
+    assert(sw != nullptr);
+    if (sw) {
+      Wsrep_async_monitor *wsrep_async_monitor{sw->get_wsrep_async_monitor()};
+      if (wsrep_async_monitor) {
+        auto seqno = sw->sequence_number();
+        assert(seqno > 0);
+        wsrep_async_monitor->skip(seqno);
+      }
     }
   }
 #endif /* WITH_WSREP */
--- a/sql/wsrep_trans_observer.h
+++ b/sql/wsrep_trans_observer.h
@@ -16,6 +16,7 @@
 #ifndef WSREP_TRANS_OBSERVER_H
 #define WSREP_TRANS_OBSERVER_H
 
+#include "debug_sync.h" /* debug_sync_set_action */
 #include "my_dbug.h"
 #include "service_wsrep.h"
 #include "sql/binlog.h"
@@ -246,6 +247,15 @@ static inline int wsrep_before_prepare(THD *thd, bool all) {
     THD_STAGE_INFO(thd, stage_wsrep_replicating_commit);
   }
 
+  DBUG_EXECUTE_IF("wsrep_before_prepare_before_async_monitor", {
+    const char act[] =
+        "now signal wsrep_before_prepare_before_async_monitor.reached wait_for "
+        "wsrep_before_prepare_before_async_monitor.continue";
+    assert(!debug_sync_set_action(thd, STRING_WITH_LEN(act)));
+  });
+
+  thd_enter_async_monitor(thd);
+
   if ((ret = thd->wsrep_cs().before_prepare()) == 0) {
     assert(!thd->wsrep_trx().ws_meta().gtid().is_undefined());
     thd->wsrep_xid.reset();
@@ -264,6 +274,7 @@ static inline int wsrep_before_prepare(THD *thd, bool all) {
       THD_STAGE_INFO(thd, stage_wsrep_write_set_replicated);
     }
   }
+  thd_leave_async_monitor(thd);
   DBUG_RETURN(ret);
 }
 
@@ -296,10 +307,19 @@ static inline int wsrep_before_commit(THD *thd, bool all) {
   WSREP_DEBUG("wsrep_before_commit: %d, %lld", wsrep_is_real(thd, all),
               (long long)wsrep_thd_trx_seqno(thd));
   int ret = 0;
+  bool async_monitor_entered = false;
 
-  /* Enter the async monitor */
-  thd_enter_async_monitor(thd);
-
+  if (!thd->wsrep_cs().transaction().ordered()) {
+    /*
+     In normal case, the transactions should have already been replicated and
+     ordered in wsrep_before_prepare. If it is still not the case (e.g. empty
+     transaction with GTID) it will be replicated and ordered as the part of
+     wsrep_cs().before_commit().
+     Enter the async monitor
+    */
+    thd_enter_async_monitor(thd);
+    async_monitor_entered = true;
+  }
   assert(wsrep_run_commit_hook(thd, all));
   if ((ret = thd->wsrep_cs().before_commit()) == 0) {
     assert(!thd->wsrep_trx().ws_meta().gtid().is_undefined());
@@ -313,6 +333,11 @@ static inline int wsrep_before_commit(THD *thd, bool all) {
     wsrep_xid_init(thd->get_transaction()->xid_state()->get_xid(),
                    thd->wsrep_trx().ws_meta().gtid());
   }
+
+  if (async_monitor_entered) {
+    thd_leave_async_monitor(thd);
+  }
+
   DBUG_RETURN(ret);
 }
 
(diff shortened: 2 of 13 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3182/

sql/wsrep_trans_observer.h
```diff
@@ -296,10 +307,19 @@ static inline int wsrep_before_commit(THD *thd, bool all) {
...
+    /*
+     In normal case, the transactions should have already been replicated and
+     ordered in wsrep_before_prepare. If it is still not the case (e.g. empty
+     transaction with GTID) it will be replicated and ordered as the part of
+     wsrep_cs().before_commit().
+     Enter the async monitor
+    */
+    thd_enter_async_monitor(thd);
+    async_monitor_entered = true;
+  }
```
satya-bodapati, line 322: in the else part, is there way to assert that async monitor is acquired?

sql/wsrep_trans_observer.h
```diff
@@ -296,10 +307,19 @@ static inline int wsrep_before_commit(THD *thd, bool all) {
   WSREP_DEBUG("wsrep_before_commit: %d, %lld", wsrep_is_real(thd, all),
               (long long)wsrep_thd_trx_seqno(thd));
   int ret = 0;
-
-  /* Enter the async monitor */
-  thd_enter_async_monitor(thd);
-
+  bool async_monitor_entered = false;
+
+  if (!thd->wsrep_cs().transaction().ordered()) {
```
satya-bodapati, line 312: this is galera ordering or the async trx ordering here? 

[optional] can move this state inside thd->acquire_async_monitor()? or something on these lines?

sql/wsrep_trans_observer.h
kamil-holubicki: No. We need to acquire async_monitor for the time of galera replication. If the transaction is already ordered (yes, this check is about galera ordering), it means it was already replicated (and async monitor was acquired and left in wsrep_before_prepare(). So here it is already left.
If the transaction was not galera replicated yet, it will be replicated here, that's why we acquire async_monitor.

sql/wsrep_trans_observer.h
kamil-holubicki: See above comment

sql/wsrep_trans_observer.h
```diff
@@ -547,8 +547,18 @@ static inline void wsrep_commit_empty(THD *thd, bool all) {
...
+
+    /* Do it only if the current transaction doesn't own any gtid.
+    It may happen that this is an empty transaction executed on async master
+    but still owning gtid (to skip transaction). Such a case will be handled in
+    MYSQL_BIN_LOG::gtid_end_transaction() when the transaction commits. We will
+    replicate GTID over galera channel. In such a case we will enter and leave
+    async monitor there.`
+    */
+    if (thd->owned_gtid.sidno == 0) {
+      thd_enter_async_monitor(thd);
```
satya-bodapati, line 584: can we assert inside monitor code that we cannot enter with same sid that is already in the queue (entered list?)?

sql/log_event.cc
```diff
@@ -11261,12 +11261,17 @@ static enum_tbl_map_status check_table_map(Relay_log_info const *rli,
...
+    // is not wsrep applier, it has to be Slave_worker.
+    assert(sw != nullptr);
+    if (sw) {
+      Wsrep_async_monitor *wsrep_async_monitor{sw->get_wsrep_async_monitor()};
+      if (wsrep_async_monitor) {
+        auto seqno = sw->sequence_number();
+        assert(seqno > 0);
+        wsrep_async_monitor->skip(seqno);
+      }
     }
```
satya-bodapati, line 11275: on release builds, what to do the else part? can we continue safely if such cast gives nullptr ? or better to abort with unireg_abort?

sql/log_event.cc
```diff
@@ -11261,12 +11261,17 @@ static enum_tbl_map_status check_table_map(Relay_log_info const *rli,
...
+        dynamic_cast<Slave_worker *>(const_cast<Relay_log_info *>(rli));
+    // It should never happen. If this is SYSTEM_THREAD_SLAVE_WORKER, but it
+    // is not wsrep applier, it has to be Slave_worker.
+    assert(sw != nullptr);
+    if (sw) {
+      Wsrep_async_monitor *wsrep_async_monitor{sw->get_wsrep_async_monitor()};
+      if (wsrep_async_monitor) {
+        auto seqno = sw->sequence_number();
+        assert(seqno > 0);
+        wsrep_async_monitor->skip(seqno);
```
satya-bodapati, line 11273: Maybe the monitor functions enter(), leave(), and skip() should also take thd as a parameter. Inside, we can assert various conditions.

There could be multiple places that use enter(),skip(), or leave(). Having low-level assertions inside monitor code will be useful.

If it is not feasible to do it in this PR, may it be done later or separately?

sql/wsrep_trans_observer.h
kamil-holubicki: We don't track the current owner of monitor. We allow entering if the requested seqno is in front of scheduled_seqnos queue. So theoretically it would be possible to re-enter with the same seqno.
My proposal is not to make revolution now that would need additional testing.

sql/log_event.cc
kamil-holubicki: it is like "this should never happen" path. `rli->info_thd->system_thread == SYSTEM_THREAD_SLAVE_WORKER &&
      !thd_is_wsrep_applier` enforces it to never be nullptr after dynamic cast. We could use static_cast here. This is just playing safe.
But if anything still goes wrong, I think it should go on, not abort. In such a case it would work as without the fix, so still low chances of hitting the deadlock.

sql/log_event.cc
kamil-holubicki: Yup, not in this PR :)

satya-bodapati (approved):
LGTM! We extend debug checks with a separate PR

sql/log_event.cc
satya-bodapati: may be ask the user to turn off the feature and restart? :thinking: async monitor can be disabled?

PR #2108, opened 2025-06-03
PXC-4673: Code refresh for - PXC 8.0.42

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4673

Corrected the test. There was a wrong WHERE condition (missed colon in column value)


aybek:
could this line also be related ? https://github.com/kamil-holubicki/percona-xtradb-cluster/blob/PXC-4673-mtr-fix/sql/wsrep_thd.cc#L261

aybek (approved):

PR #2119, opened 2025-06-06
PXC-4559 Use PXB's reduced lock feature in SST scripts

aybek:
PXB 8.4.0-1 introduces support for the --lock-ddl=REDUCED option. This commit updates PXC to leverage this feature:

1. Use --lock-ddl=REDUCED by default when supported by the installed PXB version.
2. Read and apply the --lock-ddl option from the [xtrabackup] section of my.cnf, if specified.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -836,6 +836,7 @@ read_cnf()
     fi
 
     xbstream_opts=$(parse_cnf sst xbstream-opts "")
+    lock_ddl_opt=$(parse_cnf xtrabackup lock-ddl "")
 }
 
 #
@@ -1590,6 +1591,7 @@ function initialize_pxb_commands()
 
     local disver=""
     local pxb_root pxb_bin_path pxb_plugin_dir
+    local lock_ddl="--lock-ddl"
 
     # We need to use PXB compatible with donor
     pxb_root="${XTRABACKUP_PATH_PREFIX}${donor_version_str}"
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1678,7 +1707,7 @@ function initialize_pxb_commands()
                 --target-dir=\${DATA} 2>&1 | logger -p daemon.err -t ${ssystag}innobackupex-move "
             INNOBACKUP="${pxb_bin_path} --defaults-file=${WSREP_SST_OPT_CONF} \
                 --defaults-group=mysqld${WSREP_SST_OPT_CONF_SUFFIX} $disver $iopts \
```
satya-bodapati, line 1704: we should provide a way for the user to turn off the feature. Here the cnf file is passed first and then the --lock-ddl parameter.  So even if  there is --lock-ddl=ON in cnf, we will still override because we pass --lock-ddl=ON --lock-ddl=reduced and pxb will use the last passed option value.

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1620,6 +1622,33 @@ function initialize_pxb_commands()
         disver="--no-version-check"
     fi
 
+    # Check if xtrabackup supports --lock-ddl=REDUCED
+    if ${pxb_bin_path} --help 2>/dev/null | sed -n '/^ *--lock-ddl\[=name\]/,/^ *--/p' | grep -q 'REDUCED'; then
```
satya-bodapati: I think you can simply detect based on the --version output. if it has -pro, it will have reduced option enabled.

satya-bodapati (changes requested):

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1620,6 +1622,33 @@ function initialize_pxb_commands()
...
+        supports_reduced_lock_ddl=true
+    else
+        supports_reduced_lock_ddl=false
+    fi
+
+    if [[ -z "$lock_ddl_opt" ]]; then
+        # Use --lock-ddl=REDUCED only if xtrabackup supports it 
+        # otherwise use the default --lock-ddl=ON;
+        if [[ "$supports_reduced_lock_ddl" == true ]]; then
+            lock_ddl="--lock-ddl=REDUCED"
```
satya-bodapati, line 1628: Can we also add a message to the log, that PXB PRO is detected and lock-ddl=reduced is used for SST?

scripts/wsrep_sst_xtrabackup-v2.sh
aybek: Ok so PXB requires --defaults-file to be the very first argument on the command line:
`[ERROR] [MY-011825] [Xtrabackup] --defaults-file must be specified first on the command line`
However, with the current implementation, we already validate the user-provided arguments, so an override issue should not occur.

scripts/wsrep_sst_xtrabackup-v2.sh
aybek: thanks

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1620,6 +1622,26 @@ function initialize_pxb_commands()
         disver="--no-version-check"
     fi
 
+    # Check if xtrabackup (PXB) Pro version is used (supports --lock-ddl=REDUCED)
+    if ${pxb_bin_path}  --version 2>&1 | grep -q -- '-pro'; then
+        # Use user-provided option or default to REDUCED
+        lock_ddl="--lock-ddl=${lock_ddl_opt:-REDUCED}"
+        
+        [[ -z "$lock_ddl_opt" ]] && wsrep_log_info "PXB Pro detected; using default --lock-ddl=REDUCED for SST"
```
kamil-holubicki: I think this one will print the message also in the case when we've got PXB-pro, and user provided `lock-ddl=SOMETHING_ELSE`

aybek: thanks , updated

aybek:
Created new PR with proper branch name: https://github.com/percona/percona-xtradb-cluster/pull/2122
closing this one

PR #2122, opened 2025-06-11
(8.4) PXC-4559 Use PXB's reduced lock feature in SST scripts

aybek:
PXB 8.4.0-1 introduces support for the --lock-ddl=REDUCED option. This commit updates PXC to leverage this feature:

1. Use --lock-ddl=REDUCED by default when supported by the installed PXB version.
2. Read and apply the --lock-ddl option from the [xtrabackup] section of my.cnf, if specified.

### DIFF
--- a/scripts/wsrep_sst_xtrabackup-v2.sh
+++ b/scripts/wsrep_sst_xtrabackup-v2.sh
@@ -836,6 +836,7 @@ read_cnf()
     fi
 
     xbstream_opts=$(parse_cnf sst xbstream-opts "")
+    lock_ddl_opt=$(parse_cnf xtrabackup lock-ddl "")
 }
 
 #
@@ -1590,6 +1591,7 @@ function initialize_pxb_commands()
 
     local disver=""
     local pxb_root pxb_bin_path pxb_plugin_dir
+    local lock_ddl="--lock-ddl"
 
     # We need to use PXB compatible with donor
     pxb_root="${XTRABACKUP_PATH_PREFIX}${donor_version_str}"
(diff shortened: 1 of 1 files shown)


scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1620,6 +1622,28 @@ function initialize_pxb_commands()
...
+    if ${pxb_bin_path} --version 2>&1 | grep -q -- '-pro'; then
+        if [[ -z "$lock_ddl_opt" ]]; then
+            lock_ddl="--lock-ddl=REDUCED"
+            wsrep_log_info "PXB Pro detected; using default --lock-ddl=REDUCED for SST"
+        else
+            lock_ddl="--lock-ddl=$lock_ddl_opt"
+        fi    
+    else
+        # If REDUCED is requested but not supported, exit with error
+        if [[ "$lock_ddl_opt" == "REDUCED" ]]; then
```
kamil-holubicki, line 1635: As we discussed on Slack, what if the user sets `reduced` or `ReDuCeD`? In other words, are option values case sensitive or insensitive?

kamil-holubicki (commented):
The logic looks fine. One question about case sensitivity.
Also please ensure there are no empty whitespaces at the end of lines

kamil-holubicki:
I've found that the following MTR tests are failing:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/549/

```
galera_sst_xtrabackup-v2_keyring
pxc_component_keyring_file
pxc_encrypt_rest_fpt
pxc_encrypt_rest_gt
pxc_encrypt_rest_mkey_rotate
pxc_encrypt_rest_system
```

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -1678,7 +1702,7 @@ function initialize_pxb_commands()
                 --target-dir=\${DATA} 2>&1 | logger -p daemon.err -t ${ssystag}innobackupex-move "
             INNOBACKUP="${pxb_bin_path} --defaults-file=${WSREP_SST_OPT_CONF} \
                 --defaults-group=mysqld${WSREP_SST_OPT_CONF_SUFFIX} $disver $iopts \
-                \$INNOEXTRA \$keyringbackupopt --lock-ddl --backup --galera-info \
+                \$INNOEXTRA \$keyringbackupopt $lock_ddl --backup --galera-info \
```
satya-bodapati, line 1705: I think i left the same question on old PR. 

If user provides lock-ddl=ON in the .cnf file, will it be present in the ${WSREP_SST_OPT_CONF}?

if the cnf file is passed first (which has lock-ddl=ON/OFF) then if we pass lock-ddl=REDUCED, then REDUCED  is always enabled?

kamil-holubicki: Hi @satya-bodapati , before the whole logic in read_cnf() function we've got:
`lock_ddl_opt=$(parse_cnf xtrabackup lock-ddl "")`

So if the user has my.cnf with [xtrabackup] section and in this section `lock_ddl` is specified, it will be used.
Or am I missing something?

scripts/wsrep_sst_xtrabackup-v2.sh
```diff
@@ -836,6 +836,7 @@ read_cnf()
     fi
 
     xbstream_opts=$(parse_cnf sst xbstream-opts "")
+    lock_ddl_opt=$(parse_cnf xtrabackup lock-ddl "")
```
kamil-holubicki, line 839: For the sanity, shouldn't we declare it at the beginning of the file as an empty variable (like we have for other variables)?

kamil-holubicki:
Cherry picked during 8.4.6 merge

PR #2125, opened 2025-06-18
PXC-4631: Prevent SST after failed attempt to SST/IST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4631

Problem:
If node starts and State Transfer (ST) is available through IST, but because of the network issue ST fails at the beginning, Joiner node stops which is OK. However next attempt to to start the node, forces full SST. As nothing was transferred for the 1st time, IST should be preferred.

Cause:
When ST is requested, SST script is started always. The 1st stage is setting up socat listener, sending "ready" message to the application and waiting for the message from the Donor side.
Once "ready" message is received by the application, it forms SST request string and sends to the donor. It also marks grastate.dat file as 'unsafe'. It means setting uuid:seqno pair to 0:0. It is done to force full SST after restart in case of this SST failure. However, before getting to the main SST transfer, SST script expects to receive sst-info and possibly xtrabackup_ist files. If there is no network connection between Donor and Joiner at this stage, SST script fails with timeout, grastate.dat file remains 'unsafe' and Joiner is stopped. Next start detects that there are no Galera coordiantes in grastate.dat and forces full SST.

Solution:
We can do better, because up to the point when we wipe out Joiner's datadir and start main SST transfer, if there are any errors in SST script we can move grastate.dat back to the 'safe' state.

There is no problem before sending "ready" message, because grastate.dat is marked 'unsafe' after receiving this message.
As the solution, add a way inform the applicaton, that SST script failed but didn't touch Joiner's data directory. To do this, for the time between "ready" message and datadir wipeout, if something goes wrong, exit with EAGAIN(11). This code is recognized in Galera part (ReplicatorSMM::sst_received()), and before node stop, grastate.dat is moved back to 'safe' state.


satya-bodapati:
Hi @kamil-holubicki Is it possible to delay the part that marks grastate.dat as unsafe ? or move it to the place right before data directory is deleted/modified? 

To me, it seems that the state transitions marking the seq_no as 0 is limiting factor. ie ideally there could be mapping to more states like PREPARE, READY, EXECUTING, DONE etc(may be with mapped with numbers). Then it would be easier to maintain and reason about.

kamil-holubicki:
> Hi @kamil-holubicki Is it possible to delay the part that marks grastate.dat as unsafe ? or move it to the place right before data directory is deleted/modified?
> 
> To me, it seems that the state transitions marking the seq_no as 0 is limiting factor. ie ideally there could be mapping to more states like PREPARE, READY, EXECUTING, DONE etc(may be with mapped with numbers). Then it would be easier to maintain and reason about.

Hi @satya-bodapati, It is Galera library who deals with grastate.dat file. It knows nothing about datadir. It just sees the event: "SST started - mark grastate.dat as unsafe".

scripts/wsrep_sst_common.sh
```diff
@@ -293,6 +310,14 @@ wsrep_log_debug()
     fi
 }
 
+#
+# Exit with the provided exit code or global override if defined.
+#
+safe_exit() {
```
satya-bodapati, line 316: isn't it better to just return EAGAIN at until the datadir deletion/modification point? Global override can have other side affects if it is set again somewhere in middle.

if we know for sure, we will return EAGAIN until datadir, then it is better to remove other error codes until datadir deletion code/SST start part.
currently we return codes 1,2,22,32 etc. so after your patch, we can return only 22 until SST start.

satya-bodapati: Another question, is the driver SST code taking actions based on error codes 1&2,32? now that  we override with 22, we may have to adjust if there is code written to handle 1,2 etc.

kamil-holubicki: > isn't it better to just return EAGAIN at until the datadir deletion/modification point?

Functions called before datadir deletion are called from other paths as well. Only here, in joiner path they need to return EAGAIN code. Instead of hardcoding here and there, I introduced a generic mechanism that allows forcing exit code for some time. For now it is used only in joiner path, but maybe will be useful in the future. Moreover, now everywhere we use `safe_exit` which for me is consistent.

> is the driver SST code taking actions based on error codes 1&2,32

For joiner path, if Galera finds that SST script returned 32 it calls `abort()`, but without marking grastate.dat as `safe`. That's why we need to return EAGAIN if we didn't modify datadir to inform Galera to mark grastate.dat as `safe` before calling `abort()`. Please check related changes in Galera repo.

kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3193/

mysql-test/suite/galera/t/galera_sst_timeout.inc
```diff
@@ -0,0 +1,145 @@
...
+#
+# Start node_2 and ensure it joined with IST
+#
+--connection node_2
+--let $restart_parameters = "restart: --log-error=$ofile"
+--source include/start_mysqld_no_echo.inc
+--source include/wait_until_connected_again.inc
+--source include/galera_wait_ready.inc
+
+--let $assert_text = "Check that node_2 jointed with IST"
```
satya-bodapati: joined*

satya-bodapati (approved):

mysql-test/suite/galera/t/galera_sst_timeout.inc
kamil-holubicki: fixed

PR #2126, opened 2025-06-24
PXC-4644: Inconsistency when adding a FK

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4644

Problem:
If FK constraint check fails during ALTER TABLE ... ADD FOREIGN KEY origin node is evicted from cluster by inconsistency voting protocol

Cause:
WSREP applier threads work in the context of root user. If ALTER TABLE is executed on local node in the context of user not having enough privileges for parent and child tables  (non root), the error code on local node differs from error code on replicated nodes. This is because of fix in commit f26bc31b.
Error codes are used by inconsistency voting protocol, and as they differ, local node is evicted.

Solution:
For inconsistency voting protocol ER_ROW_IS_REFERENCED and ER_ROW_IS_REFERENCED_2 have the same meaning.
The same is for ER_NO_REFERENCED_ROW and ER_NO_REFERENCED_ROW_2 pair. Detect this situation and remove disambiguity before reporting the error code to the voting logic.

### DIFF
--- a/sql/wsrep_applier.cc
+++ b/sql/wsrep_applier.cc
@@ -25,11 +25,11 @@
 #include "sql/binlog_reader.h"
 #include "sql/sql_lex.h"
 
+#include <unordered_map>
 #include "service_wsrep.h"
 #include "wsrep_priv.h"
 #include "wsrep_thd.h"
 #include "wsrep_trans_observer.h"
-
 /*
   read the first event from (*buf). The size of the (*buf) is (*buf_len).
   At the end (*buf) is shitfed to point to the following event or NULL and
@@ -78,6 +78,26 @@ Format_description_log_event *wsrep_get_apply_format(THD *thd) {
   return thd->wsrep_rli->get_rli_description_event();
 }
 
+static uint wsrep_errno_for_voting(uint err) {
+  /* errno_map keeps pairs of error codes that mean
+   the same from inconsistency voting point of view. */
+  static std::unordered_map<uint, uint> errno_map = {
+      /* Use ER_NO_REFERENCED_ROW_2 and ER_ROW_IS_REFERENCED_2 for voting
+       because commit e6dbb7d0 in 8.4 branch changed the behavior to always
+       report _2 code, so it is used for voting. This way voting in mixed
+       version cluster (8.0 and 8.4) will be compatible. */
+      {ER_NO_REFERENCED_ROW, ER_NO_REFERENCED_ROW_2},
+      {ER_ROW_IS_REFERENCED, ER_ROW_IS_REFERENCED_2}};
+
+  if (errno_map.count(err)) {
+    auto res = errno_map[err];
+    WSREP_DEBUG("Overriding ambiguous error code %u -> %u for voting", err,
+                res);
+    return res;
+  }
+  return err;
+}
+
 void wsrep_store_error(const THD *const thd, wsrep::mutable_buffer &dst) {
   Diagnostics_area::Sql_condition_iterator it =
       thd->get_stmt_da()->sql_conditions();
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3191/
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3192/

sql/wsrep_applier.cc
```diff
@@ -78,6 +78,20 @@ Format_description_log_event *wsrep_get_apply_format(THD *thd) {
   return thd->wsrep_rli->get_rli_description_event();
 }
 
+static uint wsrep_errno_for_voting(uint err) {
+  static std::map<uint, uint> errno_map = {
```
satya-bodapati: may be use unordered_map? or even std::list?

satya-bodapati: A comment on the errno_map is good to have

sql/wsrep_applier.cc
```diff
@@ -78,6 +78,20 @@ Format_description_log_event *wsrep_get_apply_format(THD *thd) {
...
+      {ER_NO_REFERENCED_ROW_2, ER_NO_REFERENCED_ROW},
+      {ER_ROW_IS_REFERENCED_2, ER_ROW_IS_REFERENCED}};
+
+  if (errno_map.count(err)) {
+    auto res = errno_map[err];
```
satya-bodapati, line 93: I am generally scared of this usage map[elem]. I have seen subtle bugs. Your usage is fine because you verify that the element exists.

If the element doesn't exist, errno_map[err] will insert a new element to the map.

Any new code might not be aware that count() should be used first. Please use find with iterator.

sql/wsrep_applier.cc
```diff
@@ -93,12 +107,14 @@ void wsrep_store_error(const THD *const thd, wsrep::mutable_buffer &dst) {
 
   auto da = thd->get_stmt_da();
   if (da->cond_count() == 0 && da->is_set()) {
+    uint const err_code = wsrep_errno_for_voting(da->mysql_errno());
+    const char *const err_str = da->message_text();
```
satya-bodapati, line 117: no reason to change da->message_text() and keep it as is (just pass da->message_text() as a parameter)

satya-bodapati (changes requested):

satya-bodapati:
@kamil-holubicki another question i have is this https://github.com/percona/percona-xtradb-cluster/commit/f26bc31bd462278866430b6b37c2b02bd2db34d0 is pretty old commit, why are we seeing PXC-4644 now?

It was discovered recently. There was no MTR test for this case. But it is an old, undiscovered problem

sql/wsrep_applier.cc
kamil-holubicki: Below we've got very similar construction in `for` loop. I made this two blocks of code to look similar for easier analysis.

sql/wsrep_applier.cc
kamil-holubicki: errno_map is local to this method. And the method is just a few lines. Let's leave it as it is.
I think from contexts it is clear what is going on, and additionally I would say that m.count() is faster than m.find() and creating of new iterator object.

sql/wsrep_applier.cc
kamil-holubicki: done

sql/wsrep_applier.cc
satya-bodapati: i asked this question as github comment, may be better to ask here. The commit you mentioned is introduced in Server on on Jun 19, 2018. Why are we seeing this now? :thinking:

satya-bodapati (approved):

PR #2142, opened 2025-08-27
PXC-4741: Code refresh for - PXC 8.4.6

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4741


satya-bodapati (approved):
LGTM! Please test reduced lock feature with PXC. Specially, test if it works with SSTs and concurrent DDLs.

PR #2143, opened 2025-09-02
(8.0) PXC-4292: Improve datadir cleanup during SST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4292

Problem:

During SST the script cleans up:
datadir
innodb_data_home_dir
innodb_log_group_home_dir
innodb_undo_directory

If above variables are set explicitly in my.cnf to the same value, the logic in xtrabackup-v2 script searched the same directory multiple times, resulting in duplicate results.
Then the result of search is passed to 'rm'. Because we use '-f' option of 'rm' it is not a problem (attempt to delete already deleted file), however we can avoid searching the same directory multiple times.

Solution:
Before searching directories that have to be cleaned up, deduplicate these directories names.

### DIFF
--- a/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_same_dirs.result
+++ b/mysql-test/suite/galera/r/galera_sst_xtrabackup-v2_same_dirs.result
@@ -1,0 +1,1 @@
+# restart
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3225/

satya-bodapati (approved):

PR #2146, opened 2025-09-03
(trunk) PXC-4144: Fix the broken links in source code

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4144

Updated obsolete web links.
Note: There is no 9.x LTS version yet, so some links are unknown. Keep as 8.4 for now. Will be updated when 9.x LTS is released.


patrickbirch (approved):
OK

alina-derkach-oaza (approved):

PR #2151, opened 2025-09-08
(8.0) PXC-4709: User is wrongly replicated to other nodes causing inconsistencies

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4709

Problem:
Node is evicted from the cluster by inconsistency voting protocol. It happens when CREATE/ALTER USER is executed by the user without AUTHENTICATION_POLICY_ADMIN privilege.

Cause:
TOI is replicated before actual execution. On source node it fails, but on replicated node, it is executed by wsrep_applier thread working in root user context, so it doesn't fail remotely. It kicks-in inconsistency voting protocol which evicts the node from the cluster.

Solution:
On the Source side validate CREATE/ALTER USER against authentication policy before replicating. However, continue the normal flow, but skip TOI if validation fails.
This is to keep original errors/warnings generated (and used for inconsistency voting). Eg. In case of policy violation and nonexistent plugin happening on the same time, we want ER_PLUGIN_IS_NOT_LOADED to be reported as the original flow does.

### DIFF
--- a/sql/auth/auth_internal.h
+++ b/sql/auth/auth_internal.h
@@ -215,7 +215,7 @@ int replace_routine_table(THD *thd, GRANT_NAME *grant_name, TABLE *table,
                           Access_bitmask rights, bool revoke_grant);
 #ifdef WITH_WSREP
 int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables,
-                      const char *db = WSREP_MYSQL_DB,
+                      bool block_toi = false, const char *db = WSREP_MYSQL_DB,
                       const char *table = nullptr);
 #else
 int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables);
@@ -322,11 +322,20 @@ Auth_id_ref create_authid_from(const LEX_CSTRING &user,
                                const LEX_CSTRING &host);
 bool roles_rename_authid(THD *thd, TABLE *edge_table, TABLE *defaults_table,
                          LEX_USER *user_from, LEX_USER *user_to);
+#ifdef WITH_WSREP
+bool set_and_validate_user_attributes(
+    THD *thd, LEX_USER *Str, acl_table::Pod_user_what_to_update &what_to_set,
+    bool is_privileged_user, bool is_role, Table_ref *history_table,
+    bool *history_check_done, const char *cmd, Userhostpassword_list &,
+    I_multi_factor_auth **mfa = nullptr, bool if_not_exists = false,
+    bool verify_passwd_history = true);
+#else
 bool set_and_validate_user_attributes(
     THD *thd, LEX_USER *Str, acl_table::Pod_user_what_to_update &what_to_set,
     bool is_privileged_user, bool is_role, Table_ref *history_table,
     bool *history_check_done, const char *cmd, Userhostpassword_list &,
     I_multi_factor_auth **mfa = nullptr, bool if_not_exists = false);
+#endif /* WITH_WSREP */
 typedef std::pair<std::string, bool> Grant_privilege;
 typedef std::unordered_multimap<Role_id, Grant_privilege, role_id_hash>
     User_to_dynamic_privileges_map;
--- a/sql/auth/sql_user.cc
+++ b/sql/auth/sql_user.cc
@@ -112,6 +112,8 @@
 
 #include <openssl/rand.h>  // RAND_bytes
 #ifdef WITH_WSREP
+#include "sql/error_handler.h"
+#include "sql/raii/sentry.h"
 #include "sql/wsrep_trans_observer.h"
 #endif /* WITH_WSREP */
 
@@ -1295,13 +1297,21 @@ error:
   @retval 0 ok
   @retval 1 ERROR;
 */
-
+#ifdef WITH_WSREP
+bool set_and_validate_user_attributes(
+    THD *thd, LEX_USER *Str, acl_table::Pod_user_what_to_update &what_to_set,
+    bool is_privileged_user, bool is_role, Table_ref *history_table,
+    bool *history_check_done, const char *cmd,
+    Userhostpassword_list &generated_passwords, I_multi_factor_auth **i_mfa,
+    bool if_not_exists, bool verify_passwd_history) {
+#else
 bool set_and_validate_user_attributes(
     THD *thd, LEX_USER *Str, acl_table::Pod_user_what_to_update &what_to_set,
     bool is_privileged_user, bool is_role, Table_ref *history_table,
     bool *history_check_done, const char *cmd,
     Userhostpassword_list &generated_passwords, I_multi_factor_auth **i_mfa,
     bool if_not_exists) {
+#endif /* WITH_WSREP */
   bool user_exists = false;
   ACL_USER *acl_user;
   plugin_ref plugin = nullptr;
@@ -1768,6 +1778,16 @@ bool set_and_validate_user_attributes(
                              std::string(Str->host.str), gen_password, 1};
       generated_passwords.push_back(p);
     }
+#ifdef WITH_WSREP
+    if (auth->generate_authentication_string(outbuf, &buflen, inbuf,
+                                             inbuflen) ||
+        (verify_passwd_history &&
+         auth_verify_password_history(thd, &Str->user, &Str->host,
+                                      Str->alter_status.password_history_length,
+                                      Str->alter_status.password_reuse_interval,
+                                      auth, inbuf, inbuflen, outbuf, buflen,
+                                      history_table, what_to_set.m_what))) {
+#else
     if (auth->generate_authentication_string(outbuf, &buflen, inbuf,
                                              inbuflen) ||
         auth_verify_password_history(thd, &Str->user, &Str->host,
@@ -1775,6 +1795,8 @@ bool set_and_validate_user_attributes(
                                      Str->alter_status.password_reuse_interval,
                                      auth, inbuf, inbuflen, outbuf, buflen,
                                      history_table, what_to_set.m_what)) {
+
+#endif
       plugin_unlock(nullptr, plugin);
       what_to_set.m_what = NONE_ATTR;
       /*
@@ -1852,7 +1874,12 @@ bool set_and_validate_user_attributes(
       Covers replication scenario too since the IDENTIFIED BY will get
       rewritten to IDENTIFIED ... WITH ... AS
     */
+#ifdef WITH_WSREP
+    if (verify_passwd_history &&
+        auth_verify_password_history(
+#else
     if (auth_verify_password_history(
+#endif
             thd, &Str->user, &Str->host,
             Str->alter_status.password_history_length,
             Str->alter_status.password_reuse_interval, auth, nullptr, 0,
@@ -2144,7 +2171,7 @@ bool change_password(THD *thd, LEX_USER *lex_user, const char *new_password,
     buff[query_length_max - 1] = 0;
     thd->set_query(buff, query_length);
 
-    if ((ret = open_grant_tables(thd, tables, &transactional_tables,
+    if ((ret = open_grant_tables(thd, tables, &transactional_tables, false,
                                  WSREP_MYSQL_DB, "user"))) {
       thd->set_query(query_save);
       return (ret != 1);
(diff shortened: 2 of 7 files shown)


sql/wsrep_applier.cc
```diff
@@ -98,6 +98,11 @@ static uint wsrep_errno_for_voting(uint err) {
   return err;
 }
 
+/* KH: We trigger inconsistency voting if there is an error (thd->is_error())
```
satya-bodapati, line 101: remove KH?

Also, I think there is a question :) is it OK? Isn't considering the last error/warning sufficient enough. I know that you need exact error code for inconsistency voting.Do you need warnings to be included to? 

If an error happens (thd->is_error() true), do we typically overwrite with new error codes?

Apart from the commentary, my request is:
1. remove KH
2. remove the question (we can discuss later elsewhere)

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+ so node will be evicted which is not correct, because on Source
+ ER_AUTHENTICATION_POLICY_MISMATCH was only warning.
+
+ That's why here we are only detecting if we should replicate or not and
+ letting the rest of the flow do the job. Let's consider a few cases. In
+ original flow which follows the call of wsrep_check_for_auth_policy() the check
+ order is: plugin then policy check. We return on first error, continue on
+ warning.
+
+ 1. Source: ER_AUTHENTICATION_POLICY_MISMATCH warning here => TOI
```
satya-bodapati, line 2939: when do we get ER_AUTHENTICATION_POLICY_MISMATCH as a warning? This looks a serious case and always should be an error?

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+        // let the original flow do the job
+        return false;
+      }
+
+      /* set_and_validate_user_attributes() will hash provided password and
+       wipe-out its plaintext version in
+       user_from_tmp.first_factor_auth_info.auth.str.
+       LEX_CSTRING holds just a pointer, so we need to do a deep copy of
+       LEX_USER for original flow to work properly. */
+      LEX_USER *user_from_tmp = deep_copy(thd, user_from);
```
satya-bodapati, line 2987: When do we the free the memory from dup/deep copy?

sql/auth/auth_internal.h
```diff
@@ -322,11 +322,20 @@ Auth_id_ref create_authid_from(const LEX_CSTRING &user,
                                const LEX_CSTRING &host);
 bool roles_rename_authid(THD *thd, TABLE *edge_table, TABLE *defaults_table,
                          LEX_USER *user_from, LEX_USER *user_to);
+#ifdef WITH_WSREP
+bool set_and_validate_user_attributes(
+    THD *thd, LEX_USER *Str, acl_table::Pod_user_what_to_update &what_to_set,
+    bool is_privileged_user, bool is_role, Table_ref *history_table,
+    bool *history_check_done, const char *cmd, Userhostpassword_list &,
+    I_multi_factor_auth **mfa = nullptr, bool if_not_exists = false,
```
satya-bodapati, line 330: Do we support other mulfi factor authentications in PXC?

sql/wsrep_applier.cc
kamil-holubicki: I would rather prefer to leave this question in place. It is something discovered during investigation, not related to the ticket. We have several such comments.

sql/auth/sql_user.cc
kamil-holubicki: It is a warning if you execute the query as a 'root' on local node. It will produce a waring. But as we discussed today, replica node is applying writesets as 'root' as well, so in case of just a warning, we want to replicate.
In other words: if it is a warning it will pass on local node with warning, and also pass on remote node.

sql/auth/sql_user.cc
kamil-holubicki: It is on mem root, so it is freed when the query ends.

sql/auth/auth_internal.h
kamil-holubicki: We support the same as PS supports. Nothing special in context of PXC.

sql/auth/sql_user.cc
```diff
@@ -1768,13 +1778,25 @@ bool set_and_validate_user_attributes(
                              std::string(Str->host.str), gen_password, 1};
       generated_passwords.push_back(p);
     }
+#ifdef WITH_WSREP
+    if (auth->generate_authentication_string(outbuf, &buflen, inbuf,
+                                             inbuflen) ||
+        (verify_passwd_history &&
+         auth_verify_password_history(thd, &Str->user, &Str->host,
```
dlenev, line 1785: There is another call to auth_verify_password_history() in this function. Should not we check verify_passwd_history flag value in its case as well?

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+    return true;
+  }
+};
+static bool wsrep_check_for_auth_policy(THD *thd, List<LEX_USER> &list,
+                                        const char *cmd) {
+  if (WSREP(thd) && !thd->wsrep_applier && list.size() > 0) {
+    Table_ref tables[ACL_TABLES::LAST_ENTRY];
+    bool transactional_tables;
+
+    if (open_grant_tables(thd, tables, &transactional_tables, true)) {
```
dlenev: Do we really need to open privileges tables here? See below.

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+       wipe-out its plaintext version in
+       user_from_tmp.first_factor_auth_info.auth.str.
+       LEX_CSTRING holds just a pointer, so we need to do a deep copy of
+       LEX_USER for original flow to work properly. */
+      LEX_USER *user_from_tmp = deep_copy(thd, user_from);
+
+      /* Skip password verification. It would append keys to writeset, but TOI
+       is not started yet, so it is not allowed. We don't need it now. Original
+       flow will handle it */
+      if (set_and_validate_user_attributes(
```
dlenev, line 2992: Taking into account that we skip password verification here anyway, do we really need to pass Table_ref * which corresponds to really open password table here? Perhaps we can pass nullptr or pointer to dummy object instead?
That would make all the opening of privilege tables above unnecessary and simplify the code quite a bit.

dlenev (changes requested):
Hello Kamil!

After taking an initial look at your patch I have a few questions/proposals.
Please see below. 

Note that I plan to continue looking at it, so more questions/proposals
might pop-up.

sql/auth/sql_user.cc
kamil-holubicki: You are right. Fixed.

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
   return false;
 }
 
+#ifdef WITH_WSREP
+static LEX_CSTRING dup_lex_cstring(THD *thd, const LEX_CSTRING *src) {
```
dlenev: Do we really need to introduce new helper function for this?
Can we perhaps use LexStringDupRoot() or one of its siblings instead?

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2886,146 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+    if (open_grant_tables(thd, tables, &transactional_tables, true)) {
+      return true;
+    }
+
+    raii::Sentry<> close_acl_tables_guard{
+        [thd]() -> void { commit_and_close_mysql_tables(thd); }};
+
+    // Get WRITE_MODE lock, because we will access ACL cache. It may happen
+    // that entry is not yet in cache and needs to be added.
+    Acl_cache_lock_guard acl_cache_lock(thd, Acl_cache_lock_mode::WRITE_MODE);
```
dlenev, line 2957: Could you please elaborate in what scenario we modify acl_cache in the code below so WRITE_MODE lock becomes necessary? Conceptually the fact that correctness check has a side effect on ACL cache bothers me. This is not how ACL cache normally works AFAIU.

sql/auth/sql_user.cc
kamil-holubicki: No. See below :)

sql/auth/sql_user.cc
kamil-holubicki: I just copied the validation code from the original flow that follows. But you are correct. As we don't check password history, opening tables is not needed, so we can skip this step and pass nullptr as a history table.

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2891,136 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+    Inconsistency voting: not triggered
+ 2. Source: ER_AUTHENTICATION_POLICY_MISMATCH error here => no TOI, continue
+            flow, return with ER_AUTHENTICATION_POLICY_MISMATCH
+    Inconsistency voting: not triggered
+ 3. Source: ER_AUTHENTICATION_POLICY_MISMATCH warning here => do TOI then return
+            with ER_PLUGIN_IS_NOT_LOADED
+    Replica: error with ER_PLUGIN_IS_NOT_LOADED
+    Inconsistency voting: triggered for both sides with ER_PLUGIN_IS_NOT_LOADED
+ */
+class Silence_any_error final : public Internal_error_handler {
```
dlenev: We have Dummy_error_handler class doing just that defined in sql/error_handler.h already.
Let us avoid creating extra entities :)

sql/auth/sql_user.cc
```diff
@@ -2864,6 +2891,136 @@ static bool check_orphaned_definers(THD *thd, List<LEX_USER> &list) {
...
+    List_iterator<LEX_USER> user_list(list);
+    Silence_any_error error_handler;
+    bool is_privileged_user = is_privileged_user_for_credential_change(thd);
+
+    thd->push_internal_handler(&error_handler);
+    raii::Sentry<> error_handler_pop_guard{
+        [thd]() -> void { thd->pop_internal_handler(); }};
+
+    while ((tmp_user_from = user_list++)) {
+      acl_table::Pod_user_what_to_update what_to_alter;
```
dlenev: Perhaps it makes sense to call this other 3 local variables like dummy_... to emphasize the fact that we are not going to use their values and just need them to be able to satisfy set_and_validate_user_attributes() calling convention.
What do you think? BTW history_check_done is not really necessary, you can pass nullptr instead of pointer to it instead.

dlenev (changes requested):
Hello Kamil!

I have a few more questions/comments about your patch.
Please see below.

sql/auth/sql_user.cc
kamil-holubicki: I somehow missed the existence of `LexStringDupRootUnlessEmpty`. Fixed

sql/auth/sql_user.cc
kamil-holubicki: This is the callstack:
```
insert_entry_in_db_cache  <--- assert for write lock here
acl_get
check_access
is_privileged_user_for_credential_change
wsrep_check_for_auth_policy
mysql_alter_user
mysql_execute_command
dispatch_sql_command
dispatch_command
do_command
handle_connection
pfs_spawn_thread
```

It is well visible in `pxc_alter_user tests`. In simple
as root:
CREATE USER User1@localhost IDENTIFIED BY 'old_pass';
ALTER USER User1@localhost IDENTIFIED BY 'new_pass';
login as User1:
ALTER USER User1@localhost IDENTIFIED BY 'absolutely_new_pass' REPLACE 'new_pass';

sql/auth/sql_user.cc
kamil-holubicki: Right. It is the leftover from my experiments. First I was experimenting with more advanced filtering, and reduced to this Dummy_error_handler. Fixed.

sql/auth/sql_user.cc
kamil-holubicki: Done

sql/auth/sql_user.cc
dlenev: OK. Thanks for the information!
How about rephrasing the above comment to something like:
// Accessing ACL cache requires lock. We get WRITE_MODE lock since the call to
// is_privileged_user_for_credential_change() below, might have to update part of ACL
// cache with DB-level privileges while it it checks if user has UPDATE privilege on 'mysql'
// database.
?

dlenev (approved):
Hello Kamil!

I think it is acceptable to push the patch after considering my comment about adjusting the comment about WRITE_MODE lock.

I am using word "acceptable" rather than LGTM/OK since I am not too happy about the whole approach with doing double checking. But I realize that trying to solve it in nicer way will probably result in too big diff against Upstream.

sql/auth/auth_internal.h
```diff
@@ -215,7 +215,7 @@ int replace_routine_table(THD *thd, GRANT_NAME *grant_name, TABLE *table,
                           Access_bitmask rights, bool revoke_grant);
 #ifdef WITH_WSREP
 int open_grant_tables(THD *thd, Table_ref *tables, bool *transactional_tables,
-                      const char *db = WSREP_MYSQL_DB,
+                      bool block_toi = false, const char *db = WSREP_MYSQL_DB,
```
satya-bodapati, line 218: One minor comment, it is better to add new "default parameters" at the end.

satya-bodapati (approved):
one minor comment on defualt parameter location.

kamil-holubicki:
> Hello Kamil!
> 
> I think it is acceptable to push the patch after considering my comment about adjusting the comment about WRITE_MODE lock.
> 
> I am using word "acceptable" rather than LGTM/OK since I am not too happy about the whole approach with doing double checking. But I realize that trying to solve it in nicer way will probably result in too big diff against Upstream.

Hi Dmitry,
Yes, it is not a perfect solution and has its own flaws. So far this approach works for several other similar cases, but I'm open to discussion. Maybe we can do it better in the future. For now unfortunately, this is the only thing we can do without diverging too much from upstream (and we definitely don't want to do so)

sql/auth/auth_internal.h
kamil-holubicki: It is on purpose, to make the caller changes minimal.

sql/auth/sql_user.cc
kamil-holubicki: Done

sql/auth/auth_internal.h
satya-bodapati: with default parameter, whether middle or end, changes are minimal, isn't it? my point is it is risky to add in middle. Imagine you already have bool, bool. and you add one more bool in middle as default.

kamil-holubicki: If I added it at the end, I would need to change the caller side in three places instead of two :). Cosmetic thing, but yes, I agree that your approach should be the rule.

PR #2152, opened 2025-09-08
PXC-4746: Code refresh for - PXC 9.4.0

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4746


sql/sp.cc
```diff
@@ -1024,7 +1028,7 @@ enum_sp_return_code sp_drop_routine(THD *thd, enum_sp_type type,
 
 #ifdef WITH_WSREP
   if (routine == nullptr) {
-    if (thd->lex->drop_if_exists) {
```
satya-bodapati: @kamil-holubicki Two questions:
1. Is the fix also applicable to PS 9.4?
2. Can't we simply initialize the thd->lex->drop_if_exists member? I assume there are other members that are thd->lex->X . Can you please check how they are handled?

kamil-holubicki: Hi @satya-bodapati 
1. I think you meant 8.4, right? No, DROP LIBRARY IF EXISTS was introduced in 9.x, so it is applicable only to 9.x
2. The 'old way' for other 'IF EXISTS' statement, like DROP PROCEDURE IF EXISTS, indeed sets lex->drop_if_exists member (parser does it). However, DROP LIBRARY IF EXISTS was implemented in a different way. Parser sets Sql_cmd_drop_library::m_if_exists member instead of lex member which is not used in DROP LIBRARY path at all. I like the "new way", but I would expect the old code to be refactored to use the new approach. Unfortunately now we have a mixture of two approaches that do the same.
I don't want to set lex->drop_if_exists for DROP LIBRARIES, because the original flow does not use this member it at all.
And yes, there are many lex->X uninitialized fields. Some of them are used only by specific statements and are initialized by the parser only for these specific statements. So this is a risky zone indeed :)

satya-bodapati: Ok, thanks for checking. Yeah, I meant 8.4 :)

satya-bodapati (approved):

PR #2154, opened 2025-09-11
(8.4) PXC-4526: Replication fails and the node becomes Inconsistent if transaction is committed on node by setting tagged gtid

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4526

Problem:
If NEXT_GTID is set to tagged GTID, replicated transaction fails triggering inconsistency voting protocol and node eviction.

Cause:
8.4 introduced tagged GTIDs. Together with this, new event type of GTID_TAGGED_LOG_EVENT was introduced. Deserialization of tagged gtid is handled by a dedicated path that uses new deserializer (for now only GTID_TAGGED_LOG_EVENT uses this new serializer/deserializer). Before passing buffer to the deserializer, buffer size is adjusted if event checksum is not disabled. (if it is not BINLOG_CHECKSUM_ALG_OFF). For async replication it can be BINLOG_CHECKSUM_ALG_OFF or BINLOG_CHECKSUM_ALG_CRC32, depending on Format_description_log_event passed from the source.

Wsrep applier thread uses Format_description_log_event with BINLOG_CHECKSUM_ALG_UNDEF because events sent through Galera channel are not followed by CRC. Because of BINLOG_CHECKSUM_ALG_UNDEF, the above logic limited the event buffer size and passed to new deserializer. Deserializer was not able to fully deserialize GTID_TAGGED_LOG_EVENT and reported the error.

Solution:
Because events in Galera writeset are not followed by CRC, and Format_description_log_event is not replicated via Galera channel, it is not clear how replica should consider incoming events. Replica can consider them as:
BINLOG_CHECKSUM_ALG_UNDEF - events from checksum-unaware servers or
BINLOG_CHECKSUM_ALG_OFF - events from servers without checksum

The approach with BINLOG_CHECKSUM_ALG_UNDEF worked so far, but together with introduction the logic around GTID_TAGGED_LOG_EVENT it's not true anymore.

As the solution, explicitly configure replica to consider all events from Galera channel as having no checksum.

Note: It is questionable if original logic around GTID_TAGGED_LOG_EVENT is correct. Probably it would be better if it limited deserializer buffer size only if event explicitly has a checksum (BINLOG_CHECKSUM_ALG_CRC32). But again, the problem is not visible for other replication channels than Galera.

### DIFF
--- a/sql/wsrep_applier.cc
+++ b/sql/wsrep_applier.cc
@@ -154,6 +154,7 @@ int wsrep_apply_events(THD *thd, Relay_log_info *rli __attribute__((unused)),
         continue;
       /* If gtid_mode=OFF then async master-slave will generate ANONYMOUS GTID.
        */
+      case mysql::binlog::event::GTID_TAGGED_LOG_EVENT:
       case mysql::binlog::event::GTID_LOG_EVENT:
       case mysql::binlog::event::ANONYMOUS_GTID_LOG_EVENT: {
         Gtid_log_event *gev = (Gtid_log_event *)ev;
--- a/sql/wsrep_high_priority_service.cc
+++ b/sql/wsrep_high_priority_service.cc
@@ -79,7 +79,9 @@ static Relay_log_info *wsrep_relay_log_init(const char *) {
     WSREP_ERROR("Failed to create Relay-Log for wsrep thread, aborting");
     unireg_abort(MYSQLD_ABORT_EXIT);
   }
-  rli->set_rli_description_event(new Format_description_log_event());
+  auto ev = new Format_description_log_event();
+  ev->footer()->checksum_alg = mysql::binlog::event::BINLOG_CHECKSUM_ALG_OFF;
+  rli->set_rli_description_event(ev);
 
   rli->current_mts_submode = new Mts_submode_wsrep();
   return (rli);
--- a/mysql-test/suite/galera/r/galera_gtid.result
+++ b/mysql-test/suite/galera/r/galera_gtid.result
@@ -8,4 +8,6 @@ UPDATE t1 SET f1 = 2;
 SET SESSION wsrep_sync_wait = 7;
 gtid_executed_equal
 1
+SET @@session.GTID_NEXT='1ed6b5f2-9091-11ef-b382-13ba123de7b3:abc124:1';
+INSERT INTO t1 VALUES (100);
 DROP TABLE t1;
--- a/mysql-test/suite/galera/t/galera_gtid.test
+++ b/mysql-test/suite/galera/t/galera_gtid.test
@@ -55,6 +55,16 @@ SET SESSION wsrep_sync_wait = 7;
 # --connection node_1
 # COMMIT;
 
+#
+# Test tagged gtid
+#
+--connection node_1
+SET @@session.GTID_NEXT='1ed6b5f2-9091-11ef-b382-13ba123de7b3:abc124:1';
+INSERT INTO t1 VALUES (100);
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM t1 WHERE f1 = 100
+--source include/wait_condition.inc
+
 
 # cleanup
 DROP TABLE t1;


satya-bodapati:
@kamil-holubicki I have a few questions:
1. Are the event checksums present for all types of events? Or only this Format_description_log_event?
2. Generally, it is better to enable checksums of the events. But in the case of PXC, we never store them on disk?
3. If we disable checksums, are there other mechanisms in Galera to detect safety & corruption of events at receiving pxc nodes?

kamil-holubicki:
Hi @satya-bodapati 
> 1. Are the event checksums present for all types of events? Or only this Format_description_log_event?

In MySql binlog file the checksum follows the event. If the event is written to binlog with or without checksum is controlled by `binlog_checksum` variable. Events are first added to cache (without checksums) and when it is a time to dump them to the binlog file, it is responsibility of the BinlogWriter stream to dump the event and add (or) not a checksum.
In Galera channel, we don't use these checksums. We use a cache of events and simply dump its content to the galera buffer, which adds its own checksums.

> 2. Generally, it is better to enable checksums of the events. But in the case of PXC, we never store them on disk?

Galera writesets are not stored to disk during replication (in a simple case). You can consider a writeset as a blob of data + keys. Data is a stream of events, not followed by MySQL checksums. Writeset has its own checksums.
 
> 3. If we disable checksums, are there other mechanisms in Galera to detect safety & corruption of events at receiving pxc nodes?

Again, I didn't disable anything. Just explicitly said "no cheksums". Previously it was said implicitly :)
And yes, as explained above, Galera cares about the data integrity on its level.

sql/wsrep_high_priority_service.cc
```diff
@@ -79,7 +79,9 @@ static Relay_log_info *wsrep_relay_log_init(const char *) {
     WSREP_ERROR("Failed to create Relay-Log for wsrep thread, aborting");
     unireg_abort(MYSQLD_ABORT_EXIT);
   }
-  rli->set_rli_description_event(new Format_description_log_event());
+  auto ev = new Format_description_log_event();
+  ev->footer()->checksum_alg = mysql::binlog::event::BINLOG_CHECKSUM_ALG_OFF;
+  rli->set_rli_description_event(ev);
```
satya-bodapati, line 84: If i understand correctly, this would fix the sender part to always set as checksum OFF. Thats good.

What if an unpatched version (before this fix) is the sender node and the receiver node still gets UNDEF. Is this a possible scenario? if yes, how is this handled?

satya-bodapati:
> Hi @satya-bodapati
> 
> > 1. Are the event checksums present for all types of events? Or only this Format_description_log_event?
> 
> In MySql binlog file the checksum follows the event. If the event is written to binlog with or without checksum is controlled by `binlog_checksum` variable. Events are first added to cache (without checksums) and when it is a time to dump them to the binlog file, it is responsibility of the BinlogWriter stream to dump the event and add (or) not a checksum. In Galera channel, we don't use these checksums. We use a cache of events and simply dump its content to the galera buffer, which adds its own checksums.
> 
> > 2. Generally, it is better to enable checksums of the events. But in the case of PXC, we never store them on disk?
> 
> Galera writesets are not stored to disk during replication (in a simple case). You can consider a writeset as a blob of data + keys. Data is a stream of events, not followed by MySQL checksums. Writeset has its own checksums.
> 
> > 3. If we disable checksums, are there other mechanisms in Galera to detect safety & corruption of events at receiving pxc nodes?
> 
> Again, I didn't disable anything. Just explicitly said "no cheksums". Previously it was said implicitly :) And yes, as explained above, Galera cares about the data integrity on its level.

Thanks for the clarification @kamil-holubicki !

sql/wsrep_applier.cc
```diff
@@ -154,6 +154,7 @@ int wsrep_apply_events(THD *thd, Relay_log_info *rli __attribute__((unused)),
         continue;
       /* If gtid_mode=OFF then async master-slave will generate ANONYMOUS GTID.
        */
+      case mysql::binlog::event::GTID_TAGGED_LOG_EVENT:
```
satya-bodapati, line 157: What does this code change do? Is it relevant to the bug fix? :thinking:

satya-bodapati:
@kamil-holubicki also jenkins links please

sql/wsrep_high_priority_service.cc
kamil-holubicki: Actually, this fixes receiver (replica side).
Source side always sends writesets without a checksum. It hasn't change.
What changed is that now replica side explicitly treats events as with checksum OFF, not as checksum UNDEF.

satya-bodapati: @kamil-holubicki Ah, Thanks for clarifying. Can you please share the path/source side that sends writesets without the checksum?

kamil-holubicki: Cached events are dumped to the writeset in `wsrep_write_cache_inc()`

sql/wsrep_applier.cc
kamil-holubicki: It makes pxc replica aware of the new event type GTID_TAGGED_LOG_EVENT.
After fixing the initial bug (parsing of GTID_TAGGED_LOG_EVENT), we go further, with parsed event and need to do something with it. Bug masked missing `case` (not discovered because of the bug)

kamil-holubicki:
> @kamil-holubicki also jenkins links please

https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/586/
All failures are instabilities

sql/wsrep_high_priority_service.cc
satya-bodapati: After our Zoom call, I understood that this Format_description_log_event() is never sent in PXC streaming of events. It is required to be set up before processing other events, as this event describes how to process other events.

satya-bodapati (approved):
LGTM!

PR #2167, opened 2025-10-06
(8.4) Add support for Percona PXC operator adding the creation of the sst_i…

Tusamarco:
…n_progress file

sst_in_Progress file containing indicating that server is executing an SST action was missed in case of clone.
Added functions to activate it as we do in XB
ticket https://perconadev.atlassian.net/browse/K8SPXC-1724

### DIFF
--- a/scripts/wsrep_sst_clone.sh
+++ b/scripts/wsrep_sst_clone.sh
@@ -200,6 +200,9 @@ sig_cleanup_joiner()
         kill -SIGKILL $NC_PID /dev/null 2>&1 ||:
     fi
 
+    wsrep_log_debug "Removing the sst_in_progress file"
+    wsrep_cleanup_progress_file
+
     rm -rf $CLONE_SOCK_DIR || :
     rm -rf $CLONE_PID_FILE || :
     rm -fr $tmp_datadir || :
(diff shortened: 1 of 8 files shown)


scripts/wsrep_sst_clone.sh
```diff
@@ -1004,6 +1011,11 @@ then
     cpat=$(parse_cnf sst cpat '.*\.pem$\|.*init\.ok$\|.*galera\.cache$\|.*gvwstate\.dat$\|.*\.err$\|.*\.log$\|.*RPM_UPGRADE_MARKER$\|.*RPM_UPGRADE_HISTORY$\|.*component_keyring_.*\.cnf$\|.*mysqld.my$')
     find $ib_home_dir $ib_log_dir $ib_undo_dir $WSREP_SST_OPT_DATA -mindepth 1  -regex $cpat  -prune  -o -exec rm -rfv {} 1>/dev/null \+
 
+    # Write the SST_PROGRESS_FILE again given we will have delete it with all the other files
```
kamil-holubicki: It is not safe. It creates a short time window when SST_PROGRESS_FILE is not present, but SST actually goes on.

Let's do it the same as it is in the case of PXB-sst. Just do not delete SST_PROGRESS_FILE (include it in `cpat` filter):

```
cpat=$(parse_cnf sst cpat '.*\.pem$\|.*init\.ok$\|.*galera\.cache$\|.*sst_in_progress$\|.*gvwstate\.dat$\|.*\.err$\|.*\.log$\|.*RPM_UPGRADE_MARKER$\|.*RPM_UPGRADE_HISTORY$\|.*component_keyring_.*\.cnf$\|.*mysqld.my$')
```

kamil-holubicki (changes requested):

scripts/wsrep_sst_clone.sh
Tusamarco: @kamil-holubicki done I used the same approach as for XB. Please review

kamil-holubicki (approved):

kamil-holubicki:
One more thing @Tusamarco . Please squash your commits before merging.

PR #2169, opened 2025-11-13
PXC-4789: DDL with FK reference causes MDL BF-BF conflict and evict the nodes out of cluster

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4789

Problem 1:
There are two tables: 'parent' and 'child'. 'child' holding the FK of the 'parent'. If we do SET foreign_key_checks=0, it is possible to drop 'parent'.
If such a DROP statement is executed in parallel to UPDATE on 'child' table, it leads to BF-BF abort conflict on replica.

Cause 1:
DROP TABLE 'parent' acquires MDL locks on 'parent' and 'child' tables. UPDATE 'child' acquires MDL lock on 'child' table. So these transactions cannot apply in parallel on replica side - they are dependent. However, the certification key related to 'child' table was not added to the writeset of DDL, so certification process considered DDL and DML not dependent.

Solution 1:
Add 'child' table to the certification keys of DDL. This way the dependency will be detected and set correctly.

Problem 2:
pt-osc creates 'parent_new' table, then drops 'parent' table and finally renames 'parent_new' to 'parent'.
Such RENAME of 'parent_new', and simultaneous UPDATE on 'child' table, cause BF-BF abort on replica node.

Cause 2:
RENAME acquires MDL locks on 'parent_new' 'parent' and 'child' tables. UPDATE acquires MDL lock on 'child' table.
So these transactions are dependent and should be ordered one after another. However, when renaming 'parent_new' to 'parent', it is not possible to get information about 'child' table from DD ('parent_new' does not have children, 'parent' table object does not exist in DD). Because of this 'child' table was not added to the certification keys of DDL. Dependency was not detected, and transactions were allowed to run in parallel on the replica side causing BF-BF conflict.

Solution 2:
Improve mechanism of table children detection. In case of no parent table object in DD, fall back to querying foreign keys table directly.

### DIFF
--- a/mysql-test/suite/galera/t/galera_bf_bf_2.test
+++ b/mysql-test/suite/galera/t/galera_bf_bf_2.test
@@ -87,3 +87,41 @@
 --let $after_test_table_check = SELECT COUNT(*) = 1 FROM t2 WHERE d=2
 --let $drop_query = DROP TABLE t1, t2
 --source galera_bf_bf_2.inc
+
+#
+# Test that
+# DROP TABLE <parent_table> and concurrent UPDATE <child_table>
+# are properly ordered.
+# Originally there is a relation between parent_table and child_table.
+# However the test creates parent_table_new from parent_table and drops parent_table.
+# At the same time child table keeps the reference to the parent_table.
+#
+--let $test_case = DROP TABLE <parent_table> and concurrent UPDATE <child_table>
+--let $setup_query = CREATE TABLE parent (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent values (0); INSERT INTO child values (0, 0, 0); CREATE TABLE parent_new LIKE parent; INSERT INTO parent_new SELECT * FROM parent;
+--let $setup_check = SELECT COUNT(*) = 1 FROM parent_new
+--let $ddl_query = SET foreign_key_checks=0; DROP TABLE parent;
+--let $dml_query = UPDATE child SET data = 1 WHERE id = 0
+# After the test we expect child table to be updated
+--let $after_test_table_check = SELECT COUNT(*) = 1 FROM child WHERE id =0
+--let $drop_query = DROP TABLE parent_new, child
+--source galera_bf_bf_2.inc
+
+#
+# Test that
+# RENAME TABLE <parent_table_new> and concurrent UPDATE <child_table>
+# are properly ordered.
+# Originally there is a relation between parent_table and child_table.
+# However the test creates parent_table_new from parent_table and drops parent_table.
+# Then it renames parent_table_new back to parent_table.
+# At the same time child table keeps the reference to the parent_table.
+#
+--let $test_case = RENAME TABLE <parent_table_new> and concurrent UPDATE <child_table>
+--let $setup_query = CREATE TABLE parent (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent values (0); INSERT INTO child values (0, 0, 0); CREATE TABLE parent_new LIKE parent; INSERT INTO parent_new SELECT * FROM parent; SET foreign_key_checks=0; DROP TABLE parent;
+--let $setup_check = SELECT COUNT(*) = 1 FROM parent_new
+--let $ddl_query = RENAME TABLE parent_new TO parent
+--let $dml_query = UPDATE child SET data = 1 WHERE id = 0
+# After the test we expect child table to be updated
+--let $after_test_table_check = SELECT COUNT(*) = 1 FROM child WHERE id =0
+--let $drop_query = DROP TABLE parent, child
+--source galera_bf_bf_2.inc
+
--- a/sql/service_wsrep.cc
+++ b/sql/service_wsrep.cc
@@ -64,9 +64,18 @@ extern "C" long long wsrep_thd_trx_seqno(const THD *thd) {
     return cs.toi_meta().seqno().get();
   } else if (cs.mode() == wsrep::client_state::m_nbo) {
     return cs.nbo_meta().seqno().get();
-  } else {
-    return cs.transaction().ws_meta().seqno().get();
   }
+  return cs.transaction().ws_meta().seqno().get();
+}
+
+extern "C" long long wsrep_thd_trx_depends_on_seqno(const THD *thd) {
+  const wsrep::client_state &cs = thd->wsrep_cs();
+  if (cs.mode() == wsrep::client_state::m_toi) {
+    return cs.toi_meta().depends_on().get();
+  } else if (cs.mode() == wsrep::client_state::m_nbo) {
+    return cs.nbo_meta().depends_on().get();
+  }
+  return cs.transaction().ws_meta().depends_on().get();
 }
 
 extern "C" void wsrep_thd_self_abort(THD *thd) {
--- a/sql/sql_admin.cc
+++ b/sql/sql_admin.cc
@@ -722,8 +722,10 @@ static bool wsrep_toi_replication(THD *thd, Table_ref *tables) {
 
   wsrep::key_array keys;
 
-  if (wsrep_append_fk_parent_table(thd, tables, &keys)) {
-    return true;
+  for (Table_ref *table = tables; table; table = table->next_global) {
+    if (wsrep_append_parent_tables(thd, table, &keys)) {
+      return true;
+    }
   }
 
   /* now TOI replication, with no locks held */
--- a/sql/sql_alter.cc
+++ b/sql/sql_alter.cc
@@ -386,17 +386,21 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
     wsrep::key_array keys;
     // append tables referenced by this table
     // append tables that are referencing this table
-    if (wsrep_append_fk_parent_table(thd, first_table, &keys) ||
-        wsrep_append_child_tables(thd, first_table, &keys)) {
-      WSREP_DEBUG("TOI replication for ALTER failed");
-      return true;
+    for (Table_ref *table = first_table; table; table = table->next_global) {
+      if (wsrep_append_parent_tables(thd, table, &keys) ||
+          wsrep_append_child_tables(thd, table, &keys)) {
+        WSREP_DEBUG("TOI replication for ALTER failed. Query: %s",
+                    WSREP_QUERY(thd));
+        return true;
+      }
     }
 
     WSREP_TO_ISOLATION_BEGIN_ALTER(
         ((lex->name.str) ? lex->query_block->db : NULL),
         ((lex->name.str) ? lex->name.str : NULL), first_table, &alter_info,
         &keys) {
-      WSREP_DEBUG("TOI replication for ALTER failed");
+      WSREP_DEBUG("TOI replication for ALTER failed. Query: %s",
+                  WSREP_QUERY(thd));
       return true;
     }
   }
--- a/sql/sql_cmd_ddl_table.cc
+++ b/sql/sql_cmd_ddl_table.cc
@@ -652,10 +652,13 @@ bool Sql_cmd_create_or_drop_index_base::execute(THD *thd) {
 
 #ifdef WITH_WSREP
   wsrep::key_array keys;
-  if (wsrep_append_fk_parent_table(thd, first_table, &keys) ||
-      wsrep_append_child_tables(thd, first_table, &keys)) {
-    WSREP_DEBUG("TOI replication for CREATE/DROP INDEX failed");
-    return true;
+  for (Table_ref *table = first_table; table; table = table->next_global) {
+    if (wsrep_append_parent_tables(thd, table, &keys) ||
+        wsrep_append_child_tables(thd, table, &keys)) {
+      WSREP_DEBUG("TOI replication for CREATE/DROP INDEX failed. Query: %s",
+                  WSREP_QUERY(thd));
+      return true;
+    }
   }
   if (keys.empty()) {
     WSREP_TO_ISOLATION_BEGIN_IF(first_table->db, first_table->table_name,
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -4313,9 +4313,7 @@ int mysql_execute_command(THD *thd, bool first_level) {
     case SQLCOM_RENAME_TABLE: {
       assert(first_table == all_tables && first_table != nullptr);
       Table_ref *table;
-#ifdef WITH_WSREP
-      wsrep::key_array keys;
-#endif
+
       for (table = first_table; table; table = table->next_local->next_local) {
         if (check_access(thd, ALTER_ACL | DROP_ACL, table->db,
                          &table->grant.privilege, &table->grant.m_internal,
@@ -4324,16 +4322,6 @@ int mysql_execute_command(THD *thd, bool first_level) {
                          &table->next_local->grant.privilege,
                          &table->next_local->grant.m_internal, false, false))
           goto error;
-
-#ifdef WITH_WSREP
-        // append tables referenced by this table
-        // append tables that are referencing this table
-        if (wsrep_append_fk_parent_table(thd, table, &keys) ||
-            wsrep_append_child_tables(thd, table, &keys)) {
-          WSREP_DEBUG("TOI replication for RENAME TABLE failed");
-          goto error;
-        }
-#endif
 
         Table_ref old_list = table[0];
         Table_ref new_list = table->next_local[0];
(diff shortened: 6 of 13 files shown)


sql/service_wsrep.cc
```diff
@@ -69,6 +69,17 @@ extern "C" long long wsrep_thd_trx_seqno(const THD *thd) {
...
+extern "C" long long wsrep_thd_trx_depends_on_seqno(const THD *thd) {
+  const wsrep::client_state &cs = thd->wsrep_cs();
+  if (cs.mode() == wsrep::client_state::m_toi) {
+    return cs.toi_meta().depends_on().get();
+  } else if (cs.mode() == wsrep::client_state::m_nbo) {
+    return cs.nbo_meta().depends_on().get();
+  } else {
+    return cs.transaction().ws_meta().depends_on().get();
+  }
+}
```
jaideepkarande: I think we can remove else part and just keep `return cs.transaction().ws_meta().depends_on().get(); ` to keep compilers on some platforms and some flags happy. Also we can consider adding a header (.h file will be better, if doxygen is run then we will have to use format).

sql/sql_alter.cc
```diff
@@ -386,10 +386,12 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
...
-        wsrep_append_child_tables(thd, first_table, &keys)) {
-      WSREP_DEBUG("TOI replication for ALTER failed");
-      return true;
+    for (Table_ref *table = first_table; table; table = table->next_global) {
+      if (wsrep_append_parent_tables(thd, table, &keys) ||
+          wsrep_append_child_tables(thd, table, &keys)) {
+        WSREP_DEBUG("TOI replication for ALTER failed");
+        return true;
+      }
     }
```
jaideepkarande, line 396: Should we wrap this code in some function, I see usage in ddl, parse also and also it has some good documentation and parse part of the code which can become common?

mysql-test/suite/galera/t/galera_bf_bf_2.test
```diff
@@ -87,3 +87,41 @@
...
+--let $test_case = RENAME TABLE <parent_table_new> and concurrent UPDATE <child_table>
+--let $setup_query = CREATE TABLE parent (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent values (0); INSERT INTO child values (0, 0, 0); CREATE TABLE parent_new LIKE parent; INSERT INTO parent_new SELECT * FROM parent; SET foreign_key_checks=0; DROP TABLE parent;
+--let $setup_check = SELECT COUNT(*) = 1 FROM parent_new
+--let $ddl_query = RENAME TABLE parent_new TO parent
+--let $dml_query = UPDATE child SET data = 1 WHERE id = 0
+# After the test we expect child table to be updated
+--let $after_test_table_check = SELECT COUNT(*) = 1 FROM child WHERE id =0
+--let $drop_query = DROP TABLE parent, child
+--source galera_bf_bf_2.inc
+
```
jaideepkarande, line 127: Should we extend the test case with more scenarios like(they are working fine):

1. grand parent <- parent <- child
--let $setup_query = CREATE TABLE parent1 (id INT PRIMARY KEY);CREATE TABLE parent2 (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent1(id) ON DELETE CASCADE ON UPDATE CASCADE); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent2(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent1 values (0);INSERT INTO parent2 values (0,0,0); INSERT INTO child values (0, 0, 0); CREATE TABLE parent_new LIKE parent1; INSERT INTO parent_new SELECT * FROM parent1;
--let $ddl_query = SET foreign_key_checks=0; DROP TABLE parent1;
--let $dml_query = UPDATE child SET data = 1 WHERE id = 0

2. 2 parents
--let $setup_query = CREATE TABLE parent1 (id INT PRIMARY KEY);CREATE TABLE parent2 (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, parent_id2 INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent1(id) ON DELETE CASCADE ON UPDATE CASCADE, FOREIGN KEY(parent_id2) REFERENCES parent2(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent1 values (0);INSERT INTO parent2 values (0); INSERT INTO child values (0, 0, 0, 0); CREATE TABLE parent_new LIKE parent1; INSERT INTO parent_new SELECT * FROM parent1;
--let $ddl_query = SET foreign_key_checks=0; DROP TABLE parent1;
--let $dml_query = UPDATE child SET data = 1 WHERE id = 0

3. 2 parents
--let $setup_query = CREATE TABLE parent1 (id INT PRIMARY KEY);CREATE TABLE parent2 (id INT PRIMARY KEY); CREATE TABLE child (id INT PRIMARY KEY, parent_id INT, parent_id2 INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent1(id) ON DELETE CASCADE ON UPDATE CASCADE, FOREIGN KEY(parent_id2) REFERENCES parent2(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent1 values (0);INSERT INTO parent2 values (0); INSERT INTO child values (0, 0, 0, 0); CREATE TABLE parent_new LIKE parent1; INSERT INTO parent_new SELECT * FROM parent1;
--let $ddl_query = SET foreign_key_checks=0; DROP TABLE parent2;
--let $dml_query = UPDATE child SET data = 1 WHERE id = 0

4. 2 child
--let $setup_query = CREATE TABLE parent1 (id INT PRIMARY KEY); CREATE TABLE child1 (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent1(id) ON DELETE CASCADE ON UPDATE CASCADE); CREATE TABLE child2 (id INT PRIMARY KEY, parent_id INT, data INT, FOREIGN KEY(parent_id) REFERENCES parent1(id) ON DELETE CASCADE ON UPDATE CASCADE); INSERT INTO parent1 values (0);INSERT INTO child1 values (0, 0, 0); INSERT INTO child2 values (0, 0, 0);CREATE TABLE parent_new LIKE parent1; INSERT INTO parent_new SELECT * FROM parent1;
--let $ddl_query = SET foreign_key_checks=0; DROP TABLE parent1;
--let $dml_query = UPDATE child1 SET data = 1 WHERE id = 0; UPDATE child2 SET data = 1 WHERE id = 0;

sql/wsrep_mysqld.cc
```diff
@@ -1788,123 +1788,142 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
+          {fk->child_schema_name().c_str(), fk->child_table_name().c_str()});
+    }
+  };
+
+  auto populate_child_tables_fallback = [thd](const Table_ref *table_ref,
+                                              const char *se) {
+    std::vector<dd::String_type> children_dbs;
+    std::vector<dd::String_type> children_names;
+
+    if (thd->dd_client()->fetch_fk_children_uncached(
```
jaideepkarande, line 1885: What I understand is there no safe method to fetch this information so we need to use this function. but we will not use it for access rather write-set and check dependencies so it should be safe in this respect.

1087      @note This is a temporary workaround until WL#6049. This function will
1088            *not* take any locks protecting against DDL changes. So the returned
1089            names could become invalid at any time - e.g. due to DROP DATABASE,
1090            DROP TABLE or DROP FOREIGN KEY.
1091    */
1092  
1093    [[nodiscard]] bool fetch_fk_children_uncached(
1094        const String_type &parent_schema, const String_type &parent_name,
1095        const String_type &parent_engine, bool uncommitted,
1096        std::vector<String_type> *children_schemas,
1097        std::vector<String_type> *children_names);

jaideepkarande (commented):
Hello Kamil,
Thanks for working on code changes, changes look fine.
I have some minor comments that may help to expand the testing, please check.

sql/service_wsrep.cc
kamil-holubicki: Yeah, you are right, it was copy-paste of `wsrep_thd_trx_seqno` function. I adjusted both.

sql/sql_alter.cc
kamil-holubicki: Let's leave it as it is. Too much hassle in my opinion. These places look similar, but they are not the same.

mysql-test/suite/galera/t/galera_bf_bf_2.test
kamil-holubicki: I've added two more scenarios. We need to stop somewhere :)
In my opinion there is no need to do tests with two/more parents/children. It is the same as with only one.
As for grand parent <- parent <- child case, yes it may become problematic one day, but so far it was not, so if that problem arises, let's handle it separately.

sql/wsrep_mysqld.cc
kamil-holubicki: Yes, this is only to retrieve information for write-set keys collection purpose

jaideepkarande (approved):
Thanks Kamil for working on code changes, changes looks fine and approved from my side.

sql/sql_admin.cc
```diff
@@ -722,8 +722,10 @@ static bool wsrep_toi_replication(THD *thd, Table_ref *tables) {
 
   wsrep::key_array keys;
 
-  if (wsrep_append_fk_parent_table(thd, tables, &keys)) {
-    return true;
+  for (Table_ref *table = tables; table; table = table->next_global) {
```
satya-bodapati, line 725: A lot has changed around this. Before your change, we iterated using next_local. Now we iterate using next_global. What is the difference? Can you add a comment on why we need to iteratoe using next_global? (To ensure nobody in future attempts

for (auto table = tables; table; table = table->next_local) {..}

Is this fixing problem 1?

Solution 1:
Add 'child' table to the certification keys of DDL. This way
the dependency will be detected and set correctly.

satya-bodapati: Instead of change to loop externally, can't we simply change wsrep_do_action_for_tables() to  use next_global? 
or add a variant called wsrep_do_action_for_tables_global()?

sql/wsrep_mysqld.cc
```diff
@@ -1788,123 +1788,142 @@ wsrep::key_array wsrep_prepare_keys_for_toi(const char *db, const char *table,
...
 }
 
 /*
- * Collect wsrep keys corresponding to child tables
+ * Collect wsrep keys corresponding to child tables of the provided table
  */
-bool wsrep_append_child_tables(THD *thd, Table_ref *tables,
-                               wsrep::key_array *keys) {
+bool wsrep_append_child_tables(THD *thd, Table_ref *table,
+                               wsrep::key_array *keys, const char **table_se) {
```
satya-bodapati: Hmm, even though wsrep_append_child_tables() looks same, it has undergone significant changes...

I think the main aim here, please correct me if I am wrong.

If we cannot fetch parents from thd->dd_client()->acquire(), we want to try fetching acquire_uncached().

To me it seems we have complicated this by joingin two things. I think the benefit of your complexity is that, we acquire MDL only once? Is this optimization really worth it?

Can we do a two phase attempts, one by one? ie attempt to fetch from DD, and then attempt using uncached_acquire of DD object? I understand that we may acquire MDLs twice but I am not sure if we will significant performance benefits?

If you think it is necessary to keep the single MDL acquire logic (?), may be we can refactor to take vector in both cases?

sql/sql_parse.cc
```diff
@@ -4355,6 +4343,41 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
+           2. 'child' table still keeps a FK from 'parent' table (but here we
+              are checking parent table)
+
+           So it is not possible to deduce that 'parent' is related to 'child'
+           table, just looking at DD objects. We need to look into
+           foreigh keys table directly (see the fallback mechanism inside
+           wsrep_append_child_tables()), but for this we need the information
+           about the storage engine used by the 'parent' table.
+           Fortunately RENAME TABLE will not change the SE, so in this loop we
+           can learn what is the SE of 'parent_new' (1st iteration) and then
```
satya-bodapati: what is the 1st iteration here? this loop? "for (table = first_table; table; table = table->next_local->next_local) {" at 4317 above?

sql/sql_parse.cc
```diff
@@ -4355,6 +4343,41 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
+           pass this information back to wsrep_append_child_tables() when
+           examining 'parent' table (2nd iteration)*/
+        if (wsrep_append_parent_tables(thd, table, &keys) ||
+            wsrep_append_child_tables(thd, table, &keys, &table_se)) {
```
satya-bodapati: How is table_se useful here in this function? I don't see it being used in mysql_execute_command(). I am missing something most likely :thinking:

sql/wsrep_mysqld.cc
satya-bodapati: WL#6049 is our @dlenev's favourite WL. Have we implemented it in 8.0? Also, what is the better way to get information about FK parent table when the parent table is DROPPED and child is about to get a new parent?

pt-osc case that is shared by Kamil:
 pt-osc does the following:
           1. CREATE TABLE parent_new LIKE parent;
           2. INSERT INTO parent_new SELECT * FROM parent;
           3. SET foreign_key_checks=0;
           4. DROP TABLE parent;
           5. RENAME TABLE parent_new TO parent;
           
 At step 5, child table DD object will not have 'parent' and 'parent_new' is not in DD (during the RENAME..)

sql/sql_admin.cc
kamil-holubicki: Loop has to be external because we want a function that does the job for one table. Why? Because sometimes it is used in 'original' loop that comes from PS.

I realized that before my change, these two functions had internal loop inside. What was the effect? Double looping - external, original one and internal. Not good.

So now, the function processes one table, finds its parents/children and that's it. If we need to process a list of tables, do it in external loop (already existing loop or created for wsrep purposes)

sql/sql_parse.cc
kamil-holubicki: I tried to explain the purpose of table_se and the flow in case of RENAME TABLE in the comment. It acts as a SE cache. It caches SE of original table and passes this value along with target table.

sql/wsrep_mysqld.cc
kamil-holubicki: To be honest I didn't think about MDL optimization.
I just extended existing function. Added a fallback action (lambda) that is executed if main action fails.

>Can we do a two phase attempts, one by one? ie attempt to fetch from DD, and then attempt using uncached_acquire of DD object?

And this is exactly what happens in this function :) if `acton` fails, execute `fallback_action`

sql/sql_parse.cc
kamil-holubicki: First iteration above is to check access to existing tables.

sql/sql_admin.cc
kamil-holubicki: As for local vs global. My understanding is that global contains all tables in the query (local tables in subquery). We iterate to collect certification keys and these should be all tables in a query.

sql/sql_alter.cc
```diff
@@ -386,10 +386,12 @@ bool Sql_cmd_alter_table::execute(THD *thd) {
...
     // append tables referenced by this table
     // append tables that are referencing this table
-    if (wsrep_append_fk_parent_table(thd, first_table, &keys) ||
-        wsrep_append_child_tables(thd, first_table, &keys)) {
-      WSREP_DEBUG("TOI replication for ALTER failed");
-      return true;
+    for (Table_ref *table = first_table; table; table = table->next_global) {
+      if (wsrep_append_parent_tables(thd, table, &keys) ||
+          wsrep_append_child_tables(thd, table, &keys)) {
+        WSREP_DEBUG("TOI replication for ALTER failed");
```
satya-bodapati: Optional: It would be nice to print the table name that ALTER failed on.

sql/sql_cmd_ddl_table.cc
```diff
@@ -652,10 +652,12 @@ bool Sql_cmd_create_or_drop_index_base::execute(THD *thd) {
...
 #ifdef WITH_WSREP
   wsrep::key_array keys;
-  if (wsrep_append_fk_parent_table(thd, first_table, &keys) ||
-      wsrep_append_child_tables(thd, first_table, &keys)) {
-    WSREP_DEBUG("TOI replication for CREATE/DROP INDEX failed");
-    return true;
+  for (Table_ref *table = first_table; table; table = table->next_global) {
+    if (wsrep_append_parent_tables(thd, table, &keys) ||
+        wsrep_append_child_tables(thd, table, &keys)) {
+      WSREP_DEBUG("TOI replication for CREATE/DROP INDEX failed");
```
satya-bodapati: Optional: better to prin the table name involved in CREATE/DROP INDEX failure..

sql/sql_parse.cc
```diff
@@ -4355,6 +4343,27 @@ int mysql_execute_command(THD *thd, bool first_level) {
...
+           1. CREATE TABLE parent_new LIKE parent;
+           2. INSERT INTO parent_new SELECT * FROM parent;
+           3. SET foreign_key_checks=0;
+           4. DROP TABLE parent;
+           5. RENAME TABLE parent_new TO parent;
+
+           Here we are when executing step 5. */
+        if (wsrep_append_parent_tables(thd, table, &keys) ||
+            wsrep_append_child_tables(thd, table, &keys)) {
+          WSREP_DEBUG("TOI replication for RENAME TABLE failed");
```
satya-bodapati: RENAME failed.. do we print the query somewhere?

satya-bodapati (approved):
LGTM! some optional comments about error messages.

sql/sql_alter.cc
kamil-holubicki: Good point. Instead of doing it here, I have added the log inside `wsrep_append_parent_tables`/`wsrep_append_child_tables` functions

sql/sql_cmd_ddl_table.cc
kamil-holubicki: likewise

sql/sql_parse.cc
kamil-holubicki: Good point. I appended the query to the debug log (here and other occurences)

kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3246/

PR #2170, opened 2025-11-14
(8.0)PXC-4765: Create trigger with definer causes inconsistency error and …

jaideepkarande:
…node to disconnect

https://perconadev.atlassian.net/browse/PXC-4765

Problem:
Executing a CREATE TRIGGER statement with a DEFINER set to another user, while connected as a user without SUPER or SET_USER_ID privileges, can cause a replication inconsistency and force the node to disconnect from the cluster.

Description:
In PXC clusters CREATE TRIGGER with a DEFINER clause could lead to inconsistent behavior across nodes.
If the DEFINER user did not exist or had mismatched privileges, the node originating the transaction could reject it, while remote applier nodes might execute it successfully. This led to inconsistency and node leaving the cluster.

Resolution:
Added explicit privilege and user existence checks for DEFINER context:
 - Verify that the executing user matches the definer.
 - Allow execution only if the user has SUPER or SET_USER_ID privileges.
 - Warn if the DEFINER user does not exist in the ACL. This ensures that both locally executed and Galera-applied transactions follow the same access control logic and produce consistent outcomes.

### DIFF
--- a/mysql-test/suite/galera/t/pxc_create_trigger_failure.test
+++ b/mysql-test/suite/galera/t/pxc_create_trigger_failure.test
@@ -1,0 +1,155 @@
+################################################################################
+# This test proves that trigger failure on 1 node causes trigger failure on all
+# nodes.
+#
+# Test:
+# 0. The test requires two servers: node1 and node2
+# 1. Create user u1 and u2 and grant required privileges. Revoke SUPER from u2.
+# 2. Connect using u1 and create table t1 and t2.
+#    Create trigger trg_t1_insert with definer u1. trg_t1_insert is created.
+# 3. Assert TRIGGER table is in synch across cluster.
+# 4. Perform some operations and ensure cluster is in synch.
+# 5. Connect using u2.
+#    Create trigger trg_t1_insert_fail with definer u1.
+#    trg_t1_insert_fail is not created.
+# 6. Assert TRIGGER table does not contain trg_t1_insert_fail in any node.
+# 7. Cleanup.
+################################################################################
+
+--source include/galera_cluster.inc
+--source include/count_sessions.inc
+
+--echo
+--echo # 1. Create user u1 and u2 and grant required privileges. Revoke SUPER from u2.
+--echo
+
+CREATE USER u1@localhost IDENTIFIED BY '';
+## u1 will be the DEFINER user for the trigger, granting system user might not be needed here.
+GRANT ALL PRIVILEGES ON test.* TO u1@localhost;
+
+
+CREATE USER u2@localhost IDENTIFIED BY '';
+## u2 will try to create the trigger, but won't have the necessary privileges
+GRANT ALL PRIVILEGES ON test.* to u2@localhost;
+## Revoke a SUPER privilege.
+REVOKE SUPER ON *.* FROM u2@localhost;
+
+
+--echo
+--echo # 2. Connect using u1 and create table t1 and t2.
+--echo #    Create trigger trg_t1_insert with definer u1. trg_t1_insert is created.
+--echo
+
+--connect (u1_connection,localhost,u1,,test, $NODE_MYPORT_1)
+CREATE TABLE t1(i INT PRIMARY KEY, n VARCHAR(9));
+
+## Create a simple table t2 which the trigger will update
+CREATE TABLE t2(log_message VARCHAR(100));
+
+
+## User u1 creates a trigger successfully (u1 is the DEFINER)
+DELIMITER $$;
+CREATE DEFINER=u1@`localhost` TRIGGER trg_t1_insert
+AFTER INSERT ON t1
+FOR EACH ROW
+BEGIN
+    INSERT INTO t2 VALUES (CONCAT('Row inserted by: ', USER()));
+END$$
+DELIMITER ;$$
+
+--echo
+--echo # 3. Assert TRIGGER table is in synch across cluster.
+--echo
+
+# Assert trigger exists on node_1
+--connection node_1
+--let $assert_text= 'Trigger trg_t1_insert exists on node_1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM information_schema.TRIGGERS WHERE TRIGGER_NAME = "trg_t1_insert" AND EVENT_OBJECT_TABLE = "t1", count, 1] = 1
+--source include/assert.inc
+
+# Assert trigger exists on node_2
+--connection node_2
+--echo 'Trigger trg_t1_insert exists on node_2'
+--let $wait_condition= SELECT COUNT(*)=1 FROM information_schema.TRIGGERS WHERE TRIGGER_NAME = "trg_t1_insert" AND EVENT_OBJECT_TABLE = "t1"
+--source include/wait_condition.inc
+
+--echo
+--echo # 4. Perform some operations and ensure cluster is in synch.
+--echo
+
+--connect (u2_connection,localhost,u2,,test, $NODE_MYPORT_1)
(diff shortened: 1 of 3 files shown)


jaideepkarande:
Created a pull request now, I had not created a pull request earlier.

jaideepkarande:
1. clang has been taken care of and will be available in next update

2. Not sure of apply error, mostly not related
============
Error: ./sql/sql_trigger.cc
============
406,408d404
<   bool has_super_or_set_user_id = sctx->check_access(SUPER_ACL) ||
<                   sctx->has_global_grant(STRING_WITH_LEN("SET_USER_ID")).first;
< 
410c406,407
<       !has_super_or_set_user_id) {
---
>       !(sctx->check_access(SUPER_ACL) ||
>         sctx->has_global_grant(STRING_WITH_LEN("SET_USER_ID")).first)) {

##[error]Bash exited with code '1'.

3. Mac has boost related env error

sql/sql_trigger.cc
```diff
@@ -406,31 +406,52 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
...
-      If WSREP is enabled, then we are ALWAYS doing binlog
-      replication of some sort, and we always require the SUPER
-      privilege (or trust_function_creators).
-      So there is no need to mention anything about the binlog.
-    */
-    if (WSREP(thd))
+  LEX *lex = thd->lex;
+  // Definer check: If a definer is specified and is different from the current
+  // user.
+  bool definer_is_not_current_user =
```
kamil-holubicki: Maybe it is only me, but I suggest to avoid naming variables like this. For analysyis it is easier to do
```
bool definer_is_current_user = ....

if (definer_is_current_user) { ...}
if (!definer_is_current_user) {...}
```
rather than:
```
bool definer_not_is_current_user = ....

if (!definer_is_not_current_user) { ...}
if (definer_is_not_current_user) {...}
```

sql/sql_trigger.cc
```diff
@@ -727,4 +748,3 @@ int wsrep_create_trigger_query(THD *thd, uchar **buf, size_t *buf_len) {
                              buf_len);
 }
 #endif /* WITH_WSREP */
```
kamil-holubicki, line 749: Unnecessary change: New line at the end of the file was removed

sql/sql_trigger.cc
```diff
@@ -406,31 +406,52 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
...
+
+  if (lex->definer && sctx->can_operate_with(lex->definer, consts::system_user,
+                                             true /* report_error */))
+    return true;
+
 #else
   if (!trust_function_creators && mysql_bin_log.is_open() &&
-      !(sctx->check_access(SUPER_ACL) ||
-        sctx->has_global_grant(STRING_WITH_LEN("SET_USER_ID")).first)) {
+      !has_super_or_set_user_id) {
```
kamil-holubicki: in `#else` part we should keep the original PS code

sql/sql_trigger.cc
```diff
@@ -406,31 +406,52 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
     applies to them too.
   */
   Security_context *sctx = thd->security_context();
+  bool has_super_or_set_user_id =
```
kamil-holubicki: PXC specific changes should be wrapped in `#ifdef WITH_WSREP` block. 
There is GitHub `Percona_Server_Diff_Check` check that checks for this, and currently it fails.

mysql-test/suite/galera/t/pxc_create_trigger_failure.test
```diff
@@ -0,0 +1,155 @@
...
+--let $assert_cond= [SELECT COUNT(*) FROM t1 WHERE i = 1] = 1
+--source include/assert.inc
+
+--let $assert_text= 'Check t2 was updated by the trigger'
+--let $assert_cond= [SELECT COUNT(*) FROM t2 WHERE log_message LIKE "Row inserted by: u2%"] = 1
+--source include/assert.inc
+
+--connection node_2
+--let $assert_text= 'Check t1 was updated.'
+--let $assert_cond= [SELECT COUNT(*) FROM t1 WHERE i = 1] = 1
```
kamil-holubicki: In my opinion we should wait for this condition on node_2 rather than assert. It may happen that the change was not applied on node_2 yet.

mysql-test/suite/galera/t/pxc_create_trigger_failure.test
```diff
@@ -0,0 +1,155 @@
...
+
+# Assert trigger exists on node_1
+--connection node_1
+--let $assert_text= 'Trigger trg_t1_insert exists on node_1'
+--let $assert_cond= [SELECT COUNT(*) AS count FROM information_schema.TRIGGERS WHERE TRIGGER_NAME = "trg_t1_insert" AND EVENT_OBJECT_TABLE = "t1", count, 1] = 1
+--source include/assert.inc
+
+# Assert trigger exists on node_2
+--connection node_2
+--let $assert_text= 'Trigger trg_t1_insert exists on node_2'
```
kamil-holubicki: In my opinion we should wait for this condition on node_2 rather than assert. It may happen that the change was not applied on node_2 yet.

In general, you can not assume that changes done on node_1 are immediately visible on node_2.

kamil-holubicki (changes requested):

mysql-test/suite/galera/t/pxc_create_trigger_failure.test
jaideepkarande: Thanks Kamil for pointing it out, corrected the test case.

mysql-test/suite/galera/t/pxc_create_trigger_failure.test
jaideepkarande: Thanks, corrected the test case.

sql/sql_trigger.cc
jaideepkarande: Thanks, I have move the code now.

sql/sql_trigger.cc
jaideepkarande: Something for me to check why this is happening. Thanks added the line.

sql/sql_trigger.cc
jaideepkarande: Agree, I have made the changes as per suggestion.

Maybe here we can keep as `definer_is_not_current_user` because of the similar usage at other places(sql_view and parser), that way consistency and search of code becomes easy. Please let me know and I will rollback to the previous change.
```
  if (lex->definer &&
      (strcmp(lex->definer->user.str,
              thd->security_context()->priv_user().str) != 0 ||
       my_strcasecmp(system_charset_info, lex->definer->host.str,
                     thd->security_context()->priv_host().str) != 0)) {
```

sql/sql_trigger.cc
jaideepkarande: I missed to reply here, but it has been taken care of when working on `#else` part.

kamil-holubicki (approved):
LGTM

kamil-holubicki (changes requested):
Force push done 4 days ago changed this branch from GCA branch to 8.0 HEAD-based one, so it can't be merged to 8.4-based branch (GCA process).
I think it is possible to use GCA here. Do we need any changes that were introduced after GCA?

jaideepkarande:
@kamil-holubicki I have re-created the branch and forced push it as per GCA. Please review the refresh patch.
There are no fresh code changes in this patch. But merge is blocked due to the refresh.

kamil-holubicki (approved):
lgtm

PR #2175, opened 2025-11-24
(8.4) PXC-4789: DDL with FK reference causes MDL BF-BF conflict and evict the nodes out of cluster.

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4789

Problem 1:
There are two tables: 'parent' and 'child'. 'child' holding the FK of
the 'parent'. If we do SET foreign_key_checks=0, it is possible to drop
'parent'.
If such a DROP statement is executed in parallel to UPDATE on 'child'
table, it leads to BF-BF abort conflict on replica.

Cause 1:
DROP TABLE 'parent' acquires MDL locks on 'parent' and 'child' tables.
UPDATE 'child' acquires MDL lock on 'child' table. So these transactions
cannot apply in parallel on replica side - they are dependent.
However, the certification key related to 'child' table was not added
to the writeset of DDL, so certification process considered DDL and DML
not dependent.

Solution 1:
Add 'child' table to the certification keys of DDL. This way
the dependency will be detected and set correctly.

Problem 2:
pt-osc creates 'parent_new' table, then drops 'parent' table and finally
renames 'parent_new' to 'parent'.
Such RENAME of 'parent_new', and simultaneous UPDATE on 'child' table,
cause BF-BF abort on replica node.

Cause 2:
RENAME acquires MDL locks on 'parent_new' 'parent' and 'child' tables.
UPDATE acquires MDL lock on 'child' table.
So these transactions are dependent and should be ordered one after
another. However, when renaming 'parent_new' to 'parent', it is not
possible to get information about 'child' table from DD ('parent_new'
does not have children, 'parent' table object does not exist in DD).
Because of this 'child' table was not added to the certification keys of
DDL. Dependency was not detected, and transactions were allowed to run
in parallel on the replica side causing BF-BF conflict.

Solution:
Improve mechanism of table children detection. In case of no parent
table object in DD, fall back to querying foreign keys table directly.

### DIFF
--- a/sql/sql_parse.cc
+++ b/sql/sql_parse.cc
@@ -4290,9 +4290,7 @@ int mysql_execute_command(THD *thd, bool first_level) {
     case SQLCOM_RENAME_TABLE: {
       assert(first_table == all_tables && first_table != nullptr);
       Table_ref *table;
-#ifdef WITH_WSREP
-      wsrep::key_array keys;
-#endif
+
       for (table = first_table; table; table = table->next_local->next_local) {
         if (check_access(thd, ALTER_ACL | DROP_ACL, table->db,
                          &table->grant.privilege, &table->grant.m_internal,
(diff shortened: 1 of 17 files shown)


satya-bodapati (approved):

sql/sql_parse.cc
```diff
@@ -4423,18 +4433,24 @@ int mysql_execute_command(THD *thd, bool first_level) {
       }
 
 #ifdef WITH_WSREP
+      wsrep::key_array keys;
+      bool replicate = false;
       for (Table_ref *table = all_tables; table; table = table->next_global) {
         if (!lex->drop_temporary &&
             (!thd->is_current_stmt_binlog_format_row() ||
              !find_temporary_table(thd, table))) {
```
jaideepkarande, line 4441: Please confirm we will need to move code to 8.0 also.

jaideepkarande (approved):
Thank Kamil for working on changes, I have some minor comment w.r.t. 8.0 which does not stop approve.

PR #2179, opened 2025-11-28
PXC-4765: Create trigger with definer causes inconsistency error and …

jaideepkarande:
…node to disconnect

https://perconadev.atlassian.net/browse/PXC-4765

Problem:
Executing a CREATE TRIGGER statement with a DEFINER set to another user, while connected as a user without SUPER or SET_USER_ID privileges, can cause a replication inconsistency and force the node to disconnect from the cluster.

Description:
In PXC clusters CREATE TRIGGER with a DEFINER clause could lead to inconsistent behavior across nodes.
If the DEFINER user did not exist or had mismatched privileges, the node originating the transaction could reject it, while remote applier nodes might execute it successfully. This led to inconsistency and node leaving the cluster.

Resolution:
Added explicit privilege and user existence checks for DEFINER context:
 - Verify that the executing user matches the definer.
 - Allow execution only if the user has SUPER or SET_USER_ID privileges.
 - Warn if the DEFINER user does not exist in the ACL. This ensures that both locally executed and Galera-applied transactions follow the same access control logic and produce consistent outcomes.


jaideepkarande:
Unexpected files are being shown as changed, will create fresh pull request.

PR #2186, opened 2025-12-02
(trunk) PXC-4770: Move "Pro" features into the community version

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4770

PXB-9.1.0-2 provides --lock-ddl=REDUCED option. It will be bundled with next versions of PXC. Use --lock-ddl=REDUCED option by default.


jaideepkarande (approved):
LGTM

PR #2187, opened 2025-12-02
(8.4) K8SPXC-1724: Operator keep killing the SST process while in execution

kamil-holubicki:
https://perconadev.atlassian.net/browse/K8SPXC-1724

Problem:
Clone SST does not create sst_in_progress file on the joiner side. Operator is not able to detect that node is receiving SST, considers it as not unhealthy and kills it.

Solution:
Create sst_in_progress file for the time of SST receive on the joiner side.


Tusamarco (approved):
lgtm

PR #2188, opened 2025-12-03
(8.0)PXC-3583: validate-config reports "ssl-ca, ssl-cert, and ssl-key must…

jaideepkarande:
… all be defined to use encrypted mode traffic"

https://perconadev.atlassian.net/browse/PXC-3583

Problem:
validate-config does not work when wsrep is enabled.

Description:
If pxc-encrypt-cluster-traffic=ON, PXC requires socket.ssl_{key,cert,ca}, but these are not yet loaded during validate-config. This causes a false failure: "ssl-ca, ssl-cert and ssl-key must all be defined".

Resolution:
PXC should skip wsrep initialization when --validate-config is used.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_validate_config_ssl.result
+++ b/mysql-test/suite/galera/r/pxc_validate_config_ssl.result
@@ -1,0 +1,52 @@
+
+# 1. Shutdown the server and initialize some configurations
+
+
+# 2. Validate simple server configuration.
+
+"ERROR not seen simple configuration."
+Pattern not found.
+
+# 3. Validate SSL with PXC Encryption ON.
+
+"ERROR not seen when SSL parameters are present and PXC Encryption is ON"
+Pattern not found.
+
+# 4. Validate SSL with PXC Encryption OFF.
+
+"ERROR not seen when SSL parameters are present and PXC Encryption is OFF"
+Pattern not found.
+
+# 5. Validate PXC Encryption OFF.
+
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3298/

kamil-holubicki (commented):
Hi @jaideepkarande , The patch looks fine. I understand that the idea is to skip check if wsrep ssl context is initialized in case of --validate-config.
Wsrep-ssl context is not initialized in such a case (see: init_server_components()).

What I'm confused of is the commit message:
```
Resolution:
PXC should skip wsrep initialization when --validate-config is used.
```
What initialization do you mean? I see that everything related to wsrep/galera is initialized (besides the above-mentioned ssl-context). For the time of config validation Galera is initialized without SSL, which seems to be fine.

Could you please clarify (or if necessary clarify the commit message)? Maybe we could also add some code comment?

kamil-holubicki:
Clang format check is failing

jaideepkarande:
Thanks for the comments, corrected the commit message and added the code comments as per suggestion.
clang not sure why it happened, thanks for the updates over slack, I have now downgraded my clang from version 18 to version 14 and hopefully clang issues may not happen again.

kamil-holubicki (approved):
LGTM. Please wait for GH checks to complete before merging.

PR #2194, opened 2025-12-12
(8.0)PXC-4799: PXC member used for backups fails with Inconsistency when p…

jaideepkarande:
…ending DDL clashes with FTRWL

Problem:

Description:
FTWRL caused MDL lock and waited for DDL to finish in the PXC. FTWRL specifically waits in desync_and_pause->pause statement waiting for concurrent pause statement to finish.

Resolution:
DDLs cannot be executed in parallel with FTWRL in local node it will fail. However remote node not aware of FTWRL on another node accepts the DDL and sends across the cluster.
So, we prevent FLUSH TABLES WITH READ LOCK when the session already holds explicit table locks.
By rejecting FLUSH TABLES WITH READ LOCK in this situation, we avoid the deadlock-like condition and allow the remote DDL to proceed. NOTE: Even with this safeguard, if the DDL waiting for MDL exceeds lock_wait_timeout, an inconsistent state may still occur. The session must execute UNLOCK for the DDL to complete.
This behavior is specific to PXC. In standalone PS/PXC servers, such conflicting operations fail naturally, but in multi-node PXC deployments, DDL may arrive from another node, making this protection necessary.


jaideepkarande:
Test case ran on PS:
################################################################################
# Post FLUSH , DDL are not allowed on same server
################################################################################
--source include/have_debug.inc
--source include/master-slave.inc

--connection master
CREATE DATABASE db1;
USE db1;
--source include/sync_slave_sql_with_master.inc

################################################################################
# Post FLUSH , DDL are not allowed on same server
################################################################################
--connection slave
LOCK INSTANCE FOR BACKUP;
FLUSH TABLES WITH READ LOCK;
--error 1223
UNLOCK TABLES;
UNLOCK INSTANCE;

################################################################################
# Post LOCK..BACKUP , DDLs are allowed on same server
################################################################################
LOCK INSTANCE FOR BACKUP;
CREATE TABLE t2 (id INT PRIMARY KEY, a INT);
UNLOCK INSTANCE;

################################################################################
# Post FLUSH , DDL are not allowed on same server
################################################################################
--connection slave
FLUSH TABLES WITH READ LOCK;
--error 1223
CREATE TABLE t3 (id INT PRIMARY KEY, a INT);
UNLOCK TABLES;

################################################################################
# Customer scenario on ASYNCH
# Slave executes LOCK and FLUSH and DDL comes from MASTER
################################################################################

--connection slave
LOCK INSTANCE FOR BACKUP;

--connection slave
--send FLUSH TABLES WITH READ LOCK;

--connection master
--send CREATE TABLE tsync (id INT PRIMARY KEY, a INT);

--connection slave
--reap

--connection slave
## FTWRL is not going to block for pause/desynch and UNLOCK will give chance for DDL to finish(assuming)
FLUSH TABLES WITH READ LOCK;

UNLOCK TABLES;
UNLOCK INSTANCE;

--connection master
--reap
DROP DATABASE db1;

--source include/rpl_end.inc

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3300/console


jaideepkarande:PXC-4799-8.0:BIN4799$ grep "fail " test.out 
[ 12%] galera_3nodes.galera_parallel_autoinc_manytrx2 'enc_on' w3  [ fail ]
[ 87%] perfschema.idx_compare_ees_by_thread_by_error w4  [ fail ]
[100%] shutdown_report                          w0  [ fail ]
jaideepkarande:PXC-4799-8.0:BIN4799$ grep retry test.out 
2025-12-13T17:59:18.138229Z 2 [Note] [MY-000000] [Galera] Requesting state transfer failed: No donor candidates temporarily available in suitable state. Will keep retrying every 1 second(s)
[ 75%] galera_3nodes.galera_parallel_autoinc_manytrx2 'enc_on' w3  [ retry-pass ]  20534
[ 75%] galera_3nodes.galera_parallel_autoinc_manytrx2 'enc_on' w3  [ retry-pass ]  20946
[ 87%] perfschema.idx_compare_ees_by_thread_by_error w4  [ retry-pass ]  119823
[ 87%] perfschema.idx_compare_ees_by_thread_by_error w4  [ retry-pass ]  120494
2025-12-13T17:59:18.138229Z 2 [Note] [MY-000000] [Galera] Requesting state transfer failed: No donor candidates temporarily available in suitable state. Will keep retrying every 1 second(s)

jaideepkarande:
Test case:
--echo #
--echo # 4.A. Hold backup lock on node-2.
--echo #
--connect node_2a, 127.0.0.1, root, , test, $NODE_MYPORT_2

--connection node_2
LOCK INSTANCE FOR BACKUP;

--echo #
--echo # 4.B. Execute DDL on node-1
--echo #

--connection node_1
--send CREATE TABLE t_synch1 (id INT PRIMARY KEY, a INT);

--echo #
--echo # 4.C. Execute FTWRL on node-2 it fails.
--echo #      UNLOCK the instance.
--echo #

--connection node_2
--send FLUSH TABLES WITH READ LOCK;

--sleep 5
--echo #
--echo # 4.D. Let DDL finish on node1 and unlock node2.
--echo #

--connection node_2a
UNLOCK TABLES;
UNLOCK INSTANCE;
--echo UNLOCK SUCCESSFUL
--connection node_2
--reap

--connection node_1
--reap

kamil-holubicki (commented):
I think I understand what is the intention of the fix (thank you for explanation over Slack).
The problem I see is that if some tool executes in one session:
```
LOCK INSTANCE FOR BACKUP;
FLUSH TABLES WITH READ LOCK;
```
which is allowed sequence, this tool will fail.

On the flip side, I'm not sure if we can figure out any better solution. We could detect if there is no ongoing DDL (TOI) and allow FTWRL, but I can imagine it is not a trivial problem to solve because of concurrency issues that will appear.

The bottom line is: I'm not sure if this behavior change can be accepted.

kamil-holubicki:
For the record:
https://github.com/percona/percona-xtradb-cluster/pull/2200

sql/lock.cc
```diff
@@ -1224,22 +1225,69 @@ bool Global_read_lock::make_global_read_lock_block_commit(THD *thd) {
...
+         may be a thread in the queue that is going to do this in the future.
+         As the consequence, executing LOCK INSTANCE FOR BACKUP -> FTWRL may fail sometimes if FTWRL
+         is executed when any writesets are waiting to be processed.
+
+         Solution for this would be try_desync_and_pause(timeout)....
+         */
+
+      // Try non-blocking desync and pause. If it would block (e.g., due to
+      // pending transactions waiting on locks), it returns undefined seqno.
+      paused_seqno = server_state.try_desync_and_pause();
```
kamil-holubicki: Would it really be possible to do try_desync_and_pause(timeout)?
We discussed it on Slack. It seems to be not so much work to do.
Just a configuration parameter, default let's say 30sec, Inside Galera I imagine it as a loop repeated every 100ms (similar to what we have in try_toi or something). 
Shouldn't be hard to implement, but in my opinion will solve many waiting issues.

PR #2198, opened 2026-01-02
PXC-4800: Fix replication of operations done from component level(8.0)

jaideepkarande:
Modify component to replication transactions across cluster.

### DIFF
--- a/sql/server_component/table_access_service.cc
+++ b/sql/server_component/table_access_service.cc
@@ -41,7 +41,9 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301  USA */
 #include "sql/table_trigger_dispatcher.h"
 #include "sql/transaction.h"
 #include "thr_lock.h"
-
+#ifdef WITH_WSREP
+#include "sql/wsrep_trans_observer.h"
+#endif /* WITH_WSREP */
 /* clang-format off */
 /**
   @page PAGE_TABLE_ACCESS_SERVICE Table Access service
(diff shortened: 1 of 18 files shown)


sql/server_component/table_access_service.cc
```diff
@@ -492,6 +493,144 @@ static String *from_api(my_h_string api) {
...
+  @class PXC_wsrep_replicate_dml
+
+  @brief Manages wsrep DML replication for a given THD.
+
+  This class is a RAII wrapper that ensures wsrep replication is correctly
+  initialized and torn down for DML operations performed through the
+  table access service. It handles starting and ending transactions,
+  and cleans up the wsrep state in the THD upon destruction.
+*/
+class PXC_wsrep_replicate_dml {
```
kamil-holubicki: Can we name this class using a noun, not a verb? Like `PXC_wsrep_dml_replicator`?

sql/server_component/table_access_service.cc
```diff
@@ -492,6 +493,144 @@ static String *from_api(my_h_string api) {
...
+ private:
+  THD *m_thd{nullptr};
+  bool m_replicate_transactions{false};
```
kamil-holubicki: Transactions should be replicated by default.

plugin/audit_log_filter/audit_table/base.cc
```diff
@@ -61,6 +65,9 @@ std::unique_ptr<TableAccessContext> AuditTableBase::open_table() noexcept {
 
   ta_context->ta_session =
       ta_factory_srv->create(ta_context->thd, kAccessedTableCount);
+#ifdef WITH_WSREP
+  ta_factory_srv->replicate(ta_context->ta_session, m_replicate_in_galera);
```
kamil-holubicki: Transactions should replicate by default. So even without calling `ta_factory_srv->replicate(..., true)' they should be replicated.
If the component wants to disable replication in a particular ta_session, it should call `ta_factory_srv->replicate(..., false)`, but it should be considered as a special case.

plugin/audit_log_filter/audit_table/base.h
```diff
@@ -109,6 +109,11 @@ class AuditTableBase {
 
  private:
   const std::string m_db_name;
+#ifdef WITH_WSREP
+ public:
+  static bool m_replicate_in_galera;
```
kamil-holubicki: I agree that we can leave replicate on/off functionality at the level of table_access_service, but I'm not sure if having it in audit_log_filter is a good idea.
My suggestion is to remove everything related to replication on/off from audit log filter. This way, replication will always be enabled as per requirement, and there will not be a single change in audit log filter

sql/server_component/table_access_service.cc
```diff
@@ -492,6 +493,144 @@ static String *from_api(my_h_string api) {
...
+
+ public:
+  /**
+    Constructor.
+    @param thd The thread context to manage.
+  */
+  explicit PXC_wsrep_replicate_dml(THD *thd) : m_thd(thd) {
+    assert(m_thd != nullptr);
+    // By default replication should be off
```
kamil-holubicki: should be `on` by default

sql/server_component/table_access_service.cc
```diff
@@ -607,7 +749,9 @@ Table_access_impl::Table_access_impl(THD *thd, size_t count)
     this->m_clear_mysys = !my_thread_is_inited();
     my_thread_init();
   }
-
+#ifdef WITH_WSREP
+  m_wsrep_replicator = new PXC_wsrep_replicate_dml(m_child_thd);
```
kamil-holubicki: Could we make it some kind of smart pointer?

sql/server_component/table_access_service.cc
```diff
@@ -818,7 +968,13 @@ int Table_access_impl::begin() {
   if (!is_access_to_tables_supported(m_table_array)) {
     return TA_ERROR_OPEN;
   }
-
+#ifdef WITH_WSREP
+  // We can keep it in set_transaction however it is better to start the trx
+  // here due to begin
+  if (m_wsrep_replicator && m_wsrep_replicator->start_trx_if_not_started()) {
```
kamil-holubicki: (Several occurrences). Is there a path when we don't have m_wsrep_replicator set?

sql/server_component/table_access_service.cc
```diff
@@ -841,6 +997,12 @@ int Table_access_impl::commit() {
   if (trans_commit_stmt(m_child_thd)) {
     return 1;
   }
+#ifdef WITH_WSREP
+  // For some reasin below does not work.
```
kamil-holubicki: ???

sql/server_component/table_access_service.cc
```diff
@@ -492,6 +493,144 @@ static String *from_api(my_h_string api) {
...
+    @param replicate_transactions If true, enables replication for the session.
+    @return 0 on success, 1 on failure.
+  */
+  int set_replication_operations(bool replicate_transactions) {
+    //assert(replicate_transactions); // Default replication is off
+    assert(m_thd != nullptr);
+    if (replicate_transactions) {
+      m_thd->variables.wsrep_on = true;  // Needed for WSREP(thd) to work
+      wsrep_open(m_thd);
+      if (wsrep_before_command(m_thd)) {
```
kamil-holubicki, line 551: Shouldn't we call it before any command? (begin, insert, commit) (before `begin()` and before each `table_update_srv->insert`)?
Ideally, this method ('set_replication_operations`) should not be called (replication enabled by default)

sql/server_component/table_access_service.cc
```diff
@@ -492,6 +493,144 @@ static String *from_api(my_h_string api) {
...
+
+  /**
+    Starts a wsrep transaction if one has not already been started.
+    @return 0 on success, non-zero on error.
+  */
+  int start_trx_if_not_started() {
+    int error = 0;
+    if (m_replicate_transactions &&
+        m_thd->wsrep_next_trx_id() == WSREP_UNDEFINED_TRX_ID) {
+      error = wsrep_before_statement(m_thd);
```
kamil-holubicki, line 575: A transaction can have more than one insert. I think this one should be called before each `table_update_srv->insert` (and begin, commit as well. In other words, before each statement). See `wsrep_before_command`

kamil-holubicki (commented):
I think the idea is fine. 
I have a few concerns (please see comments)
I didn't check MTR tests.

sql/server_component/table_access_service.cc
jaideepkarande: I ran into a limitation with table_access_factory_v1: the service templates do not allow default arguments. Any attempt to change this would have caused a significant divergence from the original service prototype, which I wanted to avoid.

To work around this, I kept the default behavior and explicitly invoked set_replication_operations() immediately after constructor to set it to ON, which in turn calls wsrep_open() etc.

In Galera, the transaction lifetime is managed using enums, and invoking certain operations at the wrong phase can lead to assertion failures. To avoid this, I placed the call to set_replication_operations() safely right after the constructor. I initially considered using enums or flags to block unwanted initialization or prevent later calls to set_replication_operations() during the transaction lifecycle. However, since the placement of these calls is fully under our control and not exposed to customers, I decided against adding that extra complexity.

My original thinking was that not all customers would want the audit log filter to be always ON. For that reason, the default behavior in the audit log filter was ON(i.e. function call you have observed later), but a UDF was provided to allow replication to be turned OFF. That said, enforcing an always-ON behavior with no option to disable it would simplify the code significantly—though I do expect that some customers may raise concerns in the future if the audit log filter sees widespread adoption.

sql/server_component/table_access_service.cc
jaideepkarande: Actually no, so it is safe to directly call m_wsrep_replicator->start_trx_if_not_started()

sql/server_component/table_access_service.cc
jaideepkarande: Fixed.

sql/server_component/table_access_service.cc
jaideepkarande: This was internal note first, but thought to make it permanent, I was expecting commit will happen here but actual call happened in destructor.

sql/server_component/table_access_service.cc
jaideepkarande: This was 1 of the scenario in my consideration so I added a test case for that:
    {"INSERT-CUSTOMER-INSERT-MULTIPLE", test_insert_customer_insert_multiple},

```
  for (int i = 2; i <= 6; ++i) {
    id_value = i;
    if (fa_integer_srv->set(access, table, ID_COL, id_value)) {
      result = "set(id) failed";
      goto cleanup;
    }

    char name_buffer[20];
    snprintf(name_buffer, sizeof(name_buffer), "Unknown Person - %d", i);
    string_converter_srv->convert_from_buffer(name_value, name_buffer,
                                              strlen(name_buffer), utf8mb4_h);

    if (fa_varchar_srv->set(access, table, NAME_COL, name_value)) {
      result = "set(name) failed";
      goto cleanup;
    }

    fa_null_srv->set(access, table, ADDRESS_COL);

    rc = ta_update_srv->insert(access, table);
    if (rc != 0) {
      result = "insert() failed";
      goto cleanup;
    }
  }
```

sql/server_component/table_access_service.cc
jaideepkarande: Fixed as of now, but I have a second thought since in destructor wind up happens and it should finish before resource cleanup. reset does call the destructor immediately but still old style would work better.

sql/server_component/table_access_service.cc
jaideepkarande: I ran into a limitation with table_access_factory_v1: the service templates do not allow default arguments. Any attempt to change this would have caused a significant divergence from the original service prototype, which I wanted to avoid.

To work around this, I kept the default behavior and explicitly invoked set_replication_operations() immediately after constructor to set it to ON, which in turn calls wsrep_open() etc.

In Galera, the transaction lifetime is managed using enums, and invoking certain operations at the wrong phase can lead to assertion failures. To avoid this, I placed the call to set_replication_operations() safely right after the constructor. I initially considered using enums or flags to block unwanted initialization or prevent later calls to set_replication_operations() during the transaction lifecycle. However, since the placement of these calls is fully under our control and not exposed to customers, I decided against adding that extra complexity.

My original thinking was that not all customers would want the audit log filter to be always ON. For that reason, the default behavior in the audit log filter was ON(i.e. function call you have observed later), but a UDF was provided to allow replication to be turned OFF. That said, enforcing an always-ON behavior with no option to disable it would simplify the code significantly—though I do expect that some customers may raise concerns in the future if the audit log filter sees widespread adoption.

jaideepkarande:
Should this changes go through QA?

kamil-holubicki:
Please check failing GH checks.
Especially Clang_Format_Check and Percona_Server_Diff_Check.
Others probably fail because of out of space problem, but please confirm.

sql/server_component/table_access_service.cc
kamil-holubicki: Please address this comment

kamil-holubicki (approved):
lgtm

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3306/consoleFull

PR #2200, opened 2026-01-09
PXC-4799 poc

kamil-holubicki:


sql/lock.cc
```diff
@@ -1151,7 +1152,7 @@ void Global_read_lock::unlock_global_read_lock(THD *thd) {
         wsrep_locked_seqno = WSREP_SEQNO_UNDEFINED;
         server_state.resume();
         pause_provider(false);
-      } else {
+      } else if (provider_paused) {
```
jaideepkarande: I’m wondering whether this is needed, because in these scenarios FTWRL would have already failed, which essentially makes the UNLOCK requirement moot.

PR #2214, opened 2026-01-22
(8.0) PXC-4823: Async PXC replica stalls replication when replicates OPTIMIZE TABLE

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4823

Problem:
If PXC node is a replica for async source, OTPIMIZE TABLE executed on async source causes async replication stuck.

Cause:
When async-replicated transaction is to be applied, its sequence number is scheduled in wsrep_async_monitor. Once scheduled, the transaction is expected to either enter-leave the monitor, or skip the scheduled sequence number.

For DDLs, entering the monitor is done in wsrep_to_isolation_begin and leaving in wsrep_to_isolation_end.

For table admin tasks, we decide if replication should be done for a given command in wsrep_toi_replication. We do so for OPTIMIZE and REPAIR.
The logic in this function skipped replication of all table admin DDLs if async replica worker was the originator.

Fix:
Allow replication of table admin DDLs if async replica worker is the originator (note that for multi threaded replica it will be SLAVE_WORKER thread, and for single threaded replicas it will be SLAVE_SQL thread)

### DIFF
--- a/mysql-test/suite/galera/r/galera_as_slave.result
+++ b/mysql-test/suite/galera/r/galera_as_slave.result
@@ -28,6 +28,18 @@ Warnings:
 Level	Warning
 Code	1287
 Message	'SHOW SLAVE STATUS' is deprecated and will be removed in a future release. Please use SHOW REPLICA STATUS instead
+CREATE TABLE t1 (a INT PRIMARY KEY);
+INSERT INTO t1 VALUES (0);
+OPTIMIZE TABLE t1;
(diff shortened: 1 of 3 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3312/
All four failed tests are instabilities.

jaideepkarande (approved):
LGTM

PR #2217, opened 2026-01-23
(trunk) PXC-4802: Misleading log message `Initiating SST cancellation` during handling fatal signal

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4802

Problem:
When the server exits because of abort or fatal signal it prints `Initiating SST cancellation` message. It is printed even if there is no SST in progress, which is misleading.

Cause:
The server installs signal handler to do the cleanup prior exit. One of cleanup actions is stopping a potentially ongoing SST. The problematic message was printed always, regardless of whether SST stop actions were executed afterwards or not.

Solution:
Print the message only if there is ongoing SST to be stopped.


jaideepkarande (approved):
LGTM

PR #2224, opened 2026-02-11
PXC-4845: Node remains up and running after IST failure, causing othe…

kamil-holubicki:
…r nodes to become unresponsive

https://perconadev.atlassian.net/browse/PXC-4845

Code changes in Galera. Server side contains only MTR test.

Problem:
After receiving malformed IST, Joiner node does not leave the cluster but continues. Donor node gets blocked after some time.

Cause:
Because of not ideal WSREP checkpointing mechanism it may happen that wsrep checkpoint stored in SE is lower than actually commited transaction seqno.
On the other hand, Galera is keeping track of its position in grastate.dat file, which always reflects the actual position. Let's say grastate.dat says we are at position 10, but SE checkpoint is 5.
If the node joins, and is able to join via IST, it will request missing sqno range in SST request. This is the range figured ot from grastate.dat file. So the Joiner will request seqnos > 10. Donor will skip SST and start serving seqnos > 10. At the same time, the Joiner will read SST seqno from SE. As it sees number 5, it starts IST receiver to receive > 5.
There is ongoing write traffic on Donor. It is replicated to the Joiner. But the 1st seqno received is 11. Joiner detects it and skips IST, but continues to operate.
After IST, the Joiner starts processing queued writesets (the ones that were executed on Donor while IST was in progress). As Joiner's Apply and Commit monitors are not properly initialized (both have last_left==-1), these transactions wait on Commit monitor causing applier threads to be blocked.
In the meantime the write traffic on the donor node continues. It is replicated to the Joiner node, but after short period of time, Joiner's receive queue grows up and Flow Control mechanism is activated. Donor is blocked.

Solution:
There is no good solution for storing wsrep checkpoints in SE. Some time ago we had wsrep_commit_queue on the server side, but it didn't solve this problem at all and caused deadlocks, so it was removed (server side commit a37608c7).
At the current stage, I introduced a proper handling of the problematic scenario on Galera side. The node is printing message and shutting down. PXC starting script will solve the problem by recovering Galera position from SE and stating the node with this position.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_malformed_ist.result
+++ b/mysql-test/suite/galera/r/pxc_malformed_ist.result
@@ -1,0 +1,14 @@
+# Populate nodes with data
+# Shut down node_2 
+# Wait for node_2 to be down
+# Execute some transactions on node_1 for node_2 to rejoin with IST
+# Setup node_1 to server minimal IST range
+SET GLOBAL wsrep_provider_options = 'dbug=d,serve_mimimal_ist';
+# Start node_2 in background. Force wrong IST starting seqno and wait after detection of malformed IST.
+# Wait for node_2 to join the cluster
+# Execute some transactions on node_1 while node_2 is blocked in IST. This will cause node_2 receive queue to grow up
+# Wait for node_1 to continue (it should abort after detecting malformed IST)
+include/assert_grep.inc [Ensure that Joiner aborted after receiving wrong IST]
+# Start node_2 without forcing malformed IST
+# restart
+DROP TABLE t1;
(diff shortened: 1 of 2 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3358/
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3359/

jaideepkarande (approved):
Thanks Kamil, LGTM.
Just a suggestion, should we add big-test to pxc_malformed_ist.test‎?

PR #2225, opened 2026-02-12
PXC-4683: Node leaves the cluster after an error in a stored procedure

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-4823

Problem:
TOI-failed inside a stored procedure e.g. ALTER TABLE on a non-existing table marked the node inconsistent

Cause:
The Galera apply-error voting payload was built from the diagnostics area and included warning codes in addition to the real error code. For cases involving deprecated integer display width (tinyint(1) -> warning 1681), nodes could produce different condition lists (1146 vs 1146+1681), leading to different vote hashes and eventually causing inconsistency.

Fix:
Make wsrep_store_error() store only SL_ERROR conditions and if no error if found check DA error), thus voting is based on the failing error(s) only.

### DIFF
--- a/mysql-test/suite/galera/r/galera_sp_toi_error_vote_warn_mismatch.result
+++ b/mysql-test/suite/galera/r/galera_sp_toi_error_vote_warn_mismatch.result
@@ -1,0 +1,36 @@
+##############################################################################
+# 1. Setup
+##############################################################################
+CREATE DATABASE IF NOT EXISTS my_database;
+USE my_database;
+CREATE PROCEDURE testCrashGaleraNode()
+BEGIN
+ALTER TABLE non_existing_table ADD somecolumn tinyint(1);
+END$$
+Warnings:
+Warning	1681	Integer display width is deprecated and will be removed in a future release.
+USE my_database;
+##############################################################################
+# 2. 1146 and 1681 error, all nodes send only 1146. Cluster stays healthy.
+##############################################################################
+ALTER TABLE non_existing_table ADD somecolumn tinyint(1);
+ERROR 42S02: Table 'my_database.non_existing_table' doesn't exist
+CALL testCrashGaleraNode();
+ERROR 42S02: Table 'my_database.non_existing_table' doesn't exist
+CALL testCrashGaleraNode();
+ERROR 42S02: Table 'my_database.non_existing_table' doesn't exist
+include/assert.inc [Assert node-1 is part of cluster.]
+include/assert.inc [Cluster size remains 2 on node1.]
+include/assert.inc [Assert node-2 is part of cluster.]
+include/assert.inc [Cluster size remains 2 on node2.]
+##############################################################################
+# 3. Assert error log. Only 1146 appears in error log. 1681 does not exist.
+##############################################################################
+include/assert.inc [Node1 vote recompute must not include warning 1681.]
+include/assert.inc [Node2 vote recompute must not include warning 1681.]
+##############################################################################
+# 4. Cleanup
+##############################################################################
+CALL mtr.add_suppression('Replica SQL: Error .*');
+CALL mtr.add_suppression('Replica SQL: Error .*');
+DROP DATABASE my_database;
--- a/mysql-test/suite/galera/t/galera_sp_toi_error_vote_warn_mismatch.test
+++ b/mysql-test/suite/galera/t/galera_sp_toi_error_vote_warn_mismatch.test
@@ -1,0 +1,108 @@
+##############################################################################
+# This test case makes sure when additional warning 1681 is logged in one of
+# the node in cluster, cluster remains healthy.
+#
+# NOTE: 1146+1681 is generated on the client side.
+# 1146 is generated at the applier side.
+# 1681(ER_WARN_DEPRECATED_SYNTAX_NO_REPLACEMENT) is deprecated with no replacement.
+#
+##############################################################################
+# 1. Setup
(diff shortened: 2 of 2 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3356/consoleFull

sql/wsrep_applier.cc
```diff
@@ -117,17 +129,35 @@ void wsrep_store_error(const THD *const thd, wsrep::mutable_buffer &dst) {
...
+    serialize the DA error code/message.
+
+    Note: This function ignores the warnings and notes, limiting the
+    voting payload to error-only conditions.
+    slider < buf_end ensures fixed-size buffer is not over flowed.
+    This may not happen since upper for loop must have have_error_condition
+    as false to enter the below if block. But it was part of for loop so care
+    has been inside if also.
+  */
+  if (!have_error_condition && da->is_error() && slider < buf_end) {
```
kamil-holubicki: We shouldn't change this, as it is not backward compatible. It changes the protocol.
Before change: vote is calculated basing on top level error + conditions errors
After change: vote is calculated on conditions errors OR top level error.

Now let's imagine a mixed version cluster.
Even if all nodes report the same top level error and the same conditions, inconsistency voting will detect a problem.

In my opinion we should not change this.

sql/wsrep_applier.cc
```diff
@@ -117,17 +129,35 @@ void wsrep_store_error(const THD *const thd, wsrep::mutable_buffer &dst) {
   const char *const buf_end = slider + max_len - 1;  // -1: leave space for \0
 
   auto da = thd->get_stmt_da();
-  if (da->cond_count() == 0 && da->is_set()) {
-    uint const err_code = wsrep_errno_for_voting(da->mysql_errno());
-    const char *const err_str = da->message_text();
+  bool have_error_condition = false;
+  for (cond = it++; cond && slider < buf_end; cond = it++) {
+    if (cond->severity() != Sql_condition::SL_ERROR) continue;
```
kamil-holubicki: In general it is a good idea. If you asked me at the beginning of Galera development if it should be like this, my answer would be "yes". But now... historical reason.
Let's imagine a mixed version cluster. Even if all nodes have the same set of errors+warings, they will vote in a different way. It will lead to inconsistency detection.

This solution should be the last-resort solution, but before going this way, I would investigate what we discussed on Slack:
1. Check commit 502079f8 (and related Galera c594e43b). We did a backward compatible change there - this is for reference
2. Looking at [THIS](https://github.com/percona/galera/blob/a430f07ec0346d6a3db73c4de342941d175c5304/gcs/src/gcs_group.cpp#L1178C25-L1178C32) part it seems that receiver part uses only string before null. Everything afterwards is ignored. So there is a chance to piggyback after null.
3. old nodes (before commit 502079f8) use vote code sent from the originator (they do not do any recalculation, message string is used only for debug)

Putting things together:
Sender side:
1. Leave it as it was (top level error + conditions). Together with warnings
2. collect all warning codes
3. when the string is created, append null, then something like `protocol_version:2, Warnings:<code1>,<code2>`. So we will have a string like `original_voting_message<nullptr>protocol_version:2, Warnings:<code1>,<code2>`
4. Calculate vote to send using `original_voting_message` (old nodes still need this)

Receiver side node prior to 502079f8
1. It will receive voting hash based on original_voting_message and use it. It will ignore string message

Receiver side node after 502079f8
1. It will receive voting hash based on original_voting_message, but will recompute it basing on error cdes from original_voting_message part. Everything after null will be ignored

Receiver side node after this fix:
1. It will receive voting hash based on original_voting_message. Then it will detect that string contains list of warings to ignore after <null>. It will filter out warnings from original_voting_message and calculate a vote. If there is nothing after <null>, it will calculate vote basing on all codes (including warnings) (the same way as without fix)

Of course it doesn't solve all combinations. Old versions are already out, and we cannot go back in time and fix them. If there are two nodes of verson 8.0.44 and they vote in a different way (PXC-4683), there is nothing we can do about it.

In another words:
1. New node votes: 
- It sends a vote that is compatible with old nodes
- It sends a vote that allows warning filtering 
2. Old node votes:
- it sends a vote compatible with old nodes
3. New node receives vote from old node
- It calculates vote basing on everything (errors + warnings)
4. New node receives vote from new node
- it calculates vote basing only on errors (after filtering out warings)
5. Old node receives vote from old node
- nothing changed :)
6. Old node receives vote from new node
- It uses sent vote code (prior to 502079f8, or recalculates vote basing on error and warning codes)

kamil-holubicki (changes requested):

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3384/consoleFull

I have changed test case to check error logs in table and some comments.
Jenkins faced issue in middle of run, so I will submit changes to jenkins again.

sql/wsrep_applier.cc
jaideepkarande: Thanks, I have made minor changes as per commit https://github.com/percona/percona-xtradb-cluster/commit/502079f8159ff87e207a46c176b6fb0496b0eed9 so resolving the issue.

kamil-holubicki (approved):
LGTM. Before merging please update galera repo pointer by force push

PR #2226, opened 2026-02-16
PXC-4799: PXC member used for backups fails with Inconsistency when p…

jaideepkarande:
…ending DDL clashes with FTRWL

Problem:

Description:
FTWRL caused MDL lock and waited for DDL to finish in the PXC. FTWRL specifically waits in desync_and_pause->pause statement waiting for concurrent pause statement to finish.

Resolution:
DDLs cannot be executed in parallel with FTWRL in local node it will fail. So on local node ER_CANT_UPDATE_WITH_READLOCK is seen post FTWRL FLUSH TABLES WITH READ LOCK;
However in PXC remote node not aware of FTWRL on another node accepts the DDL and sends across the cluster opening a pandora of unknown conditions. So, we prevent FLUSH TABLES WITH READ LOCK when the session already holds explicit table locks.
We also check that pause will block; with ongoing TO isolation, MDL locks are likely to conflict with FTWRL, so it is best to fail FTWRL. By rejecting FLUSH TABLES WITH READ LOCK in this situation, we avoid the deadlock-like condition and allow the remote DDL to proceed to avoid situations like below:
NODE-2
LOCK INSTANCE FOR BACKUP;
NODE-1
CREATE TABLE t_synch1 (id INT PRIMARY KEY, a INT); NODE-2
FLUSH TABLES WITH READ LOCK;
NODE-2
UNLOCK TABLE;
UNLOCK INSTANCE;
NOTE: Even with this safeguard, if the DDL waiting for MDL exceeds lock_wait_timeout, the DDL will fail while waiting for the maitenance operations to finish. FLUSH or DDL will not block so the session can execute UNLOCK for the DDL to complete.
This behavior is specific to PXC. In standalone PS/PXC servers, such conflicting operations fail naturally, but in multi-node PXC deployments, DDL may arrive from another node, making this protection necessary.


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3362/consoleFull

However there are some failures in Clang etc, in draft more Azure pipeline runs stopped.
So have to put PR in review again.
Likely clang changes were outside bug caused changed, since clang was ran locally before PR creation.

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3364/console

OLD RUN:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3363/

kamil-holubicki (commented):
We don't need lock order change (the patch from Dmitry?)
I also see that you skipped changes from wsrep_mysqld.cc.
How is [this scenario](https://github.com/percona/percona-xtradb-cluster/compare/8.0...kamil-holubicki:percona-xtradb-cluster:PXC-4799-poc#diff-adadb13c7a06c1a6d6be490308b4cb130848c6e0b2ccca48dde12efeda10a498R3025) handled then?

jaideepkarande:
Restored the changes in wsrep_mysqld.cc and extended the test case accordingly.
Regarding changes in mdl.*. The acquisition order we need is already in place and we will not benefit from the changes.

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3379/consoleFull

sql/lock.cc
```diff
@@ -1224,22 +1224,67 @@ bool Global_read_lock::make_global_read_lock_block_commit(THD *thd) {
...
+        (see wsrep_to_isolation_begin()) so that DDL blocks on backup lock
+        instead of holding global read lock and causing this situation when
+        possible.
+
+        Limitation: we cannot distinguish DDL (blocked on backup lock) from
+        other writesets waiting to be processed. So LOCK INSTANCE FOR BACKUP
+        followed by FTWRL may fail whenever any writesets are pending. A
+        try_desync_and_pause(timeout) could allow a bounded wait in the future.
+      */
+      paused_seqno = server_state.try_desync_and_pause();
```
kamil-holubicki: Once we have it working, would it be possible to do try_desync_and_pause(timeout)?
We discussed it on Slack. It seems to be not so much work to do.
Just a configuration parameter, default let's say 30sec, Inside Galera I imagine it as a loop repeated every 100ms (similar to what we have in try_toi or something).
Shouldn't be hard to implement, but in my opinion will solve many waiting issues.

sql/wsrep_high_priority_service.cc
```diff
@@ -441,6 +442,16 @@ int Wsrep_high_priority_service::apply_toi(const wsrep::ws_meta &ws_meta,
...
   assert(client_state.in_toi());
 
+  /*
+    Acquire BACKUP_LOCK before any GRL is taken so that TOI applier follows
+    the same lock order as the initiator (wsrep_to_isolation_begin), avoiding
+    MDL deadlock with LOCK INSTANCE FOR BACKUP + FLUSH TABLES WITH READ LOCK.
+    Lock is transaction duration and released at commit/rollback.
+  */
+  if (acquire_shared_backup_lock(thd, thd->variables.lock_wait_timeout, true)) {
+    return 1;
```
kamil-holubicki, line 453: Please add the error message explaining what happened.

sql/lock.cc
```diff
@@ -1224,22 +1224,67 @@ bool Global_read_lock::make_global_read_lock_block_commit(THD *thd) {
 
   /* If node has left the cluster no point in pausing and resuming it. */
   if (server_state.state() != Wsrep_server_state::s_disconnected) {
-    wsrep::seqno paused_seqno;
+    wsrep::seqno paused_seqno = wsrep::seqno::undefined();
```
kamil-holubicki: Suggestion: We don't need to initialize during declaration. All paths below assign a new value.

sql/lock.cc
jaideepkarande: Thanks, added timeout and test case.

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3380/
kept jenkins on run now

jaideepkarande:
Addressed failures related to variable, new run at:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3382/console

kamil-holubicki (changes requested):
I played with this PR and unfortunately it doesn't work :(

For the scenario from the bug report, FTWRL just hangs.
If we use the version of try_desync_and_pause() without timeout, it doesn't work as well. The 1st call works, but if I call FTWRL for the 2nd time, it hangs up.
The root cause of 1 and 2 is the same: Galera is desynced, and resynced which advances the local sequence number and for the 2nd time we are stuck in local monitor. Please run the scenario from the bug report under the debugger to see details.

I've fixed it. Probably it needs some posishing and cleanup, but works fine.
[Here is the fix proposal](https://github.com/percona/percona-xtradb-cluster/pull/2238/changes/df4aef0a51f745cbd6a046b10d589a4b78db5db4)
The server part remains unchanged

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3396/consoleFull

mysql-test/suite/galera/t/pxc_ftwrl_try_pause_timeout_queue.test
```diff
@@ -0,0 +1,169 @@
...
+--echo # 10. Cleanup.
+--echo #
+
+--disconnect node_2a
+
+--connection node_2
+SET @@GLOBAL.wsrep_desync_pause_retry_timeout = @wsrep_desync_pause_retry_timeout_save;
+
+--connection node_1
+DROP DATABASE db1;
```
kamil-holubicki, line 229: Suggestion: We could add one more test scenario:
1. set the timeout for 5 sec
2. do some DML, block processing
3. FTWRL
4. Wait for 10 sec
5. unblock DML
6. FTWRL should fail (timeout)

kamil-holubicki (approved):
LGTM. Please consider extending the testcase.
Before merging, please merge all submodules, then force push with proper submodule pointers.

kamil-holubicki (approved):
LGTM. Please consider extending the testcase.
Before merging, please merge all submodules, then force push with proper submodule pointers.

jaideepkarande:
[  9%] galera.pxc_ftwrl_try_pause_timeout 'enc_off' w2  [ pass ]     68
[ 18%] galera.pxc_ftwrl_try_pause_timeout_queue 'enc_off' w1  [ pass ]  10652
[ 27%] galera.galera_defaults 'enc_off'         w2  [ pass ]  10213
[ 36%] galera.pxc_ftwrl_try_pause_timeout_queue 'enc_on' w1  [ pass ]  10542
[ 45%] galera.pxc_ftwrl_try_pause_timeout 'enc_on' w1  [ pass ]     45
[ 54%] galera.pxc_ftwrl_backup_lock_deadlock 'enc_off' w2  [ pass ]  16470
[ 63%] galera.galera_defaults 'enc_on'          w1  [ pass ]  10327
[ 72%] galera.pxc_ftwrl_backup_lock_deadlock 'enc_on' w2  [ pass ]  16461
[ 81%] main.all_persisted_variables             w1  [ pass ]  41233
[ 90%] main.mysqld--help-notwin                 w2  [ pass ]    118

PR #2235, opened 2026-02-24
84->trunk nullmerge 24022026

kamil-holubicki:
This is the first null merge between branches 8.4 and trunk.
It contains changes that adjust both versions.
1. Formating things
2. Minor differences


inikep (approved):
LGTM

PR #2239, opened 2026-03-12
PXC-4837: Code refresh for - PXC 9.6.0

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4837

Code refresh for - PXC 9.6.0


storage/innobase/handler/ha_innodb.cc
```diff
@@ -12981,6 +12981,10 @@ extern dberr_t wsrep_append_foreign_key(
   int cache_key_len;
   ut_a(trx);
 
+  if (thd_is_sql_fk_checks_enabled()) {
+    return DB_SUCCESS;
+  }
```
jaideepkarande, line 12986: Please add a comment:
Post "WL#11249 - Support Foreign Key Cascading Operation in server" the server does FK, keys are now appended in the SQL layer so there is no need for InnoDB to append them again. By returning here we prevents double key appends and keeps certification consistent.

kamil-holubicki: Such a comment would not be correct. Adding a key to the writeset is an idempotent action. You can add the same key multiple times; it doesn't matter.
Here, just return if we handle FK stuff in the server layer.

jaideepkarande (approved):
LGTM

PR #2243, opened 2026-03-13
(8.0) PXC-5099: PXC node crash triggered by SELECT ... FOR UPDATE SKIP LOCKED

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-5099

Problem:
The nodes are crashing with error Unknown error code 21: Skip locked records, followed by an assertion failure in row0mysql.cc:1221.

Cause:
Native InnoDB expects sel_set_rec_lock() request to to be granted, or actual error.
However, wsrep patch in rec_lock_check_conflict() causes any type of lock request to potentially wait for high priority appliers, causing DB_SKIP_LOCKED return value here. This triggers the assert.

Solution:
Raise a deadlock error, to avoid the debug level assertion below, or fatal error in release builds.

The fix inherited from Codership's source bundle
mysql-wsrep-8.4.3-26.21.tar.gz


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3386/

storage/innobase/row/row0sel.cc
jaideepkarande, line 5137: Do we need similar fix here?

kamil-holubicki: I think you might be right that we need a fix here as well.
What I learned (of course with the help of AI) is that this branch is extremely unlikely to execute.
1. It needs the optimizer to decide to use a backward scan iterator 
`iterator = NewIterator<RefIterator<true>>(`
2. hundreds of rows carefully arranged so that some rows matching the prefix are on a page with non-matching rows
3. an index page where a matching prefix starts in the middle of the page, with non-matching records before or after it in the same page.

But looking at these two branches:
1. match_mode == ROW_SEL_EXACT  (with fix)
2. match_mode == ROW_SEL_EXACT_PREFIX (without fix)
they are very similar, so I would say 'Yes'.

I think we need InnoDB expert's opinion. @satya-bodapati ?

jaideepkarande:
1. I had a look into the stack and observed there was rollback to savepoints for other errors.
However when this error is triggered all savepoints are destroyed post changes because of transaction rollback.
The deadlock statement is a read only statement, but I am not sure what should be the correct behavior i.e. rolling back to savepoint or rolling back entire transaction.

2. What surprised me is assert was hit, was customer using debug builds? What is the behavior in release build?

kamil-holubicki:
> 1. I had a look into the stack and observed there was rollback to savepoints for other errors.
>    However when this error is triggered all savepoints are destroyed post changes because of transaction rollback.
>    The deadlock statement is a read only statement, but I am not sure what should be the correct behavior i.e. rolling back to savepoint or rolling back entire transaction.

For PXC rollback the whole transaction.

> 2. What surprised me is assert was hit, was customer using debug builds? What is the behavior in release build?
As per bug report it is a server intentional crash after `ib::fatal`

storage/innobase/row/row0sel.cc
satya-bodapati: @kamil-holubicki can you help me understand how sel_set_rec_lock()  (because of wsrep change in rec_lock_check_conflict()) returns DB_SKIP_LOCKED in PXC? what is this change, how is it propagated to here?

The condition here is simple, whenever we set a lock on record, we cannot get DB_SKIP_LOCKED or DB_NOWAIT. THis is because they contradict the lock request. 

ie you cannot ask for a lock and skip lock at the same time (original intention of the check/assert)

satya-bodapati: In short, my recommendation is do it for all the places.

Essentially we are saying, incase of PXC, high prio trx is considered to be conflicting (doesn't apply for regular record lock flow for SKIP LOCKED) and the end result for such transaction is to abort..

satya-bodapati: Why do we ignore DB_LOCK_NOWAIT? 🤔

kamil-holubicki: Done.
As for DB_LOCK_NOWAIT, it appears to be handled below. Let's not open Pandora's box and make the world better in small steps ;)

jaideepkarande (approved):
LGTM

PR #2245, opened 2026-03-13
(9.x) PXC-5099: PXC node crash triggered by SELECT ... FOR UPDATE SKIP LOCKED

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-5099

Problem:
The nodes are crashing with error Unknown error code 21: Skip locked records, followed by an assertion failure in row0mysql.cc:1221.

Cause:
Native InnoDB expects sel_set_rec_lock() request to to be granted, or actual error.
However, wsrep patch in rec_lock_check_conflict() causes any type of lock request to potentially wait for high priority appliers, causing DB_SKIP_LOCKED return value here. This triggers the assert.

Solution:
Raise a deadlock error, to avoid the debug level assertion below, or fatal error in release builds.

The fix inherited from Codership's source bundle
mysql-wsrep-8.4.3-26.21.tar.gz


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/742/

jaideepkarande (approved):
LGTM

PR #2248, opened 2026-03-18
(9.6) PXC-4965: Passwords containing ' wrongly handled

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4965

Problem:
When SET PASSWORD="a'b" is executed on the source node, the replicated node triggers inconsistency voting as the query fails on it.

Cause:
On the source node SET PASSWORD statement is rewritten to SET PASSWORD FOR. It is because on the replica node, the applier thread works in the root user context and needs to set the password for the desired user, not root. During rewriting, escaping was missing.

Solution:
1. Source side: fix escaping during the rewrite
2. Replica side: detect if the statement originates from affected node and apply proper escaping before execution

### DIFF
--- a/mysql-test/suite/galera/r/pxc_set_password.result
+++ b/mysql-test/suite/galera/r/pxc_set_password.result
@@ -1,0 +1,533 @@
+#
+# Testing for: user: testuser, password: testpassword
+#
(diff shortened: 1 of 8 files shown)


jaideepkarande (approved):
LGTM

PR #2249, opened 2026-03-19
PXC-4849: Node fails to join after successful SST due to event schedu…

jaideepkarande:
…ler and read_only

https://perconadev.atlassian.net/browse/PXC-4849

Problem: When a PXC node has read_only or super_read_only set and joins (or restarts) with events replicated from another node, the event scheduler can abort during startup.

Description:
The temporary THD used for event load has no user and no SUPER, so check_readonly() can block the commit when read_only is set. Which causes the event scheduler to abort during startup.

Fix: Skip_readonly_check has been added for the temporary THD so DD commits or any internal transaction during event load can succeed.

### DIFF
--- a/mysql-test/suite/galera/t/pxc_4849_read_only_events.test
+++ b/mysql-test/suite/galera/t/pxc_4849_read_only_events.test
@@ -1,0 +1,102 @@
+# This test case proves that when super_read_only is set on a PXC member
+# and there are events on another PRIMARY Node load_events_from_db() updates
+# them to JOINER.
+#
+# 1. Normal restart:
+#      create event on node1
+#      event replicates to node2
+#      node2 restarts with --super-read-only.
+# 2. SST:
+#      shutdown node2
+#      remove grastate(for SST)
+#      create event on node1
+#      start node2 with --super-read-only (node2 does full SST)
+#
+
+--source include/force_restart.inc
+--source include/galera_cluster.inc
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3401/consoleFull

mysql-test/suite/galera/t/pxc_4849_read_only_events.test
```diff
@@ -0,0 +1,100 @@
...
+#      remove grastate(for SST)
+#      create event on node1
+#      start node2 with --super-read-only (node2 does full SST)
+#
+
+--source include/force_restart.inc
+--source include/galera_cluster.inc
+
+--connection node_1
+CREATE DATABASE IF NOT EXISTS test;
```
kamil-holubicki: Why are we touching `test` database? It should be there created by MTR framework. 
The test should not delete this database.

jaideepkarande: Thanks, resolved it now.

jaideepkarande: Thanks, resolved it now.

kamil-holubicki (approved):

PR #2251, opened 2026-03-22
PXC-4799: PXC member used for backups fails with Inconsistency when pending DDL clashes with FTRWL(trunk)

jaideepkarande:
PXC-4799: PXC member used for backups fails with Inconsistency when pending DDL clashes with FTRWL

Problem:

Description:
FTWRL caused MDL lock and waited for DDL to finish in the PXC.
FTWRL specifically waits in desync_and_pause->pause statement waiting for
concurrent pause statement to finish.

Resolution:
DDLs cannot be executed in parallel with FTWRL in local node it will fail.
So on local node ER_CANT_UPDATE_WITH_READLOCK is seen post FTWRL
FLUSH TABLES WITH READ LOCK;
However in PXC remote node not aware of FTWRL on another node accepts the DDL    
and sends across the cluster opening a pandora of unknown conditions.
So, we prevent FLUSH TABLES WITH READ LOCK when the session already holds
explicit table locks.
We also check that pause will block; with ongoing TO isolation, MDL locks are
likely to conflict with FTWRL, so it is best to fail FTWRL.
By rejecting FLUSH TABLES WITH READ LOCK in this situation, we avoid the
deadlock-like condition and allow the remote DDL to proceed to avoid situations
like below:
NODE-2
LOCK INSTANCE FOR BACKUP;
NODE-1
CREATE TABLE t_synch1 (id INT PRIMARY KEY, a INT);
NODE-2
FLUSH TABLES WITH READ LOCK;
NODE-2
UNLOCK TABLE;
UNLOCK INSTANCE;
NOTE: Even with this safeguard, if the DDL waiting for MDL exceeds
lock_wait_timeout, the DDL will fail while waiting for the maitenance
operations to finish. FLUSH or DDL will not block so the session can execute
UNLOCK for the DDL to complete.
This behavior is specific to PXC. In standalone PS/PXC servers, such
conflicting operations fail naturally, but in multi-node PXC deployments,
DDL may arrive from another node, making this protection necessary.


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/748/consoleFull

sql/sys_vars.cc
```diff
@@ -9035,4 +9045,3 @@ static Sys_var_enum_default_table_encryption Sys_default_table_encryption(
     HINT_UPDATEABLE SESSION_VAR(default_table_encryption), CMD_LINE(OPT_ARG),
     default_table_encryption_type_names, DEFAULT(DEFAULT_TABLE_ENC_OFF),
     NO_MUTEX_GUARD, IN_BINLOG, ON_CHECK(check_set_default_table_encryption));
-
```
kamil-holubicki, line 9038: unnecessary change

kamil-holubicki (approved):

jaideepkarande:
MySQL Version 9.5.0
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_ftrw_test/BIN_T/mysql-test/var'
Installing system database
Using parallel: 2
ports_per_thread:70

[  9%] galera.pxc_ftwrl_try_pause_timeout 'enc_off' w1  [ pass ]     44
[ 18%] galera.pxc_ftwrl_try_pause_timeout_queue 'enc_off' w2  [ pass ]  10611
[ 27%] galera.galera_defaults 'enc_off'         w1  [ pass ]   9526
[ 36%] galera.pxc_ftwrl_try_pause_timeout_queue 'enc_on' w2  [ pass ]  10571
[ 45%] galera.pxc_ftwrl_try_pause_timeout 'enc_on' w2  [ pass ]     39
[ 54%] galera.pxc_ftwrl_backup_lock_deadlock 'enc_off' w1  [ pass ]  16104
[ 63%] galera.galera_defaults 'enc_on'          w2  [ pass ]  10056
[ 72%] galera.pxc_ftwrl_backup_lock_deadlock 'enc_on' w1  [ pass ]  15962
[ 81%] main.all_persisted_variables             w2  [ pass ]  34904
[ 90%] main.mysqld--help-notwin                 w1  [ pass ]     90
The servers were restarted 5 times
Spent 107.907 of 181 seconds executing testcases

Completed: All 11 tests were successful.

PR #2254, opened 2026-03-24
PXC-4887: "CREATE ROLE" not replicated in pxc when using the validate…

jaideepkarande:
… password plugin

https://perconadev.atlassian.net/browse/PXC-4887

Problem: CREATE ROLE not replicated in PXC with validate_password plugin

Description:
Role is getting lost in function call:
mysql_create_user -> wsrep_check_for_auth_policy -> set_and_validate_user_attributes
mysql_create_user has is_role variable.
set_and_validate_user_attributes also has is_role parameter but wsrep_check_for_auth_policy does not have is_role parameter so false is sent to set_and_validate_user_attributes. Which means role is getting lost.

For a role there is no password (empty string). The empty “password” was run through validate_password and it failed the password policy check blocking the TOI. The role was created only on the originating node.

Resolution: Added a is_role parameter in wsrep_check_for_auth_policy(), so that role is not lost between function call and passed the role to set_and_validate_user_attributes.

### DIFF
--- a/sql/auth/sql_user.cc
+++ b/sql/auth/sql_user.cc
@@ -2935,7 +2935,7 @@ static LEX_USER *deep_copy(THD *thd, LEX_USER *src) {
     Inconsistency voting: triggered for both sides with ER_PLUGIN_IS_NOT_LOADED
  */
 static bool wsrep_check_for_auth_policy(THD *thd, List<LEX_USER> &list,
-                                        const char *cmd) {
+                                        const char *cmd, bool is_role) {
   if (WSREP(thd) && !thd->wsrep_applier && list.size() > 0) {
     // Accessing ACL cache requires lock. We get WRITE_MODE lock since the call
     // to is_privileged_user_for_credential_change() below, might have to update
@@ -2966,6 +2966,17 @@ static bool wsrep_check_for_auth_policy(THD *thd, List<LEX_USER> &list,
         return false;
       }
 
+      /*
+        get_current_user fetches the current user information.
+        Check if the user is a role or not.
+      */
+      bool is_role_validated = is_role;
+      if (!is_role) {
+        if (ACL_USER *au =
+                find_acl_user(user_from->host.str, user_from->user.str, true))
+          is_role_validated = au->is_role;
+      }
+
       /* set_and_validate_user_attributes() will hash provided password and
        wipe-out its plaintext version in
        user_from_tmp.first_factor_auth_info.auth.str.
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3406/console

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3407/consoleFull

sql/auth/sql_user.cc
```diff
@@ -3025,7 +3042,36 @@ bool mysql_create_user(THD *thd, List<LEX_USER> &list, bool if_not_exists,
...
+  bool block_toi = wsrep_check_for_auth_policy(
+      thd, list,
+      (thd->lex->sql_command == SQLCOM_CREATE_ROLE) ? "CREATE ROLE"
+                                                    : "CREATE USER",
+      is_role);
+  /*
+    If TO Isolation is blocked and statement runs successfully,
+    it will cause inconsistency in cluster.
+  */
+  if (block_toi) {
```
kamil-holubicki: As we discussed on Slack, please remove it, because original flow below should fail as well.

kamil-holubicki (changes requested):

sql/auth/sql_user.cc
jaideepkarande: Thanks, reverted the changes.

jaideepkarande:
Logging: /home/jkarande/sandbox/pxc/pxc_role/mysql-test/mysql-test-run.pl  --parallel=4 pxc_alter_user pxc_create_role_validate_password pxc_create_user_auth_policy
MySQL Version 8.0.45
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_role/BIN/mysql-test/var'
Installing system database
Using parallel: 4

[ 14%] galera.pxc_create_role_validate_password 'enc_on' w4  [ pass ]    281
[ 28%] galera.pxc_create_role_validate_password 'enc_off' w2  [ pass ]    281
[ 42%] galera.pxc_alter_user 'enc_off'          w3  [ pass ]    668
[ 57%] galera.pxc_alter_user 'enc_on'           w1  [ pass ]    720
[ 71%] galera.pxc_create_user_auth_policy 'enc_on' w2  [ pass ]    203
[ 85%] galera.pxc_create_user_auth_policy 'enc_off' w4  [ pass ]    185
The servers were restarted 2 times
Spent 2.338 of 87 seconds executing testcases

Completed: All 7 tests were successful.

kamil-holubicki (approved):

PR #2256, opened 2026-03-27
PXC-4849: Node fails to join after successful SST due to event schedu…(8.4)

jaideepkarande:
…ler and read_only

https://perconadev.atlassian.net/browse/PXC-4849

Problem: When a PXC node has read_only or super_read_only set and joins (or restarts) with events replicated from another node, the event scheduler can abort during startup.

Description:
The temporary THD used for event load has no user and no SUPER, so check_readonly() can block the commit when read_only is set. Which causes the event scheduler to abort during startup.

Fix: Skip_readonly_check has been added for the temporary THD so DD commits or any internal transaction during event load can succeed.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_4849_read_only_events.result
+++ b/mysql-test/suite/galera/r/pxc_4849_read_only_events.result
@@ -1,0 +1,33 @@
+CREATE DATABASE IF NOT EXISTS db_super_rd_load_event_test;
+USE db_super_rd_load_event_test;
+#
+# 1. Normal restart:
+#      create event on node1
+#      event replicates to node2
+#      node2 restarts with --super-read-only.
+#
+CREATE EVENT pxc_restart_ev ON SCHEDULE EVERY 1 MONTH
+STARTS '2026-06-01 00:00:01' ON COMPLETION NOT PRESERVE ENABLE
+DO SELECT 1;
+# restart:--super-read-only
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/756/consoleFull

Logging: /home/jkarande/sandbox/pxc/pxc_sst/mysql-test/mysql-test-run.pl  galera_3nodes.inconsistency_shutdown
MySQL Version 8.4.7
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
 - Adding combinations for galera_3nodes
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_sst/BIN/mysql-test/var'
Installing system database
Using parallel: 1

[ 33%] galera_3nodes.inconsistency_shutdown 'enc_off'  [ pass ]  125654
[ 66%] galera_3nodes.inconsistency_shutdown 'enc_on'  [ pass ]  122737
The servers were restarted 1 times
Spent 248.391 of 372 seconds executing testcases

Completed: All 3 tests were successful.

Rest all failures are related to pre-release (8.4.7) issues, likely because of GCA branch.

kamil-holubicki (approved):

PR #2258, opened 2026-03-31
PXC-4887: "CREATE ROLE" not replicated in pxc when using the validate…(8.4)

jaideepkarande:
… password plugin

https://perconadev.atlassian.net/browse/PXC-4887

Problem: CREATE ROLE not replicated in PXC with validate_password plugin

Description:
Role is getting lost in function call:
mysql_create_user -> wsrep_check_for_auth_policy -> set_and_validate_user_attributes
mysql_create_user has is_role variable.
set_and_validate_user_attributes also has is_role parameter but wsrep_check_for_auth_policy does not have is_role parameter so false is sent to set_and_validate_user_attributes. Which means role is getting lost.

For a role there is no password (empty string). The empty “password” was run through validate_password and it failed the password policy check blocking the TOI. The role was created only on the originating node.

Resolution: Added a is_role parameter in wsrep_check_for_auth_policy(), so that role is not lost between function call and passed the role to set_and_validate_user_attributes.

### DIFF
--- a/mysql-test/suite/galera/r/pxc_create_role_validate_password.result
+++ b/mysql-test/suite/galera/r/pxc_create_role_validate_password.result
@@ -1,0 +1,33 @@
+INSTALL COMPONENT "file://component_validate_password";
+SET @@global.validate_password.policy = 0;
+SET @@global.validate_password.length = 16;
+SET @@global.validate_password.check_user_name = ON;
+# CREATE ROLE on node_1
+CREATE ROLE testrole;
+USE test;
+CREATE TABLE t1(a INT PRIMARY KEY);
+# GRANT on role and GRANT role to user
+GRANT SELECT ON test.* TO testrole;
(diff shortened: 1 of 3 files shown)


jaideepkarande:
MySQL Version 8.4.7
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_role/BIN/mysql-test/var'
Installing system database
Using parallel: 1

[ 33%] galera.pxc_create_role_validate_password 'enc_off'  [ pass ]    260
[ 66%] galera.pxc_create_role_validate_password 'enc_on'  [ pass ]    228

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/760/console

kamil-holubicki (approved):

PR #2260, opened 2026-03-31
PXC-4683: Node leaves the cluster after an error in a stored procedure(trunk)

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-4683

Problem: ER_WARN_DEPRECATED_SYNTAX_NO_REPLACEMENT(1681) warning causes inconsistency.
TOI-failed inside a stored procedure e.g. ALTER TABLE on a non-existing table with tinyint marked the node inconsistent due to additional 1681 warning.

Description:
TOI failure in a stored procedure (e.g. ALTER TABLE on non-existing table with tinyint) could mark the node inconsistent. The voting payload included warning 1681 (Integer display width is deprecated..) in addition to the error 1146. Warning 1681 was logged in client side while 1681 warning was not being logged via replication channels. The difference in error and warnings in different nodes caused inconsistency causing one of the node i.e. recipient of transaction to leave the cluster.

Cause:
The Galera wsrep_store_error consider the diagnostics area in addition to the real error code. Deprecated integer display width (tinyint(1) -> warning 1681), nodes produce different condition lists (1146 vs 1146+1681), leading to different vote and eventually causing inconsistency.

Fix:
Ignore the ER_WARN_DEPRECATED_SYNTAX_NO_REPLACEMENT and do not send it across cluster.

### DIFF
--- a/mysql-test/suite/galera/r/galera_sp_toi_error_vote_warn_mismatch.result
+++ b/mysql-test/suite/galera/r/galera_sp_toi_error_vote_warn_mismatch.result
@@ -1,0 +1,36 @@
+##############################################################################
+# 1. Setup
+##############################################################################
+CREATE DATABASE IF NOT EXISTS my_database;
+USE my_database;
+CREATE PROCEDURE testCrashGaleraNode()
+BEGIN
+ALTER TABLE non_existing_table ADD somecolumn tinyint(1);
+END$$
+Warnings:
+Warning	1681	Integer display width is deprecated and will be removed in a future release.
+USE my_database;
+##############################################################################
+# 2. 1146 and 1681 error, all nodes send only 1146. Cluster stays healthy.
+##############################################################################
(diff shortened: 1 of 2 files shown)


jaideepkarande:
Test only changes:

MySQL Version 9.5.0
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_stp2/BIN/mysql-test/var'
Installing system database
Using parallel: 1
ports_per_thread:70

[ 33%] galera.galera_sp_toi_error_vote_warn_mismatch 'enc_off'  [ pass ]   1710
[ 66%] galera.galera_sp_toi_error_vote_warn_mismatch 'enc_on'  [ pass ]    730

kamil-holubicki (approved):

PR #2262, opened 2026-04-03
PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=EN…

jaideepkarande:
…FORCING

Problem:

Description
Example:

Resolution:

2. Session vs global Changing 'pxc_strict_mode' may update '@@global.sql_require_primary_key' as above. The server does not rewrite '@@session.sql_require_primary_key' on 'existing' connections. Each session keeps its session value until the client changes it or opens a new connection (new sessions inherit the global default as usual).

3. Startup If Galera is active and 'pxc_strict_mode' is 'ENFORCING' or 'MASTER' but the server would otherwise start with 'global' 'sql_require_primary_key' 'OFF' (e.g. from options or config), startup forces 'global' 'ON' so the effective policy matches strict mode.

### DIFF
--- a/mysql-test/suite/galera/t/galera_pxc_sql_require_primary_key.test
+++ b/mysql-test/suite/galera/t/galera_pxc_sql_require_primary_key.test
@@ -1,0 +1,213 @@
+# This test proves that setting pxc_strict_mode to ENFORCING or MASTER
+# sets @@global sql_require_primary_key ON and test few related scenarios.
+# Session values are not updated by pxc_strict_mode.
+#
+# 1. Degrade ENFORCING -> DISABLED does not change sql_require_primary_key
+# 2. pxc_strict_mode=ENFORCING also sets global sql_require_primary_key ON
+# 3. Degrade ENFORCING -> PERMISSIVE does not change sql_require_primary_key
+# 4. pxc_strict_mode=MASTER also sets global sql_require_primary_key ON
+# sql_require_primary_key=OFF cannot be set to OFF when pxc_strict_mode is MASTER
+# 5. ALTER DROP PRIMARY KEY under ENFORCING
+# 6. restart node_2 with ENFORCING and sql-require-primary-key=OFF
+# 7. Cleanup.
+
+--source include/galera_cluster.inc
+# Due to restart, lets make this test big-test.
+--source include/big_test.inc
+#--source include/force_restart.inc
+
+--connection node_1
+call mtr.add_suppression("Cannot set sql_require_primary_key=OFF while pxc_strict_mode is");
+call mtr.add_suppression("Replica SQL: Error .Unable to create or change a table without a primary key");
+call mtr.add_suppression("Setting sql_require_primary_key=ON because pxc_strict_mode is ENFORCING or MASTER");
+--let $pxc_saved_1 = `SELECT @@global.pxc_strict_mode`
+
+--connection node_2
+call mtr.add_suppression("Cannot set sql_require_primary_key=OFF while pxc_strict_mode is");
+call mtr.add_suppression("Replica SQL: Error .Unable to create or change a table without a primary key");
+call mtr.add_suppression("Setting sql_require_primary_key=ON because pxc_strict_mode is ENFORCING or MASTER");
+--let $pxc_saved_2 = `SELECT @@global.pxc_strict_mode`
+
+--let $assert_text = sql_require_primary_key global is OFF.
+--let $assert_cond = [ SELECT @@global.sql_require_primary_key = 0 ]
+--source include/assert.inc
+
+--let $pxc_strict_baseline = `SELECT @@global.pxc_strict_mode`
+--let $assert_text = pxc_strict_mode is DISABLED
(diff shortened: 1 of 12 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3410/consoleFull

sql/sys_vars.cc
```diff
@@ -8437,6 +8445,25 @@ static Sys_var_charptr Sys_protocol_compression_algorithms(
 #include "wsrep_sst.h"
 #include "wsrep_var.h"
 
+static bool check_session_admin_and_sql_require_primary_key_on_check(
+    sys_var *self, THD *thd, set_var *var) {
+  if (check_session_admin(self, thd, var)) return true;
+  if (pxc_strict_mode < PXC_STRICT_MODE_ENFORCING) return false;
+  // Not sure if we need to check this, since variable can be set to ON/OFF
+  // causing inconsistency later.
+  // if (!(WSREP_ON)) return false;
```
kamil-holubicki: Let's remove this. We should base our logic on pxc_stric_mode. Doesn't matter if wsrep is on or off

mysql-test/suite/galera/t/galera_pxc_sql_require_primary_key.test
```diff
@@ -0,0 +1,213 @@
...
+--source include/galera_cluster.inc
+# Due to restart, lets make this test big-test.
+--source include/big_test.inc
+#--source include/force_restart.inc
+
+--connection node_1
+call mtr.add_suppression("Cannot set sql_require_primary_key=OFF while pxc_strict_mode is");
+call mtr.add_suppression("Replica SQL: Error .Unable to create or change a table without a primary key");
+call mtr.add_suppression("Setting sql_require_primary_key=ON because pxc_strict_mode is ENFORCING or MASTER");
+--let $pxc_saved_1= `SELECT @@global.pxc_strict_mode`
```
kamil-holubicki: inconsistent formating aroung `=` (multiple occurences)

mysql-test/suite/galera/t/galera_rpl_pxc_strict_mode_mismatch.test
```diff
@@ -0,0 +1,134 @@
+# This tests make sure transactions received from remote node of galera or from
```
kamil-holubicki, line 1: or from `async` source?

mysql-test/suite/galera/t/galera_rpl_pxc_strict_mode_mismatch.test
```diff
@@ -0,0 +1,134 @@
...
+
+--let $pxc_saved_2= `SELECT @@global.pxc_strict_mode`
+--let $pk_saved_2= `SELECT @@global.sql_require_primary_key`
+
+SET GLOBAL pxc_strict_mode = 'ENFORCING';
+
+--connect node_3, 127.0.0.1, root, , test, $NODE_MYPORT_3
+
+--connection node_3
+--let $pxc_saved_3= `SELECT @@global.pxc_strict_mode`
```
kamil-holubicki: inconsistent formating around `=`

mysql-test/suite/galera/t/galera_rpl_pxc_strict_mode_mismatch.test
```diff
@@ -0,0 +1,134 @@
+# This tests make sure transactions received from remote node of galera or from
+# source are applied successfully in both galera and replica even if
+# pxc_strict_mode is ENFORCING.
+#
+# Setup Summary
```
kamil-holubicki, line 5: This test essentialy checks that if we have PXC acting as an async replica, and PXC has `sql_require_primary_key=1`, but async source has `sql_require_primary_key=0` and we create a table without PK on async source, PXC will accept it.

As we spoke on Slack, please ensure that this behavior is the same for standard async replication (no PXC). I'm almost sure it is, but let's be on the safe side.

kamil-holubicki (commented):
This patch looks fine.
Just a few cosmetic changes and one thing to check before we can go ahead with the merge.

jaideepkarande:
MySQL Version 8.0.44
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_key/BIN/mysql-test/var'
Installing system database
Using parallel: 1

[ 20%] galera.galera_pxc_sql_require_primary_key 'enc_off'  [ pass ]  20338
[ 40%] galera.galera_pxc_sql_require_primary_key 'enc_on'  [ pass ]  21395
[ 60%] galera.galera_rpl_pxc_strict_mode_mismatch 'enc_off'  [ pass ]    785
[ 80%] galera.galera_rpl_pxc_strict_mode_mismatch 'enc_on'  [ pass ]    842

kamil-holubicki (approved):

PR #2264, opened 2026-04-10
(trunk) PXC-5195: Remove obsolete configuration variables 

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-5195

Removed obsolete variables:
wsrep_slave_FK_checks
wsrep_slave_UK_checks
wsrep_slave_threads
wsrep_restart_slave
wsrep_causal_reads


jaideepkarande:
Thanks Kamil for working on the changes, changes looks fine. Just some minor comments,

1. Some minor issues in Clang-format

2. mysql-test/suite/galera/r/galera_defaults.result still shows WSREP_CAUSAL_READS

Below are just observation, and I do not see impact(ignore if not applicable),
3. Support files, i.e. support-files/wsrep.cnf shows the variables wsrep_causal_reads and wsrep_slave_threads.
I am not sure if those files are still in use or not, if in use, support will have to remove the deprecated variables from file before using the file.
support-files/wsrep.cnf.sh maybe this file should not even exist because its copy of support-files/wsrep.cnf 

4. wsrep_slave_threads is present in several builds related cnf file, maybe it is not a issue since many of them will not be in use

NOTE: I have ignored below files
5. Docs/README-wsrep because it has not been updated for years
6. percona-xtradb-cluster-galera/* because it is a different repo
7. I have assumed that doc/* will be updated by the documentation team

kamil-holubicki:
>mysql-test/suite/galera/r/galera_defaults.result still shows WSREP_CAUSAL_READS

Yup, there is (or rather was) a little mess with these variables. There was a configuration variable wsrep_causal_reads = ON/OFF (bool) and a status variable wsrep_causal_reads (int). The same name, two different variables :).
The 1st one was obsolete and the same as wsrep_sync_wait = 1
The 2nd one is a counter inside the galera that reports how many times the client queries waited inside galera because of wsrep_sync_wait=0xF (any wait possible).

> Support files....

Good finding. I've removed that unused files

jaideepkarande (approved):
LGTM

PR #2266, opened 2026-04-24
PXC-4963: Discrepancy between GRANTs for in-place upgraded PXC binari…(8.0)

jaideepkarande:
…es vs fresh installed PXC binaries

https://perconadev.atlassian.net/browse/PXC-4963

Problem:
After an in-place upgrade from an older PXC, the ACL rows for the internal role 'mysql.pxc.sst.role'@'localhost' retain the narrower privilege set defined by the older release, while a fresh 8.0.x install creates the role with the full set required by current SST code paths (CREATE USER, SHUTDOWN, EXECUTE, CREATE ROLE, DROP ROLE, GRANT OPTION, BACKUP_ADMIN WITH GRANT OPTION, and Select/Insert/Update/Grant on performance_schema.*) via mysql_system_users.sql.

mysql_system_tables_fix.sql -- the script mysql_upgrade runs on existing installs -- only contains INSERT IGNORE statements for this role. INSERT IGNORE is a no-op when the rows already exist, so upgraded installs never receive the missing privileges and end up with a role from an older PXC.

Resolution:
Added UPDATE statements to the scripts mysql_system_tables_fix.sql

### DIFF
--- a/mysql-test/r/pxc_sst_role_grants_upgrade_force.result
+++ b/mysql-test/r/pxc_sst_role_grants_upgrade_force.result
@@ -1,0 +1,104 @@
+#
+# 1. Save the fresh-install state for the role 'mysql.pxc.sst.role'.
+#    We need to diff grants of fresh install with upgrade.
+#
+CREATE TABLE test.t_pxc_sst_user_fresh        AS SELECT * FROM mysql.user          WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3411/consoleFull

kamil-holubicki (approved):

PR #2267, opened 2026-04-24
PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=EN…(8.4)

jaideepkarande:
…FORCING

Problem:

Description
Example:

Resolution:

2. Session vs global Changing 'pxc_strict_mode' may update '@@global.sql_require_primary_key' as above. The server does not rewrite '@@session.sql_require_primary_key' on 'existing' connections. Each session keeps its session value until the client changes it or opens a new connection (new sessions inherit the global default as usual).

3. Startup If Galera is active and 'pxc_strict_mode' is 'ENFORCING' or 'MASTER' but the server would otherwise start with 'global' 'sql_require_primary_key' 'OFF' (e.g. from options or config), startup forces 'global' 'ON' so the effective policy matches strict mode.


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/771/console

kamil-holubicki (approved):

PR #2269, opened 2026-04-28
(8.0) PXC-4844: Galera applier stalls when an empty write set follows a failed TOI

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-4844

Problem:
After a TOI fails on the applier (e.g. a REVOKE of a non-existent grant failing with ER_NONEXISTING_TABLE_GRANT), the very next ordered write set processed by the same applier THD can stall the applier and surface in the error log as a "ha_rollback_trans" carrying the previous TOI's error message ("There is no such grant defined for user ..."), even though the second write set is unrelated. The applier's commit-order monitor for that seqno is never released, blocking subsequent write sets and effectively hanging replication on that node.

Cause:
THD::get_stmt_da() is normally cleared at the start of each statement by mysql_reset_thd_for_next_command(), called from Rows_log_event::do_apply_event() for the first event of an RBR statement. A failed TOI does not clear the Diagnostics_area: the TOI path's cleanup is conditional on success, and trans_rollback_stmt() does not touch the DA either. The error therefore stays in the THD. When the next ordered write set arrives empty (buf_len == 0 in wsrep_apply_events()) - common with empty-result CTAS, replicated no-op DML, etc. - no row events run, so
mysql_reset_thd_for_next_command() is never invoked and the stale DA survives into Wsrep_high_priority_service::commit(). cleanup_context() -> slave_close_thread_tables() then sees thd->is_error() == true, takes the trans_rollback_stmt() path, logs the prior TOI's error against the empty write set, and Galera flips the wsrep transaction state from s_executing to s_aborted. That breaks the "Commit not finished for applier" recovery branch in Wsrep_high_priority_service::commit() (which only fires while the state is s_executing), so the commit-order monitor for the empty seqno is never released and the applier hangs.

Solution:
In wsrep_apply_events(), when buf_len == 0, call
mysql_reset_thd_for_next_command(thd) before falling through to commit, mirroring the DA reset that the first row event of a non-empty RBR statement would perform. This guarantees a clean THD for the empty write set regardless of what the previous statement on this applier left behind, so cleanup_context() takes the normal trans_commit_stmt() path, the wsrep transaction state stays at s_executing, the recovery branch advances the commit-order monitor, and the applier proceeds.

An MTR test (galera/pxc-4844) reproduces the scenario deterministically using a new wsrep_force_empty_apply DBUG knob in wsrep_apply_events() that forces the next non-TOI apply to take the empty-buffer path. The preceding failed REVOKE TOI is broadcast by wsrep_to_isolation_begin() before local execution, so it fails identically on every node with ER_NONEXISTING_TABLE_GRANT and the inconsistency voting protocol does not evict the applier node.

### DIFF
--- a/mysql-test/suite/galera/r/pxc-4844.result
+++ b/mysql-test/suite/galera/r/pxc-4844.result
@@ -1,0 +1,19 @@
+CALL mtr.add_suppression("There is no such grant defined for user");
+CALL mtr.add_suppression("There is no such grant defined for user");
+CALL mtr.add_suppression("Slave SQL: Error .ER_NONEXISTING_TABLE_GRANT");
+CALL mtr.add_suppression("Event apply failed");
+CALL mtr.add_suppression("Failed to apply TOI write set");
+SET GLOBAL wsrep_applier_threads = 1;
+CREATE TABLE t1 (id INT PRIMARY KEY) ENGINE=InnoDB;
+CREATE USER 'dataloader'@'%';
+REVOKE SELECT, INSERT ON test.t1 FROM 'dataloader'@'%';
+ERROR 42000: There is no such grant defined for user 'dataloader' on host '%' on table 't1'
+# Adding debug point 'wsrep_force_empty_apply' to @@GLOBAL.debug
+INSERT INTO t1 VALUES (1);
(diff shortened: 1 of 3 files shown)


jaideepkarande (approved):
LGTM

PR #2273, opened 2026-04-29
PXC-4963: Discrepancy between GRANTs for in-place upgraded PXC binaries vs fresh installed PXC binaries(trunk)

jaideepkarande:
PXC-4963: Discrepancy between GRANTs for in-place upgraded PXC binaries vs fresh installed PXC binaries

https://perconadev.atlassian.net/browse/PXC-4963

Problem:
After an in-place upgrade from an older PXC, the ACL rows for the
internal role 'mysql.pxc.sst.role'@'localhost' retain the narrower
privilege set defined by the older release, while a fresh 8.0.x install
creates the role with the full set required by current SST code paths
(CREATE USER, SHUTDOWN, EXECUTE, CREATE ROLE, DROP ROLE, GRANT OPTION,
BACKUP_ADMIN WITH GRANT OPTION, and Select/Insert/Update/Grant on
performance_schema.*) via mysql_system_users.sql.

mysql_system_tables_fix.sql -- the script mysql_upgrade runs on
existing installs -- only contains INSERT IGNORE statements for this
role. INSERT IGNORE is a no-op when the rows already exist, so upgraded
installs never receive the missing privileges and end up with a role
from an older PXC.

Resolution:
Added UPDATE statements to the scripts mysql_system_tables_fix.sql

### DIFF
--- a/mysql-test/r/pxc_sst_role_grants_upgrade_force.result
+++ b/mysql-test/r/pxc_sst_role_grants_upgrade_force.result
@@ -1,0 +1,104 @@
+#
+# 1. Save the fresh-install state for the role 'mysql.pxc.sst.role'.
+#    We need to diff grants of fresh install with upgrade.
+#
+CREATE TABLE test.t_pxc_sst_user_fresh        AS SELECT * FROM mysql.user          WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+CREATE TABLE test.t_pxc_sst_db_fresh          AS SELECT * FROM mysql.db            WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+CREATE TABLE test.t_pxc_sst_tables_priv_fresh AS SELECT * FROM mysql.tables_priv   WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+CREATE TABLE test.t_pxc_sst_procs_priv_fresh  AS SELECT * FROM mysql.procs_priv    WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+CREATE TABLE test.t_pxc_sst_global_fresh      AS SELECT * FROM mysql.global_grants WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+SHOW GRANTS FOR 'mysql.pxc.sst.role'@'localhost';
+Grants for mysql.pxc.sst.role@localhost
+GRANT RELOAD, SHUTDOWN, PROCESS, SUPER, LOCK TABLES, EXECUTE, REPLICATION CLIENT, CREATE USER, CREATE ROLE, DROP ROLE ON *.* TO `mysql.pxc.sst.role`@`localhost` WITH GRANT OPTION
+GRANT CLONE_ADMIN,CONNECTION_ADMIN,INNODB_REDO_LOG_ARCHIVE,SYSTEM_USER,SYSTEM_VARIABLES_ADMIN ON *.* TO `mysql.pxc.sst.role`@`localhost`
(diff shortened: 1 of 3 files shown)


mysql-test/r/pxc_sst_role_grants_upgrade_force.result
```diff
@@ -0,0 +1,106 @@
...
+2	install
+3	data
+#
+# 3. Remove the privileges from the ACL rows to simulate the old version
+#    PXC members.
+#
+UPDATE mysql.user SET Shutdown_priv = 'N', Execute_priv = 'N', Grant_priv = 'N', Create_user_priv = 'N', Create_role_priv = 'N', Drop_role_priv = 'N' WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost';
+UPDATE mysql.global_grants SET WITH_GRANT_OPTION = 'N' WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost' AND priv = 'BACKUP_ADMIN';
+UPDATE mysql.db SET Insert_priv = 'N', Update_priv = 'N', Grant_priv = 'N' WHERE user = 'mysql.pxc.sst.role' AND host = 'localhost' AND Db = 'performance_schema';
+FLUSH PRIVILEGES;
```
kamil-holubicki, line 47: As it generates warning, is obsolete, and in general not needed, maybe it is a good idea just to not do flush?

jaideepkarande: Lets keep it to avoid any random failures in the jenkins since mysql.* tables are being updated, and the docs still say statements are needed:
https://dev.mysql.com/doc/refman/9.7/en/mysqladmin.html
https://dev.mysql.com/doc/refman/9.7/en/privilege-changes.html

"If you modify the grant tables directly using statements such as [INSERT], [UPDATE], or [DELETE] (which is not recommended), the changes have no effect on privilege checking until you either tell the server to reload the tables or restart it. Thus, if you change the grant tables directly but forget to reload them, the changes have no effect until you restart the server. This may leave you wondering why your changes seem to make no difference!"

The reason of deprecation might be
https://dev.mysql.com/doc/refman/9.7/en/grant-tables.html#grant-tables-user-db
"Direct modification of grant tables using statements such as [INSERT], [UPDATE], or [DELETE] is discouraged and done at your own risk. The server is free to ignore rows that become malformed as a result of such modifications.

For any operation that modifies a grant table, the server checks whether the table has the expected structure and produces an error if not. To update the tables to the expected structure, perform the MySQL upgrade procedure. See [Chapter 3, Upgrading MySQL]."

"The server reads the contents of the grant tables into memory when it starts. You can tell it to reload the tables by issuing a [FLUSH PRIVILEGES] statement or executing a [mysqladmin flush-privileges] or  command. Changes to the grant tables take effect as indicated in [Section 8.2.13, “When Privilege Changes Take Effect”]."

kamil-holubicki: Interesting, so what is the reason for depreciation? 
Anyways, I agree, let's leave it for now as it is.

kamil-holubicki (approved):

mysql-test/r/pxc_sst_role_grants_upgrade_force.result
jaideepkarande: I wanted to comment to agree its very strange and confusing to me also. Because its odd, deprecation without alternative and documentation itself is conflicting.

jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/773/consoleFull

PR #2287, opened 2026-06-03
PXC-5229: WSREP applier threads should never have tx_read_only=true(8.0)

jaideepkarande:
WSREP applier threads may inherit global `transaction_read_only=1` after changing `wsrep_applier_threads`, leading to apply failures such as:
  - BF applier errors
  - Cluster inconsistency `transaction_read_only` is a SESSION+GLOBAL variable. NOTE:
transaction_read_only has NO_CMD_LINE means there is no command-line/option-file entry registered for it.
The only way to set the global is SET GLOBAL transaction_read_only=1 at runtime. (But persist works, which can be used as a hack.)

`SET GLOBAL wsrep_applier_threads=N` spawns brand new applier THDs (`start_wsrep_THD -> init_wsrep_thread`) *after* `transaction_read_only=1` was set, so those new appliers inherit `tx_read_only=true` and fail to apply writes.

The read-only-transaction flag is not just a SQL-layer gate. It is enforced at `open_tables`/`lock_tables` (`sql/sql_base.cc`, raising `ER_CANT_EXECUTE_IN_READ_ONLY_TRANSACTION`) and it propagates into InnoDB. Therefore a fix that only bypasses the SQL-layer check is insufficient/unsafe: InnoDB would still start a read-only `trx` and the replicated write would fail at the storage engine.

Resolution:

### DIFF
--- a/mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
+++ b/mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
@@ -1,0 +1,80 @@
+# === Purpose ===
+# Verify wsrep applier threads never run with transaction_read_only=1.
+#
+# === Reference ===
+# PXC-5229: WSREP applier threads should never have tx_read_only=true
+
+--source include/galera_cluster.inc
+
+--connection node_2
+SET @transaction_read_only_save = @@GLOBAL.transaction_read_only;
+SET @wsrep_applier_threads_save = @@GLOBAL.wsrep_applier_threads;
+
+SET GLOBAL transaction_read_only = 1;
+SET GLOBAL wsrep_applier_threads = 10;
+# Confirm 10 applier threads
+--let $wait_condition = SELECT COUNT(*) = @@wsrep_applier_threads + 1 FROM performance_schema.processlist WHERE USER = 'system user'
+--source include/wait_condition.inc
+
+# Have 2 tables a lock one of the table
+--connection node_1
+CREATE TABLE test.pxc_5229_t1(id INT PRIMARY KEY, v INT);
+CREATE TABLE test.pxc_5229_t2(id INT PRIMARY KEY, v INT);
+
+--connection node_2
+--let $wait_condition = SELECT COUNT(*) = 1 FROM information_schema.tables WHERE table_schema = 'test' AND table_name = 'pxc_5229_t1'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 1 FROM information_schema.tables WHERE table_schema = 'test' AND table_name = 'pxc_5229_t2'
+--source include/wait_condition.inc
+
+LOCK TABLE test.pxc_5229_t1 READ;
+
+--connection node_1
+# Generate workload to distribute among applier threads
+--disable_query_log
+--let $i = 1
+while ($i <= 50)
+{
+  --eval INSERT INTO test.pxc_5229_t1 VALUES ($i, $i)
+  --eval INSERT INTO test.pxc_5229_t2 VALUES ($i, $i)
+  --inc $i
+}
+UPDATE test.pxc_5229_t1 SET v = v + 5;
+UPDATE test.pxc_5229_t2 SET v = v + 5;
+--enable_query_log
+
+# Sleep to let workload distribute and UNLOCK tables
+--connection node_2
+
+--sleep 2
+UNLOCK TABLES;
+
+--let $wait_condition = SELECT COUNT(*) = 50 FROM test.pxc_5229_t2
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT COUNT(*) = 50 FROM test.pxc_5229_t1
+--source include/wait_condition.inc
+
+--let $assert_text = node_2 wsrep_cluster_status must stay Primary
+--let $assert_cond = "[SELECT VARIABLE_VALUE FROM performance_schema.global_status WHERE VARIABLE_NAME = \'wsrep_cluster_status\', VARIABLE_VALUE, 1]" = "Primary"
+--source include/assert.inc
+
+--let $assert_text = node_2 wsrep_connected must stay ON
+--let $assert_cond = "[SELECT VARIABLE_VALUE FROM performance_schema.global_status WHERE VARIABLE_NAME = \'wsrep_connected\', VARIABLE_VALUE, 1]" = "ON"
+--source include/assert.inc
+
+--let $assert_text = node_2 wsrep_ready must stay ON
+--let $assert_cond = "[SELECT VARIABLE_VALUE FROM performance_schema.global_status WHERE VARIABLE_NAME = \'wsrep_ready\', VARIABLE_VALUE, 1]" = "ON"
+--source include/assert.inc
+
+--let $assert_text = node_2 wsrep_local_state_comment must stay Synced
+--let $assert_cond = "[SELECT VARIABLE_VALUE FROM performance_schema.global_status WHERE VARIABLE_NAME = \'wsrep_local_state_comment\', VARIABLE_VALUE, 1]" = "Synced"
+--source include/assert.inc
+
+SET @@GLOBAL.transaction_read_only = @transaction_read_only_save;
+SET @@GLOBAL.wsrep_applier_threads = @wsrep_applier_threads_save;
+
+--connection node_1
+DROP TABLE test.pxc_5229_t1;
+DROP TABLE test.pxc_5229_t2;
--- a/mysql-test/suite/galera/r/pxc_5229_wsrep_applier_tx_read_only.result
+++ b/mysql-test/suite/galera/r/pxc_5229_wsrep_applier_tx_read_only.result
@@ -1,0 +1,16 @@
+SET @transaction_read_only_save = @@GLOBAL.transaction_read_only;
+SET @wsrep_applier_threads_save = @@GLOBAL.wsrep_applier_threads;
+SET GLOBAL transaction_read_only = 1;
+SET GLOBAL wsrep_applier_threads = 10;
+CREATE TABLE test.pxc_5229_t1(id INT PRIMARY KEY, v INT);
+CREATE TABLE test.pxc_5229_t2(id INT PRIMARY KEY, v INT);
+LOCK TABLE test.pxc_5229_t1 READ;
+UNLOCK TABLES;
+include/assert.inc [node_2 wsrep_cluster_status must stay Primary]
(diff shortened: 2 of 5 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3427/consoleFull

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
```diff
@@ -0,0 +1,54 @@
+# === Purpose ===
+# Verify wsrep applier threads never run with transaction_read_only=1.
+#
+# === Reference ===
+# PXC-5229: WSREP applier threads should never have tx_read_only=true
+
+--source include/galera_cluster.inc
+
+--connection node_2
+SET @wsrep_applier_threads_save= @@GLOBAL.wsrep_applier_threads;
```
kamil-holubicki: inconsistent formatting around `=`

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
```diff
@@ -0,0 +1,54 @@
...
+SET @transaction_read_only_save= @@GLOBAL.transaction_read_only;
+SET @wsrep_applier_threads_save= @@GLOBAL.wsrep_applier_threads;
+
+SET GLOBAL wsrep_applier_threads = 1;
```
kamil-holubicki: I think this sequence will be valid only if the initial wsrep_applier_threads = 1. 
If the initial count is 10, this sequence will do nothing. wsrep_appliers die only after applying a writeset, so if the initial count is 10, we need to process 9 writesets to decrease number to 1.
Please check galera_var_applier_threads.test

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
```diff
@@ -0,0 +1,54 @@
...
+SET GLOBAL transaction_read_only = 1;
+SET GLOBAL wsrep_applier_threads = 10;
+--let $wait_condition = SELECT COUNT(*) = @@wsrep_applier_threads + 1 FROM performance_schema.processlist WHERE USER = 'system user'
+--source include/wait_condition.inc
+
+--connection node_1
+CREATE TABLE test.pxc_5229_t1(id INT PRIMARY KEY, v INT);
+INSERT INTO test.pxc_5229_t1 VALUES (1, 10), (2, 20);
+UPDATE test.pxc_5229_t1 SET v = v + 5 WHERE id = 2;
+DELETE FROM test.pxc_5229_t1 WHERE id = 1;
```
kamil-holubicki: How do we know that any of them is processed by newly created applier threads? It may be the case that all of them were handled by the applier that stayed there after shrinking above.

kamil-holubicki (commented):
Fix is fine. A few concerns about the test.

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
jaideepkarande: Thanks, I have added cnf file to make sure we always start with value 1 for wsrep_applier_threads.

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
jaideepkarande: I have updated the workload to:
```
# Generate workload to distribute among applier threads
--disable_query_log
--let $i = 1
while ($i <= 50)
{
  --eval INSERT INTO test.pxc_5229_t1 VALUES ($i, $i)
  --inc $i
}
UPDATE test.pxc_5229_t1 SET v = v + 5;
DELETE FROM test.pxc_5229_t1 WHERE id != 1;
--enable_query_log
```

jaideepkarande:
Test only changes in last commit,
Logging: /home/jkarande/sandbox/pxc/pxc_applier/mysql-test/mysql-test-run.pl  pxc_5229_wsrep_applier_tx_read_only --parallel=4 --repeat=10
MySQL Version 8.0.45
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_applier/BIN/mysql-test/var'
Installing system database
Using parallel: 4

[  4%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]   1186
[  9%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w2  [ pass ]   1428
[ 14%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w1  [ pass ]    949
[ 19%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w4  [ pass ]   1248
[ 23%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]    693
[ 28%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]   1219
[ 33%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]   1228
[ 38%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]   3885
[ 42%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]   1111
[ 47%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w3  [ pass ]    882
[ 52%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w2  [ pass ]    849
[ 57%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w2  [ pass ]    771
[ 61%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w1  [ pass ]    784
[ 66%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w4  [ pass ]    769
[ 71%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w1  [ pass ]   1002
[ 76%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w2  [ pass ]    970
[ 80%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w4  [ pass ]    875
[ 85%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w2  [ pass ]   1012
[ 90%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w1  [ pass ]   1201
[ 95%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w3  [ pass ]    567
The servers were restarted 4 times
Spent 22.629 of 157 seconds executing testcases

Completed: All 21 tests were successful.

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
```diff
@@ -0,0 +1,58 @@
...
+
+SET GLOBAL transaction_read_only = 1;
+SET GLOBAL wsrep_applier_threads = 10;
+# Confirm 10 applier threads
+--let $wait_condition = SELECT COUNT(*) = @@wsrep_applier_threads + 1 FROM performance_schema.processlist WHERE USER = 'system user'
+--source include/wait_condition.inc
+
+--connection node_1
+CREATE TABLE test.pxc_5229_t1(id INT PRIMARY KEY, v INT);
+# Generate workload to distribute among applier threads
```
kamil-holubicki, line 34: I'm afraid that they still will be processed by the same applier thread. Writesets from a single client connection depend on each other (are serialized, even if they do not depend on each other). Parallelization is only for writesets that come from another user's connections.
We need two connections with commiting transactions. Maybe look at other tests how they force parallel execution on the replica side?

kamil-holubicki: Actually I have another idea. What if we set  transaction_read_only=1 in config file? This way the global var will be set even when the 1st applier is created, so we will not have to worry about "which applier applied after setting  transaction_read_only=1, the old one or new one". Maybe this way?

kamil-holubicki:
One more thing: Wsrep_client_service::replay() creates new THD as well. Then it passes it to Wsrep_replayer_service::Wsrep_replayer_service(THD *replayer_thd, THD *orig_thd). At the end of the constructor there is
```
  wsrep_assign_from_threadvars(replayer_thd);
  wsrep_reset_threadvars(orig_thd);
  wsrep_store_threadvars(replayer_thd);
```
but it seems it doesn't copy session variables (bug?)
I think the replayer thread should have the same session vars to be set as the applier thread. Replayer is kind of 'local applier'. It is fired out if the local transactions gets aborted after certification. In such a case we start applier-like thread and apply locally the writeset that was replicated. I think it should call init_wsrep_thread().

A second thought: Actually in our case if the user session was read_only, we will never get to replay path, because the transaction will be rejected. But the comment is still valid. Replayer THD needs to be able to do the same as the applier... It should at least inherit session variables from the user session. For now it looks like it works by chance.

mysql-test/suite/galera/t/pxc_5229_wsrep_applier_tx_read_only.test
jaideepkarande: Thanks, Modified the test case to improve the distribution of workload.

jaideepkarande:
jaideepkarande:PXC-5229-8.0:BIN$ MTR pxc_5229_wsrep_applier_tx_read_only --parallel=4
Logging: /home/jkarande/sandbox/pxc/pxc_applier/mysql-test/mysql-test-run.pl  pxc_5229_wsrep_applier_tx_read_only --parallel=4
MySQL Version 8.0.45
Checking supported features
 - Binaries are debug compiled
Using 'all' suites
Collecting tests
Checking leftover processes
Removing old var directory
Creating var directory '/home/jkarande/sandbox/pxc/pxc_applier/BIN/mysql-test/var'
Installing system database
Using parallel: 2

[ 33%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_off' w2  [ pass ]   3650
[ 66%] galera.pxc_5229_wsrep_applier_tx_read_only 'enc_on' w1  [ pass ]   3234
The servers were restarted 0 times
Spent 6.884 of 59 seconds executing testcases

Completed: All 3 tests were successful.

kamil-holubicki (approved):

PR #2288, opened 2026-06-03
PXC-5209: Corrupted Galera Caches Causes PXC to OOM(8.0)

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-5209

When a node restarts with gcache.recover=yes, RingBuffer::scan() walks the on-disk galera.cache and rebuilds the seqno2ptr_ map. The scan trusted the BufferHeader metadata read from disk: BH_test() only checks that the recorded size is at least sizeof(BufferHeader), and seqno_g is not validated. A truncated or otherwise corrupted cache could therefore present a buffer header whose size advertised a payload larger than the whole cache, or a seqno_g outside range.

Two failure modes follow from this:

  - A large bh->size makes ptr + bh->size look like the start of the next valid header, so the scan keeps "recovering" garbage and a bad pointer is later handed to GCache::free_common()

  - A corrupted seqno_g (e.g. 10^12) is inserted into the seqno2ptr_ deque, which then tries to allocate an enormous index range. On a memory-constrained host this may reported OOM kill or may throw std::bad_alloc

Fix (percona-xtradb-cluster-galera):

  - Before inserting into seqno2ptr_, reject a header whose bh->size exceeds the cache, whose seqno_g falls outside the [seqno_min, seqno_max] range
  - Forward the preamble's seqno_min / seqno_max bounds from recover() into scan() so the preamble-anchored check can run even for the very first buffer, when seqno2ptr_ is still empty.
  - log the corruption in such cases, clear the recovered state and force a full reset

### DIFF
--- a/mysql-test/suite/galera/r/pxc_gcache_corrupt_seqno.result
+++ b/mysql-test/suite/galera/r/pxc_gcache_corrupt_seqno.result
@@ -1,0 +1,35 @@
+
+# 1. Create table and populate 10 rows on the single node.
+
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY, f2 VARCHAR(64)) ENGINE=InnoDB;
+
+# 2. Cleanly shut the node down so the synced preamble is written to galera.cache.
+
+
+# 3. Corrupt the first buffer header's seqno_g on disk to an implausible value.
+
+
+# 4. Restart with gcache.recover=yes; scan() flags the implausible seqno.
+
+# restart
+
+# 5. GCache recovery is aborted and a full reset is forced.
+
+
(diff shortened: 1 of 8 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3430/consoleFull

kamil-holubicki:
The following commit message is obsolete:
```
Tighten GCACHE_SCAN_BUFFER_TEST so a header is accepted only when bh->size is greater than zero and does not exceed either the cache size or the bytes remaining in the current segment
```

kamil-holubicki (approved):
LGTM.
Please remember about reverting changes in .gitmodules and attaching proper galera submodule pointer after merging galera PR

PR #2297, opened 2026-06-15
PXC-5240: Port MDEV-39721 fix for wsrep_notify_cmd shell injection

kamil-holubicki:
Port of MariaDB commit ed8404a73f93 ("MDEV-39721: wsrep_notify.cc: reject shell-unsafe characters in joiner-supplied member fields"). See https://jira.mariadb.org/browse/MDEV-39721 and https://github.com/MariaDB/server/commit/ed8404a73f93e79aff7227bb4560a44cbfdb5d1c

wsp::process spawns the configured wsrep_notify_cmd via /bin/sh -c, so any unescaped shell metacharacter in a remote joiner's wsrep_node_name or wsrep_node_incoming_address reaches the donor's shell as code. A malicious joiner could thus execute arbitrary commands on every node in the cluster on view change.

On view change, validate each member's name and incoming address against a conservative whitelist (alnum + -_. for the name; additionally :[]/ for the address, to permit host:port and [ipv6]). If any member fails the check, abort the notification with an error log entry.

This degrades gracefully: existing clusters whose member names contain characters outside the whitelist remain operational; only the notify callback is skipped for affected views, and the operator is informed via the error log.

Add galera.galera_wsrep_notify_cmd_injection: node_2 starts with a malicious wsrep_node_name containing ';touch PWN;#', and the test asserts node_1 logs the rejection and the injected payload does not execute. Positive end-to-end notify coverage remains in galera.galera_var_notify_cmd.

### DIFF
--- a/mysql-test/suite/galera/t/galera_wsrep_notify_cmd_injection.test
+++ b/mysql-test/suite/galera/t/galera_wsrep_notify_cmd_injection.test
@@ -1,0 +1,29 @@
+#
+# Joiner-supplied wsrep_node_name must not inject shell commands into
+# the donor's wsrep_notify_cmd. Fails if "Notification command failed"
+# co-occurs with ";touch" in node_1's error log.
+#
+
+--source include/galera_cluster.inc
+
+SELECT 1;
+
+--connection node_1
+call mtr.add_suppression('Unsafe characters in cluster member');
+--let $datadir=`select @@datadir`
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 'Synced' FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_local_state_comment'
+--source include/wait_condition.inc
+
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+# Wait for node_1 to attempt the notify and reject the unsafe member name.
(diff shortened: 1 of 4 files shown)


kamil-holubicki:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3433/
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.0-pipeline-parallel-mtr/3434/

mysql-test/suite/galera/t/galera_wsrep_notify_cmd_injection.test
```diff
@@ -0,0 +1,29 @@
...
+--let $wait_condition = SELECT VARIABLE_VALUE = 2 FROM performance_schema.global_status WHERE VARIABLE_NAME = 'wsrep_cluster_size'
+--source include/wait_condition.inc
+
+# Wait for node_1 to attempt the notify and reject the unsafe member name.
+# The notify is invoked from a background thread after the view change, so
+# polling the error log is more robust than a fixed sleep.
+--let $grep_file = $MYSQLTEST_VARDIR/log/mysqld.1.err
+--let $grep_pattern = Unsafe characters in cluster member
+--let $wait_timeout = 60
+--source include/wait_for_pattern_in_file.inc
```
jaideepkarande, line 27: Maybe we can use performance_schema.error_log here to avoid grep in file.

jaideepkarande (approved):
Just a minor comment that may help to improve the test case. Patch LGTM.

PR #2299, opened 2026-06-16
(8.4) PXC-5240: Port MDEV-39721 fix for wsrep_notify_cmd shell injection

kamil-holubicki:
Port of MariaDB commit ed8404a73f93 ("MDEV-39721: wsrep_notify.cc: reject shell-unsafe characters in joiner-supplied member fields"). See https://jira.mariadb.org/browse/MDEV-39721 and https://github.com/MariaDB/server/commit/ed8404a73f93e79aff7227bb4560a44cbfdb5d1c

wsp::process spawns the configured wsrep_notify_cmd via /bin/sh -c, so any unescaped shell metacharacter in a remote joiner's wsrep_node_name or wsrep_node_incoming_address reaches the donor's shell as code. A malicious joiner could thus execute arbitrary commands on every node in the cluster on view change.

On view change, validate each member's name and incoming address against a conservative whitelist (alnum + -_. for the name; additionally :[]/ for the address, to permit host:port and [ipv6]). If any member fails the check, abort the notification with an error log entry.

This degrades gracefully: existing clusters whose member names contain characters outside the whitelist remain operational; only the notify callback is skipped for affected views, and the operator is informed via the error log.

Add galera.galera_wsrep_notify_cmd_injection: node_2 starts with a malicious wsrep_node_name containing ';touch PWN;#', and the test asserts node_1 logs the rejection and the injected payload does not execute. Positive end-to-end notify coverage remains in galera.galera_var_notify_cmd.

### DIFF
--- a/mysql-test/suite/galera/r/galera_wsrep_notify_cmd_injection.result
+++ b/mysql-test/suite/galera/r/galera_wsrep_notify_cmd_injection.result
@@ -1,0 +1,5 @@
+SELECT 1;
+1
+1
+call mtr.add_suppression('Unsafe characters in cluster member');
+include/wait_for_pattern_in_file.inc [Unsafe characters in cluster member]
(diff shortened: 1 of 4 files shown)


jaideepkarande (approved):
LGTM

PR #2301, opened 2026-06-16
(8.4) PXC-5244: Change xtrabackup default lock_ddl to ON during SST

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-5244

Due to PXB-3818 switch the default locking mode to lock-ddl=ON. Once lock-ddl=REDUCED Problems are solved, revert this commit.


satya-bodapati (commented):
I believe user can still force lock-ddl=reduced? we should allow override incase user wants to take some risks :)

kamil-holubicki:
> I believe user can still force lock-ddl=reduced? we should allow override incase user wants to take some risks :)

Yes, just the default changed

satya-bodapati (approved):

PR #2307, opened 2026-06-24
(trunk)PXC-5229: WSREP applier threads should never have tx_read_only=true

jaideepkarande:
PXC-5229: WSREP applier threads should never have tx_read_only=true

WSREP applier threads may inherit global `transaction_read_only=1` after
changing `wsrep_applier_threads`, leading to apply failures such as:
  - BF applier errors
  - Cluster inconsistency
`transaction_read_only` is a SESSION+GLOBAL variable.
NOTE:
transaction_read_only has NO_CMD_LINE means there is no
command-line/option-file entry registered for it.
The only way to set the global is SET GLOBAL transaction_read_only=1 at
runtime. (But persist works, which can be used as a hack.)

`SET GLOBAL wsrep_applier_threads=N` spawns brand new applier THDs
(`start_wsrep_THD -> init_wsrep_thread`) *after* `transaction_read_only=1` was
set, so those new appliers inherit `tx_read_only=true` and fail to apply writes.

The read-only-transaction flag is not just a SQL-layer gate. It is enforced
at `open_tables`/`lock_tables` (`sql/sql_base.cc`, raising
`ER_CANT_EXECUTE_IN_READ_ONLY_TRANSACTION`) and it propagates into InnoDB.
Therefore a fix that only bypasses the SQL-layer check is insufficient/unsafe:
InnoDB would still start a read-only `trx` and the replicated write would fail
at the storage engine.

Resolution:

### DIFF
--- a/mysql-test/suite/galera/r/galera_var_sst_donor.result
+++ b/mysql-test/suite/galera/r/galera_var_sst_donor.result
@@ -1,0 +1,25 @@
+SET @sst_donor_orig=@@GLOBAL.wsrep_sst_donor;
+set global wsrep_sst_donor='127.0.0.1;';
+ERROR 42000: Variable 'wsrep_sst_donor' can't be set to the value of '127.0.0.1;'
+select @@wsrep_sst_donor;
+@@wsrep_sst_donor
+
+set global wsrep_sst_donor='node1\'';
+ERROR 42000: Variable 'wsrep_sst_donor' can't be set to the value of 'node1''
+select @@wsrep_sst_donor;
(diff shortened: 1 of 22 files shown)


jaideepkarande:
3 failing test cases have been handled in 9.7.1(due to missing variable gache failing test cases)
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/829/consoleFull

kamil-holubicki (approved):

PR #2308, opened 2026-06-24
(8.4)PXC-5229: WSREP applier threads should never have tx_read_only=true

jaideepkarande:
PXC-5229: WSREP applier threads should never have tx_read_only=true

WSREP applier threads may inherit global `transaction_read_only=1` after
changing `wsrep_applier_threads`, leading to apply failures such as:
  - BF applier errors
  - Cluster inconsistency
`transaction_read_only` is a SESSION+GLOBAL variable.
NOTE:
transaction_read_only has NO_CMD_LINE means there is no
command-line/option-file entry registered for it.
The only way to set the global is SET GLOBAL transaction_read_only=1 at
runtime. (But persist works, which can be used as a hack.)

`SET GLOBAL wsrep_applier_threads=N` spawns brand new applier THDs
(`start_wsrep_THD -> init_wsrep_thread`) *after* `transaction_read_only=1` was
set, so those new appliers inherit `tx_read_only=true` and fail to apply writes.

The read-only-transaction flag is not just a SQL-layer gate. It is enforced
at `open_tables`/`lock_tables` (`sql/sql_base.cc`, raising
`ER_CANT_EXECUTE_IN_READ_ONLY_TRANSACTION`) and it propagates into InnoDB.
Therefore a fix that only bypasses the SQL-layer check is insufficient/unsafe:
InnoDB would still start a read-only `trx` and the replicated write would fail
at the storage engine.

Resolution:

### DIFF
--- a/sql/wsrep_thd.cc
+++ b/sql/wsrep_thd.cc
@@ -448,6 +448,24 @@ void wsrep_delete_threadvars() {
 #endif
 }
 
+/*
+  Copy session context fields from src to dst.
+*/
+void wsrep_copy_session_from_thd(THD *dst, const THD *src) {
+  /*
+    Ref: set_slave_thread_default_charset
+  */
+  dst->variables.character_set_client = src->variables.character_set_client;
+  dst->variables.collation_connection = src->variables.collation_connection;
+  dst->variables.collation_server = src->variables.collation_server;
+  dst->variables.sql_mode = src->variables.sql_mode;
+  dst->variables.time_zone = src->variables.time_zone;
+  dst->update_charset();
+
+  dst->variables.transaction_read_only = false;
+  dst->tx_read_only = false;
+}
+
 void wsrep_assign_from_threadvars(THD *) {
 #if 0
   if (thread_handling == SCHEDULER_TYPES_COUNT) {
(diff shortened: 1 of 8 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/828/consoleFull

But log size is very big due to failing test case innodb.dblwr_zlib_encrypt_recv.
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/827/consoleFull

jaideepkarande:
Due to GCA branch was old, post null merge branch was updated to latest, so this patch is now based in latest 8.4 code changes

https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/834/consoleFull

sql/wsrep_thd.cc
```diff
@@ -448,6 +448,24 @@ void wsrep_delete_threadvars() {
 #endif
 }
 
+/*
+  Copy session context fields from src to dst.
+*/
+void wsrep_copy_session_from_thd(THD *dst, const THD *src) {
+  /*
+    Ref: set_slave_thread_default_charset
```
kamil-holubicki, line 456: Let's remove this comment. `set_slave_thread_default_charset` copies from global.

kamil-holubicki (approved):

PR #2321, opened 2026-07-10
PXC-5252: wsrep_sst_clone fails when its password does not meed validate_password component requirements

kamil-holubicki:
https://perconadev.atlassian.net/browse/PXC-5252

Problem:
Galera Clone SST fails when validate_password component is enabled with a strict policy. Donor-side creation of mysql.pxc.sst.user is rejected because clone SST script generates a password that does not meet the configured requirements.

Cause:
1. wsrep_create_sst_user() runs through normal DDL on the donor, so validate_password applies to the internal SST account. This affects all SSTs.
2. Clone SST script builds its password as wsrep_gen_secret() plus a fixed Aa suffix, which is too short and too weak for stricter length, mixed-case, digit, and special-character rules.

Solution:
1. Bypass password validation for the ephemeral donor mysql.pxc.sst.user in wsrep_create_sst_user() via m_disable_password_validation. It is safe because it is short living user, restricted to localhost access only.

2. Add configurable [sst] sst-password-suffix to wsrep_sst_clone.sh, append it to the generated clone password, and validate the suffix against an allowlist aligned with donor SST request validation.

Q: Why not parse cnf validate_password.* config in clone SST script? A1: That would couple SST with this particular component. A2: If component is configured not in cnf file but by SET PERSIST for example, that configuration is not easy accessible from SST script. A3: When adimistrator configures validate_password-like mechanism, he exactly knows what password suffix will satisfy validation.


sql/wsrep_sst.cc
```diff
@@ -1297,6 +1303,8 @@ static int wsrep_create_sst_user(bool initialize_thread, const char *password) {
   // Overwrite query (clear out any sensitive data)
   ::memset(auth_buf, 0, auth_len);
 
+  thd->m_disable_password_validation = saved_disable_password_validation;
+
```
jaideepkarande, line 1307: Is this code changes still needed because with $SST_PASSWORD_SUFFIX password validation should pass.
I do not see this code as a problem since validation is switched of temporarily and help us with future such issues, but then we still bypass enforced validation which may not be needed.

scripts/wsrep_sst_clone.sh
```diff
@@ -124,6 +130,12 @@ OS=$(uname)
 JOINER_TIMEOUT_WAIT_DONOR_MESSAGE=$(parse_cnf sst joiner-timeout-wait-donor-message "60")
 JOINER_TIMEOUT_WAIT_CLONE_INSTANCE=$(parse_cnf sst joiner-timeout-clone-instance "90")
 DONOR_TIMEOUT_WAIT_JOINER_CLONE_INSTANCE=$(parse_cnf sst donor-timeout-wait-joiner "200")
+SST_PASSWORD_SUFFIX=$(parse_cnf sst sst-password-suffix "")
+
+if [[ ! "$SST_PASSWORD_SUFFIX" =~ ^[A-Za-z0-9._/-]*$ ]]; then
```
jaideepkarande, line 135: 1. Should we allow /?
2. Is empty string allowed "sst-password-suffix="?

jaideepkarande (commented):
Thanks Kamil for working on changes. Code changes looks fine, I just have some queries.

sql/wsrep_sst.cc
kamil-holubicki: Hi, yes, it is needed. This user is created for PXB access on the donor (localhost-only access). It is created by the server. The purpose  is to allow Donor-side SST script to access the server (PXB access in case of pxb-sst, admin access in case of clone-SST).

The password generated by clone-SST joiner side script is another one. It has to be known by both sides: Joiner and Donor. Joiner sends it to the Donor in SST request string, then Donor creates the clone user, then Joiner connects to the donor using this password. This path does not touch the code in question.

Long story short, both passwords go through validata_password and need to fulfill its requirements.

scripts/wsrep_sst_clone.sh
kamil-holubicki: 1. It will go to the Donor side and go through SST request validation regex (wsrep_sst.cc, is_sst_request_valid). The logic there allows exactly the same special character subset. Do you see any problem with `/`. If so, we can limit this set. 
2. Yes, it is. By default, we add `readonly PSWD_POLICY="Aa1."` suffix for the password. `sst-password-suffix=` is to extend it if needed.

jaideepkarande: You are right, thought came because of recent galera security related to wsrep_notify_cmd, not with password.

https://github.com/MariaDB/server/pull/5113
```
The current blacklist of unsafe characters is too narrow and fails to block many common shell metacharacters that could be used for command injection. To properly secure the sh -c execution, you should also reject characters like ;, &, |, ", $, \, and others.
```
https://bugs.launchpad.net/ubuntu/+source/dash/+bug/259671
https://bugs.mysql.com/bug.php?id=69044 (both treated as bug and solved)

jaideepkarande (approved):
LGTM

PR #2325, opened 2026-07-16
[DISCARDED PXC-5291 will have fix](8.4)PXC-5292: CREATE TRIGGER with nonexistent DEFINER by SET_ANY_DEFINER …

jaideepkarande:
…user can cause node inconsistency/eviction

The early privilege gate added by PXC-4765 in Sql_cmd_create_trigger::execute() runs before wsrep_to_isolation_begin() so the origin rejects an invalid DEFINER before replicating it. That gate mirrored only the definer-mismatch stage of check_valid_definer() (WL#15874), not the orphan-definer stage. As a result a user with SET_ANY_DEFINER but without SUPER or ALLOW_NONEXISTENT_DEFINER could create a trigger with a non-existent DEFINER: the early gate passed, the statement replicated, then the origin rejected it during creation (after TOI) while the appliers created it successfully -- leaving the origin inconsistent and evicting it from the cluster.

Add the orphan-definer check (require SUPER or ALLOW_NONEXISTENT_DEFINER when the DEFINER is not an existing ACL user) to the early pre-TOI gate, mirroring stage 2 of check_valid_definer(). The informational ER_NO_SUCH_USER note is left to the authoritative post-TOI check_valid_definer() call to avoid a duplicate warning.

### DIFF
--- a/sql/sql_trigger.cc
+++ b/sql/sql_trigger.cc
@@ -449,6 +449,33 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
                                              true /* report_error */))
     return true;
 
+  /*
+    A DEFINER referring to a non-existent user (orphan object) requires SUPER or
+    ALLOW_NONEXISTENT_DEFINER. This mirrors stage 2 of check_valid_definer() in
+    sql/auth/sql_authorization.cc. Doing it here, before
+    wsrep_to_isolation_begin(), makes the origin reject the statement before it
+    is replicated; otherwise the origin would reject it only during creation
+    (after TOI) while the appliers create it successfully, leaving this node
+    inconsistent and evicting it from the cluster (PXC-5292, a PXC-4765-class
+    issue). The informational ER_NO_SUCH_USER warning is intentionally left to
+    the authoritative check_valid_definer() call during trigger creation.
+    is_acl_user() self-locks the ACL cache (READ, non-blocking) and returns true
+    when it cannot lock or ACL is not yet initialized; in that rare case this
+    gate is skipped and we simply fall back to the pre-existing post-TOI
+    check_valid_definer() behavior -- no worse than before this fix.
+  */
+  const bool has_super =
+      sctx->check_access(SUPER_ACL);  // This will come from PXC-5291
+  if (lex->definer &&
+      !is_acl_user(thd, lex->definer->host.str, lex->definer->user.str) &&
+      !has_super &&
+      !sctx->has_global_grant(STRING_WITH_LEN("ALLOW_NONEXISTENT_DEFINER"))
+           .first) {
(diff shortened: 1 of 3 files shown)


jaideepkarande:
https://pxc.cd.percona.com/view/8.0%20parallel%20MTR/job/pxc-8.x-pipeline-parallel-mtr/856/consoleFull

sql/sql_trigger.cc
```diff
@@ -449,6 +449,33 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
                                              true /* report_error */))
     return true;
 
+  /*
```
kamil-holubicki, line 452: I think both prechecks (This one and the one introduced in PXC-4765) can be collapsed into a single call to `check_valid_definer()` + binlog specific checks, like

```
Security_context *sctx = thd->security_context();
LEX *lex = thd->lex;

const bool definer_is_current_user =
    !lex->definer ||
    (strcmp(lex->definer->user.str, sctx->priv_user().str) == 0 &&
     my_strcasecmp(system_charset_info,
                   lex->definer->host.str, sctx->priv_host().str) == 0);

// WSREP-specific binlog gate: creating a trigger under your OWN name still
// requires SUPER/SET_ANY_DEFINER when trust_function_creators is off and
// binlog is on. check_valid_definer() does not cover this (its stage 1 is
// skipped when definer == current user).
if (definer_is_current_user &&
    !trust_function_creators &&
    (WSREP_EMULATE_BINLOG(thd) || mysql_bin_log.is_open()) &&
    !(sctx->check_access(SUPER_ACL) ||
      sctx->has_global_grant(STRING_WITH_LEN("SET_ANY_DEFINER")).first)) {
  my_message(ER_BINLOG_CREATE_ROUTINE_NEED_SUPER,
             "You do not have the SUPER privilege (you *might* want to use "
             "the less safe log_bin_trust_function_creators variable)",
             MYF(0));
  return true;
}

// Everything else: differing-definer privilege, system_user protection,
// orphan-definer privilege - delegate to the authoritative check.
if (lex->definer && check_valid_definer(thd, lex->definer)) return true;
```

NOTE:

The wrinkle: ER_NO_SUCH_USER warning duplication on the origin.

If a caller with ALLOW_NONEXISTENT_DEFINER creates a trigger with an orphan definer:

- Pre-TOI check_valid_definer passes stage 2 and pushes ER_NO_SUCH_USER (sql_authorization.cc:7832-7834).
- Post-TOI check_valid_definer at table_trigger_dispatcher.cc:207 runs again on the origin during actual creation and pushes it again.

Client sees the note twice.

To fix that:
Add a bool quiet (or bool report_no_such_user_warning) param to check_valid_definer(). Minimal, does the job; touches sql/auth/sql_authorization.cc

kamil-holubicki (changes requested):

PR #2339, opened 2026-07-29
(trunk) PXC-5252: wsrep_sst_clone fails when its password does not meet valid…

kamil-holubicki:
…ate_password component requirements

https://perconadev.atlassian.net/browse/PXC-5252

Problem:
Galera Clone SST fails when validate_password component is enabled with a strict policy. Donor-side creation of mysql.pxc.sst.user is rejected because clone SST script generates a password that does not meet the configured requirements.

Cause:
1. wsrep_create_sst_user() runs through normal DDL on the donor, so validate_password applies to the internal SST account. This affects all SSTs.
2. Clone SST script builds its password as wsrep_gen_secret() plus a fixed Aa suffix, which is too short and too weak for stricter length, mixed-case, digit, and special-character rules.

Solution:
1. Bypass password validation for the ephemeral donor mysql.pxc.sst.user in wsrep_create_sst_user() via m_disable_password_validation. It is safe because it is short living user, restricted to localhost access only.

2. Add configurable [sst] sst-password-suffix to wsrep_sst_clone.sh, append it to the generated clone password, and validate the suffix against an allowlist aligned with donor SST request validation.

Q: Why not parse cnf validate_password.* config in clone SST script? A1: That would couple SST with this particular component. A2: If component is configured not in cnf file but by SET PERSIST for example, that configuration is not easy accessible from SST script. A3: When adimistrator configures validate_password-like mechanism, he exactly knows what password suffix will satisfy validation.


jaideepkarande (approved):
LGTM

PR #2353, opened 2026-08-26
(8.4)PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=EN…

jaideepkarande:
…FORCING

Problem:

Description
Example:

Resolution:

2. Session vs global Changing 'pxc_strict_mode' may update '@@global.sql_require_primary_key' as above(if setting to ON). The server does not rewrite '@@session.sql_require_primary_key' on 'existing' connections. Each session keeps its session value until the client changes it or opens a new connection (new sessions inherit the global default as usual).

3. Startup If Galera is active and 'pxc_strict_mode' is 'ENFORCING' or 'MASTER' and the server starts with 'global' 'sql_require_primary_key' 'OFF' (e.g. from options or config), startup forces 'global' 'sql_require_primary_key' 'ON' so the effective policy matches strict mode.

4. Upgrade Do not apply the pxc_strict_mode guard to statements run by the server's upgrade threads.

### DIFF
--- a/sql/sys_vars.cc
+++ b/sql/sys_vars.cc
@@ -428,6 +428,12 @@ static bool check_session_admin_no_super(sys_var *self, THD *thd,
   @param thd the session context
   @param setv the SET operations metadata
  */
+#ifdef WITH_WSREP
+static bool check_session_admin_and_sql_require_primary_key_on_check(sys_var *,
+                                                                     THD *,
+                                                                     set_var *);
+#endif /* WITH_WSREP */
+
 static bool check_session_admin(sys_var *self, THD *thd, set_var *setv) {
   Security_context *sctx = thd->security_context();
 
@@ -7946,7 +7952,11 @@ static Sys_var_bool Sys_sql_require_primary_key{
     DEFAULT(false),
     NO_MUTEX_GUARD,
     IN_BINLOG,
+#ifdef WITH_WSREP
+    ON_CHECK(check_session_admin_and_sql_require_primary_key_on_check)};
+#else
     ON_CHECK(check_session_admin)};
+#endif
 
 static Sys_var_bool Sys_sql_generate_invisible_primary_key(
     "sql_generate_invisible_primary_key",
@@ -8203,6 +8213,29 @@ static Sys_var_charptr Sys_protocol_compression_algorithms(
 #include "wsrep_binlog.h"
 #include "wsrep_sst.h"
 #include "wsrep_var.h"
+
+static bool check_session_admin_and_sql_require_primary_key_on_check(
+    sys_var *self, THD *thd, set_var *var) {
+  if (check_session_admin(self, thd, var)) return true;
+  if (pxc_strict_mode < PXC_STRICT_MODE_ENFORCING) return false;
+  /*
+    Allow the server upgrade thread to bypass pxc_strict_mode when temporarily
+    disabling sql_require_primary_key for system tables without primary keys.
+    These are internal statements compiled into the server; rejecting them can
+    abort upgrades and prevent the node from starting (PXC-5296).
+  */
+  if (thd->is_server_upgrade_thread()) return false;
+  if (var->save_result.ulonglong_value == 0) {
+    const char *strict_name =
+        (pxc_strict_mode == PXC_STRICT_MODE_MASTER) ? "MASTER" : "ENFORCING";
+    WSREP_ERROR(
+        "Cannot set sql_require_primary_key=OFF while pxc_strict_mode is %s.",
+        strict_name);
+    my_error(ER_WRONG_VALUE_FOR_VAR, MYF(0), self->name.str, "OFF");
+    return true;
+  }
+  return false;
+}
 
 static PolyLock_mutex PLock_wsrep_cluster_config(&LOCK_wsrep_cluster_config);
 static Sys_var_charptr Sys_wsrep_provider(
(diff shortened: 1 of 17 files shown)


jaideepkarande:
## Jenkins results

Job: [`pxc-8.x-pipeline-parallel-mtr`](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/) · branch `PXC-5106_2-8.4` @ `4f203a74122`

| Run | Configuration | Result | Failures |
|---|---|---|---|
| [884](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/884/) | **baseline** (`PXC-5267-8.4-ci`, Debug, no ASAN) | UNSTABLE | 8 |
| [890](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/890/) | **this branch**, parameters identical to 884 (only `BRANCH` differs) | UNSTABLE | 5 |

Run 890 is the like-for-like comparison: `ubuntu:noble`, `CMAKE_BUILD_TYPE=Debug`, `ANALYZER_OPTS=""`, `GALERA_PARALLEL_RUN=2`, `FULL_MTR=yes`, `MTR_ARGS="--unit-tests-report --big-test --mem --testcase-timeout=120"`.

(An earlier run, [889](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/889/), showed 16 failures but is **not** a valid comparison — it was triggered without `--testcase-timeout=120` and on `oraclelinux:8`. 7 of its 8 extra failures were literally `timeout after 600 seconds`, and the 8th was an OpenSSL error-string difference caused by the OS change.)

### Tests added/touched by this PR — all pass, both combinations

| Test | 890 `enc_off` | 890 `enc_on` |
|---|---|---|
| `galera.galera_pxc_5296_upgrade_strict_mode` | pass 60.3s | pass 57.0s |
| `galera.galera_pxc_sql_require_primary_key` | pass 17.4s | pass 19.6s |
| `galera.galera_rpl_pxc_strict_mode_mismatch` | pass | pass |
| `galera.pxc_install_gr` | pass | pass |

### Failure analysis for run 890

| Test | Present in baseline 884? | Assessment |
|---|---|---|
| `galera.galera_bf_bf_2.enc_off` | yes | Pre-existing. Its recorded `.result` contains a baked-in `wait_condition.inc` timeout line that did not occur, so the diff is a *missing* timeout. |
| `shutdown_report` (galera-small) | yes | Not a test — MTR reporting a server that failed to shut down in time. |
| `shutdown_report` (galera-big) | same class | As above. |
| `galera.galera_sst_allowed_methods` (`enc_off` + `enc_on`) | no | **Only failure not in the baseline** — see below. |

Failures present in baseline 884 but **not** on this branch: `percona.acl_mdl_scalability`, `galera_bf_bf_2.enc_on`, `main.all_persisted_variables`, `galera-x.performance_schema_stmt`, `shutdown_report` (innodb, galera_3nodes).

No occurrence anywhere in either console log of any error this change can produce — `ER_TABLE_WITHOUT_PK` (3750), `ER_WRONG_VALUE_FOR_VAR` (1231) on `sql_require_primary_key`, `Cannot set sql_require_primary_key=OFF while pxc_strict_mode is ...`, `Setting sql_require_primary_key=ON because ...`, or `Failed to upgrade server`.

### `galera_sst_allowed_methods` — open item, not yet proven fragile

```
mysqltest: At line 101: Command "shutdown_server" failed with error 2. my_errno=175.
  include/restart_mysqld.inc:102
  <- suite/galera/t/galera_sst_allowed_methods_garbd.inc:26
  <- suite/galera/t/galera_sst_allowed_methods.test:82
```
It fails in the garbd section that deliberately sends malformed SST requests after killing node #2 to free ports. Notes:

* In baseline 884 it passed, but slowly — 394825 ms and 364250 ms (~6.5 min per combination). It is a heavy test close to its limits.
* The failing machinery (garbd, SST method validation, `shutdown_server`) is not touched by this change.
* Leftover state from the new tests is ruled out: they ran on the same worker (`WORKER_5_galera-big`) at 67–68%, this failed at 85%, and 15+ tests passed in between — including `galera_rsu_add_pk` (a primary-key test) and `galera_ist_xtrabackup-v2 'enc_on'` immediately before.

I would like one targeted rerun of just this test before labelling it fragile.

### Note on running `enc_on` locally

The galera suite's `enc_on` combination only works if the server binary is named `mysqld`, because `suite/galera/t/suite-setup.sh:9` hardcodes the local keyring manifest as `$datadir/mysqld.my`. A `-DBUILD_CONFIG=mysql_release` debug build produces `mysqld-debug`, which then looks for `$datadir/mysqld-debug.my` and never loads `component_keyring_file`, so GCache encryption aborts at startup. Jenkins avoids this by copying the debug binary to `bin/mysqld`. Arguably a harness bug worth fixing separately.

With that addressed locally, `galera.galera_kill_connection` passes both combinations on this branch (796s / 820s) — confirming its run-889 timeout was purely the missing `--testcase-timeout`.

sql/sys_vars.cc
```diff
@@ -8976,4 +9009,3 @@ static Sys_var_enum_default_table_encryption Sys_default_table_encryption(
     HINT_UPDATEABLE SESSION_VAR(default_table_encryption), CMD_LINE(OPT_ARG),
     default_table_encryption_type_names, DEFAULT(DEFAULT_TABLE_ENC_OFF),
     NO_MUTEX_GUARD, IN_BINLOG, ON_CHECK(check_set_default_table_encryption));
-
```
kamil-holubicki, line 8979: Removed empty line at the end of file. Please revert.

kamil-holubicki (approved):
LGTM

sql/sys_vars.cc
jaideepkarande: Reverted — the trailing empty line at the end of `sql/sys_vars.cc` is restored. The file's last bytes are now byte-identical to `pxc/8.4` (`;\n\n`), and the only remaining removal in this file is the intended `ON_CHECK(pxc_strict_mode_check), ON_UPDATE(0));` line, which becomes `ON_UPDATE(pxc_strict_mode_update)`.

Applied on the propagation branches too, so the three lines stay in parity:
* `PXC-5106_2-8.4` → `2140bcccac5` (this PR)
* `PXC-5106_2-trunk` → `6b15bc82da8` (#2356)
* `PXC-5106_2-9.7` — no change needed; that base never had the trailing blank line, and the branch removes nothing from this file (#2357)

PR #2320, opened 2026-07-10
(8.4)PXC-5208: Force SST after inconsistency eviction

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-5208

When a node is voted out of the cluster due to a data inconsistency, remove grastate.dat on shutdown so the node is forced to perform a full SST on the next start instead of IST, which would otherwise resurrect the inconsistent dataset via --wsrep-recover.

PR #2355, opened 2026-08-27
[trunk]PXC-5208: Force SST after inconsistency eviction

jaideepkarande:
# PXC-5208 — port to 9.7 and trunk

**Ticket**: [PXC-5208: Add an option to force SST after a node is restarted following an inconsistency](https://perconadev.atlassian.net/browse/PXC-5208)

---

## Problem Summary

When a node hits a data inconsistency it is voted out of the cluster and marks its state
corrupt — `grastate.dat` is rewritten with `uuid = 00000000-...-0` and `seqno = -1`. The node
is deliberately left running so the inconsistency can be investigated.

A `seqno = -1` state file looks like it should force SST, but it does not. `mysqld_safe`'s
`wsrep_recover_position()` sees `seqno == -1`, decides the file is unusable, and therefore
runs `mysqld --wsrep_recover`, which reads the **last committed position out of InnoDB** and
prints a perfectly valid `uuid:seqno`. The node then starts with a known position, the
cluster offers it **IST**, and the write sets are replayed on top of the still-inconsistent
local dataset. The inconsistency is never repaired.

The only reliable way to force SST is for the node to start with no usable position at all,
which is what `mysqld_safe` does when `grastate.dat` is absent. Hence the long-standing
manual workaround of `rm grastate.dat`, which most operators do not know about.

## Fix

The node removes `grastate.dat` on shutdown when it left the cluster because of an
inconsistency, so the next start can only be a full SST. Implemented in the Galera
provider, entirely inside `#ifdef PXC`:

- `SavedState::set_remove_on_close(bool)` sets a `remove_on_close_` flag;
  `~SavedState()` unlinks the file after closing it (idempotent — `ENOENT` ignored, other
  errors logged with a hint to remove the file manually).
- `SavedState::try_remove_on_close()` unlinks immediately, for the `recv_IST()` path that
  ends in `abort()` and so never runs the destructor.
- `SavedState::mark_uncorrupt()` clears the flag, so a node that legitimately recovers a
  consistent state (e.g. via SST) keeps its state file.
- Wired in at every "leaving because of an inconsistency" exit:
  `ReplicatorSMM::mark_corrupt_and_close()` (covers `on_inconsistency()` and the fatal
  apply path), `async_recv()` on `INCONSISTENCY_CODE`, and `process_IST_writeset()` /
  `recv_IST()`.

Behaviour is unconditional — the `repl.force_sst_after_inconsistency` provider option from
an earlier iteration was deliberately dropped. The corrupt `grastate.dat` is still available
for inspection for as long as the evicted node is kept up; it only disappears at shutdown.

---

## Compatibility & Upgrade Assessment

| Dimension | Impact? | Notes |
|---|---|---|
| In-place upgrade (on-disk / DD / redo-undo / system tables) | **No** | No format change. `grastate.dat`'s own format is untouched; only its lifetime changes — it is unlinked on shutdown after an inconsistency eviction. |
| Downgrade | **No** | An older binary starting on a datadir with no `grastate.dat` takes the pre-existing "no state file" path and requests SST. That is the desired outcome, and it is exactly what the documented manual `rm grastate.dat` workaround already produces. |
| Rolling / mixed-version cluster (wsrep, IST/SST) | **No** | The change is entirely node-local and post-eviction. No wsrep/Galera protocol, certification, state-exchange or SST/IST wire change. A mixed cluster simply has some nodes that force SST after their own eviction and some that do not — no interaction between them. |
| Replication (binlog format, GTID, cross-version N-1↔N) | **No** | Nothing in the binlog, GTID or applier path is touched. |
| Client-server wire protocol | **No** | No protocol, capability flag or packet change. |
| System variables / config / defaults | **No new variable** — but a **behaviour change with no opt-out** | The `repl.force_sst_after_inconsistency` provider option from an earlier iteration was removed, so the new behaviour is unconditional. Operators who relied on the corrupt `grastate.dat` surviving a restart lose that; the file is still inspectable for as long as the evicted node stays up. Worth calling out in the release notes. |
| Cross-version behaviour (8.0 / 8.4 / 9.7 / trunk) | **Consistent** | Same galera change on every line; the five touched files are byte-identical between the 9.7 and trunk/8.4 galera bases, and the MTR test is byte-identical across branches. |

Overall: **no upgrade, downgrade, format or protocol impact.** The one thing reviewers
should weigh is that the SST-forcing behaviour is now unconditional.
---

## Related PRs

| Line | PR |
|---|---|
| galera (`4.x-8.0`, shared by all lines) | https://github.com/percona/galera/pull/315 |
| 8.4 | https://github.com/percona/percona-xtradb-cluster/pull/2320 |
| 9.7 | https://github.com/percona/percona-xtradb-cluster/pull/2354 |
| trunk | (this PR / see below) |

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "percona-xtradb-cluster-galera"]
 	path = percona-xtradb-cluster-galera
-	url = https://github.com/percona/galera
+	url = https://github.com/jaideepkarande/galera
 	branch = 4.x-8.0
 [submodule "wsrep-lib"]
 	path = wsrep-lib
--- a/mysql-test/suite/galera/r/pxc_5208_force_sst_after_inconsistency.result
+++ b/mysql-test/suite/galera/r/pxc_5208_force_sst_after_inconsistency.result
@@ -1,0 +1,6 @@
+DELETE FROM t1 WHERE f1 = 2;
+# Verify grastate.dat was removed on inconsistency shutdown
+# restart
+node_2 t1 after SST: 1:0,3:0,4:0,5:0,6:0,7:0,8:0
+node_1 t1 contents: 1:0,3:0,4:0,5:0,6:0,7:0,8:0
+datasets consistent: 1
--- a/mysql-test/suite/galera/t/pxc_5208_force_sst_after_inconsistency.test
+++ b/mysql-test/suite/galera/t/pxc_5208_force_sst_after_inconsistency.test
@@ -1,0 +1,97 @@
+#
+# PXC-5208: Force SST after a node is restarted following an inconsistency.
+#
+# When a node is voted out of the cluster because of a data inconsistency it
+# marks its state as corrupt.  On the next start, during recovery IST can
+# bring back the inconsistent dataset.
+#
+# This test provokes an inconsistency on node_2, waits for the cluster to evict
+# it, shuts it down and verifies that grastate.dat has been removed. The node
+# is then restarted and expected to rejoin via SST with a consistent dataset.
+#
+
+--source include/galera_cluster.inc
+
+--connection node_1
+--disable_query_log
+# pc.weight=2 lets node_1 maintain quorum alone after node_2 leaves (2-node cluster).
+SET GLOBAL wsrep_provider_options='pc.weight=2';
+CREATE TABLE t1 (f1 INTEGER PRIMARY KEY, f2 INT);
+INSERT INTO t1 VALUES (1, 0),(2, 0),(3, 0),(4, 0),(5, 0),(6, 0),(7, 0),(8, 0);
+--enable_query_log
+
+--file_exists $MYSQLTEST_VARDIR/mysqld.2/data/grastate.dat
+#
+# Provoke inconsistency: delete row f1=2 locally on node_2 (not replicated),
+# then delete it on node_1 (replicated).  When the replicated DELETE arrives
+# at node_2 and can't find the row, the apply error triggers inconsistency
+# voting and node_2 is evicted.
+#
+--connection node_2
+--let $dummy = `SELECT COUNT(*) FROM t1`
+--disable_query_log
+SET wsrep_on=OFF;
+DELETE FROM t1 WHERE f1 = 2;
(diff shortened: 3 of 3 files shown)


jaideepkarande:
### Jenkins results — trunk

**[AI job build 5](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr-AI/5/)** · Debug · ubuntu:noble · commit `87f467945cc`
PXB: 9.7.1-rc1 → `pxb-9.7`, 8.4.0-6 → `pxb-8.4`, 9.1.0 → `pxb-9.1`
Result: **UNSTABLE** — 11042 passed, 8 failed.

#### All four PXC-5208 tests passed, both encryption combinations

| test | enc_off | enc_on | worker |
|---|---|---|---|
| `pxc_5208_force_sst_after_inconsistency` | pass | pass | galera-small |
| `pxc_5208_force_sst_after_inconsistency_default` | pass | pass | galera-small |
| `pxc_5208_force_sst_after_ist_receiver_failure` | pass | pass | galera-big |
| `pxc_5208_force_sst_after_ist_failure` | pass | pass | galera-big |

8/8. The `[ skipped ] Not a big test` / `Test needs 'big-test'` lines in the logs are
just the galera-small vs galera-big split — each test runs on exactly one worker.

#### Failures — none related to this change

| test | combos | assessment |
|---|---|---|
| `galera_fk_selfreferential` | enc_off, enc_on | **pre-existing** — also fails on the clean trunk baseline (build 902, run without any of this work) |
| `main.join_cache_bnl` | — | unrelated optimizer/join-cache test, no connection to SST, IST or state-file handling |

Notably the gcache failures seen in earlier runs (`galera_gcache_pagestore`,
`galera_gcache_recover_full_gcache`, `galera_gcache_recover_manytrx`) did **not**
recur here, though they are present in the trunk baseline.

No `pxc_5208_*` test failed. `shutdown_report` entries are MTR's end-of-worker
report, raised whenever anything else in that worker failed, not real tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

kamil-holubicki (approved):

PR #899, opened 2019-12-30
- PXC-2894: stablizing pquery3 against PXC-8.0

kbauskar:
- If a high priority thread kills a local transaction that is currently
    executing ROLLBACK TO SAVEPOINT statement then transaction is not
    rolled back. This is because of special handling in MySQL rollback
    flow that ignore command code to SQLCOM_ROLLBACK_TO_SAVEPOINT.

  - Given local transaction is forcefully aborted by a background thread
    state of transaction is marked as must_abort. This state could be
    check along with command code == SQLCOM_ROLLBACK_TO_SAVEPOINT to kick-off
    rollback in the case described above.

  (Other fixes were related to accidental merge miss).

PR #890, opened 2019-12-16
Pxc rel 5.7.28

EvgeniyPatlan:

PR #2357, opened 2026-08-28
(9.7)PXC-5296: PXC upgrade fails when pxc_strict_mode is ENFORCING

jaideepkarande:
Ports the fix from #2353 (8.4) to **9.7**.

https://perconadev.atlassian.net/browse/PXC-5296

## What this is

Unlike 8.4 and trunk, the `9.7` branch **still carries the original PXC-5106** — the revert in `fe0dac6dbd2` never reached it (verified: `check_session_admin_and_sql_require_primary_key_on_check`, `pxc_strict_mode_update` and the startup forcing are all present at `5dcb2246f13`). This line therefore has the upgrade bug, and this PR applies **only the PXC-5296 fix delta**, not a re-land.

Derived from `PXC-5106_2-8.4` @ `fb8f1d2ed99` (the head of #2353), applying only the PXC-5296 delta on top of the PXC-5106 already present here.

## The bug being fixed

`pxc_strict_mode=ENFORCING/MASTER` forces `sql_require_primary_key=ON` and rejects turning it back off. The server upgrade thread runs the compiled-in upgrade script — `mysql_system_tables.sql` concatenated with `mysql_system_tables_fix.sql` (see `scripts/CMakeLists.txt`) — which must relax that policy to rebuild system tables that have no primary key. Both halves were rejected, so the upgrade aborted and the node could not start:

```
[Server] Server upgrade from '80045' to '80046' started.
[WSREP]  Cannot set sql_require_primary_key=OFF while pxc_strict_mode is ENFORCING.
[Server] Execution of server-side SQL statement
         'SET @@session.sql_require_primary_key = 0; ' failed with error code = 1231
[Server] Failed to upgrade server.
[Server] Aborting
```

Since `pxc_strict_mode` defaults to `ENFORCING`, this hit every upgrade of a node left in a strict mode.

## Changes

* `sql/sys_vars.cc` — exempt the **server upgrade thread** from the strict-mode guard so the script's internal `SET` is accepted. A normal client `SET sql_require_primary_key=OFF` under ENFORCING is still rejected. Deliberately narrow: `is_dd_system_thread()` / `is_initialize_system_thread()` are *not* exempted because they are unreachable here (`--initialize` forces `pxc_strict_mode=DISABLED`, and the DD restart thread only runs the `PXC_SECTION` block, which never touches this variable).
* `sql/bootstrap.cc` — start the upgrade thread with `sql_require_primary_key=false` (under `#ifdef WITH_WSREP`). The script only relaxes the policy around the log tables, so this covers the primary-key-less firewall DDL it does not wrap (`ALTER TABLE mysql.firewall_whitelist ENGINE=InnoDB`, `CREATE TABLE mysql.firewall_membership`).
* `scripts/mysql_system_tables.sql`, `scripts/mysql_system_tables_fix.sql` — relax `pxc_strict_mode` around the log-table blocks, gated on it not already being `DISABLED` so a non-cluster node and `--initialize` are unaffected (`pxc_strict_mode` can only be changed on a cluster node; setting it elsewhere fails with `ER_UNKNOWN_ERROR` and would itself abort the upgrade).

Both C++ changes are required, verified by reverting each in isolation: without `sys_vars.cc` the failure is error **1231** on the internal `SET`; without `bootstrap.cc` it is error **3750** `ER_TABLE_WITHOUT_PK` on the firewall DDL.

## Tests

`mysql-test/suite/galera/t/galera_pxc_5296_upgrade_strict_mode.test` seeds a node with a legacy primary-key-less `mysql.firewall_whitelist`, restarts it with `--pxc_strict_mode=ENFORCING --upgrade=FORCE`, and asserts the node returns, rejoins, stays ENFORCING with `sql_require_primary_key=ON`, that the primary-key-less system tables were rebuilt, and that pre-upgrade data survived. It also guards that a client `SET ... = OFF` remains blocked.

Verified the fix regions are byte-identical across `PXC-5106_2-8.4`, `PXC-5106_2-9.7` and this branch.

## Jenkins

[pxc-8.x-pipeline-parallel-mtr #899](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/899/) — Debug, `ubuntu:noble`, `FULL_MTR=yes`, `MTR_ARGS="--unit-tests-report --big-test --mem --testcase-timeout=120"`, PXB `pxb-9.6`/`pxb-9.5`/`pxb-8.4` per this line's `wsrep_sst_xtrabackup-v2.sh`. **Running — results to follow.**

For reference, the 8.4 equivalent (#2353) ran as [#890](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/890/): 5 failures against a matched baseline's 8, with all four affected tests passing in both `enc_off` and `enc_on`.

## Note on the PXB parameters used

`.ai_helper.md`'s "Jenkins 9.7 parameters" specify target dirs `pxb-9.7` (THIS) / `pxb-9.6` (PREV), but that matches `release-9.7.1-1`'s `wsrep_sst_xtrabackup-v2.sh`. The `9.7` branch this PR targets expects `pxb-9.6` / `pxb-9.5` / `pxb-8.4`, so the run above uses those (identical to trunk's mapping) — parameters must match the code under test. Worth reconciling that config block.

Also included here: `galera_rpl_pxc_strict_mode_mismatch.test` on this line still referenced the pre-8.4 include names `include/wait_for_slave_io_to_start.inc` / `wait_for_slave_sql_to_start.inc`, which no longer exist; they are repointed at `include/rpl/wait_for_receiver_to_start.inc` / `wait_for_applier_to_start.inc`. Without that the test fails on every run.

### DIFF
--- a/mysql-test/suite/galera/r/galera_pxc_5296_upgrade_strict_mode.result
+++ b/mysql-test/suite/galera/r/galera_pxc_5296_upgrade_strict_mode.result
@@ -1,0 +1,68 @@
+#
+# 1. Baseline.
+#
+call mtr.add_suppression("Cannot set sql_require_primary_key=OFF while pxc_strict_mode is");
+call mtr.add_suppression("Setting sql_require_primary_key=ON because pxc_strict_mode is ENFORCING or MASTER");
+call mtr.add_suppression("Cannot set sql_require_primary_key=OFF while pxc_strict_mode is");
+call mtr.add_suppression("Setting sql_require_primary_key=ON because pxc_strict_mode is ENFORCING or MASTER");
+include/assert.inc [pxc_strict_mode is DISABLED]
+include/assert.inc [sql_require_primary_key global is OFF.]
+#
+# 2. Seed node_2 with a legacy primary-key-less system table and user data.
+#
+SET SESSION wsrep_on = OFF;
+CREATE TABLE mysql.firewall_whitelist (USERHOST VARCHAR(80) NOT NULL,
+RULE text NOT NULL) ENGINE=InnoDB;
+SET SESSION wsrep_on = ON;
+include/assert.inc [seeded mysql.firewall_whitelist has no primary key]
+CREATE TABLE t1 (id INT PRIMARY KEY, val VARCHAR(32)) ENGINE=InnoDB;
+INSERT INTO t1 VALUES (1, 'before'), (2, 'upgrade'), (3, 'data');
+SELECT * FROM t1 ORDER BY id;
+id	val
+1	before
+2	upgrade
+3	data
+#
+# 3. Regression guard: a client SET sql_require_primary_key=OFF is still
+#    blocked under ENFORCING.
+#
+SET GLOBAL pxc_strict_mode = 'ENFORCING';
+SET GLOBAL sql_require_primary_key = OFF;
+ERROR 42000: Variable 'sql_require_primary_key' can't be set to the value of 'OFF'
+SET SESSION sql_require_primary_key = OFF;
+ERROR 42000: Variable 'sql_require_primary_key' can't be set to the value of 'OFF'
+SET GLOBAL pxc_strict_mode = 'DISABLED';
+SET GLOBAL sql_require_primary_key = OFF;
+#
+# 4. Restart node_2 with --pxc_strict_mode=ENFORCING --upgrade=FORCE.
+#
+# restart:--pxc_strict_mode=ENFORCING --upgrade=FORCE
+#
+# 5. Post-upgrade checks.
+#
+include/assert.inc [node_2 is up and pxc_strict_mode is still ENFORCING]
+include/assert.inc [global sql_require_primary_key is ON under ENFORCING]
+include/assert.inc [mysql.firewall_whitelist gained its ID primary key]
+include/assert.inc [mysql.firewall_membership was created by the upgrade]
+include/assert.inc [mysql.firewall_membership got its composite primary key]
+include/assert.inc [mysql.general_log and mysql.slow_log survived without a primary key]
+SELECT * FROM t1 ORDER BY id;
+id	val
+1	before
+2	upgrade
+3	data
+#
+# 6. Cleanup.
+#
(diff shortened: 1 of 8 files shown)


jaideepkarande:
**Closing the ticket because this ticket is wrong
Details:**
This is wrong, 9.7.0 was missing rollback commit. Rollback commit is present in 9.7.1.
So this PR is blocked till 9.7.1 is merged back to 9.7.
once merge is done, rollback will be visible and complete code change can be made.
```
commit fe0dac6dbd23ee87d080cac583158a23cc238cec (origin/PXC-5296-trunk)
Author: Jaideep Karande <jaideep.karande@percona.com>
Date:   Wed Jul 22 14:43:04 2026 +0530

    Revert "PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=ENFORCING"
    
    This reverts commit fa84530038f20409f6265a32eb3fa8882e69af5f.
```

jaideepkarande:
## Jenkins status — not yet validated

Run [#899](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/899/) against this branch is **invalid** and should be ignored. SST could not run: the PXB 9.1.0 tarball is staged into `pxc_extra/pxb-9.6` while this line's `scripts/wsrep_sst_xtrabackup-v2.sh` sets `XB_THIS_REQUIRED_VERSION="9.6.0-1"`, so `verify_pxb_version()` rejects it and effectively the whole galera suite fails. Unrelated to the code here.

A valid run needs the temporary CI-only PXB-version relaxation for this line (kept out of the PR by convention, on a separate `-ci` branch). That patch is not available for 9.7 at the moment, so **this PR has no Jenkins validation yet** and should stay in draft until it does.

The equivalent change is validated on 8.4 — see #2353, run [#890](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/890/): 5 failures against a matched baseline's 8, with all four affected tests passing in both `enc_off` and `enc_on`. The fix regions on this branch are byte-identical to that one.

PR #2366, opened 2026-09-14
[8.4]PXC-5140: Invisible PK is generated even after sql_generate_invisible…

jaideepkarande:
…_primary_key is turned OFF globally

https://perconadev.atlassian.net/browse/PXC-5140

Problem:
sql_generate_invisible_primary_key is a session variable. During wsrep TOI, DDL is replicated as raw statement text without the originating session's variable state. Consequently, a PK-less CREATE TABLE can produce different table definitions across nodes when the global setting is OFF. ALTER TABLE operations can also fail to apply when applier session settings differ from the originating node.

Description:
WSREP TOI does not preserve session variables that affect DDL execution. Applier threads may also retain stale values inherited from their creation-time global settings, causing inconsistent behavior when global settings change.

Resolution:

Introduce WSREP_SESSION_FLAGS_EVENT, a wsrep-private Intvar control event carrying a bitmask of session flags that affect replicated DDL. Propagate the originating session's effective setting for CREATE TABLE, ALTER TABLE, CREATE INDEX, and DROP INDEX. Apply the event before deferred-event collection so the state is available when the associated DDL executes.
Re-seed applier session variables from the current global settings for each write set.

Preserve compatibility: no on-disk, wire, or wsrep protocol changes; older nodes ignore the event and retain existing behavior.


jaideepkarande:
## Testing results

Diff under test: 20 files (4 source + 16 test) against merge base `247589d70d1`.

### Local MTR

| Suite | Result |
|---|---|
| `galera.galera_gipk_global_on` / `_global_off` / `_mixed_global` / `_as_master` (× `enc_off`/`enc_on`) | **pass** — and **3 of 4 fail on the unfixed tree**, so they reproduce the bug |
| `galera.galera_nbo_binlog_control_intvar` | **pass 7/7**; **fails 7/7 with 12 server crashes** without the relay-log initialisation |
| `galera_nbo` (full suite) | **63/63 pass** |
| 24 upstream GIPK replication tests — 12 `rpl`, 1 `rpl_nogtid`, 2 `group_replication` | **all pass, unchanged** |
| full `galera` suite (757 tests), two sweeps + unfixed-baseline comparison | only `galera_gra_log` attributable — expected write-set content change, `.result` re-recorded |

The new per-node assertions (`suite/galera/include/galera_gipk_assert.inc`) check, on **every**
node: `my_row_id` present, that it is the PRIMARY key, that it is AUTO_INCREMENT + INVISIBLE,
and that the node's column definition is identical to node_1's. Per-node checks catch "all
nodes wrong the same way"; the cross-node comparison catches divergence. Verified to bite —
against the unfixed tree they fail on the exact node and value
(`node_2: t4 my_row_id column present = 0` → `1 = 0`).

### Jenkins

| Run | Commit tested | Result |
|---|---|---|
| [`pxc-8.x-pipeline-parallel-mtr` #924](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/924/) | earlier revision | UNSTABLE — 10882 pass / 8 fail; **all new tests pass** |
| [`pxc-8.x-pipeline-parallel-mtr-AI` #1](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr-AI/1/) | earlier revision | UNSTABLE — 7225 pass / 3 fail |
| [`pxc-8.x-pipeline-parallel-mtr-AI`](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr-AI/) (queued) | **this head** | running — Debug, FULL_MTR, ubuntu:noble |

Neither completed run tested this exact head, hence the fresh one.

AI #1's only genuine failure was `galera_gra_log` (deterministic, both combinations, 2/2
retries): an `assert(0)` on the `WSREP_SESSION_FLAGS_EVENT` case in the **mysqlbinlog** print
path aborted the tool (SIGABRT, status 134). The event rides in every DDL write set and PXC
dumps failed write sets as GRA logs precisely so they can be decoded, so asserting there is
wrong. **Already fixed in this head**; the test passes both combinations locally.

Everything else in both runs reproduces on baseline #923 (identical parameters, unrelated
branch) or passes on a serial re-run — `galera_bf_bf_2` fails **3/3 on the unfixed tree**
locally; `galera_truncate`, `galera_alter_table_big`, `pxc_inconsistency_voting`, `galera_wan`,
`galera_var_node_address` are load-sensitive under `--parallel`.

### Real mixed-version cluster

MTR cannot build a mixed-binary cluster, so this was done with two real builds, **with an
unpatched-on-both control run** so every result has attribution. node_old 8.4.6 (unpatched)
bootstraps; node_new 8.4.10 (patched) joins and performs a genuine in-place upgrade.

| | passed | failed | crashes |
|---|---|---|---|
| baseline — unpatched on both nodes | 18 | 15 | **0** |
| mixed — patch on the joiner | 22 | 11 | **0** |

- **Fixes 3**: `GIPK ON, DDL from node_old`, `GIPK ON, CTAS from node_old`,
  `DROP PRIMARY KEY from node_old`.
- **Regresses 1**: `old=OFF new=ON, DDL from node_old`. In baseline the nodes agree *by luck*
  (the applier's stale value happens to match); with the fix the applier re-seeds from its own
  current global and diverges. This is the deliberate "current global vs hard `false`" fallback
  choice — `false` would break a GIPK=ON cluster for the whole upgrade window, which is worse.
- **7 unchanged**, all originating on the unpatched node, which cannot honour a value it does
  not understand.

### NBO across versions with mixed GIPK

| Scenario | baseline | mixed | Verdict |
|---|---|---|---|
| NBO `ADD COLUMN`, old=ON new=OFF, from either node (4 cases) | pass | pass | fine |
| NBO `ADD COLUMN`, old=OFF new=ON, from node_new (2 cases) | pass | pass | fine |
| NBO `ADD COLUMN` on PK-less table, old=OFF new=ON, from node_old | pass | **fail** | same single fallback regression as above |
| NBO `DROP PRIMARY KEY`, GIPK ON on originator | **fail, cluster size 0** | **fail, cluster size 0** | **pre-existing** — identical in the control |
| crashes | **0** | **0** | |

**No crashes in any cross-version run**, on any node — verified by four independent markers
(`got signal`, `Assertion … failed`, `SIGSEGV`, stacktrace) across all node logs. A real 8.4.6
node ignores `WSREP_SESSION_FLAGS_EVENT` exactly as the design intends: `Intvar_event`'s
constructor does not validate the type byte, and the pre-fix `do_apply_event()` switch has no
`default:` branch, so an unknown type is a no-op.

### Compatibility

| Dimension | Impact |
|---|---|
| In-place upgrade / on-disk format | No — no DD, redo/undo, tablespace or system-table change |
| Downgrade | No — one extra ordinary `Intvar` event; unknown types are ignored |
| Rolling / mixed-version cluster | **Improves, does not cure** — 15 → 11 failures, 0 crashes; divergence remains wherever the un-patched node originates DDL |
| Replication (binlog, GTID, cross-version) | No — the event enters only the wsrep write-set buffer, never the binlog. Verified by `galera_gipk_as_master`, which inspects the applier node's binary log directly, and by the 24 unchanged upstream rpl/GR tests |
| Wire protocol / system variables | No |

**Suggested release note**: during a rolling upgrade the cluster is not GIPK-safe. Either keep
`sql_generate_invisible_primary_key` consistent and unchanged for the whole upgrade, or avoid
PK-less `CREATE TABLE` and `ALTER ... DROP PRIMARY KEY` until every node carries the fix.
`ALTER ... DROP PRIMARY KEY` with GIPK ON can evict a node — demonstrated, and **pre-existing**
(it fails identically with no patch anywhere).

### Not covered

- **No sanitizer run.** `ANALYZER_OPTS` was empty in both Jenkins runs, so the deliberately
  unlocked `global_system_variables` read is unverified by TSAN, and leak-class defects are
  invisible to these suites. An ASAN/UBSAN run before merge is recommended.
- **9.7 / trunk propagation** not done against current bases.
- `rpl.rpl_gipk_global_off_async` and `group_replication.gr_gipk_global_off` were written and
  passing (they assert the reported scenario does **not** reproduce outside the wsrep layer)
  but are not included in this PR.

### Related pre-existing defect found on the way

`Relay_log_info::deferred_events_collecting` is declared in `rpl_rli.h` and initialised nowhere
in `rpl_rli.cc`. The NBO worker builds its own rli via `wsrep_relay_log_init()`, and
`wsrep_apply_events()` applies against `thd->wsrep_rli`, so an `Intvar` event reaching an NBO
worker read an indeterminate flag and dereferenced a null `deferred_events` — SIGSEGV.
Reachable on an unmodified tree (`SET sql_log_bin=OFF` + NBO `ALTER`), just rare enough never to
have surfaced. Addressed here by the `wsrep_relay_log_init()` initialisation and covered by
`galera_nbo_binlog_control_intvar`; **the underlying uninitialised member in `Relay_log_info`
deserves its own ticket.**

PR #2369, opened 2026-09-14
[CANCELLED][9.7]PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=ENFORCING

jaideepkarande:
PXC-5106: Set sql_require_primary_key=1 when using pxc_strict_mode=ENFORCING

Problem:

Description
When pxc_strict_mode=ENFORCING or MASTER is set we get ERROR when doing
operations on table without primary key.
Example:
ERROR 1105 (HY000): Percona-XtraDB-Cluster prohibits use of DML command on a
table (test.table_1) without an explicit primary key with
pxc_strict_mode = ENFORCING or MASTER
However pxc_strict_mode=ENFORCING or MASTER does not restrict table creation
without primary keys.
Another system variable sql_require_primary_key enforce the requirement that
tables have a primary key for both creation of new tables or altering the
structure of existing tables
pxc_strict_mode=ENFORCING or MASTER should also enforce the requirement that
tables have a primary key for both creation of new tables or altering the
structure of existing tables

Resolution:

1. Policy consistency
   Before, operators could set 'sql_require_primary_key=OFF' while the cluster
   was in a strict PXC mode, which weakened the intended “primary key required”
   policy.
   The server now enforces: under 'ENFORCING'/'MASTER', turning
   'sql_require_primary_key' 'ON' (via 'SET GLOBAL' or 'SET SESSION') is
   rejected when moving into strict mode, if global 'sql_require_primary_key'
   was 'OFF', the server now sets 'global' to 'ON'.

2. Session vs global
   Changing 'pxc_strict_mode' may update '@@global.sql_require_primary_key'
   as above(if setting to ON).
   The server does not rewrite '@@session.sql_require_primary_key' on
   'existing' connections.
   Each session keeps its session value until the client changes it or opens
   a new connection (new sessions inherit the global default as usual).

3. Startup
   If Galera is active and 'pxc_strict_mode' is 'ENFORCING' or 'MASTER' and the
   server starts with 'global' 'sql_require_primary_key' 'OFF'
   (e.g. from options or config), startup forces
   'global' 'sql_require_primary_key' 'ON' so the effective policy matches
   strict mode.

4. Upgrade
   Do not apply the pxc_strict_mode guard to statements run by the server's
   upgrade threads.


jaideepkarande:
### Jenkins results — 9.7

**[AI job build 9](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr-AI/9/)** · Debug · ubuntu:noble · commit `abd1dd97728`
PXB: 9.7.1-rc1 → `pxb-9.7`, 8.4.0-6 → `pxb-8.4`, 9.1.0 → `pxb-9.1`
Result: **UNSTABLE** — 1104x passed, 6 failed.

#### This commit's own tests all passed

| test | enc_off | enc_on | worker |
|---|---|---|---|
| `galera.galera_pxc_sql_require_primary_key` | pass | pass | galera-big |
| `galera.galera_rpl_pxc_strict_mode_mismatch` | pass | pass | galera-small |
| `galera.pxc_block_write_while_in_rolling_upgrade` | pass | pass | galera-small |
| `galera.pxc_install_gr` | pass | pass | galera-small |

8/8. Worth noting `galera_rpl_pxc_strict_mode_mismatch` — this test is added/modified
by this commit, and it **passes here**, whereas it was among the failures on runs of
branches that did *not* carry this change.

#### Failures — none related to this change

| test | assessment |
|---|---|
| `galera_fk_selfreferential` (enc_off, enc_on) | **pre-existing** — also fails on the clean 9.7 baseline (build 900) and on the PXC-5208 9.7 branch (build 4) |
| `innodb_undo.purge_on_replica` | unrelated InnoDB undo/purge test |
| `test_service_sql_api.test_sql_stmt` | unrelated; also failed on baseline build 900 |

`shutdown_report` entries are MTR's end-of-worker report, raised whenever anything
else in that worker failed; they are not real tests.

PR #2324, opened 2026-07-15
[8.4]PXC-5291: User without super can create trigger even without log_bin_…

jaideepkarande:
…trust_function_creators enabled

https://perconadev.atlassian.net/browse/PXC-5291

The PXC-4765 change folded the trigger DEFINER-mismatch check and the binlog-safety (log_bin_trust_function_creators) check into a single privilege test that accepted SUPER or SET_ANY_DEFINER. As a result a user holding only SET_ANY_DEFINER could create a trigger while log_bin_trust_function_creators = 0, diverging from PS and PXC 8.4.7 which reject it with ER_BINLOG_CREATE_ROUTINE_NEED_SUPER.

Split the two checks in Sql_cmd_create_trigger::execute(): the DEFINER-mismatch check keeps SUPER or SET_ANY_DEFINER, while the binlog-safety check now requires SUPER only. SET_ANY_DEFINER no longer bypasses the binlog-safety gate.

PR #2370, opened 2026-09-14
[CANCELLED][9.7]PXC-5291: User without super can create trigger even without log_bin_trust_function_creators enabled

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-5291

Propagation of the 8.4 fix (#2324) to the `9.7` line.

## Problem

The PXC-4765 change combined trigger DEFINER validation with the
`log_bin_trust_function_creators` check and allowed either SUPER or
SET_ANY_DEFINER to satisfy both. A user holding only SET_ANY_DEFINER could
therefore create a trigger with `log_bin_trust_function_creators=0`,
diverging from PS and PXC 8.4.7, which reject it with
`ER_BINLOG_CREATE_ROUTINE_NEED_SUPER`.

While fixing this, a second issue was found (PXC-5292): the early validation
covered a *differing* DEFINER but not a *nonexistent* one. A SET_ANY_DEFINER
user could pass the pre-TOI check, after which the origin rejected the trigger
during creation while the appliers accepted it — leaving the origin
inconsistent and liable to eviction.

## Resolution

Replace the combined privilege test in `Sql_cmd_create_trigger::execute()`
with two independent checks:

* Call `check_valid_definer()` before TOI to validate differing and
  nonexistent DEFINER accounts, including the SYSTEM_USER rule. It is gated on
  `WSREP(thd)` so that outside a cluster — where there is no TOI and so no
  reason to check early — PS error precedence is preserved (the
  table-privilege check reports first, as in PS).
* Keep the binlog-safety check separate and require SUPER when
  `log_bin_trust_function_creators=0`. SET_ANY_DEFINER no longer satisfies it.
  It still applies under emulated binlogging because Galera replicates the
  trigger body.

`check_valid_definer()` gains a PXC-only `report_no_such_user_warning`
parameter (guarded by `WITH_WSREP`) so the pre-TOI call suppresses the
informational `ER_NO_SUCH_USER` note, which the authoritative post-TOI call
still raises — otherwise the client would see it twice.


jaideepkarande:
## Jenkins — 9.7

[pxc-8.x-pipeline-parallel-mtr #953](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/953/) — **UNSTABLE**
Branch `PXC-5291-9.7` @ `508e01a67fc`, base `upstream/9.7` `db0fe4004d0`. Debug, ubuntu:noble, full MTR.

`16731 total / 10885 passed / 11 failed / 5835 skipped`

### New tests — both pass, both encryption combos

```
[ 52%] galera.pxc_create_trigger_binlog_super 'enc_off'        [ pass ]  1094
[ 52%] galera.pxc_create_trigger_nonexistent_definer 'enc_off' [ pass ]   947
[ 73%] galera.pxc_create_trigger_binlog_super 'enc_on'         [ pass ]  1070
[ 73%] galera.pxc_create_trigger_nonexistent_definer 'enc_on'  [ pass ]   930
```

### Failures — 4 real, none related to this change

7 of the 11 are `shutdown_report` worker meta-entries. The 4 real ones:

| Test | Area |
|---|---|
| `galera_3nodes.galera_parallel_autoinc_manytrx2.enc_on` | parallel autoinc |
| `galera_3nodes.inconsistency_shutdown.enc_off` | inconsistency-eviction shutdown |
| `galera.galera_bf_bf_2.enc_on` | BF-BF conflict handling |
| `galera.galera_many_rows.enc_on` | large transaction |

None create a trigger or exercise DEFINER privilege checks.

PR #2371, opened 2026-09-14
[trunk]PXC-5291: User without super can create trigger even without log_bin_trust_function_creators enabled

jaideepkarande:
https://perconadev.atlassian.net/browse/PXC-5291

Propagation of the 8.4 fix (#2324) to the `trunk` line.

## Problem

The PXC-4765 change combined trigger DEFINER validation with the
`log_bin_trust_function_creators` check and allowed either SUPER or
SET_ANY_DEFINER to satisfy both. A user holding only SET_ANY_DEFINER could
therefore create a trigger with `log_bin_trust_function_creators=0`,
diverging from PS and PXC 8.4.7, which reject it with
`ER_BINLOG_CREATE_ROUTINE_NEED_SUPER`.

While fixing this, a second issue was found (PXC-5292): the early validation
covered a *differing* DEFINER but not a *nonexistent* one. A SET_ANY_DEFINER
user could pass the pre-TOI check, after which the origin rejected the trigger
during creation while the appliers accepted it — leaving the origin
inconsistent and liable to eviction.

## Resolution

Replace the combined privilege test in `Sql_cmd_create_trigger::execute()`
with two independent checks:

* Call `check_valid_definer()` before TOI to validate differing and
  nonexistent DEFINER accounts, including the SYSTEM_USER rule. It is gated on
  `WSREP(thd)` so that outside a cluster — where there is no TOI and so no
  reason to check early — PS error precedence is preserved (the
  table-privilege check reports first, as in PS).
* Keep the binlog-safety check separate and require SUPER when
  `log_bin_trust_function_creators=0`. SET_ANY_DEFINER no longer satisfies it.
  It still applies under emulated binlogging because Galera replicates the
  trigger body.

`check_valid_definer()` gains a PXC-only `report_no_such_user_warning`
parameter (guarded by `WITH_WSREP`) so the pre-TOI call suppresses the
informational `ER_NO_SUCH_USER` note, which the authoritative post-TOI call
still raises — otherwise the client would see it twice.


jaideepkarande:
## Jenkins — trunk

[pxc-8.x-pipeline-parallel-mtr #954](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr/954/) — **UNSTABLE**
Branch `PXC-5291-trunk` @ `3679ef48698`, base `pxc/trunk` `124b735f1c0`. Debug, ubuntu:noble, full MTR.

`17076 total / 11031 passed / 11 failed / 6034 skipped`

### New tests — both pass, both encryption combos

```
[ 47%] galera.pxc_create_trigger_binlog_super 'enc_off'        [ pass ]  1248
[ 48%] galera.pxc_create_trigger_nonexistent_definer 'enc_off' [ pass ]  1092
[ 79%] galera.pxc_create_trigger_binlog_super 'enc_on'         [ pass ]  1231
[ 79%] galera.pxc_create_trigger_nonexistent_definer 'enc_on'  [ pass ]  1062
```

### Failures — 7 real, none related to this change

4 of the 11 are `shutdown_report` worker meta-entries. The 7 real ones:

| Test | Area |
|---|---|
| `innodb_undo.purge_on_replica` | undo purge |
| `innodb.attachable_trx` | attachable transactions |
| `perfschema.idx_compare_metadata_locks` | perfschema index compare |
| `perfschema.idx_compare_mutex_instances` | perfschema index compare |
| `galera.galera_fk_selfreferential.enc_off` | self-referential FK |
| `galera.galera_fk_selfreferential.enc_on` | self-referential FK |
| `galera.galera_sst_allowed_methods.enc_on` | SST method validation |

None create a trigger or exercise DEFINER privilege checks.

sql/auth/sql_authorization.cc
```diff
@@ -7885,6 +7890,9 @@ bool check_valid_definer(THD *thd, LEX_USER *definer) {
                "SUPER or ALLOW_NONEXISTENT_DEFINER");
       return true;
     } else
+#ifdef WITH_WSREP
+        if (report_no_such_user_warning)
```
kamil-holubicki, line 7894: Concern: skipping  curly brackets can be error prone in the future

sql/sql_trigger.cc
```diff
@@ -409,25 +410,21 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
...
-      !lex->definer || /* If definer is not specified, consider current user */
-      (strcmp(lex->definer->user.str, sctx->priv_user().str) == 0 &&
-       my_strcasecmp(system_charset_info, lex->definer->host.str,
-                     sctx->priv_host().str) == 0);
-  bool binlog_requires_super =
+
+  /*
+    check_valid_definer() is the authorization check introduced in WL#15874.
+  */
+  if (WSREP(thd) && lex->definer &&
```
kamil-holubicki, line 417: Observation: I started wondering if this is the best place to do the pre-check. Shouldn't we do it just before TOI?
Let's say that original flow will reject at `find_temporary_table` below. With the precheck we will have another error returned to the user.

sql/sql_trigger.cc
```diff
@@ -409,25 +410,21 @@ bool Sql_cmd_create_trigger::execute(THD *thd) {
...
-      sctx->has_global_grant(STRING_WITH_LEN("SET_ANY_DEFINER")).first;
-
-  // Definer check: If a definer is specified and is different from the current
-  // user, then we need to check for SUPER or SET_ANY_DEFINER privileges.
-  if ((!definer_is_current_user || binlog_requires_super) &&
-      !has_super_or_set_user_id) {
-    if (!definer_is_current_user) {
-      my_error(ER_SPECIFIC_ACCESS_DENIED_ERROR, MYF(0), "SUPER or SET_ANY_DEFINER");
-    } else if (WSREP(thd)) {
+  if (binlog_requires_super && !sctx->check_access(SUPER_ACL)) {
```
kamil-holubicki, line 426: Observation: Actually, it would be good to do the pre-check only for wsrep-applier, so it should be guarded by `!thd->wsrep_applier`.
But on the other hand, it won't hurt the applier either, because it operates in the root context. As I see, other prechecks are not consistent in this.

kamil-holubicki (approved):
One concern and two observations. Nothing blocking.

sql/sql_trigger.cc
jaideepkarande: Addressed the comments, different error message will be seen but behavior is not impacted. Since 8.4 is out, I have made extra commit on top of cherry-pick which I will add to 8.4 in next release(8.4.12).

jaideepkarande:
## Jenkins — trunk

[pxc-8.x-pipeline-parallel-mtr-AI #22](https://pxc.cd.percona.com/job/pxc-8.x-pipeline-parallel-mtr-AI/22/) — **UNSTABLE**
Branch `PXC-5291-trunk` @ `ebf1b02dddc`, base `pxc/trunk` `8454bf57757`.
Debug, ubuntu:noble, full MTR, `GALERA_PARALLEL_RUN=2`.

`17115 total / 11062 passed / 6 failed / 6047 skipped`

### New tests — pass, both encryption combos

```
[ 50%] galera.pxc_create_trigger_binlog_super 'enc_off'        [ pass ]  1377
[ 50%] galera.pxc_create_trigger_nonexistent_definer 'enc_off' [ pass ]  1024
[ 73%] galera.pxc_create_trigger_binlog_super 'enc_on'         [ pass ]  1104
[ 73%] galera.pxc_create_trigger_nonexistent_definer 'enc_on'  [ pass ]  1127
```

This covers the two re-recorded expectations from the review fix — the cases
where the binlog gate now reports instead of the DEFINER check.

### Failures — 3 real, none related to this change

3 of the 6 are `shutdown_report` worker meta-entries. The 3 real ones:

| Test | Prior trunk run (#954) | Classification |
|---|---|---|
| `galera.galera_fk_selfreferential.enc_off` | also failed | pre-existing |
| `galera.galera_fk_selfreferential.enc_on` | also failed | pre-existing |
| `galera.galera_bf_bf_2.enc_on` | — (failed on 8.4 #952 and 9.7 #953) | fragile across lines |

`galera_fk_selfreferential` failed in both encryption combos in the previous
trunk run as well, so it is pre-existing on this line. `galera_bf_bf_2` has now
failed on all three version lines but in differing encryption variants, which
is the signature of a fragile test rather than a deterministic regression.
Neither creates a trigger or exercises DEFINER privilege checks.

### Compared with the previous trunk run

Real failures dropped from 7 (#954) to 3. These now pass:
`innodb_undo.purge_on_replica`, `innodb.attachable_trx`,
`perfschema.idx_compare_metadata_locks`, `perfschema.idx_compare_mutex_instances`,
`galera.galera_sst_allowed_methods.enc_on`.

### SST

Ran with PXB 9.7.1-rc1 as THIS (`pxb-9.7`) and the TESTING PXB 9.1.0 build as
PREV (`pxb-9.6`). No donor failure, and `galera_sst_allowed_methods` passed, so
the PXC-5208 CI-ONLY workaround remains unnecessary on this line.

Note: this run is on the `-AI` job; the earlier #954 figures quoted above come
from `pxc-8.x-pipeline-parallel-mtr`.

